PR #16, opened 2019-06-17
alertlog: graphql2 endpoint

arurao:

**Description:**
This PR migrates `alertlogs` endpoint to graphql2. 
They can be retrieved by querying `alert` which has a field `recentEvents`.

Alerts query now has a field `recentEvents` which returns a list of log entries for that alert. It has cursor support for pagination and a limit that can be specified as inputs.


graphql2/schema.graphql
```diff
@@ -525,6 +533,19 @@ input AlertSearchOptions {
   omit: [Int!]
 }
 
+input AlertLogSearchOptions {
+	first: Int = 15
+	after: String = ""
+	filterByAlertID: Int
+	filterByServiceID: String
+	filterByUserID: String
+	filterByIntegrationKeyID: String
```
mastercactapus: Should these take arrays instead of single values? Especially for `event`, I could see `[created, duplicate]` being a valid query.

Updating the SQL query to use `any( ... )` should make that work.

What do you think?

graphql2/graphqlapp/alertlog.go
```diff
@@ -0,0 +1,129 @@
...
+		err = search.ParseCursor(*opts.After, &s)
+		if err != nil {
+			return nil, errors.Wrap(err, "parse cursor")
+		}
+	}
+
+	if opts.FilterByAlertID != nil {
+		s.AlertID = *opts.FilterByAlertID
+	} else {
+		s.AlertID = 0
```
mastercactapus: Zero is already the default state, so we can lose the `else` block.

graphql2/graphqlapp/alertlog.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	if opts.FilterByServiceID != nil {
+		s.ServiceID = *opts.FilterByServiceID
+	}
+
+	s.SortDesc = *opts.SortDesc
```
mastercactapus: Nil check needed or this can panic.

graphql2/graphqlapp/alertlog.go
```diff
@@ -0,0 +1,129 @@
...
+
+	if opts.First != nil {
+		s.Limit = *opts.First
+	}
+	if s.Limit == 0 {
+		s.Limit = 15
+	}
+
+	err = validate.Many(
+		validate.Range("First", s.Limit, 1, 100),
```
mastercactapus: Use `search.MaxResults`

graphql2/graphqlapp/alertlog.go
```diff
@@ -0,0 +1,129 @@
...
+		case graphql2.EventEscalationRequest:
+			s.Event = "escalation_request"
+		}
+	}
+
+	if opts.First != nil {
+		s.Limit = *opts.First
+	}
+	if s.Limit == 0 {
+		s.Limit = 15
```
mastercactapus: Use `search.DefaultMaxResults`

graphql2/graphqlapp/alertlog.go
```diff
@@ -0,0 +1,129 @@
...
+
+	s.SortDesc = *opts.SortDesc
+
+	if opts.Event != nil {
+		switch *opts.Event {
+		case graphql2.EventCreated:
+			s.Event = "created"
+		case graphql2.EventClosed:
+			s.Event = "closed"
+		case graphql2.EventNotificationSent:
```
mastercactapus: Can we map `alertlog.Type` into graphql directly (inside `gqlgen.yml`) instead of using strings here?

graphql2/graphqlapp/alertlog.go
```diff
@@ -0,0 +1,129 @@
...
+	return graphql2.Event(log.Event), nil
+}
+
+func (l *AlertLog) AlertID(ctx context.Context, log *alert.Log) (int, error) {
+	var s alertlog.SearchOptions
+	s.Start = log.Timestamp
+	s.Event = alertlog.Type(log.Event)
+	s.Limit = 1
+
+	entries, _, err := l.LogStore.Search(ctx, &s)
```
mastercactapus: Is there a way we can get this (e.g. from the `alert.Log`) without doing a query?

graphql2/gqlgen.yml
```diff
@@ -30,6 +30,8 @@ models:
     model: github.com/target/goalert/alert.Alert
   AlertState:
     model: github.com/target/goalert/alert.State
+  AlertLog:
+    model: github.com/target/goalert/alert.Log
```
mastercactapus: Should this be `log.Entry` instead of `alert.Log`? I think the `alert.Log` is only intended for "already acknowledged" type things.

arurao:
> I propose we limit the scope of this PR to the minimum required for the details page and expand on it when log search actually goes in. Then we can take the time to walk through a proper proposal for how alert log searching will behave in the new API.
> 
I like the idea of limiting the scope of this PR to including only the basic endpoint for alert logs. The search logic could be done in a separate PR. 

> Alternatively, we could add a field to `alert` like `recentEvents` with an optional limit and no pagination.
> 
> ```graphql
> type Alert {
>   // ... existing fields...
> 
>   recentEvents(limit: Int = 50): [AlertLogEntry!]
> }
> 
> type AlertLogEntry {
>   id: Int!
>   timestamp: ISOTimestamp!
>   message: String!
> }
> ```
I like this approach. Keeps it super short and simple. Building on top of this piece-by-piece should be easy too.

alert/log/store.go
```diff
@@ -637,7 +644,8 @@ func (db *DB) Search(ctx context.Context, opts *SearchOptions) ([]Entry, int, er
 	(coalesce(a.timestamp < cast($4 as timestamp with time zone), true)) and
 	($5 = '' or a.event = $5::enum_alert_log_event)and
 	($6 = '' or a.sub_user_id = cast($6 as UUID)) and 
-	($7 = '' or a.sub_integration_key_id = cast($7 as UUID))`
+	($7 = '' or a.sub_integration_key_id = cast($7 as UUID)) and
+	($8 = '0' OR a.id < $8 :: int)`
```
mastercactapus: I think we should revert this method to the original, and rename it to `LegacySearch` as we've done elsewhere (when switching to cursor pagination) and then implement a new `Search` method for cursor pagination.

This one is using count + limit/offset and wont give what we are expecting if we combine the two, or in the least will make it more confusing to follow.

graphql2/schema.graphql
```diff
@@ -543,6 +543,25 @@ type Alert {
 
   # Escalation Policy State for the alert.
   state: AlertState
+
+  # recent log entries for the alert.	
+  recentEvents(input: AlertRecentEventsOptions): AlertLogEntryConnection!
+}
+
+input AlertRecentEventsOptions {
+  limit: Int = 50
```
mastercactapus: `DefaultMaxResults` that we're using is 15, we should reflect that here, or perhaps better, omit the "default" since they are all the same. That way the schema doesn't get out of date if we change `DefaultMaxResults`.

graphql2/graphqlapp/alert.go
```diff
@@ -155,6 +176,63 @@ func (a *Alert) Service(ctx context.Context, raw *alert.Alert) (*service.Service
...
+
+	var s alertlog.SearchOptions
+	if opts.Limit != nil {
+		s.Limit = *opts.Limit
+	}
+	if s.Limit == 0 {
+		s.Limit = search.DefaultMaxResults
+	}
+	err := validate.Many(
+		validate.Range("Limit", s.Limit, 15, search.MaxResults),
```
mastercactapus: Minimum should be `1`, but we can drop this entirely as the `Search` method should do the validation already.

graphql2/graphqlapp/alert.go
```diff
@@ -12,12 +13,32 @@ import (
 	"github.com/target/goalert/validation/validate"
```
mastercactapus: nit:  remove extra whitespace

mastercactapus: nit: remove extra whitespace and have goimports run (should be automatic)

graphql2/graphqlapp/alert.go
```diff
@@ -155,6 +176,56 @@ func (a *Alert) Service(ctx context.Context, raw *alert.Alert) (*service.Service
...
+	}
+
+	if opts.Limit != nil {
+		s.Limit = *opts.Limit
+	}
+	if s.Limit == 0 {
+		s.Limit = search.DefaultMaxResults
+	}
+
+	aID, err := a.AlertID(ctx, obj)
```
mastercactapus: Can we just use `obj.ID`?

alert/log/search.go
```diff
@@ -0,0 +1,146 @@
...
+	"text/template"
+
+	"github.com/lib/pq"
+	"github.com/pkg/errors"
+)
+
+// SearchOptions contains criteria for filtering alert logs.
+type SearchOptions struct {
+	// FilterAlertIDs restricts the log entries belonging to specific alertIDs only.
+	FilterAlertIDs []int
```
mastercactapus: Add a JSON tag for this

alert/log/search.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// Limit restricts the maximum number of rows returned. Default is 15.
+	Limit int
```
mastercactapus: Add `json:"-"` so this value doesn't get encoded into the cursor.

alert/log/search.go
```diff
@@ -0,0 +1,146 @@
...
+		u.name,
+		a.sub_integration_key_id,
+		i.name,
+		a.sub_hb_monitor_id,
+		hb.name,
+		a.sub_channel_id,
+		nc.name,
+		a.sub_classifier,
+		a.meta
+	FROM alert_logs a
```
mastercactapus: Maybe call this `log` instead of `a` for clairity.

alert/log/search.go
```diff
@@ -0,0 +1,146 @@
...
+	{{if .FilterAlertIDs }}
+		LEFT JOIN alerts ON alerts.id = a.alert_id
```
mastercactapus: It doesn't look like we are using this join on the `alerts` table anywhere, can we drop it?

alert/log/search.go
```diff
@@ -0,0 +1,146 @@
...
+	{{end}}
+	LEFT JOIN users u ON u.id = a.sub_user_id
+	LEFT JOIN integration_keys i ON i.id = a.sub_integration_key_id
+	LEFT JOIN heartbeat_monitors hb ON hb.id = a.sub_hb_monitor_id 
+	LEFT JOIN notification_channels nc ON nc.id = a.sub_channel_id
+	WHERE TRUE
+	{{if .FilterAlertIDs}}
```
mastercactapus: nit: we should use consistent spacing, add `-` to match the `if` statement below.

alert/log/search.go
```diff
@@ -0,0 +1,146 @@
...
+
+type renderData SearchOptions
+
+func (opts renderData) Normalize() (*renderData, error) {
+	if opts.Limit == 0 {
+		opts.Limit = search.DefaultMaxResults
+	}
+
+	var err error
+	if len(opts.FilterAlertIDs) <= 0 {
```
mastercactapus: These don't need to be required, the query will work fine without and filter IDs. We impose a max limit so the cost is already bounded.

The only thing we should validate is that the length of `FilterAlertIDs` isn't astronomical (since we send the list to the DB). Maybe limit it to 50 or something beyond what is realistic, but well below an amount that could cause problems.

graphql2/graphqlapp/alert.go
```diff
@@ -3,21 +3,41 @@ package graphqlapp
...
+	"github.com/target/goalert/alert/log"
 	"github.com/target/goalert/assignment"
 	"github.com/target/goalert/graphql2"
 	"github.com/target/goalert/permission"
 	"github.com/target/goalert/search"
 	"github.com/target/goalert/service"
 	"github.com/target/goalert/validation/validate"
-
-	"github.com/pkg/errors"
+	"time"
```
mastercactapus: nit: Run `goimports` on this file, these should be sorted.

graphql2/graphqlapp/alert.go
```diff
@@ -155,6 +175,52 @@ func (a *Alert) Service(ctx context.Context, raw *alert.Alert) (*service.Service
...
+	}
+
+	if opts.Limit != nil {
+		s.Limit = *opts.Limit
+	}
+	if s.Limit == 0 {
+		s.Limit = search.DefaultMaxResults
+	}
+
+	s.FilterAlertIDs = append(s.FilterAlertIDs, obj.ID)
```
mastercactapus: Search options (other than `Limit` and `After`) should be set before `ParseCursor` otherwise the cursor won't behave properly.

app/initgraphql.go
```diff
@@ -23,6 +23,7 @@ func (app *App) initGraphQL(ctx context.Context) error {
 		NRStore:           app.NotificationRuleStore,
 		NCStore:           app.NCStore,
 		AlertStore:        app.AlertStore,
+		AlertLogStore:     app.AlertlogStore,
```
mastercactapus: Can we do a find & replace for `AlertlogStore` -> `AlertLogStore`?

smoketest/graphql2alertlogs_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/target/goalert/smoketest/harness"
+	"testing"
+)
+
+// TestGraphQL2AlertLogs tests that logs are properly generated for an alert.
+func TestGraphQL2AlertLogs(t *testing.T) {
```
mastercactapus: Let's drop the 2 from the name and file, it's only temporary while we transition to the new API.

smoketest/graphql2alertlogs_test.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	sid := h.UUID("sid")
+	h.CreateAlert(sid, "alert1")
+
+	// Acknowledging alert
+	doQL(fmt.Sprintf(`
+		mutation {
+			updateAlerts(input: {
+				alertIDs: [%d],
```
mastercactapus: Would it be more efficient to just add the entries into the DB through init and assert on the output (single GraphQL call)?

arurao: I like the idea of testing log generation too, i.e. performing actions and ensuring it generates log entries in the db and then retrieving them. Although this file is specifically for the graphql mutation for logs, it could be a nice-to-have. What do you think?

mastercactapus: Hmm. That is a fair point. Let's stick with it for now.

graphql2/schema.graphql
```diff
@@ -545,6 +545,25 @@ type Alert {
 
   # Escalation Policy State for the alert.
   state: AlertState
+
+  # recent log entries for the alert.
```
m17ch: nit: capitalize **R**ecent

PR #18, opened 2019-06-18
ui dependencies: non-breaking change package updates

Auchindoun:

**Description:**
This PR updates several packages and libraries that do _not_ include breaking changes.

**Additional Info:**
It also removes the following packages that are no longer being used: `express`, `express-static-gzip`, and `react-sortable-hoc`.

`this.context.router` was deprecated a very long time ago, and looks to be no longer supported. This PR switches the few places we still used `context.router` to the `withRouter` HOC to get the `history` prop injected into the relevant classes.

`chance.word()` doesn't actually support `min` and `max` props. Changes to use `length` with the average of the two (caught by the `@types/chance` package upgrade).


Auchindoun:
Tests passed.

PR #23, opened 2019-06-19
switchover: hardening & refactor instrumentation

mastercactapus:

**Description:**
The switchover implementation had become out of date and not all tables had correct triggers to allow logical replication. This PR updates the switchover shell code to create triggers dynamically to prevent that from happening in the future.

- Removed migration-created triggers in favor of a pragmatic approach
- Fixed some cancelation and progress bar bugs
- Added a README covering some of the basics of the feature
- Changed how config validation works so different URLs can be used as long as they result in connections to the correct database(s)


switchover/dbsync/dbid.go
```diff
@@ -0,0 +1,15 @@
+package dbsync
+
+import (
+	"crypto/rand"
+	"encoding/hex"
+)
+
+func newDBID() string {
```
m17ch: Would you mind walking me through how this is leveraged?  From `switchover/dbsync/sync.go` it looks like it's being used to create both `oldDbID` and `newDbID`, so I'm confused how it's leveraged in the context of config verification.

mastercactapus: Hmm. Yeah, I suppose that is a little confusing. The "new" in the context of this function is just referring to creation, like `http.NewServeMux` creates a new `http.ServeMux`.

Inside `sync.go` oldDB things refer to the `--db-url` database, the one being switched away from. newDB things refer to the `--db-url-next`, or the one we are switching _to_.

Before we call it an official/hardened feature, it might be worth coming up with a better naming scheme for tracking the two databases, rather than old and new.

As far as the purpose of this function, each time the shell starts it generates a random ID for each DB, and publishes that ID to a channel on only that DB. It serves as a functional test that what the shell thinks is the next/new DB is the same one as what each GoAlert instance thinks is the next/new DB. Same for the old/prev DB.

Does that help?

migrate/migrate.go
```diff
@@ -112,7 +113,11 @@ func ensureTableQuery(ctx context.Context, db *sql.DB, fn func() error) error {
 	}
 	// 42P01 is undefined_table
 	// https://www.postgresql.org/docs/9.6/static/errcodes-appendix.html
-	if pErr, ok := err.(*pq.Error); !ok || pErr.Code != "42P01" {
+	if pqErr, ok := err.(*pq.Error); ok && pqErr.Code == "42P01" {
```
JohnRoesler: Worth making the pq error code a const? Only two references at this point, but just a thought

mastercactapus: Hmm. Not a bad idea, maybe a package with Postgres errors that we use in-app to hold those.

`pqerr.UndefinedTable` or something like that?

We are using 2 different drivers (pgx for switchover and data generation, pq for everything else), but it seems like they both play well in keeping the code the same format/string.

I like the idea of something like `pqerr.IsCode(pqerr.CodeUndefinedTable, err)` or simply `pqerr.IsCode("42P01", err)` that could handle the two drivers transparently. There's a lot of stuff in `maperror.go` that could be more portable/cleaner that way.

---

I think that might be better as a PR on its own though -- there are quite a few places we are doing these checks, and I'd rather not introduce a new style in this particular spot without updating the rest to match.

@m17ch @arurao What do you guys think?

JohnRoesler: We do something similar in an API i'm working on - keeps it transparent and allows us to do any massaging of the errors we may want to do.

switchover/dbsync/dbid.go
m17ch: Yes - thank you!

switchover/dbsync/shell.go
```diff
@@ -130,43 +133,81 @@ func RunShell(oldURL, newURL string) error {
...
 			sh.Println(status)
-			sh.Println("change_log enabled.")
-
+			sh.Println("change_log disabled")
 			return nil
 		},
 	})
 	sh.AddCmd(ctxCmd{
 		Name: "disable",
 		Help: "Enable change_log",
```
Auchindoun: Enable -> Disable?

switchover/dbsync/shell.go
```diff
@@ -294,12 +356,19 @@ func RunShell(oldURL, newURL string) error {
...
+				err = s.Sync(ctx, false, false)
+				if err != nil {
+					return err
+				}
+				dur := time.Since(start).Truncate(time.Second / 10)
+				sh.Printf("Completed sync in %s\n", dur.String())
+				if dur < maxSync {
+					break
+				}
+				sh.Println("Took longer than max sync time, re-syncing")
```
Auchindoun: potential to become an infinite loop here?

mastercactapus: Yes, intentionally. It will keep re-syncing the DB until it catches up, or the user uses CTRL+C to stop the process.

It's hard to predict how many times it may take, and if it were to stop at an arbitrary number, a busy system could fall behind again.

Since it's safe to run multiple times and can be interrupted at any time, it will keep going as long as is needed this way, rather than having the user re-run `execute` and confirm over and over.

PR #27, opened 2019-06-20
Favorite rotations

dayander:

**Description:**
- Added favoriting functionality to rotations. Ordering on rotations is by favorites first. In search selects for rotations, favorites are first.


web/src/app/main/components/ToolbarTitle.js
```diff
@@ -32,6 +32,10 @@ const styles = {
   title: {
     padding: '0 4px 0 4px',
     flex: 1, // pushes toolbar actions to the right
+    fontWeight: 500,
```
Auchindoun: it would make more sense to add this to a separate PR, especially one where we can tackle some more accessibility issues all at once.

web/src/app/rotations/components/SetFavoriteButton.js
```diff
@@ -0,0 +1,109 @@
...
+  }
+`
+
+const mutation = gql`
+  mutation setFav($input: SetFavoriteInput!) {
+    setFavorite(input: $input)
+  }
+`
+
+export default class SetFavoriteButton extends React.PureComponent {
```
Auchindoun: I would recommend that we, instead of having multiple components named `SetFavoriteButton` under multiple directories within GoAlert with the majority of the same code between them, we consolidate these files into one (probably under `app/util`).

It could accept props that are more generic, such as `targetID: p.string.isRequired` as well as `targetType: p.oneOf(['service, rotation'])` which can be used to determine the data key, as well as updating the cache.

This file can also be written as a React hook, now that we're on the latest React version.

web/src/app/rotations/components/SetFavoriteButton.js
```diff
@@ -0,0 +1,109 @@
...
+        mutation={mutation}
+        client={graphql2Client}
+        update={cache => {
+          const variables = { id: this.props.rotationID }
+          // get cached service
+          const { rotation } = cache.readQuery({
+            query,
+            variables,
+          })
+          console.log(variables, rotation)
```
Auchindoun: del console.log

web/src/cypress/integration/favorites.ts
```diff
@@ -0,0 +1,107 @@
+import { testScreen } from '../support'
+import { Chance } from 'chance'
+const c = new Chance()
+
+testScreen('Favorites', testFavorites)
+
+function testFavorites(screen: ScreenFormat) {
```
Auchindoun: If we create a new test integration file just for favorites, we will want to move the service favorites tests over here as well, and since the tests between these will be mostly the same, we can have a function that accepts a couple params that handles doing everything, then within each `describe` we would have something like

(pseudo code):

```
describe Rotation Favorites {
  it should have favored rotations move to first on rotation list:
    testSortOrder(rot.name)

  it should be able to set as a favorite
    testSettingFavorite(rot.id)

  // other rotation specific tests
}

describe Service Favorites {
  . . .
}
```

web/src/cypress/integration/favorites.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    it('should have favored rotations move to first on rotation list', () => {
+      cy.pageSearch(rot.name)
+      cy.get('#app')
+        .contains(rot.name)
+        .click()
+      cy.get('button[aria-label="Set as a Favorite Rotation"]')
+        .click()
+        .visit('/rotations')
+      cy.get('#app').contains(rot.name)
```
Auchindoun: In order to make sure it's first in the list, we need to assert its position within the list. I believe this can be achieved with `eq`.

e.g.: `cy.get('SomeListReference').eq(0)`

web/src/cypress/integration/favorites.ts
```diff
@@ -0,0 +1,107 @@
...
+    })
+
+    it('should have favorites first in RotationSelect on escaltion policy steps', () => {
```
Auchindoun: typo: `escaltion` -> `escalation`

web/src/cypress/integration/favorites.ts
```diff
@@ -0,0 +1,107 @@
...
+        .contains(rot.name)
+        .click()
+      cy.get('button[aria-label="Set as a Favorite Rotation"]')
+        .click()
+        .visit('/escalation-policies')
+
+      cy.pageFab()
+
+      cy.get('div[role=dialog]').as('dialog')
+      cy.get('@dialog').should('contain', 'Create Escalation Policy')
```
Auchindoun: we can utilize some cypress support functions we have created in the past to create an escalation policiy by directly querying Graphql, which should speed up this test's runtime. Before the test runs you can do something along the lines of:

```
let ep = null
cy.createEP().then(e => {
  ep = e
  cy.visit(`/escalation-policies/${ep.id}`)
  cy.reload()
})

// open dialog and check dropdown
```

web/src/cypress/integration/favorites.ts
```diff
@@ -0,0 +1,107 @@
...
+        .contains(rot.name)
+        .click()
+      cy.get('button[aria-label="Set as a Favorite Rotation"]')
+        .click()
+        .visit('/schedules')
+
+      const name = c.word({ length: 8 })
+      const description = c.sentence({ words: 5 })
+
+      cy.pageFab()
```
Auchindoun: You can do the same here as well, to create a schedule directly.

web/src/cypress/integration/rotations.ts
```diff
@@ -103,6 +103,17 @@ function testRotations(screen: ScreenFormat) {
       }),
     )
 
+    it('should allow setting and unsetting as a favorite rotation', () => {
```
Auchindoun: If we move the rest of the favorites tests into one favorites file, this test can be moved there as well as the service's favorites tests

web/src/cypress/integration/rotations.ts
```diff
@@ -252,7 +263,9 @@ function testRotations(screen: ScreenFormat) {
         .clear()
         .type('5')
 
-      cy.get('button[type=submit]').click()
+      cy.get('button[type=submit]')
+        .eq(1)
```
Auchindoun: Determining which submit button based off of order within the DOM can be fragile in future tests if some code ever changes, which could cause regressions in the future. I would suggest adding a `data-cy=''` attribute to the submit button you're trying to press, and in cypress code you can reference it by `cy.get('button[data-cy="my-attribute-name"]').click()`, which will guarantee that it will always be the correct submit button clicked on.

web/src/cypress/integration/favorites.ts
dayander: From the tests running there could be anywhere from 1- 15 rotations in the rotation list, so I don't think testing for the position in the list would work well. I am not sure what would be the best practice to test for something like this.

web/src/app/rotations/components/SetFavoriteButton.js
dayander: What would be the best practice for the queries and mutations when moving this to util? Should we ask for a query and mutation prop?

Auchindoun: Yeah, that's what we do in other locations. For example we have a generic `FormDialog` component, but we send it the form as a prop

web/src/cypress/integration/favorites.ts
Auchindoun: If you ensure that only one rotation is a favorite (going from 0 favorites -> 1 favorite), we can guarantee that asserting based off the first rotation in the list will be successful

web/src/app/rotations/components/RotationSetFavoriteButton.js
```diff
@@ -0,0 +1,77 @@
...
+  render() {
+    return (
+      <Query
+        query={query}
+        variables={{ id: this.props.rotationID }}
+        render={({ data }) => {
+          if (!data || !data.rotation) return null
+
+          return this.renderMutation(
+            data && data.rotation && data.rotation.isFavorite,
```
Auchindoun: since we are returning out if data.rotation is null, we can safely assume that it exists here and instead `return this.renderMutation(data.rotation.isFavorite)`

web/src/app/rotations/components/RotationSetFavoriteButton.js
```diff
@@ -0,0 +1,77 @@
...
+          )
+        }}
+      />
+    )
+  }
+
+  renderMutation(isFavorite) {
+    return (
+      <Mutation
+        key='main'
```
Auchindoun: a key shouldn't be needed here as we aren't rendering an array

web/src/app/rotations/components/RotationSetFavoriteButton.js
```diff
@@ -0,0 +1,77 @@
...
+  }
+`
+
+const mutation = gql`
+  mutation setFav($input: SetFavoriteInput!) {
+    setFavorite(input: $input)
+  }
+`
+
+export default class RotationSetFavoriteButton extends React.PureComponent {
```
Auchindoun: since this is a new component, i would prefer if we wrote this in hooks. i can help you with this if you want

Auchindoun: @m17ch @arurao @mastercactapus Should a generic `SetFavoriteButton` include both the query and the mutations, or just include the icon/relevant styling?

web/src/app/util/SetFavoriteButton.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
+import p from 'prop-types'
+import IconButton from '@material-ui/core/IconButton'
+import FavoriteFilledIcon from '@material-ui/icons/Star'
+import FavoriteBorderIcon from '@material-ui/icons/StarBorder'
+
+export default class SetFavoriteButton extends React.PureComponent {
```
Auchindoun: if we could also write this in hooks

web/src/cypress/integration/rotations.ts
```diff
@@ -252,7 +252,9 @@ function testRotations(screen: ScreenFormat) {
         .clear()
         .type('5')
 
-      cy.get('button[type=submit]').click()
+      cy.get('button[type=submit]')
```
mastercactapus: Can we add something like `[data-cy=edit]` or maybe reference an aria label on the edit button? Selecting all submit buttons, then excluding `set-fav` seems fragile (e.g. if we add an another one)

web/src/cypress/integration/alerts.ts
```diff
@@ -62,6 +62,42 @@ function testAlerts(screen: ScreenFormat) {
       cy.get('body').should('not.contain', 'No results') // mock alerts should show again
     })
 
+    it('should filter through alerts, active, unacknowledged, acknowledged, closed, and all', () => {
```
mastercactapus: This test seems a little too dependant on DB state. To be fair, the one above it is too. We should probably truncate the user favorites table before it runs to ensure a clean/known state.

But for this one testing filters, we need to create known alerts of each type to ensure the filter is working properly. Though it may or may not be worth it with the redesign work pending.

migrate/migrations/20190613111645-add-rotation-favorite.sql
```diff
@@ -0,0 +1,17 @@
+-- +migrate Up
+ALTER TABLE user_favorites ADD COLUMN tgt_rotation_id UUID REFERENCES rotations (id);
+
+ALTER TABLE user_favorites ALTER COLUMN tgt_service_id DROP NOT NULL;
+
+ALTER TABLE user_favorites ADD CONSTRAINT user_fav_rotations UNIQUE (user_id, tgt_rotation_id);
```
mastercactapus: I think we can put these all into a single statement. That way we only run a single round-trip to Postgres while holding the table lock instead of 3.

As for the name, these have to be globally unique. I'd suggest just going with the Postgres default: `user_favorites_user_id_tgt_rotation_id_key` to match what we have for services.

```sql
-- Table Definition ----------------------------------------------

CREATE TABLE user_favorites (
    user_id uuid NOT NULL REFERENCES users(id) ON DELETE CASCADE,
    tgt_service_id uuid NOT NULL REFERENCES services(id) ON DELETE CASCADE,
    id BIGSERIAL UNIQUE,
    CONSTRAINT user_favorites_user_id_tgt_service_id_key UNIQUE (user_id, tgt_service_id)
);

-- Indices -------------------------------------------------------

CREATE UNIQUE INDEX user_favorites_user_id_tgt_service_id_key ON user_favorites(user_id uuid_ops,tgt_service_id uuid_ops);
CREATE UNIQUE INDEX user_favorites_uniq_id ON user_favorites(id int8_ops);
```

schedule/rotation/search.go
```diff
@@ -22,33 +22,65 @@ type SearchOptions struct {
...
+
+	//FavoritesUserID is the userID associated with the rotation favorite
+	FavoritesUserID string `json:"u,omitempty"`
 }
 
 // SearchCursor is used to indicate a position in a paginated list.
 type SearchCursor struct {
-	Name string `json:"n,omitempty"`
+	Name       string `json:"n,omitempty"`
+	IsFavorite bool   `json:"f"`
```
mastercactapus: Add `omitempty`

schedule/rotation/search.go
```diff
@@ -22,33 +22,65 @@ type SearchOptions struct {
...
 	{{if .SearchStr}}
 		AND (rot.name ILIKE :search OR rot.description ILIKE :search)
 	{{end}}
 	{{if .After.Name}}
-		AND lower(rot.name) > lower(:afterName)
+		AND
+		{{if not .FavoritesFirst}}
+			lower(rot.name) > lower(:afterName)
+		{{else if .After.IsFavorite}}
+			((fav notnull AND lower(rot.name) > lower(:afterName)) OR fav isnull)
```
mastercactapus: Should the first part be `fav is distinct from null` like above?

service/search.go
```diff
@@ -47,7 +47,7 @@ var searchTemplate = template.Must(template.New("search").Parse(`
 		svc.name,
 		svc.description,
 		svc.escalation_policy_id,
-		fav notnull
+		fav IS DISTINCT FROM NULL
```
mastercactapus: I think we'll need to update the other `notnull` below

web/src/cypress/integration/alerts.ts
dayander: If this is pending updates, is this something I should remove till the updates happen?

web/src/cypress/integration/rotations.ts
```diff
@@ -218,7 +217,7 @@ function testRotations(screen: ScreenFormat) {
         .should('not.contain', 'Active')
     })
 
-    it('should allow deleting the rotation', () => {
+    it.only('should allow deleting the rotation', () => {
```
mastercactapus: Remove `.only` (it will prevent all other tests from running).

web/src/app/util/QuerySetFavoriteButton.js
```diff
@@ -0,0 +1,88 @@
...
+          },
+        })
+      }}
+    />
+  )
+}
+
+QuerySetFavoriteButton.propTypes = {
+  id: p.string.isRequired,
+  typeName: p.oneOf(['rotation', 'service']),
```
mastercactapus: Instead of requiring both type and ID from all callers, can we just use `serviceID` and `rotationID` fields?

node_modules/.yarn-integrity
```diff
@@ -0,0 +1,10 @@
+{
```
mastercactapus: How did this get committed? (`node_modules` should be ignored)

web/node_modules/.yarn-integrity
```diff
@@ -0,0 +1,10 @@
+{
```
mastercactapus: Same as the other, this file should be removed

user/favorite/store.go
```diff
@@ -41,14 +41,15 @@ func NewDB(ctx context.Context, db *sql.DB) (*DB, error) {
...
-			VALUES ($1, $2)
+			INSERT INTO user_favorites (user_id, tgt_service_id, tgt_rotation_id)
+			VALUES ($1, $2, $3)
 			ON CONFLICT DO NOTHING
 		`),
 		delete: p.P(`
 			DELETE FROM user_favorites
 			WHERE user_id = $1 and
-			tgt_service_id = $2
+			tgt_service_id  = $2 OR tgt_rotation_id = $3
```
mastercactapus: nit: get rid of the extra space before `=`

user/favorite/store.go
```diff
@@ -108,13 +114,25 @@ func (db *DB) Unset(ctx context.Context, userID string, tgt assignment.Target) e
...
 		validate.UUID("UserID", userID),
-		validate.OneOf("TargetType", tgt.TargetType(), assignment.TargetTypeService),
+		validate.OneOf("TargetType", tgt.TargetType(), assignment.TargetTypeService, assignment.TargetTypeRotation),
 	)
 	if err != nil {
 		return err
 	}
 
-	_, err = db.delete.ExecContext(ctx, userID, tgt.TargetID())
+	var rotationID, serviceID sql.NullString
```
mastercactapus: nit: Use the same ordering throughout the file `serviceID`, `rotationID`

user/favorite/store.go
```diff
@@ -108,13 +114,25 @@ func (db *DB) Unset(ctx context.Context, userID string, tgt assignment.Target) e
...
+
+	switch tgt.TargetType() {
+	case assignment.TargetTypeRotation:
```
mastercactapus: same as above

user/favorite/store.go
```diff
@@ -148,6 +166,8 @@ func (db *DB) FindAll(ctx context.Context, userID string, filter []assignment.Ta
 			switch f {
 			case assignment.TargetTypeService:
 				allowServices = true
+			case assignment.TargetTypeRotation:
+				allowServices = true
```
mastercactapus: If we ask for rotations we shouldn't return services.

A second bool should be added, `allowRotations`, instead and the `findAll` query updated.

smoketest/graphql2rotations_test.go
```diff
@@ -0,0 +1,153 @@
+package smoketest
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/target/goalert/smoketest/harness"
+	"testing"
+)
+
+func TestGraphQL2RotationsFavorite(t *testing.T) {
```
mastercactapus: Just call it `TestGraphQLFavoriteRotations` also update the filename to match the test name.

Putting favorite first will group all the favorite gql tests together.

The `2` bit is only temporary while we transition to the new library.

smoketest/graphql2rotations_test.go
```diff
@@ -0,0 +1,153 @@
...
+	t.Parallel()
+	sql := `
+	insert into users (id, name, email, role) 
+	values 
+		({{uuid "user"}}, 'bob', 'joe', 'admin');
+
+	insert into rotations (id, name, description, type, start_time, time_zone)
+	values
+		({{uuid "r1"}}, 'test', 'test', 'daily', now(), 'UTC'),
+		({{uuid "r2"}}, 'test2', 'test2', 'daily', now(), 'UTC');
```
mastercactapus: We can get rid of the extra rotation since there's no functional difference between the two.

It should be enough to test setting & unsetting favorite on a single rotation.

service/search.go
```diff
@@ -78,7 +78,7 @@ var searchTemplate = template.Must(template.New("search").Parse(`
 		{{if not .FavoritesFirst}}
 			lower(svc.name) > lower(:afterName)
 		{{else if .After.IsFavorite}}
-			((fav notnull AND lower(svc.name) > lower(:afterName)) OR fav isnull)
+			((fav NOTNULL AND lower(svc.name) > lower(:afterName)) OR fav isnull)
```
mastercactapus: Doesn't this need to be `IS DISTINCT FROM NULL` as we did above?

migrate/migrations/20190613111645-add-rotation-favorite.sql
```diff
@@ -0,0 +1,13 @@
+-- +migrate Up
+ALTER TABLE user_favorites
+    ADD COLUMN tgt_rotation_id UUID REFERENCES rotations (id),
+    ALTER COLUMN tgt_service_id DROP NOT NULL,
+    ADD CONSTRAINT user_favorite_rotations UNIQUE (user_id, tgt_rotation_id);
```
mastercactapus: Update name to postgres default

graphql2/graphqlapp/rotation.go
```diff
@@ -185,6 +199,14 @@ func (q *Query) Rotations(ctx context.Context, opts *graphql2.RotationSearchOpti
 		searchOpts.Limit = 15
 	}
 
+	if opts.FavoritesFirst != nil {
+		searchOpts.FavoritesFirst = *opts.FavoritesFirst
```
mastercactapus: These need to be before the `ParseCursor` call

graphql2/graphqlapp/rotation.go
```diff
@@ -185,6 +199,14 @@ func (q *Query) Rotations(ctx context.Context, opts *graphql2.RotationSearchOpti
...
+	}
+
+	if opts.FavoritesOnly != nil {
+		searchOpts.FavoritesOnly = *opts.FavoritesOnly
```
mastercactapus: Same as above

web/src/app/rotations/RotationCreateDialog.js
```diff
@@ -33,6 +33,7 @@ export default class RotationCreateDialog extends React.PureComponent {
         .startOf('hour')
         .toISO(),
       shiftLength: 1,
+      favorite: true,
```
mastercactapus: Add to wizard query too

web/src/app/util/QuerySetFavoriteButton.js
```diff
@@ -0,0 +1,88 @@
...
+          props.id,
+          queries[props.typeName],
+          props.typeName,
+        )
+      }}
+    />
+  )
+}
+
+function renderMutation(isFavorite, id, query, typeName) {
```
mastercactapus: Looks like `query` isn't used here, we should drop it.

web/src/cypress/integration/favorites.ts
```diff
@@ -1,5 +1,6 @@
 import { testScreen } from '../support'
 import { Chance } from 'chance'
+import { cyan } from '@material-ui/core/colors'
```
mastercactapus: Why is this imported?

node_modules/.yarn-integrity
dayander: Not sure, Interesting.

web/src/app/main/components/ToolbarTitle.js
```diff
@@ -45,10 +45,10 @@ const mapSingular = {
 
 const nameQuery = typeName => gql`
   query($id: ID!) {
-    data: ${typeName}(id: $id) {
-      id
-      name
-    }
+  data: ${typeName}(id: $id) {
```
Auchindoun: nit: revert spacing here (might need to run both linters as i saw some spacing in the go code not match as well)

web/src/app/util/QuerySetFavoriteButton.js
```diff
@@ -0,0 +1,92 @@
...
+export function QuerySetFavoriteButton(props) {
+  let typeName = ''
+  let id = ''
+  if (props.rotationID) {
+    typeName = 'rotation'
+    id = props.rotationID
+  } else if (props.serviceID) {
+    typeName = 'service'
+    id = props.serviceID
+  }
```
Auchindoun: if id/typeName isn't one of `serviceID` or `rotationID` it will throw a proptype error, but it may also cause the page to crash because of an apollo error in the query and variables. we should default to returning `null` if either of these isn't set

migrate/inline_data_gen.go
```diff
@@ -774,7 +775,8 @@ db8EThCvl6l0oXwy-rVnvhPcYvwhueRbSpXvfwEAAP__xk_2yLkaBAA=
 		{Data: dataRange(253205, 253408), Name: "migrations/20190313125552-slack-user-link.sql"},
 		{Data: dataRange(253408, 253724), Name: "migrations/20190404105850-nc-no-meta.sql"},
 		{Data: dataRange(253724, 254211), Name: "migrations/20190517144224-trigger-config-sync.sql"},
-		{Data: dataRange(254211, 267892), Name: "migrations/20190613114217-remove-switchover-triggers.sql"},
-		{Data: dataRange(267892, 268985), Name: "migrations/20190613120345-drop-switchover-resources.sql"},
+		{Data: dataRange(254211, 254668), Name: "migrations/20190613111645-add-rotation-favorite.sql"},
```
mastercactapus: Migration needs to be renamed to appear at the bottom of the list

web/src/cypress/support/rotation.ts
```diff
@@ -48,69 +49,60 @@ type Part = { pos: number }
...
-            }
+            
           }
       }`
 
   return cy.fixture('users').then(users => {
     if (!rot) rot = {}
-    const ids = c.pickset(users, rot.count).map((usr: any) => usr.id)
-    const parts = ids.map(id => ({ rotation_id: 'rot', user_id: id }))
+    const ids = users.map((user: any) => user.id)
```
mastercactapus: This should respect the `count` field of `RotationOptions`. Use `c.pickset` to select a random sample of the provided number of users.

web/src/app/main/components/ToolbarTitle.js
```diff
@@ -46,10 +46,10 @@ const mapSingular = {
 const nameQuery = typeName => gql`
   query($id: ID!) {
     data: ${typeName}(id: $id) {
-      id
-      name
+        id
```
mastercactapus: Spacing is still off in this. I thought prettier used to handle this automatically though. Anybody remember?

user/favorite/store.go
```diff
@@ -41,20 +41,22 @@ func NewDB(ctx context.Context, db *sql.DB) (*DB, error) {
...
+			INSERT INTO user_favorites (user_id, tgt_service_id, tgt_rotation_id)
+			VALUES ($1, $2, $3)
 			ON CONFLICT DO NOTHING
 		`),
 		delete: p.P(`
 			DELETE FROM user_favorites
 			WHERE user_id = $1 and
-			tgt_service_id = $2
+			tgt_service_id = $2 OR tgt_rotation_id = $3
+
```
mastercactapus: nit: cleanup blank line

web/src/app/main/components/ToolbarTitle.js
dayander: It does not work for whatever reason. Cook couldn't get it to work yesterday either.

PR #28, opened 2019-06-24
profile: show warning and option to reactivate when contact method is disabled

Auchindoun:

**Description:**
This PR adds a warning to a contact method if it is currently disabled. The warning icon should have a tooltip directing the user on how to reactivate said contact method.

**Additional Info:**
This PR also refactors the verification form/dialog to use React hooks, as well as updates the verification mutations to use our latest graphql library.


graphql/contactmethod.go
```diff
@@ -170,77 +170,3 @@ func (h *Handler) sendContactMethodTest() *g.Field {
 		},
 	}
 }
-
-func (h *Handler) sendContactMethodVerification() *g.Field {
```
Auchindoun: deleting these as the return types have changed, and isn't worth refactoring to bring it to a usable state.

Auchindoun:
Tests passed.

notification/store.go
```diff
@@ -266,41 +268,41 @@ func (db *DB) SendContactMethodVerification(ctx context.Context, id string, rese
...
-		return nil, validation.NewFieldError("Code", "unrecognized code")
+		return validation.NewFieldError("Code", "unrecognized code")
 	}
 	if err != nil {
-		return nil, err
+		return err
 	}
 
-	rows, err := db.enableContactMethods.QueryContext(ctx, userID, cmValue)
+	_, err = db.enableContactMethods.QueryContext(ctx, userID, cmValue)
```
mastercactapus: Let's update this to only enable the single `cmID` instead of all. That also means we can update the verify query to not return the contact method info.

Lastly, these queries should be executed in the transaction with `tx.StmtContext()`. Somehow we missed that before.

notification/store.go
```diff
@@ -266,41 +268,41 @@ func (db *DB) SendContactMethodVerification(ctx context.Context, id string, rese
...
 
-func (db *DB) VerifyContactMethod(ctx context.Context, cmID string, code int) ([]string, error) {
+func (db *DB) VerifyContactMethod(ctx context.Context, cmID string, _code string) error {
 	userID, err := db.cmUserID(ctx, cmID)
 	if err != nil {
-		return nil, err
+		return err
+	}
+
+	code, err := strconv.ParseInt(_code, 10, 64)
```
mastercactapus: We can do more granular error checking after we change these to int, but the Store's job is to just wrap the DB, so let's move the `Range` check to the GraphQL layer.

web/src/app/contact-methods/components/ContactMethodForm.js
```diff
@@ -191,23 +191,14 @@ class ContactMethodForm extends Component {
...
-        value={contactMethod.value}
-        type={contactMethod.type}
-        open={showVerifyForm}
-        userId={this.props.userId}
-        handleRequestClose={() => this.setState({ showVerifyForm: false })}
-        onSuccess={() => this.onVerificationSuccess()}
+  renderVerificationForm = () =>
+    this.state.showVerifyForm && (
+      <ContactMethodVerificationDialog
+        key='verify-dialog'
```
mastercactapus: Is key used?

graphql2/schema.graphql
```diff
@@ -748,6 +750,16 @@ input UpdateUserContactMethodInput {
   value: String
 }
 
+input SendContactMethodVerificationInput {
+  contactMethodID: ID!
+  resend: Boolean
```
mastercactapus: Drop this, new or resent code should be dictated by the backend.

web/src/app/users/ContactMethodVerificationForm.js
```diff
@@ -0,0 +1,107 @@
...
+  fieldGridItem: {
+    flexGrow: 1,
+  },
+  sendGridItem: {
+    display: 'flex',
+    alignItems: 'center',
+  },
+})
+
+export default function ContactMethodVerificationForm(props) {
```
mastercactapus: This should be `UserContactMethodVerificationForm`

web/src/app/users/ContactMethodVerificationDialog.js
```diff
@@ -0,0 +1,122 @@
...
+    return `+91-${n.slice(3, 5)}-${n.slice(5, 8)}-${n.slice(8)}`
+  }
+  if (n.startsWith('+44')) {
+    return `+44 ${n.slice(3, 7)} ${n.slice(7)}`
+  } else {
+    return <span>{n}</span>
+  }
+}
+
+export default function ContactMethodVerificationDialog(props) {
```
mastercactapus: Add `User` prefix to match the other components

web/src/app/users/UserContactMethodList.js
```diff
@@ -34,6 +37,8 @@ const testCM = gql`
   }
 `
 
+const DISABLED_CM_TOOLTIP = `Number disabled. See contact method's options to reactivate`
```
mastercactapus: We should use something more generic than `Number` as future contact methods (e.g. email and/or webhooks) won't always be phone numbers.

web/src/app/icons/components/Icons.js
```diff
@@ -26,7 +26,7 @@ export class Trash extends Component {
 @withStyles(styles)
 export class Warning extends Component {
   render() {
-    const { classes, details } = this.props
+    const { classes, tooltip } = this.props
```
mastercactapus: Would calling it `text`, `message` or `label` be more appropriate? Also let's add a string prop-type for it.

web/src/app/contact-methods/components/ContactMethodForm.js
Auchindoun: it's returned as an array in this (older) component, so it will throw an error without it

Auchindoun:
Comments addressed. Outstanding work is to create a new migration for the `user_contact_methods` table to:
- Remove `user_id` and `contact_method_value` columns
- Add a `contact_method_id` column
- Replace `send_to` with a bool `sent`

web/src/app/users/UserContactMethodVerificationDialog.js
```diff
@@ -0,0 +1,122 @@
...
+      type
+      value
+    }
+  }
+`
+
+/*
+ * Used for the subtitle of the dialog
+ */
+function formatNumber(n) {
```
m17ch: Is there a reason we can't use the existing function `formatPhoneNumber` in `web/src/app/users/util.js` for this?  There also appears to be formatting differences between this new function and existing.

web/src/app/users/UserContactMethodList.js
```diff
@@ -34,6 +37,9 @@ const testCM = gql`
   }
 `
 
+const DISABLED_CM_TOOLTIP =
+  'Contact method disabled. See its options to reactivate.'
```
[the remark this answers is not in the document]
Auchindoun: For sure -- I wasn't able to find a sentence that felt good, so I went with this for the time being. Since it's a tooltip we will want to keep it as concise as possible, since this is an overlay on the rest of the page.

If we can't think of good wording for the tooltip in just a few words perhaps we can discuss showing the warning in-line under or near the warning icon.

web/src/app/users/UserContactMethodVerificationDialog.js
Auchindoun: There's not! I didn't know we had that function

Auchindoun:
@arurao Looks like I forgot to change the type of the textfield to `type='number'` which would disallow typing anything but numbers. Thanks for catching, I will get this fixed

mastercactapus:
Couple quick things I think might be helpful on the verification dialog:
- Set the title to just "Verify Contact Method" (due to next point)
- Add text above the form saying `Send the verification code to +1 (612) 555-0123 (SMS)`
- Add text below the form, only for SMS, saying `If you do not receive a code, try sending UNSTOP to  <twilio fromNumber, from config, here> and then hitting Resend`

The above text is so that the user always has the context to what is on the screen, and what the resend button is going to do. Sort of like what we started doing with filters on list pages.

The SMS text at the bottom will help a lot of the questions we get (as it will silently fail otherwise), especially since the send-error stuff is out of scope. If someone unsubscribes, Twilio will drop new messages and if we have instructions in the UI, it would be a lot more likely users could fix the issue themselves.

@arurao @m17ch thoughts?

web/src/app/users/UserContactMethodList.js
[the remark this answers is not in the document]
Auchindoun: On second thought I think we should approach this differently after chatting with @dayander for a bit. A tooltip might not even be the way to go, as it basically forces the user to first find hover over the tooltip to get the information, click away, find the options menu, click it, then open the verification dialog from that menu. That's a total of **3 clicks, and a hover** (over the tooltip) to get to the dialog. Maybe we can get the team together and sketch a few designs out (like we did with Sylas) and go from there. @m17ch @arurao @mastercactapus @Katy-M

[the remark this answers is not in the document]
m17ch: Yeh - that sounds good @Auchindoun -- I did think clicking "Verify / Reactivate" from menu on CM then clicking on "Send" also felt a bit unnecessary.

I think having a wholistic design/flow discussion would be very worthwhile!

Auchindoun:
Test passed round 2.

migrate/migrations/20190701130233-verification-codes-update.sql
```diff
@@ -0,0 +1,54 @@
+-- +migrate Up
+UPDATE engine_processing_versions
+SET "version" = 2
+WHERE type_id = 'verify';
+
+-- add new columns
+ALTER TABLE user_verification_codes
+    ADD COLUMN contact_method_id UUID REFERENCES user_contact_methods (id) UNIQUE,
```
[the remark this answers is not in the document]
Auchindoun: If I recall correctly @mastercactapus and I didn't add `CASCADE` as it could potentially delete outgoing messages as well, if any existed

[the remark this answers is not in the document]
arurao: If i'm not wrong, `ON DELETE CASCADE` on contact_method_id column means that if the foreign table reference (user_contact_methods (id)) is deleted, it will also delete the column from user_verification_codes too. I am not sure how that would delete outgoing_messages. Is there a trigger somewhere that does this? 
Also, if the contact method itself is deleted, we do want the corresponding outgoing_messages to be deleted too.
What do you think?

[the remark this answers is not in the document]
Auchindoun: Whoops, I was thinking of something else 😅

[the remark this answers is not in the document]
mastercactapus: The not wanting to delete the row here was specifically during the initial migration so that we don't drop in-flight messages.

Deleting a contact method, however, would be an appropriate reason to cleanup verification codes. Consequently deleting the CM will cleanup everything, including `outgoing_messages` for that CM.

Good catch

web/src/package.json
```diff
@@ -89,6 +89,7 @@
     "moment": "^2.22.2",
     "react": "16.8.6",
     "react-apollo": "2.5.6",
+    "react-apollo-hooks": "^0.4.5",
```
mastercactapus: I think we were going to lock versions in `package.json`
```suggestion
    "react-apollo-hooks": "0.4.5",
```

web/src/app/dialogs/FormDialog.js
```diff
@@ -64,6 +64,9 @@ export default class FormDialog extends React.PureComponent {
     onClose: p.func,
     onSubmit: p.func,
 
+    // disables form submit button
+    submitDisabled: p.bool,
```
mastercactapus: Do we have a use-case where we need to disable submitting? Everywhere else we allow submit and then display errors.

If they shouldn't submit, then when clicking display an error message (e.g. "You must click Send first before you can submit a code") so the user isn't left wondering why they can't submit the form.

In the case of CM though, it's possible the user already has a code (from having the form open previously) that they want to enter without sending again.

web/src/app/users/UserContactMethodVerificationForm.js
```diff
@@ -0,0 +1,100 @@
...
+  sendGridItem: {
+    display: 'flex',
+    alignItems: 'center',
+  },
+})
+
+export default function UserContactMethodVerificationForm(props) {
+  const classes = useStyles()
+
+  function getTitle() {
```
arurao: Is this needed now? Since with the new design, clicking on either the warning icon or Reactivate button does the action of sending the code.
Is there a scenario where we could get into a state with the dialog open but code not sent yet?

m17ch:
Can we treat the `readOnly` bits that control rendering a bit differently?  I think rendering the warning icon for any disabled contact (so that you can view disabled status for others) would be a good thing to display for everyone.  This would allow someone on a team to proactively identify / troubleshoot why someone did not receive a notification.

I understand if we want to make the re-activation process guarded to only YOUR contact methods, but how do people feel if we always have a visual indicator for a disabled contact method?

We may have to solve how we're going to explain what indicator means (instead of just clicking on it to discover), but I feel like it would be better to have a visual indicator that anyone could see.

Let me know your thoughts.

notification/store.go
```diff
@@ -256,51 +243,42 @@ func (db *DB) SendContactMethodVerification(ctx context.Context, id string, rese
 		return validation.NewFieldError("ContactMethod", "test message rate-limit exceeded")
```
arurao: We should probably change this to be something like `rate limit exceeded` or something more specific to codes and not test messages

web/src/app/lists/FlatList.js
```diff
@@ -46,8 +54,10 @@ export default class FlatList extends React.PureComponent {
           title: p.node.isRequired,
           subText: p.node,
           action: p.element,
+          button: p.element, // a secondary button element to be placed to the left of the action
```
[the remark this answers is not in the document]
Auchindoun: The first action is already a "secondaryAction" by material (as clicking the list item itself is the primary action), so I wanted to avoid any discrepancy there.

web/src/app/users/UserContactMethodList.js
```diff
@@ -43,6 +46,7 @@ export default class UserContactMethodList extends React.PureComponent {
   state = {
     edit: null,
     delete: null,
+    isVerifyDialogOpen: false,
```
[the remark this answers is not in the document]
Auchindoun: I prefer more verbose flag names when using booleans because when looking at a file for the first time, I don't really know what `edit` or `delete` actually does without spending extra time looking through the file. I would be okay with altering the edit/delete names instead. I've always posed my booleans as `isSomethingTrue` or `hasSomething` so it's clear that what someone is looking at is a boolean.

web/src/app/users/UserContactMethodList.js
```diff
@@ -98,19 +106,90 @@ export default class UserContactMethodList extends React.PureComponent {
...
+          onClose={() => setShowVerifyDialog(false)}
+        />
+      )}
+    </React.Fragment>
+  )
+}
+
+ReactivateButton.propTypes = {
+  contactMethodID: p.string.isRequired,
+  ButtonComponent: p.object.isRequired,
```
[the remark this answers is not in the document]
Auchindoun: I was just trying to make props more clear, but can remove if needed.

web/src/app/users/UserContactMethodList.js
```diff
@@ -98,19 +106,90 @@ export default class UserContactMethodList extends React.PureComponent {
...
-            cmID={this.state.delete}
-            onClose={() => this.setState({ delete: null })}
-          />
-        )}
       </Grid>
     )
   }
 }
+
+function ReactivateButton(props) {
```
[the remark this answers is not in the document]
Auchindoun: We can do an `icon` bool, but it would result in the same amount of logic (if not more) on whether or not `IconButton` or `Button` should be rendered.

And only one dialog would ever be able to be opened at a time (would be unable to click the other button while the dialog is open), so I'm not sure I follow how this could be an issue

web/src/app/dialogs/FormDialog.js
Auchindoun: It was a suggestion by @arurao. I'm open to discussion -- I don't really have a preference.

web/src/app/lists/FlatList.js
```diff
@@ -46,8 +54,10 @@ export default class FlatList extends React.PureComponent {
...
           url: p.string,
-          icon: p.element,
+          icon: p.element, // renders a list item icon
```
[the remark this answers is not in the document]
Auchindoun: I can look into if any warnings are actually thrown, but per the material ui docs `ListItemAvatar` is for avatars, and `ListItemIcon` is for icons.

arurao:
> Can we treat the `readOnly` bits that control rendering a bit differently? I think rendering the warning icon for any disabled contact (so that you can view disabled status for others) would be a good thing to display for everyone. This would allow someone on a team to proactively identify / troubleshoot why someone did not receive a notification.
> 
> I understand if we want to make the re-activation process guarded to only YOUR contact methods, but how do people feel if we always have a visual indicator for a disabled contact method?
> 
> We may have to solve how we're going to explain what indicator means (instead of just clicking on it to discover), but I feel like it would be better to have a visual indicator that anyone could see.
> 
> Let me know your thoughts.

I think this could definitely be a nice-to-have. Visually being able to spot somebody else's disabled contact methods would be useful in troubleshooting.

Auchindoun:
Sure, we can leave the warning icon in (and just not a button) if it's not the user, and show some text like we discussed before such as `Mobile - (612) 555-5555 - Disabled` or something

web/src/app/dialogs/FormDialog.js
arurao: That's correct. I did make this suggestion, but since we changed our design a little after that - sending the code on 1-clicking of the icon or Reactivate button, it seems that we could actually no longer need it. Meaning we could just remove the disabling Submit button entirely now. Since clicking on one of those two will send the code and then open the dialog, in which case when the dialog is open, there would already have had a code sent. So Submit can always be enabled.

notification/store.go
Auchindoun: for sure

web/src/app/users/UserContactMethodVerificationForm.js
Auchindoun: I did want to keep Send/Resend in just in the case that if the send fails on the initial dialog opening, it would show the error message and an option to try sending it manually. From there it would switch to resend (once the message is actually sent at least once). If we feel this isn't necessary I can take it out and always have the button read as "Resend"

web/src/app/lists/FlatList.js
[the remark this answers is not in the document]
mastercactapus: `FlatList` has it's own API/props that should be consistent for the component as a whole, regardless of current implementation details under the covers. We could rename `action` to `secondaryAction` and then add `tertiaryAction` if `action` is already misleading though.

But it also might make more sense to handle this outside of `FlatList` entirely and instead just treat the extra stuff as content. Does material and/or material UI define how a third action should render in lists?

web/src/app/lists/FlatList.js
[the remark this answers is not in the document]
mastercactapus: Are we using icons for lists? I thought they were all avatars in the first place (everything under util is wrapped in `<Avatar>`). 

I think `icon` still makes sense for the prop name though, AFAIK avatar is just a subset/type of icon.

If we do need to support both maybe we can instead just check on the type of element and render `ListItemAvatar` wrapper for `Avatar`-type elements and `ListItemIcon` for anything else. But we can probably get away with just swapping `ListItemIcon` for `ListItemAvatar` until we have a new need.

web/src/app/users/UserContactMethodList.js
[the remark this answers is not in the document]
mastercactapus: I'd be fine with updating names to indicate what they are. I suggest `editContactMethodID`, `deleteContactMethodID`, and `verifyContactMethodID`.

web/src/app/users/UserContactMethodList.js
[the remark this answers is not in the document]
mastercactapus: I think the API of a component is more important as it's implementation. I'd rather refactor or replace a single component than have to refactor every call site with a bunch of props.

The more "flexible" the component, the more work that lives everywhere else that is using it to maintain. It's easier to replace or refactor a component that takes a string and bool, than one that takes 2 components and a string. It's also less context for the caller to only need to provide "should this be an icon?" and a contact method ID, than needing to read ReactivateButton to see what the two passed-in components need to be for it to work properly.

web/src/app/users/UserContactMethodVerificationForm.js
mastercactapus: Wouldn't that just always flicker then? (if when opening the dialog we always send it first), also the mutation returns once the send is queued but not when it's actually sent -- we don't yet have that info (or send errors) on the UI.

notification/store.go
mastercactapus: We could probably have the UI display a nicer message if we get `rate limit exceeded` too. We haven't really solved structured errors yet though.

It would be nice to let the user know they should try again after a few seconds (or a minute, whatever we have it set to).

web/src/app/lists/FlatList.js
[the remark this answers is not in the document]
Auchindoun: They don't :( It's not "material" standard as far as material.io is concerned

web/src/app/lists/FlatList.js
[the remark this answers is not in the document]
Auchindoun: We use avatars everywhere, but in this case I felt an icon was more appropriate for the warning symbol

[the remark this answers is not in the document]
Auchindoun: I'm not sure how we would determine if a component is a MUI avatar vs icon, though

web/src/app/users/UserContactMethodVerificationForm.js
Auchindoun: So do we want to just default to "Resend" then, and do away with the `sendAttempted` props?

web/src/app/users/UserContactMethodList.js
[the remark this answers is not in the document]
Auchindoun: I could understand where you're coming from if we were to export this component, but it's a private functional component just to save lines of code in this file, not to make something very generic for several use-cases. if we were to pass in an `icon` prop we would then have to determine what higher level component to use (`Button` vs `IconButton`) as well as what children to render (a string vs an icon), which we would then have to add additional functionality to allow an icon to be passed as a prop (which is what `buttonChild` is currently doing), instead of having it be a bool like in the first suggestion.

If this is something you feel strongly towards, I can make a generic "button" component under Util where we can specify an icon or a string to switch between button types, but personally I don't find it necessary

web/src/app/dialogs/FormDialog.js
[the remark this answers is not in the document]
Auchindoun: The "Too many messages" error message is actually from the UI that we show if it errors out when you manually hit send since the backend error messages look bad. I think it would be better to remove that though, and make the backend error message better

web/src/package.json
```diff
@@ -89,6 +89,7 @@
     "moment": "^2.22.2",
     "react": "16.8.6",
     "react-apollo": "2.5.6",
+    "react-apollo-hooks": "0.4.5",
```
mastercactapus: Should we reconsider using the apollo beta?

This package recommends using that for new work, as it is planned to be deprecated. https://github.com/trojanowski/react-apollo-hooks

web/src/app/lists/PaginatedList.js
```diff
@@ -147,7 +148,8 @@ export class PaginatedList extends React.PureComponent {
         title: p.string.isRequired,
         subText: p.string,
         isFavorite: p.bool,
-        icon: p.element,
+        icon: p.element, // renders a list item icon (or avatar)
+        avatar: p.element, // renders a list item avatar with a backdrop
```
mastercactapus: No longer used?

web/src/app/lists/FlatList.js
```diff
@@ -90,21 +97,31 @@ export default class FlatList extends React.PureComponent {
...
-        {item.action && (
-          <ListItemSecondaryAction>{item.action}</ListItemSecondaryAction>
+        {(item.secondaryAction || item.tertiaryAction) && (
+          <ListItemSecondaryAction>
+            <Grid
+              container
+              spacing={2}
+              className={this.props.classes.actionGrid}
+            >
+              {item.tertiaryAction && <Grid item>{item.tertiaryAction}</Grid>}
```
mastercactapus: So this is my bad, but I didn't realize these were both being rendered as the `ListItemSecondaryAction` in the same place when I suggested using `secondaryAction` and `tertiaryAction`.

If these are always rendered in the same place, I don't think we actually need two and can just use `action` as was previous.

We can use the original `<ListItemSecondaryAction>{item.action}</ListItemSecondaryAction>` code here and then in the one place we do something different, just pass in <Grid> and it's children as the `action` prop and we'll get the same result without introducing additional props.

web/src/app/dialogs/components/DialogTitleWrapper.js
```diff
@@ -58,21 +61,32 @@ export default class DialogTitleWrapper extends Component {
...
+              {toolbarItems}
+              {menu}
+            </Toolbar>
+          </AppBar>
+          <DialogContent style={{ overflowY: 'unset' }}>
+            <Typography variant='subtitle1'>{subTitle}</Typography>
+          </DialogContent>
+        </React.Fragment>
       )
     } else {
```
mastercactapus: No need for else if we're returning above

web/src/app/users/UserContactMethodList.js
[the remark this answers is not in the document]
mastercactapus: That's understandable. And no, it's not a huge deal. My thinking was that we should render the dialog the same as edit and delete (and not render multiple dialogs for verify separately) to keep things consistent rather than diverging to save a few lines of code.

It might not be an issue today (referring to multiple dialogs opening), but it's more to debug tomorrow. Not having an example of how a user would get into said state, doesn't mean it won't happen either, or that it won't happen in CI tests which operate faster than users. 

So with that idea, having the dialog raised out of this component, `ReactivateButton` serves little use since it just renders whatever you pass it, and that's where the idea of moving the `icon` logic to it. My primary concern was only rendering this dialog in a different way than all the others.

It's more important to be consistent and clear than to save lines of code. Code is read far more often than it is written.

web/src/package.json
Auchindoun: 👍

web/src/app/lists/PaginatedList.js
Auchindoun: Yep, should just be `icon` now. I'll remove

web/src/app/lists/FlatList.js
Auchindoun: Oh that's a good idea. So just leave action as is, but have it be a grid that we pass from the parent.  🎉 for collaboration!

web/src/app/users/UserContactMethodList.js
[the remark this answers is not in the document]
Auchindoun: That sounds good. I'll adjust such that the dialog logic is rendered in the parent component

web/src/app/users/UserContactMethodDeleteDialog.js
```diff
@@ -13,7 +13,8 @@ const mutation = gql`
 `
 export default class UserContactMethodDeleteDialog extends React.PureComponent {
   static propTypes = {
-    cmID: p.string.isRequired,
+    contactMethodID: p.string.isRequired,
+    onClose: p.func.isRequired, // passed to FormDialog
```
arurao: If a user has two contact methods, and on trying to delete the first contact method (doesn't matter disabled or not), the dialog for deleting contact method does not close on successful deletion. 
This is not observed when there is only one contact method.

notification/store.go
```diff
@@ -18,11 +18,12 @@ import (
 )
 
 const minTimeBetweenTests = time.Minute
+const minTimeBetweenTestsFmt = "1 minute"
```
arurao: Can we remove this and just reuse the above const (minTimeBetweenTests)

notification/store.go
```diff
@@ -253,54 +241,45 @@ func (db *DB) SendContactMethodVerification(ctx context.Context, id string, rese
 		return err
 	}
 	if rows != 1 {
-		return validation.NewFieldError("ContactMethod", "test message rate-limit exceeded")
+		return validation.NewFieldError("ContactMethod", fmt.Sprintf("Too many messages! Please try again in %s.", minTimeBetweenTestsFmt))
```
arurao: Just reuse minTimeBetweenTests.Minutes() here.

web/src/app/users/UserContactMethodList.js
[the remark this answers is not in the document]
Auchindoun: I'll just remove it.

web/src/app/users/UserContactMethodVerificationForm.js
```diff
@@ -0,0 +1,95 @@
...
+      <Grid container spacing={2}>
+        <Grid item className={classes.sendGridItem}>
+          <LoadingButton
+            color='primary'
+            loading={status.loading}
+            disabled={props.disabled}
+            buttonText={'Resend Code'}
+            noSubmit
+            onClick={() =>
+              sendCode().catch(err => props.setSendError(err.message))
```
mastercactapus: Should we just define this in a const and use it here and above instead of duplicating?

```js
const doSendCode = () => sendCode.catch( ... )

useEffect(doSendCode, [])
// ...
onClick={doSendCode}
```

web/src/app/users/util.test.js
```diff
@@ -6,12 +6,12 @@ import {
 } from './util'
 
 test('formatPhoneNumber', () => {
-  expect(formatPhoneNumber('+17635550100')).toBe('+1 763-555-0100')
-  expect(formatPhoneNumber('+12085550105')).toBe('+1 208-555-0105')
-  expect(formatPhoneNumber('+15165550184')).toBe('+1 516-555-0184')
-  expect(formatPhoneNumber('+911400000000')).toBe('+91 14 0000 0000')
-  expect(formatPhoneNumber('+911401234567')).toBe('+91 14 0123 4567')
-  expect(formatPhoneNumber('+911409876543')).toBe('+91 14 0987 6543')
+  expect(formatPhoneNumber('+17635550100')).toBe('+1 (763) 555-0100')
```
mastercactapus: International format should be `+1 763-555-0123`

image

web/src/app/users/UserContactMethodVerificationForm.js
```diff
@@ -0,0 +1,95 @@
...
+    sendCode().catch(err => props.setSendError(err.message))
+  }, [])
+
+  return (
+    <FormContainer optionalLabels {...props}>
+      <Grid container spacing={2}>
+        <Grid item className={classes.sendGridItem}>
+          <LoadingButton
+            color='primary'
+            loading={status.loading}
```
mastercactapus: Where is `status` defined?

web/src/app/users/UserContactMethodList.js
```diff
@@ -34,60 +37,133 @@ const testCM = gql`
...
+  const classes = useStyles()
+
+  const [showVerifyDialogByID, setShowVerifyDialogByID] = useState(null)
+  const [showEditDialogByID, setShowEditDialogByID] = useState(null)
+  const [showDeleteDialogByID, setShowDeleteDialogByID] = useState(null)
+  const [testID, setTestID] = useState(null)
+
+  const [sendTest] = useMutation(testCM, {
+    variables: {
+      id: testID,
```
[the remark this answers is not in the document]
Auchindoun: If sendTest accepts that as a parameter we can, but when defining the variables in the mutation's initialization, a hook can't be nested inside of a second function. I can check the apollo docs.

web/src/app/users/util.test.js
Auchindoun: This is a US format though, isn't it? We only support US, India, and the UK. We will be deleting and replacing these tests once we support international phone numbers from previous discussions we've had.

web/src/app/users/UserContactMethodVerificationForm.js
Auchindoun: Typo. Need to grab status from the mutation hook initialization

migrate/inline_data_gen.go
```diff
@@ -779,6 +785,7 @@ dpHSHLldtNrA7SIlOvtfAAAA__80AwSB6R0EAA==
 		{Data: dataRange(254211, 267892), Name: "migrations/20190613114217-remove-switchover-triggers.sql"},
 		{Data: dataRange(267892, 268985), Name: "migrations/20190613120345-drop-switchover-resources.sql"},
 		{Data: dataRange(268985, 269442), Name: "migrations/20190701111645-add-rotation-favorite.sql"},
-		{Data: dataRange(269442, 269801), Name: "migrations/20190702161722-add-schedule-favorites.sql"},
+		{Data: dataRange(269442, 271133), Name: "migrations/20190701130233-verification-codes-update.sql"},
```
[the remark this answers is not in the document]
Auchindoun: Is there a command to regenerate the specific date stamp?

web/src/app/users/UserContactMethodList.js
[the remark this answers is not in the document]
Auchindoun: If a second parameter is specified, `useEffect` will only run if the variable given in that parameter has changed.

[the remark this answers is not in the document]
Auchindoun: Looks like that is unsupported:

```
UserContactMethodList.js:99 - Network error: HTTP Response 422: Unprocessable Entity
{"errors":[{"message":"cannot be null","path":["variable","id"]}],"data":null}
```

migrate/inline_data_gen.go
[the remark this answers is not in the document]
mastercactapus: I don't think so, yet anyway, you can use this though: `date +%Y%m%d%H%M%S`

web/src/app/users/UserContactMethodList.js
```diff
@@ -34,60 +37,133 @@ const testCM = gql`
...
+    },
+  })
+
+  // send test to CM after testID is set
+  useEffect(() => {
+    if (testID) {
+      sendTest().catch(err => console.error(err.message))
```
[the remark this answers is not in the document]
Auchindoun: My IDE was telling me to resolve the promise on an error. We can ignore then

web/src/app/users/UserContactMethodList.js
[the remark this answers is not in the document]
mastercactapus: I know it technically works, but it's not super clear what is happening.

If we can't pass in args (their documentation looks like it _should_ work though), then maybe we can move the `useMutation` to a separate component that's fed a CM ID directly so we can still fire the mutation on-click instead of post-render via `useEffect`.

image

That's an odd type signature... I'll see if I can figure out locally why it wouldn't be using the options param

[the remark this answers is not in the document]
Auchindoun: I think it will be more clear as more people start using hooks (since doing something in a hook based off a variable rather than a prop I can see as a common use case).

I can either add a comment to the `useEffect` we define, or as you're saying create a sort of higher order component that wraps it, if preferred.

web/src/app/users/util.test.js
mastercactapus: It's not the format outside of the US though. It would be incorrectly formatted for users in the UK or India. That's why we format all numbers as international, as it is the same in and from all locations, everyone sees the same standard format.

For example, the national format of an India number is `123 123 1234` which isn't used in the US.

image

And in any case, if we change it here (thus in the cypress tests) we will just be changing it back in a few days when the CC PR lands as formatting will happen by `libphonenumber`.

Auchindoun: Sure.

web/src/app/users/UserContactMethodList.js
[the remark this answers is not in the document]
mastercactapus: This seems to work fine:
image

notification/store.go
```diff
@@ -253,54 +215,35 @@ func (db *DB) SendContactMethodVerification(ctx context.Context, id string, rese
...
-		return nil, err
-	}
-	defer tx.Rollback()
-
-	var cmValue string
-	err = db.verifyVerificationCode.QueryRowContext(ctx, cmID, code).Scan(&cmValue)
-	if err == sql.ErrNoRows {
-		return nil, validation.NewFieldError("Code", "unrecognized code")
+func (db *DB) VerifyContactMethod(ctx context.Context, cmID string, code int) error {
+	res, err := db.verifyAndEnableContactMethod.ExecContext(ctx, cmID, code)
```
mastercactapus: This still needs the call to `db.cmUserID()` -- you can throw away the user ID into `_` but it still performs validation and permissions checks we need.

notification/store.go
```diff
@@ -33,12 +32,11 @@ type DB struct {
...
 	getCMUserID            *sql.Stmt
 	setVerificationCode    *sql.Stmt
-	verifyVerificationCode *sql.Stmt
-	enableContactMethods   *sql.Stmt
+	verifyAndEnableContactMethod   *sql.Stmt
 	insertTestNotification *sql.Stmt
 	updateLastSendTime     *sql.Stmt
-	codeExpiration         *sql.Stmt
 	getCode                *sql.Stmt
+	isDisabled				*sql.Stmt
```
mastercactapus: Formatting

PR #31, opened 2019-06-25
services: favorites test code improvements

mastercactapus:

**Description:**
This PR makes new services favorited by default and refactors the Cypress tests for service favorites to be more generalized for adding new favorite types.

The `service.ts` support file was also updated to use the new GraphQL endpoint.

### DIFF
--- a/web/src/cypress/support/service.ts
+++ b/web/src/cypress/support/service.ts
@@ -38,7 +38,7 @@ declare global {
     description?: string
     epID?: string
     ep?: EPOptions
-    isFavorite?: boolean
+    favorite?: boolean
   }
 
   interface Label {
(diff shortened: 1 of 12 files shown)


web/src/cypress/support/service.ts
```diff
@@ -142,10 +143,12 @@ function createLabel(label?: LabelOptions): Cypress.Chainable<Label> {
...
 
   return cy
-    .graphql(query, {
+    .graphql2(query, {
       input: {
-        target_id: label.svcID,
-        target_type: 'service',
+        target: {
+          type: 'service',
+          id: label.svcID,
```
arurao: Can we do `id: svcID` instead?

arurao:
> Did we forget to add favorite when creating new service in wizard (`make test` failed):
> 
I think it was added in the later commits. it was there when I tested functionally.

arurao:
Looks like making service as favorite was added to the wizard, but was missed out in the util test, because of which `make test` was failing.
Adding `favorite: true,` to util.test.js on line 16 should fix that.

PR #39, opened 2019-07-02
dev: easily run integration tests by cli

mastercactapus:

**Description:**

This PR intends to make it easier to run integration tests locally, by adding `make check-all` that will perform all checks and tests without interaction.

**Out of Scope:**
- Parallel tests
- CI updates

**Additional Info:**
- Added `--log-errors-only` flag to the backend
- Added `make cy-wide-prod-run` and `make cy-mobile-prod-run` commands for running tests
- Added `make check-all` that runs all integration & smoketests, as well as `make check`
- Updated the `runjson` tool to support env-params
- Use chrome by default to run tests


m17ch:
I assume post-merge, we'll want to update our PR template item to reference `make check-all`?

mastercactapus:
I don't think so, at least not yet. I'd rather have CI do that heavy-lifting for now because it's so slow. `make check` should be fast (ideally anyway), run often and catch common things, but not be too painful for someone to run before opening their PR.

`make check-all` will take quite a long time to run, so until/unless we can speed it up `make check` is fine at that stage IMO. `check-all` is just like the final catch-all to make sure we don't introduce regressions.

@arurao @Auchindoun What do you guys think?

arurao:
`make check-all` does take a long time to run, but it also covers everything. It would be super easy for PR reviewers to review code knowing that it has passed all tests, backend and front-end as well as all formatting specifics. At this point, we can not guarantee that a developer submitting a PR ran all tests- like make check, make test, make smoketest and then the cypress ones. 
I think we could include `make check-all` in our PR template. In the future, we will make changes to it so that it speeds up.

m17ch:
functionally validated / looks good

PR #40, opened 2019-07-02
ui: accessibility upgrades

dayander:

**Description:**
Added accessibility upgrades to the UI.
* Added alt-text to images
* Increased color contrast on the cancel button
* Added jsx-a11y linter - recommend to provide suggestion but not errors.
* Added language to the page, removed user-no-scale
* rearranged navigation to be a page region
* added titles to pagination buttons
* added skip-link for the repetitive page content 
* added aria-expanded to filter buttons
* Document title updates on page change

Creating a more accessible UI

**Screenshots:**
Screen Shot 2019-07-03 at 9 52 31 AM

Skip to content button is only shown when focused
image


mastercactapus:
Can we add some screenshots to this for changed UI elements?

web/src/app/alerts/components/AlertDetails.js
```diff
@@ -115,7 +115,9 @@ export default class AlertDetails extends Component {
       <Card className={this.getCardClassName()}>
         <div style={{ display: 'flex' }}>
           <CardContent style={{ flex: 1, paddingBottom: 0 }}>
-            <Typography variant='h5'>Event Log</Typography>
+            <Typography variant='h5' variantMapping={{ h5: 'h3' }}>
```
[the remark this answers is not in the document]
dayander: I am not sure that changing the global variant mapping will solve the heading structure. There are times where h6's shouldn't be a heading and times where h5's should be an h2 or h3 and same with h6. 

In the future with a11y testing implemented, it would catch not consecutive heading structure.

web/src/app/details/DetailsPage.js
```diff
@@ -125,12 +132,24 @@ export default class DetailsPage extends React.PureComponent {
...
+                variant='subtitle1'
+                variantMapping={{ subtitle1: 'div' }}
+              >
                 <Markdown value={details} />
               </Typography>
               {titleFooter && (
-                <Typography variant='subtitle1'>{titleFooter}</Typography>
+                <Typography
+                  variant='subtitle1'
+                  data-cy='time-zone'
```
Auchindoun: this should be deleted or renamed if it is being used

web/src/app/main/components/SideBarDrawerList.js
```diff
@@ -154,56 +155,58 @@ export default class SideBarDrawerList extends React.PureComponent {
     const { classes } = this.props
 
     return (
-      <List className={classes.list} data-cy='nav-list'>
-        <div className={classes.logoDiv}>
+      <div>
```
Auchindoun: if this div is in here only to encapsulate the rest of the elements, you can use `<React.Fragment>` instead

web/src/app/main/components/ToolbarTitle.js
```diff
@@ -116,12 +120,13 @@ const mapStateToProps = state => {
 @connect(mapStateToProps)
 export default class ToolbarTitle extends React.Component {
   renderTitle = title => {
+    document.title = `Go Alert || ${title}`
```
Auchindoun: as this is a string, is showing the `||` intentional, or is this supposed to be an `or` expression?

web/src/app/rotations/RotationUserList.js
```diff
@@ -50,7 +50,7 @@ export default class RotationUserList extends React.PureComponent {
     return (
       <React.Fragment>
         <Card>
-          <CardHeader title='Users' />
+          <CardHeader style={{ margin: 0 }} component='h3' title='Users ' />
```
Auchindoun: when adding new styles we prefer to use `className` with mui styles to be consistent (styles can always be found at the top of the file)

web/src/app/services/components/OnCallForService.js
```diff
@@ -97,7 +97,11 @@ export default class OnCallForService extends Component {
 
     return (
       <Card>
-        <CardHeader title='On Call Users' />
+        <CardHeader
+          style={{ margin: 0 }}
```
Auchindoun: same here for style

web/src/app/users/UserContactMethodList.js
```diff
@@ -75,7 +75,11 @@ export default class UserContactMethodList extends React.PureComponent {
     return (
       <Grid item xs={12}>
         <Card>
-          <CardHeader title='Contact Methods' />
+          <CardHeader
+            style={{ margin: 0 }}
```
Auchindoun: same here for style

web/src/app/users/UserNotificationRuleList.js
```diff
@@ -50,7 +50,11 @@ export default class UserNotificationRuleList extends React.PureComponent {
     return (
       <Grid item xs={12}>
         <Card>
-          <CardHeader title='Notification Rules' />
+          <CardHeader
+            style={{ margin: 0 }}
```
Auchindoun: same here for style

web/src/app/util/Options.js
```diff
@@ -142,6 +147,7 @@ export default class Options extends Component {
           data-cy='other-actions'
           color='inherit'
           onClick={onClick}
+          aria-expanded={this.state.ariaExpanded}
```
Auchindoun: you could set this is just whether or not `show` is true

web/src/app/util/FilterContainer.js
```diff
@@ -62,7 +70,14 @@ export default class FilterContainer extends React.PureComponent {
...
           color='inherit'
-          onClick={e => this.setState({ anchorEl: e.target })}
+          onClick={e =>
+            this.setState({
+              anchorEl: e.target,
+              ariaExpanded: !this.state.ariaExpanded,
+            })
+          }
+          title='filter'
+          aria-expanded={this.state.ariaExpanded}
```
Auchindoun: you can set this to whether or not `anchorEl` is set, as we do on line 117 here

web/src/app/util/OtherActions.js
```diff
@@ -39,20 +39,27 @@ export default class OtherActions extends React.PureComponent {
...
+      this.setState({ anchorEl: null, ariaExpanded: !this.state.ariaExpanded }),
+    )
     return (
       <React.Fragment>
         <span ref={this.ref}>
           {React.cloneElement(this.props.icon, {
             'aria-label': 'Other Actions',
             'data-cy': 'other-actions',
             color: 'inherit',
+            'aria-expanded': this.state.ariaExpanded,
```
Auchindoun: same here, with using `anchorEl` as a boolean

web/src/app/util/SkipToContentLink.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+import Link from '@material-ui/core/Link'
+import { makeStyles } from '@material-ui/core'
+
+const useStyles = makeStyles({
+  skipLink: {
+    position: 'absolute !important',
```
Auchindoun: we should avoid using `!important` here and on line 16

web/src/app/main/components/ToolbarTitle.js
dayander: This is intentional. I've used it before but can change it to anything or remove it altogether.

web/src/app/util/FilterContainer.js
dayander: For this, we need to keep it because anchorEl uses null and not bool values.

web/src/app/main/components/Login.js
```diff
@@ -160,7 +160,7 @@ export default class Login extends Component {
     // create login button
     let loginButton = null
     const loginIcon = logoUrl ? (
-      <img src={logoUrl} className={classes.loginIcon} />
+      <img alt='Go Alert' src={logoUrl} className={classes.loginIcon} />
```
mastercactapus: GoAlert

web/src/app/rotations/RotationUserList.js
```diff
@@ -36,6 +38,7 @@ const mutation = gql`
   }
 `
 
+@withStyles(styles)
```
mastercactapus: Instead of adding everything, just pull the pieces needed. This will result in dumping all of material styles on the page instead of only `cardHeader`

web/src/app/services/components/OnCallForService.js
```diff
@@ -10,7 +10,10 @@ import Typography from '@material-ui/core/Typography'
 import { UserAvatar } from '../../util/avatar'
 import { Link } from 'react-router-dom'
 import Spinner from '../../loading/components/Spinner'
+import { withStyles } from '@material-ui/core'
+import { styles } from '../../styles/materialStyles'
 
+@withStyles(styles)
```
mastercactapus: Same as above, only add in the styles being used instead of everything.

web/src/app/users/UserContactMethodList.js
```diff
@@ -34,6 +36,7 @@ const testCM = gql`
   }
 `
 
+@withStyles(styles)
```
mastercactapus: Same

web/src/app/users/UserNotificationRuleList.js
```diff
@@ -26,6 +34,7 @@ const query = gql`
   }
 `
 
+@withStyles(styles)
```
mastercactapus: Same

web/src/app/util/FilterContainer.js
mastercactapus: Instead of adding a separate/duplicate flag, use `Boolean(this.state.anchorEl)`

web/src/app/util/Options.js
```diff
@@ -45,24 +45,28 @@ export default class Options extends Component {
...
     showErrorDialog: false,
     showOptions: false,
+    ariaExpanded: false,
   }
 
   handleOpenMenu = event => {
     this.setState({
       anchorEl: event.currentTarget,
       show: true,
+      ariaExpanded: !this.state.ariaExpanded,
```
mastercactapus: Same as above, this is redundant, just check `anchorEl` wrap in `Boolean()` or prefix with `!!` if a bool is needed.

web/src/app/util/OtherActions.js
```diff
@@ -39,20 +39,27 @@ export default class OtherActions extends React.PureComponent {
 
   state = {
     anchorEl: null,
+    ariaExpanded: false,
```
mastercactapus: Same

web/src/app/util/avatar/BaseAvatar.js
```diff
@@ -57,11 +57,11 @@ export default class BaseAvatar extends Component {
     const { userID, fallback, ...props } = this.props
 
     if (this.state.valid) {
-      return <Avatar src={this._image.src} {...props} />
+      return <Avatar alt='' src={this._image.src} {...props} />
```
mastercactapus: Why the empty alt string, isn't that the default?

web/src/app/wizard/WizardForm.js
```diff
@@ -57,7 +57,9 @@ export default class WizardForm extends React.PureComponent {
             <StepIcon icon='1' />
           </Grid>
           <Grid item xs={10}>
-            <Typography variant='h6'>Team Details</Typography>
+            <Typography variant='h6' variantMapping={{ h6: 'h2' }}>
```
mastercactapus: Can we have a quick group discussion about the variantMapping?

I feel like there needs to be a better way than having to specify this everywhere. Also, it's not obvious what it should be set too for future work that others will do.

web/src/cypress/integration/schedules.ts
```diff
@@ -22,8 +22,8 @@ function testSchedules(screen: ScreenFormat) {
         .click()
 
       // verify on details by content headers
-      cy.get('h5').contains(name)
-      cy.get('h6').contains(description)
+      cy.get('h2').contains(name)
```
mastercactapus: We should probably use a `data-cy` attribute instead of relying on tag names.

web/src/cypress/integration/schedules.ts
```diff
@@ -22,8 +22,8 @@ function testSchedules(screen: ScreenFormat) {
...
+      cy.get('div[name=details]').contains(description)
```
mastercactapus: I didn't think `name` was a valid attribute for divs?

web/src/app/util/avatar/BaseAvatar.js
dayander: It means that the image is decorative so the screen reader should ignore it.

web/src/app/util/FilterContainer.js
```diff
@@ -31,6 +31,7 @@ const style = theme => {
 export default class FilterContainer extends React.PureComponent {
   state = {
     anchorEl: null,
+    ariaExpanded: false,
```
mastercactapus: Switch to using `Boolean(anchorEl)` instead of this

web/src/app/util/Options.js
```diff
@@ -45,24 +45,28 @@ export default class Options extends Component {
     errorMessage: '',
     showErrorDialog: false,
     showOptions: false,
+    ariaExpanded: false,
```
mastercactapus: Same

web/src/app/util/OtherActions.js
```diff
@@ -39,20 +39,30 @@ export default class OtherActions extends React.PureComponent {
...
```
mastercactapus: Drop this and use `anchorEl`

web/src/cypress/integration/schedules.ts
```diff
@@ -89,9 +89,9 @@ function testSchedules(screen: ScreenFormat) {
 
       // verify changes occurred
       cy.reload()
-      cy.get('h5').contains(newName)
-      cy.get('h6').contains(newDesc)
-      cy.get('h6').contains(newTz)
+      cy.get('h2').contains(newName)
+      cy.get('div[name=details]').contains(newDesc)
```
mastercactapus: I think this was meant to be `div[data-cy=details]`

web/src/app/util/FilterContainer.js
```diff
@@ -49,7 +49,14 @@ export default class FilterContainer extends React.PureComponent {
           {this.props.onReset && (
             <Button onClick={this.props.onReset}>Reset</Button>
           )}
-          <Button onClick={() => this.setState({ anchorEl: null })}>
+          <Button
+            onClick={() =>
+              this.setState({
+                anchorEl: null,
+                ariaExpanded: Boolean(!this.state.anchorEl),
```
mastercactapus: This line can be dropped (no need for `ariaExpanded` in state)

web/src/cypress/integration/schedules.ts
```diff
@@ -89,9 +89,9 @@ function testSchedules(screen: ScreenFormat) {
 
       // verify changes occurred
       cy.reload()
-      cy.get('h5').contains(newName)
-      cy.get('h6').contains(newDesc)
-      cy.get('h6').contains(newTz)
+      cy.get('h2').contains(newName)
```
mastercactapus: Should this use a `[data-cy=]` selector instead?

web/src/app/details/DetailsPage.js
```diff
@@ -30,6 +30,13 @@ const styles = theme => ({
     [theme.breakpoints.up('md')]: { float: 'left' },
     margin: 20,
   },
+  mainHeading: {
+    fontSize: '1.5rem',
+    fontFamily: '"Roboto", "Helvetica", "Arial", sans-serif',
```
Auchindoun: We should be good to remove this and stick with whatever default MUI gives us

web/src/app/details/DetailsPage.js
```diff
@@ -30,6 +30,13 @@ const styles = theme => ({
...
+    fontWeight: 400,
+    lineHeight: 1.33,
+    letterSpacing: '0em',
```
Auchindoun: same with line weight, height and spacing, unless the defaults are way off when sizing the font down

web/src/app/details/DetailsPage.js
```diff
@@ -125,12 +132,20 @@ export default class DetailsPage extends React.PureComponent {
...
+                variant='h2'
+              >
+                {title}
+              </Typography>
+              <Typography data-cy='details' variant='subtitle1'>
                 <Markdown value={details} />
               </Typography>
               {titleFooter && (
-                <Typography variant='subtitle1'>{titleFooter}</Typography>
+                <Typography variant='subtitle1' data-cy='time-zone'>
```
Auchindoun: I see that you're using `data-cy=time-zone` in a test for the schedules here, but since this is the main details component that can be reused in many places, I'd suggest we name it something a little more generic, such as `data-cy='title-footer'` or something

web/src/app/main/components/ToolbarTitle.js
```diff
@@ -32,6 +32,10 @@ const styles = {
   title: {
     padding: '0 4px 0 4px',
     flex: 1, // pushes toolbar actions to the right
+    fontWeight: 500,
+    lineHeight: 1.6,
+    fontSize: '1.25rem',
+    letterSpacing: '0.0075em',
```
Auchindoun: same here with the weight, height, and spacing, unless needed we should fall back to MUI's default

web/src/app/util/SkipToContentLink.js
```diff
@@ -0,0 +1,40 @@
...
+      height: 'auto',
+      padding: '15px 23px 14px',
+      fontWeight: 'bold',
+      fontSize: '14px',
+      textDecoration: 'none',
+      lineHeight: 'normal',
+      color: '#CD1931',
+      backgroundColor: '#ffffff',
+      borderRadius: '3px',
+      boxShadow: '0 0 2px 2px rgba(0, 0, 0, 0.6)',
```
Auchindoun: It may be useful to use a MUI button here so that so much legwork doesn't need to be done for it to render properly. You should be able to accomplish the same functionality with a Link with a Button with `onClick="window.location.href = '#content';"`

Let me know what you think.

web/src/cypress/integration/schedules.ts
```diff
@@ -89,9 +89,9 @@ function testSchedules(screen: ScreenFormat) {
 
       // verify changes occurred
       cy.reload()
-      cy.get('h5').contains(newName)
-      cy.get('h6').contains(newDesc)
-      cy.get('h6').contains(newTz)
+      cy.get('[data-cy=details-heading]').contains(newName)
+      cy.get('[data-cy=details]').contains(newDesc)
+      cy.get('[data-cy=time-zone]').contains(newTz)
```
Auchindoun: same here with the data-cy tag pointing to `time-zone`

web/src/app/main/NewApp.js
```diff
@@ -83,6 +84,7 @@ export default class App extends React.PureComponent {
             data-cy='app-bar'
             data-cy-format={cyFormat}
           >
+            <SkipToContentLink />
```
Auchindoun: I'm a little confused as to how this is being rendered. I don't see the link above the sidebar when in normal or screen reader mode. Can you touch on how to get this functionality showing?

web/src/package.json
```diff
@@ -132,6 +132,7 @@
     "eslint-config-standard-jsx": "^6.0.2",
     "eslint-plugin-cypress": "^2.1.2",
     "eslint-plugin-import": "2.17.3",
+    "eslint-plugin-jsx-a11y": "^6.2.1",
```
Auchindoun: We are starting to use explicit versioning in our `package.json` as we've running into issues with our `yarn.lock` file with duplicate declarations of packages if there are peer dependency conflicts. You can fix this pretty easily by removing the `^` and rerunning `yarn install`

web/src/app/main/NewApp.js
dayander: It only appears when it's focused, so if you're a keyboard-only user and you navigate the page using mostly the tab key, it allows you to skip over the repetitive content quickly.

web/src/app/util/SkipToContentLink.js
dayander: I will try and find a way to make it look like an MUI button, but for accessibility/semantics, it should be an anchor tag.

web/src/app/details/DetailsPage.js
dayander: Screen Shot 2019-07-12 at 8 23 02 AM

These are the default styles MUI gives for an h2. I think the best approach if we want to keep the visuals the same as before and not add the CSS is to use the variant prop, and pass what was previously being rendered visually but still with the component prop for semantics.

dayander: I will remove everything and move all this work to the headings branch I started.

web/src/app/util/SkipToContentLink.js
dayander: I changed it using the variant as Button, but I have reservations using it because it transforms text to uppercase which different screenreaders will interact with differently. Some will read out the text normally, whereas others will read each character individually. If you prefer to do this, I would want to override that property.

dayander: I wonder if I could write a SkipContent component or variant to links for material-ui that is added to their repo🧐

node_modules/.yarn-integrity
```diff
@@ -0,0 +1,10 @@
+{
```
Auchindoun: this file shouldn't be committed

web/src/app/util/SkipToContentLink.js
```diff
@@ -0,0 +1,39 @@
...
+      borderRadius: '3px',
+    },
+  },
+})
+
+export function SkipToContentLink() {
+  const classes = useStyles()
+  return (
+    <Link className={classes.skipLink} href='#content'>
+      Skip to content
```
Auchindoun: Can you match the MUI typography here? You should be able to wrap `Skip to content` with Typography but retain the `href`.

Also, having it fixed in the top left corner may cause issues with overlap with the Toolbar title. If you render it within the toolbar it should move the Title to the right while it's rendered

node_modules/.yarn-integrity
dayander: I wonder how it got committed, I will remove it.

web/src/app/util/SkipToContentLink.js
dayander: Let's render it outside the toolbar because of its a header landmark, and we don't want the user to have to navigate into the header landmark to skip repetitive content. I can move it to the right, though.

Auchindoun: I think either way we will fall into the issue of it covering content, whether it be the title of the page or the actions on the right. Without it being inline with the toolbar I'm not sure where a good place to render it will be. Perhaps covering the GoAlert logo in (or over) the sidebar? Is this only being rendered on desktop?

dayander: It's mobile as well if you're using a screen reader on mobile it's helpful to jump down to the content.

I understand your concern for covering content. To be able to access the page actions was a keyboard user they would need to tab again to reach the page actions which would hide the skip link.

dayander: Githubs looks like this

image

dayander: Targets approach is to push down the bar 
image

Auchindoun: I was going to suggest perhaps we push down the bar to make it feel relatively less invasive like how we have on our main website. @arurao @mastercactapus @m17ch @Katy-M what do you think?

dayander: image

mastercactapus: I don't feel super strongly, it seems pretty split across the net.

My preference would be the floating one because it doesn't cause content to jump or move when reading/tabbing -- (like GitHub). 

image
image
image
image
image

dayander: @mastercactapus If you don't feel strongly, either way, I think the team wants to go with how Target.com does it. The branch currently reflects this approach and should be ready to merge.

mastercactapus: Yep, saw that in the chatroom, sounds good!

Issue #42, opened 2019-07-08
QuerySelect: queries containing spaces are flaky


mastercactapus:
**Describe the Bug:**
When searching (via search-select or the search bar) if the query contains a space things can get into an odd/broken state.

**Steps to Reproduce:**
1. Go to the services list page
1. Click create new
1. In the EP search box, start searching for a policy with a space in the name (enter the name with space)
1. Gets stuck at "Loading..."
1. Clear the form and enter other info, results don't update

**Expected Behavior:**
No errors, appropriate results displayed.

**Observed Behavior:**
Drop-down displays trailing-space error message and/or gets stuck at "Loading..."

**Screenshots/Stack Traces:**
![image](https://user-images.githubusercontent.com/595010/60827752-8456ba80-a176-11e9-84c3-be3ca3f9e2bb.png)

**Application Version:**

```
$ goalert version
Version:   v0.22.0
GitCommit: 49153123b11243a5c539fbf598915aefbc44d7bd (clean)
BuildDate: 2019-06-10T16:52:26Z
GoVersion: go1.12.4 (gc)
Platform:  linux/amd64
Migration: trigger-config-sync (#179)
```

**Desktop:**
 - OS: MacOS
 - Browser: Chrome
 - Version: `Version 75.0.3770.100 (Official Build) (64-bit)`

**Additional Context**
Seems to have an issue if the search string contains a trailing space at any point. Stays broken until the page is refreshed.

PR #47, opened 2019-07-10
Search: update validation for search strings containing spaces

arurao:

**Description:**
This PR adds a separate validation function for search strings. It is less strict and removes constraints like not allowing a space at the beginnning/end of the search string. 

Fixes https://github.com/target/goalert/issues/42.

**Additional Info:**
This helps solve the problem in QuerySelects when searching (via search-select or the search bar) for queries containing spaces, resulting in the result of QuerySelect to get locked and go into a broken state for any more characters entered.


validation/validate/search.go
```diff
@@ -0,0 +1,34 @@
...
+// If body is empty, the input is considered valid.
+func Search(fname, body string) error {
+	if body == "" {
+		return nil
+	}
+
+	r := []rune(body)
+	l := len(r)
+
+	if l == 0 {
```
mastercactapus: This is redundant, we do an empty check at the top, and the `range` below would be a no-op anyways. We can move the `len()` bit to the `if` statement below that way too (since we only need to check it once).

PR #48, opened 2019-07-10
http: request queue and timeouts

mastercactapus:

**Description:**
This PR introduces request queueing to the concurrency limiter.

The concurrency limiter was designed to limit the number of simultaneous integration keys, service, and/or user-requests that are processed concurrently per-instance.

The present-day implementation limits concurrency, effectively isolating busy consumers, but all "pending" requests are then processed in random order. This results in latency spikes under load (per individual key) and 502 responses due to "backend timeout".

This change introduces a FIFO queue for all pending requests instead of the random group.
- For requests beyond the maximum (250 per key), they are immediately returned with status code 429 (too many requests)
- Requests stuck in the queue for longer than 25s are also returned with 429

Rejected requests are still scoped per integration key/service/user as before but will now allow a distinction between actual hung requests and throttling.

Throttled requests will now return status code 429 instead of timing out and returning 502.

### DIFF
--- a/app/concurrencylimiter.go
+++ b/app/concurrencylimiter.go
@@ -40,17 +40,31 @@ func (list *pendingList) newReq() *pendingReq {
 
 	return req
 }
+
 func (list *pendingList) remove(req *pendingReq) bool {
+	if req == nil || list == nil {
+		return false
+	}
+
+	defer func() { req.prev, req.next = nil, nil }()
+
 	if list.head == req {
 		list.head = req.next
+		if list.head != nil {
+			list.head.prev = nil
+		}
+		if list.tail == req {
+			list.tail = nil
+		}
 		return true
 	}
 
 	if list.tail == req {
 		list.tail = req.prev
-		if req.prev != nil {
-			req.prev.next = nil
+		if list.tail != nil {
+			list.tail.next = nil
 		}
+
 		return true
 	}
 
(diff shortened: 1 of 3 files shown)


app/concurrencylimiter.go
```diff
@@ -3,70 +3,155 @@ package app
...
+		return true
+	}
+
+	if req.next != nil {
+		req.next.prev = req.prev
+	}
+	if req.prev != nil {
+		req.prev.next = req.next
+		return true
+	}
```
arurao: The above two sections only handles removing links for prev and next of `req` to `req`. Should we also handle removing same for `req` itself?
For example: List is `a = b = c`. And if we want to remove b from the list.
We shoud also say b.prev = nil  and b.next = nil. 
Meaning `req.prev = nil` and `req.next = nil`.

This way, we will remove all links from `b` to any other nodes. From my understanding, if there are no links, that would be equivalent to pointing to nil, But since links do exist here (since it's not a net new node), we should make sure they point to nil.

That way, in the future, if we ever introduce code involving `req.next` or `req.prev` it would not break.

app/concurrencylimiter.go
```diff
@@ -3,70 +3,155 @@ package app
...
+}
+func (list *pendingList) remove(req *pendingReq) bool {
+	if list.head == req {
+		list.head = req.next
+		return true
+	}
+
+	if list.tail == req {
+		list.tail = req.prev
+		if req.prev != nil {
```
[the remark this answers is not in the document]
mastercactapus: Makes sense

app/concurrencylimiter.go
mastercactapus: 👍 actually uncovered a possible bug anyway. Added a test for the list specifically.

PR #50, opened 2019-07-11
ui dependencies: upgrade lodash-es to 4.17.14

m17ch:

**Description:**
This PR upgrades `lodash-es` package from `4.17.11` to `4.17.14`.

https://github.com/lodash/lodash/issues/4348


web/src/package.json
```diff
@@ -83,7 +83,7 @@
     "graphql": "^14.3.1",
     "graphql-tag": "2.10.1",
     "history": "^4.7.2",
-    "lodash-es": "^4.17.10",
+    "lodash-es": "^4.17.14",
```
Auchindoun: let's remove the `^` so we don't fall into discrepancies with the lockfile anymore (like how we had to Cypress packages at one point)

PR #53, opened 2019-07-16
contact methods: allow all country codes and return formatted numbers

dayander:

**Description:**
For user contact methods, when adding a new contact method, the user needs to type their country code along with their phone number to submit a contact method. Using lib phone number, it validates whether the country code and corresponding phone number are valid. Once the number is saved, we now provide a formatted phone number from the backend to be displayed in the contact method list.

**Screenshots:**
image
image

Users now type their phone number with country code instead of having a drop-down of supported country codes.

Formated value is not saved in the DB, but when querying for contact methods, the backend returns a formatted value using lib phone number.


graphql2/schema.graphql
```diff
@@ -769,6 +769,7 @@ type UserContactMethod {
   name: String!
 
   value: String!
+  formattedValue: String
```
mastercactapus: Should be "required", add `!`

user/contactmethod/contactmethod.go
```diff
@@ -8,12 +8,13 @@ import (
...
-	Value    string `json:"value"`
-	Disabled bool   `json:"disabled"`
-	UserID   string `json:"-"`
+	ID             string `json:"id"`
+	Name           string `json:"name"`
+	Type           Type   `json:"type"`
+	Value          string `json:"value"`
+	Disabled       bool   `json:"disabled"`
+	UserID         string `json:"-"`
+	FormattedValue string `json:"formattedValue"`
```
mastercactapus: Rather than adding a field here, add a `FormattedValue` method and then call it from graphql

user/contactmethod/contactmethod_test.go
```diff
@@ -27,10 +27,35 @@ func TestContactMethod_Normalize(t *testing.T) {
 	invalid := []ContactMethod{
 		{Name: "abcd", Type: TypeSMS, Value: "+15555555555"},
 	}
+	validIndia := []ContactMethod{
```
mastercactapus: Add all these to the `valid` slice instead of separate ones. You can add an in-line comment for a hint. Same for `invalid`

```go
valid := []ContactMethod{
  { ... } // US
  { ... } // India
```

web/src/app/users/UserContactMethodForm.js
```diff
@@ -52,40 +51,25 @@ export default class UserContactMethodForm extends React.PureComponent {
...
-            >
-              <MenuItem value='+1'>+1 (USA)</MenuItem>
-              <MenuItem value='+91'>+91 (India)</MenuItem>
-              <MenuItem value='+44'>+44 (UK)</MenuItem>
-            </TextField>
-          </Grid>
-          <Grid item xs={12} sm={12} md={6}>
+          <Grid item xs={12} sm={12}>
             <FormField
+              placeholder={'+1 123-456-7891'}
```
mastercactapus: I don't think this would be valid input, do we need it?

Also, if using an example number use a fake one with 555 like `123-555-0123` so we don't embed someones actual number by random chance.

user/contactmethod/store.go
```diff
@@ -249,6 +249,12 @@ func (db *DB) FindOne(ctx context.Context, id string) (*ContactMethod, error) {
 	if err != nil {
 		return nil, err
 	}
+	num, err := libphonenumber.Parse(c.Value, "")
```
arurao: Since this is being used more than once, I think we should put them in a separate function.

mastercactapus: Yeah, when we move the struct field to the `FormattedValue` method, it should happen there.

Make sure to add a switch on type so we only format phone types

Something like
```go
func (cm *ContactMethod) FormattedValue() (string, error) {
  switch cm.Type {
  case TypeSMS, TypeVoice:
   // libphonenumber
  }
  return cm.Value, nil
}
```

web/src/yarn.lock
```diff
@@ -1,11981 +0,0 @@
-# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
Auchindoun: we don't want this deleted, you can regenerate with `yarn install` from within `web/src`

mastercactapus: Might be a yarn version discrepancy too:

```bash
$ yarn --version
1.17.3
```

web/src/app/users/UserContactMethodForm.js
```diff
@@ -52,40 +51,25 @@ export default class UserContactMethodForm extends React.PureComponent {
...
-              }
-              select
-            >
-              <MenuItem value='+1'>+1 (USA)</MenuItem>
-              <MenuItem value='+91'>+91 (India)</MenuItem>
-              <MenuItem value='+44'>+44 (UK)</MenuItem>
-            </TextField>
-          </Grid>
-          <Grid item xs={12} sm={12} md={6}>
+          <Grid item xs={12} sm={12}>
```
Auchindoun: you can delete `sm={12}`, as setting `xs` will apply `12` across all break points if it's the only one

web/src/app/users/UserContactMethodForm.js
```diff
@@ -52,40 +51,25 @@ export default class UserContactMethodForm extends React.PureComponent {
...
-                getCountryCode(this.props.value.value) +
-                value.replace(/[^0-9]/g, '').slice(0, 14)
-              }
             />
+            <Typography
+              variant='caption'
+              component='p'
+              id='countryCodeIndicator'
+            >
+              Please provide your country code eg. +1 (USA), +91 (India) +44
```
Auchindoun: `eg.` should be `e.g.`

PR #54, opened 2019-07-16
contact methods: disable and require verification after creating or editing

Katy-M:

**Description:**
Creating a new contact method in user profile sets the number to disabled by default and brings up a dialog for verification. Changing the number associated with an existing contact method also sets the method to disabled and requires re-verification.

Created contact methods will be set to disabled by default until they are verified. When creating, a verification form will be brought up and a verification code will be automatically sent to users.


web/src/app/users/UserContactMethodCreateDialog.js
```diff
@@ -16,69 +18,121 @@ const createMutation = gql`
...
 
-  state = {
-    value: {
-      name: '',
-      type: 'SMS',
-      value: '+1',
-    },
-    errors: [],
+  const onComplete = (data, errors) => {
+    if (!contactMethodID) {
```
mastercactapus: Instead of these `if` switches at every level, it really feels like we should have 2 separate dialogs, one for verifying a contact method, and one for creating. We can just transition from one to the other on completion.

We can solve for the animation issues (if it looks odd) with props if needed, but I think it would make the flow easier to follow.

web/src/app/users/UserDetails.js
```diff
@@ -62,6 +63,7 @@ export default class UserDetails extends React.PureComponent {
   state = {
     createCM: false,
     createNR: false,
+    verifyCMID: null,
```
Auchindoun: if we could rename this to `showVerifyDialogByID` to be more clear as to what it's being used for

web/src/app/users/UserDetails.js
```diff
@@ -98,7 +100,22 @@ export default class UserDetails extends React.PureComponent {
         {this.state.createCM && (
           <UserContactMethodCreateDialog
             userID={this.props.userID}
-            onClose={() => this.setState({ createCM: false })}
+            onClose={result => {
+              console.log(result)
```
mastercactapus: Remove `console.log`

web/src/cypress/integration/profile.ts
```diff
@@ -37,15 +37,32 @@ function testProfile(screen: ScreenFormat) {
...
+        .get('input[name=name]')
+        .type(name)
+        .get('input[name=type]')
+        .selectByLabel(type)
+        .get('input[name=value]')
+        .type(value)
+        .get('button[type=submit]')
+        .click()
+      cy.get(`[data-cy='verify-form']`)
+        .find('button[type=button]')
```
mastercactapus: This can be replaced/merged with the below line as `.contains('button[type=button]', 'Cancel')`

web/src/cypress/integration/profile.ts
```diff
@@ -54,12 +71,20 @@ function testProfile(screen: ScreenFormat) {
...
+        .clear()
+        .type(value)
       cy.get('button[type=submit]').click()
 
       cy.get('ul[data-cy=contact-methods]').should(
         'contain',
         `${name} (${cm.type})`,
       )
+      cy.get('ul[data-cy="contact-methods"] li')
+        .contains(`${name} (${cm.type})`)
```
mastercactapus: Can we try simplifying this a bit?

Maybe this would work:
```js
cy.get('ul[data-cy=contact-methods]')
  .contains('li', '${name} (${cm.type})')
  .find('[data-cy=cm-disabled]')
```

- Get the list
- Find the `li` that contains the text
- Find the `cm-disabled` element within the `li`

web/src/app/users/UserContactMethodCreateDialog.js
```diff
@@ -17,68 +16,54 @@ const createMutation = gql`
...
+    />
+  )
+  return (
+    <FormDialog
+      data-cy='create-form'
+      title='Create New Contact Method'
+      loading={loading}
+      errors={nonFieldErrors(error)}
+      onClose={props.onClose}
+      onSubmit={() => createCM()}
```
Auchindoun: you can reference `createCm` here without the wrapper function:

`onSubmit={createCM}`

mastercactapus: Won't that pass the `onSubmit` event handler arguments as options to the mutation function?

mastercactapus: *Update* Discussed offline, let's go back to the wrapper, but add a comment noting why (we don't want the SyntheticEvent being passed to `createCM` or something to that effect).

PR #60, opened 2019-07-17
contact methods: hide reactivate button on mobile/always have option within other actions

Auchindoun:

**Description:**
Hides the `Reactivate` button on mobile as it overlaps ListItem text.

**Additional Info:**
Adds a reactivate button the a contact method's other actions menu if disabled, regardless of the current breakpoint.

### DIFF
--- a/web/src/app/users/UserContactMethodList.js
+++ b/web/src/app/users/UserContactMethodList.js
@@ -4,6 +4,7 @@ import Query from '../util/Query'
 import gql from 'graphql-tag'
 import FlatList from '../lists/FlatList'
 import { Button, Card, CardHeader, Grid, IconButton } from '@material-ui/core'
+import withWidth, { isWidthUp } from '@material-ui/core/withWidth'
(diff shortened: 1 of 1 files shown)


web/src/app/users/UserContactMethodList.js
```diff
@@ -4,6 +4,7 @@ import Query from '../util/Query'
 import gql from 'graphql-tag'
 import FlatList from '../lists/FlatList'
 import { Button, Card, CardHeader, Grid, IconButton } from '@material-ui/core'
+import withWidth, { isWidthUp } from '@material-ui/core/withWidth'
 import { formatCMValue, sortContactMethods } from './util'
```
dayander: Are we still using format value?

Auchindoun: We shouldn't be.. Maybe that got missed

Auchindoun: Fixed with latest from master

PR #61, opened 2019-07-17
navigation: adding submenu

dayander:

**Description:**
Adding a submenu component to handle the expansion of main routes. Currently, no new pages are added, but admin is now ready for more routes.

**Screenshots:**
image
image

When clicking on the admin route it now reroutes to /admin/config, which is the current admin page. Admin now has a submenu that opens when the route is active and is closed when inactive.


web/src/app/main/components/NavSubMenu.js
```diff
@@ -0,0 +1,107 @@
...
+  function activeLink(match) {
+    if (!match) {
+      return handleClick(false)
+    }
+
+    return handleClick(true)
+  }
+  return (
+    <React.Fragment>
+      <NavLink to={path} className={classes.nav} isActive={activeLink}>
```
mastercactapus: This shouldn't result in a changing state, and does not represent a click. The default behavior is to match the path though, does it work if we simply omit it entirely?

dayander: Are you talking about the isActive? We need it to change the state of the collapse component.

mastercactapus: The `Collapse in={}` bit shouldn't be using a state param, it should be calculated by the path directly.

Currently, this equates to the flow:
1. render
1. check if route matched
1. set state based on match
1. re-render with new state
1. re-check match
1. set-state (should be the same)
1. update collapsed

That causes a minimum of 2 renders and probably a flicker on slower systems. There's also no guarantee that `isActive` runs every time, or isn't cached -- as it is not a render hook, it's only meant to determine active state for the `NavLink`

Instead, just calculate the value based on the current route directly:
```js
<Collapse in={isThisOurRoute()}
```

arurao:
Two things I noticed while validating functionally : 
1. The spacing before `Admin` and before `Config` is a little goofy. Could we align both of them?
Screen Shot 2019-07-22 at 3 35 21 PM

2. When I manually route to `/admin`, an unhappy face appears for just a second (probably will show up as a flicker on slower systems), and then reroutes to `/admin/config`. Not quite sure what is happening there.

Auchindoun:
@dayander

> When I manually route to /admin, an unhappy face appears for just a second (probably will show up as a flicker on slower systems), and then reroutes to /admin/config. Not quite sure what is happening there.

This is an overarching problem with the admin page. I'll have a ticket represented for it soon.

web/src/app/admin/AdminRouter.js
```diff
@@ -1,14 +1,15 @@
 import React from 'react'
-import { Switch, Route } from 'react-router-dom'
+import { Switch, Route, Redirect } from 'react-router-dom'
 import { GenericError, PageNotFound } from '../error-pages/Errors'
 import AdminConfig from './AdminConfig'
 import RequireConfig from '../util/RequireConfig'
 
 export const AdminRouter = () => (
   <Switch>
+    <Route exact path='/admin' render={() => <Redirect to='/admin/config' />} />
```
mastercactapus: Can we move this to `routes.js`?

We already handle most redirects inside `renderRoutes`, and could add something like this:
```js
if (cfg.subRoutes && cfg.subRoutes.length) {
    routes.push(
      <Redirect
        key={`redir_sub_${idx}`}
        strict
        exact
        from={path.replace(/\/$/, '')}
        from={cfg.subRoutes[0].path}
      />,
    )
}
```

That way all subroute-items will behave the same and all redirect logic is in a single place.

Auchindoun:
@mastercactapus if you pull the most recent commits the colors have been updated

mastercactapus:
Awesome, thanks! Looks like hover state is still broken though

web/src/app/main/components/NavSubMenu.js
```diff
@@ -0,0 +1,105 @@
...
+      transform: 'rotate(0)',
+    },
+    dropdownClosed: {
+      transform: 'rotate(-90deg)',
+    },
+  }
+})
+
+export default function NavSubMenu(props) {
+  const { parentIcon, parentTitle, path, subMenuRoutes } = props
```
Auchindoun: can we add proptypes for everything this component can expect?

dayander: 🤘

web/src/app/main/components/NavSubMenu.js
```diff
@@ -0,0 +1,113 @@
...
+          <IconComponent />
+        </ListItemIcon>
+        <ListItemText
+          className={classes.parentItem}
+          disableTypography
+          primary={<Typography variant='subtitle1'>{label}</Typography>}
+        />
+        <ExpandMoreIcon
+          className={
+            (classes.dropdown,
```
mastercactapus: We need to join these together in a string -- as committed it will only add the open/closed class and omit the `dropdown` class thus no animation.

I'm assuming it was meant to be:
```suggestion
            classnames(classes.dropdown,
```

Don't forget to add `import classnames from 'classnames'`

PR #62, opened 2019-07-17
ui: grid item container consistency

Auchindoun:

**Description:**
This PR removes any inconsistencies between `Grid` components being rendered with both the `item` and `container` properties, when in most use cases it should be one or the other. In some cases here, a grid was not needed at all has been subsequently removed.

**Additional Info:**
Also updates the few places where we use `spacing={3}` to `spacing={2}`, which is what we mainly use across GoAlert.

### DIFF
--- a/web/src/app/admin/AdminConfig.js
+++ b/web/src/app/admin/AdminConfig.js
@@ -89,7 +89,8 @@ export default class AdminConfig extends React.PureComponent {
           {groups.map((groupID, index) => (
             <Grid
               key={index}
-              item
+              container // contains title above card/card itself
+              item // for each admin config section
               xs={12}
               className={this.props.classes.gridItem}
             >
--- a/web/src/app/admin/AdminConfig.js
+++ b/web/src/app/admin/AdminConfig.js
@@ -83,7 +83,7 @@ export default class AdminConfig extends React.PureComponent {
       <React.Fragment>
         <Grid
           container
-          spacing={3}
+          spacing={2}
           className={this.props.classes.gridContainer}
         >
           {groups.map((groupID, index) => (
--- a/web/src/app/admin/AdminConfig.js
+++ b/web/src/app/admin/AdminConfig.js
@@ -89,7 +89,6 @@ export default class AdminConfig extends React.PureComponent {
           {groups.map((groupID, index) => (
             <Grid
               key={index}
-              container
               item
               xs={12}
               className={this.props.classes.gridItem}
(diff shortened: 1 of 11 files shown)


web/src/app/escalation-policies/PolicyStepForm.js
```diff
@@ -119,7 +119,7 @@ export default class PolicyStepForm extends React.Component {
 
     return (
       <FormContainer {...this.props}>
-        <Grid container spacing={3}>
+        <Grid container spacing={2}>
```
dayander: image
I think this change clicks the data badge at the top of the EP step form.

web/src/app/admin/AdminConfig.js
```diff
@@ -83,13 +83,12 @@ export default class AdminConfig extends React.PureComponent {
...
         <Grid
           container
-          spacing={3}
+          spacing={2}
           className={this.props.classes.gridContainer}
         >
           {groups.map((groupID, index) => (
             <Grid
               key={index}
-              container
```
mastercactapus: This should be both a container and item because it contains a nested grid item and is an item of a parent container.

https://material-ui.com/components/grid/#nested-grid

web/src/app/details/DetailsPage.js
```diff
@@ -123,7 +123,7 @@ export default class DetailsPage extends React.PureComponent {
       classes,
     } = this.props
     return (
-      <Grid item container>
```
mastercactapus: I think the details page is rendered directly under a `Grid` container and so should be an item. `NewApp.js` may need to be cleaned up a bit though.

web/src/app/escalation-policies/PolicyStepForm.js
Auchindoun: good catch

web/src/app/admin/AdminConfig.js
Auchindoun: It looks like that first Grid container may not be needed (A grid container holding only a grid container as its child seems at face value redundant). I'll take a look and see what's happening.

web/src/app/details/DetailsPage.js
Auchindoun: I'll take a look. I didn't get any warnings or errors when I made this change, but I'll dig deeper and see if there are any parent containers above DetailsPage.

Auchindoun: Looks like the parent components calling `<DetailsPage />` have no reference to grids, so we should be fine here.

web/src/app/escalation-policies/PolicyStepForm.js
Auchindoun: Ah, it looks like badges are being cut off at the top regardless of spacing. It must be clipping into the Dialog Title. I'll take note of that for a future issue to resolve

PR #63, opened 2019-07-18
profile: contact method management fixes

mastercactapus:

**Description:**
This PR fixes issues around contact method management in preparation for the next release.

- Fixes "A verification code has been sent to undefined" message for new contact methods
- Fixes a regression that prevented pasting pre-formatted numbers into the contact method field
- Fixes the "too many messages" error when opening the reactivate dialog
- Fixes the "Retry" button appearing in the reactivate dialog when it had not been submitted

It is no longer possible to edit a contact method value. This was done to ensure consistency in logs and verification.

image

### DIFF
--- a/graphql2/graphqlapp/contactmethod.go
+++ b/graphql2/graphqlapp/contactmethod.go
@@ -71,9 +71,6 @@ func (m *Mutation) CreateUserContactMethod(ctx context.Context, input graphql2.C
 }
 
 func (m *Mutation) UpdateUserContactMethod(ctx context.Context, input graphql2.UpdateUserContactMethodInput) (bool, error) {
-	if input.Value != nil {
-		return false, validation.NewFieldError("Value", "changing contact method value is not allowed")
-	}
 
 	err := withContextTx(ctx, m.DB, func(ctx context.Context, tx *sql.Tx) error {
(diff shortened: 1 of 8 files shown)


graphql2/graphqlapp/contactmethod.go
```diff
@@ -71,6 +71,10 @@ func (m *Mutation) CreateUserContactMethod(ctx context.Context, input graphql2.C
 }
 
 func (m *Mutation) UpdateUserContactMethod(ctx context.Context, input graphql2.UpdateUserContactMethodInput) (bool, error) {
+	if input.Value != nil {
+		return false, validation.NewFieldError("Value", "changing contact method value is not allowed")
```
Auchindoun: Should we also be doing this for Type then — since this is disabled on the UI? From what it looks like here, they should behave the same both here and on the UI, since they would both be disabled fields, and both be un-editable when conversing with GraphQL

Also, is this validation needed since we are covering the field errors within the store below? Only one would ever return

mastercactapus: Type isn't in the schema so it isn't needed here. But good catch on the value.

The old UI is going to send the `value` field always, so we should delegate to the store and only reject it if it changes.

PR #72, opened 2019-07-30
app: fix panic in listenevents

arurao:

**Description:**
This PR fixes a panic in the listener channel caused by receiving nil values. A dropped DB connection will no longer cause a crash.

If the listener channel (listenevents) received a nil value (due to dropped DB connection), it would panic and the application would crash.


app/listenevents.go
```diff
@@ -23,17 +23,19 @@ func (app *App) listenEvents(ctx context.Context) error {
 				return
 			}
 
-			log.Debugf(log.WithFields(ctx, log.Fields{
-				"Channel": n.Channel,
-				"PID":     n.BePid,
-				"Extra":   n.Extra,
-			}), "NOTIFY")
+			if n!= nil {
```
mastercactapus: Let's invert and flatten this, also formatting

```go
if n == nil {
  continue
}
```

PR #73, opened 2019-07-30
cypress: fix pagination tests

mastercactapus:

**Description:**
This PR fixes the pagination Cypress tests.

The issue was actually a bug, if you were on any page past the first, and clicked on a nav link the page would be stuck in a loading state.

For example:
1. Click on rotations
1. Navigate to the next page
1. Click the rotations link again
1. Observe broken list

**Additional Info:**
The fix was to change the `PaginatedList` key value from the search param to the router location key (which updates on any navigation change). In doing so discovered that the search bar was causing a double-page-navigation on load which caused intermittent failures. That was fixed by updating the URL actions to only dispatch a navigation action if the parameter changes.

Also:
- Updated `cypress` to 3.4.1, resolving the lodash warning
- Updated `react-apollo`

### DIFF
--- a/web/src/app/actions/main.js
+++ b/web/src/app/actions/main.js
@@ -13,13 +13,22 @@ export function resetURLParams(...keys) {
     keys.forEach(key => {
       q.delete(key)
     })
-    if (q.sort) q.sort()
 
-    const search = q.toString()
-    dispatch(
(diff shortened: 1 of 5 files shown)


dayander:
LGTM

PR #79, opened 2019-08-01
hooks: add useConfig and useSessionInfo

mastercactapus:

**Description:**
This PR adds hooks `useConfig` and `useSessionInfo` to eventually replace the `Config` and `RequireConfig` higher-order components.

**Additional Info:**
- Existing hooks components have been updated to use the new hooks
- Fixes the admin page flicker (on load)

### DIFF
--- a/web/src/app/util/RequireConfig.js
+++ b/web/src/app/util/RequireConfig.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useContext } from 'react'
 import p from 'prop-types'
 import Query from './Query'
 import gql from 'graphql-tag'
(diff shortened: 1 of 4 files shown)


web/src/app/util/RequireConfig.js
```diff
@@ -71,6 +71,32 @@ function isTrue(value) {
...
 
+const mapConfig = value =>
+  _.chain(value.config)
+    .groupBy('id')
+    .mapValues(v => parseValue(v[0].type, v[0].value))
+    .value()
+
+// useSessionInfo returns an object with `isAdmin` true if the current
+// session is an admin, and a `userID` value set to the current users ID.
+export function useSessionInfo() {
```
Auchindoun: would `getSessionInfo`/config make more sense here? i was a little confused as to what it was doing until i got to the file and was able to read through it

mastercactapus: Since it is a hook they recommend to always start with a `use` prefix to separate it from non-hooks functions, even if `get` makes more sense. :)

PR #82, opened 2019-08-06
new user setup: contact method verification logic added to the first time login flow

arurao:

**Description:**
This PR introduces verification logic to the new user first-time login flow. 
When a user logs in for the first time, they will now be shown the contact method verification dialog after adding their first contact method.

When a user logs into GoAlert for the first time, they were being shown the set up a contact method dialog. As soon as they get past that point, they will be shown with the verification dialog, which will assist the user in verifying their contact method, and not leaving it as disabled. Previously, users would have to go to their profile page and reactivate their contact method in order to enable it.


web/src/app/main/components/NewUserSetup.js
```diff
@@ -1,79 +1,45 @@
...
+      <UserContactMethodVerificationDialog
+        contactMethodID={contactMethodID.contactMethodID}
+        onClose={clearIsFirstLogin}
       />
     )
   }
+
+  return (
+    <UserContactMethodCreateDialog
+      title={'Welcome to GoAlert!'}
```
mastercactapus: No need for `{` if it's just a string

web/src/app/users/UserContactMethodCreateDialog.js
```diff
@@ -43,18 +43,22 @@ export default function UserContactMethodCreateDialog(props) {
 
   const { loading, error } = createCMStatus
   const fieldErrs = fieldErrors(error)
+  var title = 'Create New Contact Method'
```
mastercactapus: How about:

```js
const {title = 'Create New Contact Method', subtitle} = props
```

Then we can remove the ternary expressions inside the return for both.

web/src/app/users/UserContactMethodCreateDialog.js
```diff
@@ -43,18 +43,22 @@ export default function UserContactMethodCreateDialog(props) {
...
+
   const form = (
     <UserContactMethodForm
       disabled={loading}
       errors={fieldErrs}
       onChange={cmValue => setCmValue(cmValue)}
       value={cmValue}
+      disclaimer={props.disclaimer}
```
mastercactapus: Add these as optional props to `propTypes`

web/src/app/users/UserContactMethodForm.js
```diff
@@ -84,6 +84,9 @@ export default class UserContactMethodForm extends React.PureComponent {
               </Typography>
             )}
           </Grid>
+          <Grid item xs={12}>
+            <Typography variant='caption'>{this.props.disclaimer}</Typography>
```
mastercactapus: Add `disclaimer` as an optional prop type /w a comment

web/src/app/users/UserContactMethodCreateDialog.js
```diff
@@ -68,4 +72,5 @@ export default function UserContactMethodCreateDialog(props) {
 UserContactMethodCreateDialog.propTypes = {
   userID: p.string.isRequired,
   onClose: p.func,
+  disclaimer: p.string,
```
mastercactapus: Missing `title` and `subtitle` prop types

arurao:
> Could you add the disclaimer text to the dialog that appears when the user creates a new contact method from their profile as well?
> 
> image
> image

We definitely could do that. Although we do have the same disclaimer text on the contact methods list too (on the profile page). So it might be redundant to add it to the dialog and keep it  in the list too. What do the others think? @m17ch @mastercactapus @Auchindoun @dayander ?

web/src/app/users/UserContactMethodForm.js
```diff
@@ -11,6 +11,8 @@ export default class UserContactMethodForm extends React.PureComponent {
       name: p.string.isRequired,
       type: p.oneOf(['SMS', 'VOICE']).isRequired,
       value: p.string.isRequired,
+      // config value for disclaimer is used if navigated here from first login attempt
```
mastercactapus: Let's keep this general (e.g. "disclaimer text to display at the bottom of the form") so it's more clear what it does, rather than where we expect it to be used.

mastercactapus:
As per a quick discussion, we'll keep the disclaimer text as-is on profile, continue to omit it from the Create Contact Method dialog, and add it to the New User dialog.

We'll revisit this in future work.

PR #85, opened 2019-08-08
system limits: set default values

arurao:

**Description:**
This PR introduces a migration to set defaults for all the system limits. It does not modify any existing values. This also adds an exception for the specific migration in the migrations-test, so that it does not fail on migrate-down.


migrate/migrations/20190807210857-set-default-system-limits.sql
```diff
@@ -0,0 +1,15 @@
+-- +migrate Up
+INSERT INTO config_limits (id, max)
+    VALUES
```
Katy-M: When I run `make regendb` I get this error:
```
ERRO[0002] perform migration after resettting: pq: syntax error at or near "    VALUES" 
exit status 1
```
Can anyone replicate this? @dayander , @m17ch , @mastercactapus

migrate/migrations/20190807210857-set-default-system-limits.sql
```diff
@@ -0,0 +1,15 @@
+-- +migrate Up
+INSERT INTO config_limits (id, max)
+    VALUES 
+        ('notification_rules_per_user', 15),
+        ('contact_methods_per_user', 10),
```
mastercactapus: These are indented with non-breaking spaces (`0xa0c2`) somehow, which is causing a syntax error. 

To be consistent with other migrations, remove the indent from `VALUES` and then indent these with 4 spaces (`0x20`)

Double check with `hexdump -C` to see the hex codes to double check, but a simple `make regendb` should tell you if it's fixed.

migrate/migrations/20190807210857-set-default-system-limits.sql
dayander: Yes I get the same error

migrate/migrations/20190807210857-set-default-system-limits.sql
```diff
@@ -0,0 +1,15 @@
...
+        ('ep_steps_per_policy', 10),
+        ('ep_actions_per_step', 20),
+        ('participants_per_rotation', 50),
+        ('rules_per_schedule', 30),
+        ('integration_keys_per_service', 30),
+        ('unacked_alerts_per_service', 200),
+        ('targets_per_schedule', 10)
+ON CONFLICT DO NOTHING;
```
mastercactapus: Missing default values for `heartbeat_monitors_per_service` and `user_overrides_per_schedule`.

I'd say 30 for heartbeats (match integration keys) and for overrides may be a little higher at 35. (note the limit only affects current and future overrides). They can always be adjusted manually later (in DB or UI once that launches).

PR #94, opened 2019-08-15
service details: fix padding for empty list message

Auchindoun:

**Description:**
This PR updates the service details on call list to use FlatList such that its padding issues are resolved.


web/src/app/services/ServiceOnCallList.js
```diff
@@ -75,33 +70,22 @@ export default class OnCallForService extends Component {
 
   renderUsers() {
     const usersDict = this.getUsersDict(this.props.onCallUsers)
+    const items = Object.keys(usersDict).map(id => {
+      const step = usersDict[id]['steps'].length > 1 ? 'Steps' : 'Step'
```
mastercactapus: Can these just be `usersDict[id].steps.length`? (same for `steps` and `name` below)

Though, admittedly, it seems like something `eslint` (or `prettier`) should be handling:
https://eslint.org/docs/rules/dot-notation

PR #105, opened 2019-08-22
cypress: more assertions for admin and page search tests

Auchindoun:

**Description:**
This PR adds more assertions (e.g. `cy.get('x').should('have.value', val)`) such that retry logic is hardened as cypress navigates throughout the page.

### DIFF
--- a/web/src/app/admin/AdminConfirmDialog.js
+++ b/web/src/app/admin/AdminConfirmDialog.js
@@ -72,7 +72,7 @@ export default class AdminConfirmDialog extends React.PureComponent {
         form={
           <List data-cy='confirmation-diff'>
             {changes.map(c => (
-              <ListItem divider key={c.id} data-cy={c.id}>
+              <ListItem divider key={c.id} data-cy={'diff-' + c.id}>
                 <ListItemText
                   disableTypography
                   secondary={
--- a/web/src/app/admin/AdminConfirmDialog.js
+++ b/web/src/app/admin/AdminConfirmDialog.js
@@ -70,9 +70,9 @@ export default class AdminConfirmDialog extends React.PureComponent {
           })),
         )}
         form={
-          <List>
+          <List data-cy='confirmation-diff'>
             {changes.map(c => (
-              <ListItem divider key={c.id}>
+              <ListItem divider key={c.id} data-cy={c.id}>
                 <ListItemText
                   disableTypography
                   secondary={
--- a/web/src/cypress/integration/admin.ts
+++ b/web/src/cypress/integration/admin.ts
@@ -32,27 +32,37 @@ function testAdmin(screen: ScreenFormat) {
       const newVal = 'http://' + c.domain()
       cy.get('input[name="General.PublicURL"]')
         .clear()
+        .should('be.empty')
         .type(newVal)
+        .should('have.value', newVal)
       cy.get('button[data-cy="save"]').click()
+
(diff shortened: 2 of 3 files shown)


web/src/cypress/support/page-search.ts
```diff
@@ -6,21 +6,17 @@ declare namespace Cypress {
...
-        .click({ force: true }) // since we're running tests, it's ok if it is already open
+      cy.get('[data-cy=app-bar] button[data-cy=open-search]').click({
+        force: true,
+      }) // since we're running tests, it's ok if it is already open
     }
 
-    cy.get('@container')
-      .find('input')
-      .type(`{selectall}${s}{enter}`)
+    cy.get('[data-cy=app-bar] input').type(`{selectall}${s}`)
```
mastercactapus: Should have value assertion?

web/src/cypress/integration/admin.ts
```diff
@@ -63,23 +75,31 @@ function testAdmin(screen: ScreenFormat) {
...
         .type(newVal1)
-      cy.get('input[name="Mailgun.EmailDomain"]').type(newVal2)
+        .should('have.value', newVal1)
+      cy.get('input[name="Mailgun.EmailDomain"]')
+        .type(newVal2)
+        .should('have.value', newVal2)
       cy.get('button[data-cy="save"]').click()
 
+      // save dialog
       cy.get('li')
```
mastercactapus: Should we use `.should('contain'` before the `.contains` for these?

Auchindoun: > `.contains()` will automatically retry until the element(s) exist in the DOM.

I think as long as we verify that the list has loaded and is present before we move on, we should be good

web/src/app/admin/AdminConfirmDialog.js
```diff
@@ -70,9 +70,9 @@ export default class AdminConfirmDialog extends React.PureComponent {
           })),
         )}
         form={
-          <List>
+          <List data-cy='confirmation-diff'>
             {changes.map(c => (
-              <ListItem divider key={c.id}>
+              <ListItem divider key={c.id} data-cy={c.id}>
```
mastercactapus: Since we search for `data-cy` attrs globally, we should qualify this one, maybe `data-cy={'confirmation-diff-'+c.id}` or `'diff-'+c.id`, just so that there's never a possible conflict.

web/src/cypress/integration/admin.ts
```diff
@@ -30,29 +30,49 @@ function testAdmin(screen: ScreenFormat) {
...
+        .should('be.empty')
         .type(newVal)
+        .should('have.value', newVal)
       cy.get('button[data-cy="save"]').click()
-      cy.get('p[data-cy="old"]').should('contain', cfg.General.PublicURL)
-      cy.get('p[data-cy="new"]').should('contain', newVal)
+
+      // save dialog
+      cy.get(
+        'ul[data-cy="confirmation-diff"] li[data-cy="General.PublicURL"] p[data-cy="old"]',
```
mastercactapus: nit: should we drop the double quotes?

Auchindoun: It doesn't work with `General.PublicURL` because of the period. I've found not needing double quotes to be pretty random on certain characters, so I always use quotes so I don't have to think/test what will and won't work

PR #106, opened 2019-08-22
ui: Upgrade react to 16.9.0

mastercactapus:

**Description:**
Upgrades React to 16.9.0 which fixes a hooks lifecycle issue resulting in some intermittent failing tests.

- Updated avatars to use hooks to resolve some of the warnings for deprecated APIs

**Out of Scope**
There are more warnings for deprecated React APIs in use by some larger components, and some dependencies, that are not addressed by this PR.

**Additional Info**
Search input would occasionally get dropped as the `useEffect` cleanup didn't always happen on state change. This caused the debounce timeout to sometimes override what was entered in the field.


web/src/app/util/avatar/types.js
```diff
@@ -1,54 +1,67 @@
...
 import { Layers, RotateRight, Today, VpnKey, Person } from '@material-ui/icons'
+import { useSessionInfo } from '../RequireConfig'
+import { Avatar } from '@material-ui/core'
 
-import { graphql } from 'react-apollo'
-import gql from 'graphql-tag'
+function useImage(srcURL) {
+  const [valid, setValid] = useState(false)
 
-export class UserAvatar extends BaseAvatar {
```
Auchindoun: Can `BaseAvatar` be deleted with these changes?

PR #107, opened 2019-08-22
cypress: add more assertions for epSteps test

arurao:

**Description:**
This PR adds more assertions to the cypress test for escalation policy steps.


web/src/cypress/integration/escalationPolicySteps.ts
```diff
@@ -40,7 +40,7 @@ function testSteps(screen: ScreenFormat) {
 
         cy.pageFab()
         cy.get('div[role=dialog]').as('dialog')
```
mastercactapus: We can drop this if `@dialog` is no longer used.

PR #125, opened 2019-09-03
db: switch to pgx driver for listener

arurao:

**Description:**
This PR handles migrating to the pgx driver from pq for the listener logic. 
A Listener has been added to `sqlutil` package which handles listener connections.


sqltrace/stmt.go
```diff
@@ -3,6 +3,7 @@ package sqltrace
 import (
 	"context"
 	"database/sql/driver"
+
```
m17ch: Do we want this space?

switchover/dbsync/status.go
```diff
@@ -97,7 +97,7 @@ func (s *Sync) status(ctx context.Context) (string, error) {
 	buf.WriteString(table.String() + "\n\n")
 
 	if len(nodes) == 0 {
-		issues = append(issues, "No nodes detected.")
+		issues = append(issues, "No nodes detected.(Have you set `db-url-next` for running GoAlert instances?)")
```
mastercactapus: Maybe this one could be ` (make sure GoAlert is running)`

And for the `Non-switchover GoAlert connection` error, we can add ` (try setting --db-url-next)`

The other hints are super short & lower case, also don't forget the space before `(`

Issue #101, opened 2019-08-21
Chore: Transition lodash-es imports to lodash


m17ch:
Following front end unit tests that import lodash-es appear to take a substantially longer time to execute than the others.  
Confirmed that replacing `lodash-es` import with `lodash` speeds up the unit tests.

`app/users/util.test.js`
`app/selectors/url.test.js`

We should adjust all front end app code imports to just use `lodash`.

PR #145, opened 2019-09-19
refetch/rotations: convert to hooks

arurao:

**Description:**
This PR converts the rotations details page to hooks.


web/src/app/rotations/RotationDeleteDialog.js
```diff
@@ -1,11 +1,10 @@
 import React from 'react'
-import { useDispatch } from 'react-redux'
 import p from 'prop-types'
 import gql from 'graphql-tag'
 import Spinner from '../loading/components/Spinner'
 import FormDialog from '../dialogs/FormDialog'
 import { useQuery, useMutation } from 'react-apollo'
-import { push } from 'connected-react-router'
+import { get } from 'lodash-es'
```
m17ch: Do you think if we can use `get` from `lodash` instead of `lodash-es` here?  Related to https://github.com/target/goalert/issues/101 where we witnessed slowness with the `lodash-es` imports  -- thoughts?

Maybe this doesn't matter any more (so I should delete the issue)?

mastercactapus: We should only use one or the other for now, so we don't embed both libraries in our build. Though importing `_` and using `_.get` might be more clear.

PR #146, opened 2019-09-19
refetch/services: convert to hooks

mastercactapus:

**Description:**
This PR updates the service details page and dialogs to use hooks, and makes some improvements to loading behavior (less flicker, render while fetching data, etc..)

- Edit/Delete dialogs now show spinners and handle loading (instead of delay before opening)
- Edit/Delete dialogs should now show error messages for data failures
- Details page loads from the cache, while fetching EP & on-call in the background (if needed)
- Redirect moved to details page (from delete dialog) to fix getting stuck with a 404 page on delete
- The on-call list now shows errors/loading inside instead of flickering and putting a spinner in the center of the screen


web/src/app/services/ServiceDeleteDialog.js
```diff
@@ -23,7 +22,11 @@ function DeleteForm({ epName, error, value, onChange }) {
             value='delete-escalation-policy'
           />
         }
-        label={`Also delete escalation policy: ${epName}`}
+        label={
+          <React.Fragment>
+            `Also delete escalation policy: {epName}
```
Auchindoun:
```suggestion
            Also delete escalation policy: {epName}
```

Auchindoun: Any benefit to change this to use a fragment instead of the 1 line we had before?

web/src/app/services/ServiceDeleteDialog.js
```diff
@@ -61,34 +63,42 @@ export default function ServiceDeleteDialog({ serviceID, onClose }) {
...
   })
   const input = [{ type: 'service', id: serviceID }]
-  const dispatch = useDispatch()
   const [deleteService, { loading, error }] = useMutation(mutation, {
     variables: { input },
-    onCompleted: () => dispatch(push('/services')),
   })
 
-  if (dataLoading) return <Spinner />
+  const epID = _.get(data, 'service.ep.id')
```
Auchindoun: is `_.get` necessary for this one since there is no fallback if it's undefined? As opposed to simply `const epID = data.service.ep.id`

web/src/app/services/ServiceDeleteDialog.js
```diff
@@ -61,34 +63,42 @@ export default function ServiceDeleteDialog({ serviceID, onClose }) {
     variables: { id: serviceID },
   })
   const input = [{ type: 'service', id: serviceID }]
-  const dispatch = useDispatch()
   const [deleteService, { loading, error }] = useMutation(mutation, {
```
Auchindoun: I'd much prefer calling out `deleteServiceStatus` here instead of scraping the variables we need, as it results in easier to read code later in the component, i.e. situations like on L95 where we are doing `loading || dataLoading`. IMO `loading` should always refer to loading data in, and we can use `status` to identify that we are looking at mutation updates

web/src/app/services/ServiceDetails.js
```diff
@@ -42,123 +53,96 @@ const query = gql`
...
-        partialQuery={titleQuery}
-        variables={{ serviceID: this.props.serviceID }}
-        render={this.renderData}
+      </PageActions>
+      <DetailsPage
+        title={data.service.name}
+        details={data.service.description}
+        titleFooter={
+          <div>
+            Escalation Policy:&nbsp;
```
Auchindoun: Let's decide on using `&nbsp;` or `{' '}`, since we have both

web/src/app/services/ServiceEditDialog.js
```diff
@@ -29,80 +28,45 @@ const mutation = gql`
...
-            disabled={loading}
-            value={this.state.value || defaultValue}
-            onChange={value => this.setState({ value })}
-          />
+export default function ServiceEditDialog({ serviceID, onClose }) {
+  const [value, setValue] = useState(null)
+  const { data, loading: dataLoading, error: dataError } = useQuery(query, {
+    variables: { id: serviceID },
+  })
+  const [save, { loading, error }] = useMutation(mutation, {
```
Auchindoun: Same as above

web/src/app/services/ServiceEditDialog.js
```diff
@@ -29,80 +28,45 @@ const mutation = gql`
...
+    variables: { input: { ...value, id: serviceID } },
+    onCompleted: onClose,
+  })
+  const defaults =
+    data && data.service && data.service.id
+      ? {
+          ..._.pick(data.service, ['name', 'description']),
```
Auchindoun: nit: add comment to what is happening here

Auchindoun: nit: add comment to what is happening here

Auchindoun: nit: add comment to what is happening here

web/src/app/services/ServiceDeleteDialog.js
```diff
@@ -61,34 +63,42 @@ export default function ServiceDeleteDialog({ serviceID, onClose }) {
...
+      id: epID,
     })
   }
 
   return (
     <FormDialog
       title='Are you sure?'
       confirm
-      subTitle={`This will delete the service: ${data.service.name}`}
+      subTitle={
```
Auchindoun: This looks to be throwing a warning if the spinner is returned: ```Failed prop type: Invalid prop `subTitle` of type `object` supplied to `FormDialog`, expected `string`.```. So we should update that to a `p.node` propType.

May have to make further adjustments though as this bubbles down to `DialogTitleWrapper` and `subTitle` is wrapped by `Typography`:
```
<DialogContent className={classes.subtitle}>
  <Typography variant='subtitle1' component='p'>
    {subTitle}
  </Typography>
</DialogContent>
```

web/src/app/services/ServiceDeleteDialog.js
mastercactapus: Aside from the ``` ` ``` typo, this was needed so that `epName` could be text or a React node. Putting it in the string/line like before will render `[object Object]`

web/src/app/services/ServiceDeleteDialog.js
mastercactapus: The get is an alternative to `data && data.service && data.service.ep && data.service.ep.id` since `data.service` can exist with`ep` being null, and `data` without `service`.

No fallback is OK in this case, because without ep the form can't get submitted/is disabled, however we still want to continue rendering.

web/src/app/services/ServiceDeleteDialog.js
Auchindoun: Will `epName` ever not be a string?

web/src/app/services/ServiceDeleteDialog.js
Auchindoun: Do you thinking pulling in the lodash package here is worth it over the `&&`s?

web/src/app/services/ServiceDeleteDialog.js
mastercactapus: The dialogs primary action is the mutation, and some need data and some don't, which is why I went with `loading` and `dataLoading` without thinking about it, but clear is always better.

I disagree with just using `status` for the query though as it seems just as ambiguous (status of what?). Maybe when there is more than one query and/or mutation, we should always make it clear?

What do you think of:
```js
const [data, dataStatus] = //...
const [deleteService, deleteServiceStatus] = //...
```

Just follow the first arg name for the second, unless we have a single query, and no mutation, then the shorthand `[data, {loading, error}]` would be appropriate. 

What do you think?

web/src/app/services/ServiceDetails.js
mastercactapus: I hadn't noticed, might be a bug. Does `{' '}` translate to a non-breaking space even if it's a normal (non-breaking) space?

Auchindoun: I believe it adds to the DOM the same way as `&nbsp;`. I think the linter adds `{' '}` when it tries to break a long string in JSX, so using that might be easier to maintain in the long run, unless we change that rule

web/src/app/services/ServiceDeleteDialog.js
Auchindoun: I mean keeping loading/error/data reserved for the main `query` if ever needed, and the `status` _suffix_ (didn't mean as a whole variable, mb) for mutations like `updateRotationStatus`, etc. But I see what you mean here, and that works. When mutations are present I like being explicit with `xxStatus.loading`, though

web/src/app/services/ServiceDeleteDialog.js
mastercactapus: Yep, propTypes need to be updated to `p.node.isRequired` -- will do

web/src/app/services/ServiceDeleteDialog.js
mastercactapus: I think it's clearer to read and easier maintain. Pulling in lodash doesn't cost anything other than the line of the import statement, since we're already using lodash in the application. Webpack puts it in a single global/shared "module", so the code is always downloaded and only "executed" once for the entire page on first import.

Might be worth a quick discussion if we should always use it going forward though (At least for >1 level deep). It also supports the use of a `default` value which is handy elsewhere.

 `foo && foo.bar && foo.bar.baz && foo.bar.baz.bin || _default`
vs
`_.get(foo, 'bar.baz.bin', _default)`

Esp. since the first one only works because `&&` takes precedence over `||`, otherwise you'd need to put parens. in there too.

web/src/app/services/ServiceDeleteDialog.js
mastercactapus: Awesome, will make the change

web/src/app/services/ServiceDetails.js
mastercactapus: I think in this instance it may not even matter (since `&nbsp` is just so multiple spaces don't get squished into one). I think `{' '}` will do wherever we just need a single space.

If we need multiple, I think `&nbsp;` is clearer since `{' '}` and `{' '}` look exactly the same otherwise.

Auchindoun: It doesn't really matter which way we go with, just wanted to pick one since both are present

mastercactapus: Updated to use `{' '}` for regular spaces

PR #147, opened 2019-09-20
flat list: use same drag and drop methods for eps/rotations

Auchindoun:

**Description:**
This PR adjusts EP steps and Rotation participants to use the same drag and drop logic within FlatList. This PR should treat `PolicyStepsCard`, `RotationUsersList`, and `FlatList` as net new files to review instead of line-by-line changes, as they have been refactored into hooks.

**Additional Info:**
- Converts `RotationUsersList` to hooks (as well as a renaming `RotationUserList` -> `RotationUsersList`)
- Converts `PolicyStepsCard` to hooks and utilizes `FlatList`
- Converts `FlatList` to hooks
- Adds `FlatList` support for a custom `ListItem` element
- Removes requirement for `fullScreen` prop passed to `DialogTitleWrapper`


web/src/cypress/integration/escalationPolicySteps.ts
```diff
@@ -170,7 +171,8 @@ function testSteps(screen: ScreenFormat) {
       cy.get('div[data-cy=slack-chip]').should('contain', '#foobar')
 
       // open edit step dialog
-      cy.get('ul[data-cy=steps-list] :nth-child(1) li')
+      cy.get('ul[data-cy=steps-list] li')
+        .eq(1)
```
mastercactapus: I think we changed to `:nth-child` to avoid using `eq` due to the retry issues.

web/src/app/rotations/RotationUsersList.js
```diff
@@ -0,0 +1,262 @@
...
+  mutation updateRotation($input: UpdateRotationInput!) {
+    updateRotation(input: $input)
+  }
+`
+
+const useStyles = makeStyles(theme => ({
+  cardHeader: globalStyles(theme).cardHeader,
+}))
+
+export default function RotationUsersList(props) {
```
mastercactapus: This should be `RotationUserList` (singular as it's a list of User items).

web/src/app/rotations/RotationDetails.js
```diff
@@ -85,7 +92,12 @@ export default class RotationDetails extends React.PureComponent {
           title={data.rotation.name}
           details={data.rotation.description}
           titleFooter={summary}
-          pageFooter={<RotationUserList rotationID={this.props.rotationID} />}
+          pageFooter={
+            <RotationUsersList
+              rotationID={this.props.rotationID}
+              rotation={data.rotation}
```
mastercactapus: Shouldn't this be handled by the nested component?

web/src/app/dialogs/components/DialogTitleWrapper.js
```diff
@@ -36,10 +37,11 @@ const styles = theme => {
  * Renders a fullscreen dialog with an app bar if on a small
  * or mobile screen, and a standard dialog title otherwise.
  */
+@isFullScreen()
 @withStyles(styles)
 export default class DialogTitleWrapper extends Component {
   static propTypes = {
-    fullScreen: p.bool.isRequired,
+    fullScreen: p.bool,
```
mastercactapus: Is this related to drag-and-drop or fixing another issue?

web/src/app/lists/FlatList.js
```diff
@@ -1,228 +1,225 @@
...
+    backgroundColor: '#ebebeb',
+  },
   highlightedItem: {
     borderLeft: '6px solid #93ed94',
     background: '#defadf',
   },
-  participantDragging: {
-    backgroundColor: '#ebebeb',
+  listItem: {
+    width: '100',
```
mastercactapus: Where does `100` come from?

web/src/app/lists/FlatList.js
```diff
@@ -1,228 +1,225 @@
...
-      )
+  /*
+   * Handles rendering either as empty message text,
+   * standard list items, list items that support
+   * drag and drop, or as a subheader list item.
+   */
+  function renderListItems() {
+    // render as empty message
+    if (!props.items.length) {
+      return <ListSubheader>{props.emptyMessage}</ListSubheader>
```
mastercactapus: Does this change how it's displayed?

web/src/app/lists/FlatList.js
```diff
@@ -1,228 +1,225 @@
...
                   className={draggingBackground}
                 >
-                  {this.renderItem(item, idx)}
+                  {renderListItem(item, idx)}
                 </div>
               )
             }}
           </Draggable>
         )
+      } else if (item.subHeader) {
```
mastercactapus: This can just be an `if` on its own line since we are returning above.

web/src/app/lists/FlatList.js
```diff
@@ -1,228 +1,225 @@
...
+              component='h2'
+              variant='subtitle1'
+              color='textSecondary'
+              data-cy='flat-list-item-subheader'
+            >
+              {item.subHeader}
+            </Typography>
+          </ListSubheader>
+        )
+      } else {
```
mastercactapus: No need for else block since we are returning in both above cases.

web/src/cypress/integration/escalationPolicySteps.ts
Auchindoun: > .eq() will automatically retry until the element(s) exist in the DOM.

We should be good on that front. `:nth-child` stopped working here, for some reason, but `eq` worked as intended

web/src/app/rotations/RotationDetails.js
Auchindoun: From what we discussed last week, we were to pass that data down from the parent so things would load together

web/src/app/rotations/RotationUsersList.js
Auchindoun: I had changed it here with the inverse in my mind (plural as it is a list of users).

web/src/app/dialogs/components/DialogTitleWrapper.js
Auchindoun: Yeah, in my Additional Info section, I added 
> Removes requirement for `fullScreen` prop passed to `DialogTitleWrapper`
to make implementing something easier in relation to fixing the DND issue

web/src/app/lists/FlatList.js
Auchindoun: Should be with a `%` iirc

web/src/app/lists/FlatList.js
Auchindoun: It shouldn't! Using this how the material-ui.com docs recommend

web/src/app/lists/FlatList.js
Auchindoun: I feel that it adds clarity to the operations performed here, since they are all relevant to the same `if`/`if else`/`else` block.

Auchindoun: I feel that it adds clarity to the operations performed here, since they are all relevant to the same `if`/`if else`/`else` block.

web/src/app/rotations/RotationDetails.js
Auchindoun: Also, it made converting to hooks much easier, so I went with it

Auchindoun:
I'm kind of in favor to leave this as one PR, since the main proponent behind this PR is to fix drag and drop, and the rest came with that refactoring. It would take more effort to extract this info into three separate PRs than I think it's worth at this point, imo. If the three main files feel too large to review together (`FlatList`, `RotationUsersList`, and `PolicyStepsCard`) then I can separate them out, though.

mastercactapus:
It's mostly due to the overall size and scope. Could do a feature branch to keep it together, but for review purposes, there's a lot of separate/independent things going on each of which need their own attention IMO.

Auchindoun:
I'll work on moving them over, then.

Auchindoun:
Re-opening as code is old and will be easier to manage as one unified codebase to review (too many moving parts, scoping to small PRs proved harder to manage).

arurao:
On functionally testing on this branch, it seems like adding a user to a rotation does not reflect immediately and instead reflects after the next refetch.

Auchindoun:
@arurao Thanks for taking a look. Something may have gotten out of sync since this PR is a bit old. I'll take a look when I get a chance!

Auchindoun:
Closing until story is moved into sprint again

PR #148, opened 2019-09-23
app: tls support

m17ch:

**Description:**
This PR introduces TLS (HTTPS) support within GoAlert by listening on a 2nd socket for TLS connections which are then passed to the existing HTTP server.

- Introduces `--listen-tls`/ `-t` flag to specify the `address:port` of the tls listener (i.e. `localhost:8443`)
- Introduces flags for specifying path of PEM-encoded certificate files  `--tls-cert-file`and `--tls-key-file` (i.e. `cert.pem` and `key.pem`)
- Introduces flags for PEM-encoded certificate data `--tls-cert-data` and `--tls-key-data` (i.e. certificate data provided as value of ENV var `GOALERT_TLS_CERT_DATA`)
- `app/tlsconfig.go` handles appropriate flag usage and helpful error messages
- Enables HTTP/2 support on tls listener.


app/app.go
```diff
@@ -48,7 +49,7 @@ type App struct {
 	mgr *lifecycle.Manager
 
 	db     *sql.DB
-	l      net.Listener
+	l      *multiListener
```
mastercactapus: This can remain `net.Listener` as `multiListener` should implement that interface.

app/app.go
```diff
@@ -127,6 +127,33 @@ func NewApp(c appConfig, db *sql.DB) (*App, error) {
 		}
 	}
 
+	if c.TLSListenAddr != "" {
```
mastercactapus: This can go above line 114 with the other listener

app/app.go
```diff
@@ -127,6 +127,33 @@ func NewApp(c appConfig, db *sql.DB) (*App, error) {
...
+
+		cfg := &tls.Config{Certificates: []tls.Certificate{cert}}
+
+		l2, err := tls.Listen("tcp", c.TLSListenAddr, cfg)
+		if err != nil {
+			return nil, errors.Wrapf(err, "bind address %s", c.TLSListenAddr)
+		}
+		app.l = newMultiListener(l, l2)
+	} else {
+		app.l = newMultiListener(l)
```
mastercactapus: Changing the app to just use `net.Listener` for it's struct field eliminates the need for this `else`

app/app.go
```diff
@@ -127,6 +127,33 @@ func NewApp(c appConfig, db *sql.DB) (*App, error) {
 		}
 	}
 
+	if c.TLSListenAddr != "" {
+		var cert tls.Certificate
+		if c.TLSCertFile != "" && c.TLSKeyFile != "" {
```
mastercactapus: Can we put this stuff in a separate file & method?

```go
func (cfg appConfig) tlsConfig() (*tls.Config, error) {
 // do  `*tls.Config` stuff here
}
```

That will also make it easier to add ACME stuff later.

app/appconfig.go
```diff
@@ -7,11 +7,16 @@ import (
...
 
 type appConfig struct {
-	ListenAddr  string
-	Verbose     bool
-	JSON        bool
-	LogRequests bool
-	APIOnly     bool
+	ListenAddr    string
+	TLSListenAddr string
+	TLSCertFile   string
```
mastercactapus: nit: put the `TLS` fields in their own "section" (just separated by whitespace) like `Jaeger`, `Tracing`, `DBMax` etc..

app/cmd.go
```diff
@@ -479,7 +479,12 @@ func getConfig() (appConfig, error) {
 
 		DisableHTTPSRedirect: viper.GetBool("disable-https-redirect"),
 
-		ListenAddr: viper.GetString("listen"),
+		ListenAddr:    viper.GetString("listen"),
```
mastercactapus: nit: same as above, separate them on their own

app/cmd.go
```diff
@@ -515,6 +520,12 @@ func getConfig() (appConfig, error) {
 		return cfg, validation.NewFieldError("db-url", "is required")
 	}
 
+	if cfg.TLSListenAddr != "" {
+		if cfg.TLSCertFile == "" && cfg.TLSKeyFile == "" && cfg.TLSCert == "" && cfg.TLSKey == "" {
```
mastercactapus: Hmm. I'm wondering if `appConfig` should just have `*tls.Config` as a single field, and handle reading/loading files here (or in it's own func).

In the future, that would allow the app to simply:
- bind TLS if the listen field is specified
- use `*tls.Config` directly, or default to ACME support if it's nil

Let's have a quick chat on this.

app/cmd.go
```diff
@@ -524,6 +535,12 @@ func getConfig() (appConfig, error) {
 
 func init() {
 	RootCmd.Flags().StringP("listen", "l", "localhost:8081", "Listen address:port for the application.")
+	RootCmd.Flags().StringP("listen-tls", "t", "", "HTTPS listen address:port for the application.  Requires setting --tls-cert and --tls-key OR --tls-cert-file and --tls-key-file.")
+
+	RootCmd.Flags().String("tls-cert-file", "", "If using listen-tls, specify path of PEM-encoded certificate file.")
```
mastercactapus: nit: use descriptive phrasing instead of instructional.

Something like "Specifies a path to a PEM-encoded certificate. Has no effect if `--listen-tls` is unset.)"

app/cmd.go
```diff
@@ -524,6 +535,12 @@ func getConfig() (appConfig, error) {
...
+	RootCmd.Flags().String("tls-key-file", "", "If using listen-tls, specify path of PEM-encoded private key file.")
+	RootCmd.Flags().String("tls-cert", "", "If using listen-tls, specify PEM-encoded certificate.")
```
mastercactapus: Same as above, also, should we call this `tls-cert-data` to be a little clearer?

app/multilistener.go
```diff
@@ -0,0 +1,74 @@
...
+type multiListener struct {
+	listeners []net.Listener
+
+	ch    chan net.Conn
+	errCh chan error
+}
+
+func newMultiListener(ln ...net.Listener) *multiListener {
+	ml := multiListener{listeners: ln}
+	ml.ch = make(chan net.Conn)
```
mastercactapus: nit: Define these in the struct declaration:

```go
ml := multiListener{
  listeners:: //...
  ch: //...
}
```

app/multilistener.go
```diff
@@ -0,0 +1,74 @@
...
+	ml.errCh = make(chan error)
+
+	for _, l := range ln {
+		go ml.listen(l)
```
mastercactapus: Maybe add a `sync.WaitGroup` so that when `Close` is called, we wait for the background goroutines to finish. That way we can add a unit test that will validate proper cleanup too.

app/multilistener.go
```diff
@@ -0,0 +1,74 @@
...
+
+// listen waits for and returns the next connection for the listener.
+func (ml multiListener) listen(l net.Listener) {
+	for {
+		c, err := l.Accept()
+		if err != nil {
+			ml.errCh <- err
+			return
+		}
+		ml.ch <- c
```
mastercactapus: This could hang if we accept a connection right before `Close` is called and `Accept` is never called again.

Probably needs to be a `select` that listens for some sort of `shutdown` channel, and then calls `Close` on `c` if we are shutting down the listener. Otherwise, we could get a stalled goroutine and a hung connection.

app/multilistener.go
```diff
@@ -0,0 +1,74 @@
...
+
+// Close ranges through listeners closing all of them and and returns an error if any listener encountered an error while closing.
+// It will log all individual listener errors and return an error in the end in the case of error(s).
+func (ml multiListener) Close() error {
+	hasErr := false
+	for _, l := range ml.listeners {
+		err := l.Close()
+		if err != nil {
+			hasErr = true
+			log.Log(context.Background(), errors.Wrap(err, "Listener error"))
```
mastercactapus: Error wrap message should be "action"-esque. `errors.Wrap(err, "close listener")`

app/multilistener.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+	}
+
+	if hasErr {
+		return errors.New("Multiple listeners failed")
```
mastercactapus: Should be lower-case and a description of what was being done. e.g. `close listeners: multiple errors`

app/tlsconfig.go
```diff
@@ -0,0 +1,52 @@
+package app
+
+import (
+	"crypto/tls"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/viper"
+)
+
+func getTLSConfig() (*tls.Config, error) {
```
Auchindoun: Can we add comments in here? It's difficult to follow what these values + cases + functions mean/are doing without already being knowledgable in the area

PR #149, opened 2019-09-23
engine/cleanup: option to clear old alerts

mastercactapus:

**Description:**
This PR adds the config value `Maintenance.AlertCleanupDays`. Alerts that are closed, that were created more than `Maintenance.AlertCleanupDays` days prior, will be deleted in increments of 100.

This will cause data for the alert, sent messages, logs/responses, to all be purged from the DB.

100 each cyle can handle roughly 72k alerts per hour with one engine instance running. There is a limit of 3 seconds to the query in the cleanup module, though observed times were around 50ms, max of 400ms with large datasets.

The default behavior is no cleanup (same as today) so this change should be safe to all existing configurations. Admins will need to manually set a retention period to enable this feature.

**Out of Scope**
- In the future, a strategy could be implemented to allow archiving data. With this change, DB snapshots/backups will still contain the data should it need to be revisited.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -25,6 +25,10 @@ type Config struct {
 		DisableLabelCreation   bool   `public:"true" info:"Disables the ability to create new labels for services."`
 	}
 
+	Maintenance struct {
+		AlertCleanupDays int `public:"true" info:"Closed alerts will be deleted after this many days (0 means disable cleanup)."`
+	}
+
 	Auth struct {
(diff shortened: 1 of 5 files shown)


Auchindoun:
Should this be closed/put on hold until the relevant pgx updates are finalized and merged?

mastercactapus:
Yeah. Closing until pgx v4 is merged

PR #151, opened 2019-09-24
rotations: fix dnd

Auchindoun:

**Description:**
Fixes drag and drop functionality on rotation users.

**Additional Info:**
- Converts FlatList to hooks
- Supports for custom ListItem elements


web/src/app/lists/FlatList.js
```diff
@@ -1,228 +1,230 @@
...
+              component='h2'
+              variant='subtitle1'
+              color='textSecondary'
+              data-cy='flat-list-item-subheader'
+            >
+              {item.subHeader}
+            </Typography>
+          </ListSubheader>
+        )
+      } else {
```
mastercactapus: We should flatten these since we're exiting the function in each body. IIRC eslint has a rule to catch this specifically we should probably enable.

```js
if () return
if () return
return
```

web/src/cypress/integration/rotations.ts
```diff
@@ -146,7 +146,7 @@ function testRotations(screen: ScreenFormat) {
         .should('contain', rot.users[1].name)
     })
 
-    it('should allow re-ordering participants', () => {
+    it.skip('should allow re-ordering participants', () => {
```
mastercactapus: Why is this skipped?

Auchindoun: it wasn't working, and i'd rather not spend the time trying to debug since it will be working with all 3 components being updated (this will be merged into a feature branch).

web/src/app/lists/FlatList.js
Auchindoun: Sure, like I said before though: I prefer the clarity of an `if`, `if else`, and `else`, but I can make this change.

Auchindoun:
@mastercactapus

> Re-ordering on the rotations list throws a bunch of errors in the console

This is expected, as rotation DND will not work until the rotation PR is in. I don't want to spend time fixing rotations DND in here because I already have fixed it before I split it out into a feature branch. If we want to fix it all together, then I don't think we should have split things apart in the first place. This isn't going into master, so I don't think we have anything to worry about.

For the styling changes, I can add back in the italics, but I'm using MUI's recommended way of using a list subheader text. Just because we may be doing it one way in production, doesn't mean it's the right way.

Before, we were doing the following:
```
<ListItem>
  <ListItemText
    disableTypography
    secondary={
      <Typography color='textSecondary'>{headerNote}</Typography>
    }
  style={{ fontStyle: 'italic' }}
/>
```
When we should be doing:
```
<ListSubheader>{headerNote}</ListSubheader>
```
I add the list subheader prop `component='p'` such that our tests would remain backwards compatible.

mastercactapus:
The primary goal of separating large bodies of work with the "feature branch" approach is to make reviewing easier by taking it in meaningful pieces so that when the branch is merged we don't need to re-review the code, but can focus on the high-level functionality, as going back-and-forth on the code involved would be a massive endeavor.

It makes for some extra work for the submitter in most cases, but to make it quicker and easier for the multiple reviewers to give feedback faster, and clearer/easier to catch errors.

The work required here (to my understanding) is a refactor of `FlatList`, `RotationUserList`, and `PolicyStepsCard`, but that the `FlatList` API is not changing.

If we cannot refactor `FlatList` in an API-compatible way (e.g. without also refactoring `RotationUserList`) then we should have a discussion about what changes need to be made and then put in the PR description so it can be reviewed & tested.

There's no benefit to splitting up PRs if the work cannot be broken up in a way that allows us to verify each piece being completed along the way.

Auchindoun:
Sure. I will move forward with fixing it again and push it to this branch.

PR #152, opened 2019-09-24
twilio: allow sending to all country codes

arurao:

Description:
This PR removes country code validation for a contact number in GoAlert.
As a result, all valid international phone numbers will now be accepted by GoAlert.
Users should confirm with their provider for actual delivery of notifications to those phone numbers.

Out of Scope:
GoAlert will accept any valid phone numbers as contact methods. But the delivery of notifications to international phone numbers depends on the settings at the provider's side.


smoketest/simplenotification_india_test.go
```diff
@@ -1,15 +1,14 @@
...
+
+	"github.com/target/goalert/smoketest/harness"
 )
 
 // TestSimpleNotifications_India tests that SMS and Voice goes out for
 // 1 alert -> service -> esc -> step -> user. 2 rules (1 of each) immediately.
-//
-// Currently, country code '+222' is used as a negative test. If we support
-// 222 in the future, this test will need to be updated.
+// Since with a recent PR, GoAlert accepts all international phone numbers, the negative test for '+222' has been adjusted accordingly.
```
mastercactapus: Future readers of this won't have the context of what used to be here, how about we rename this to `NonUS` instead of just `India`, and we can test India and Mauritania here.

Or, alternatively, just add all 3 to `simplenotification` and get rid of this one.

smoketest/simplenotification_internationalcountrycodes_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+import (
+	"testing"
+
+	"github.com/target/goalert/smoketest/harness"
+)
+
+// TestSimpleNotifications_InternationalCountries tests that SMS and Voice goes out for international country codes for
+// 1 alert -> service -> esc -> step -> user. 2 rules (1 of each) immediately.
+func TestSimpleNotifications_InternationalCountryCodes(t *testing.T) {
```
mastercactapus: Same as above, maybe we just update simplenotification to have all these?

PR #154, opened 2019-09-25
tcpa compliance: move notification disclaimer to add contact method form

Auchindoun:

**Description:**
To meet internal compliance needs, we need to move the Notification Disclaimer from the Contact Method card on the profile page to the "Add Contact Method" form.

**Additional Info:**
- Converts `UserDetails` to hooks

### DIFF
--- a/web/src/app/users/UserDetails.js
+++ b/web/src/app/users/UserDetails.js
@@ -1,7 +1,6 @@
-import React from 'react'
+import React, { useState } from 'react'
 import p from 'prop-types'
 import gql from 'graphql-tag'
-import Query from '../util/Query'
 import DetailsPage from '../details/DetailsPage'
 import StatusUpdateNotification from './UserStatusUpdatePreference'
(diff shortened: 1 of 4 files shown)


web/src/app/users/UserDetails.js
```diff
@@ -53,128 +57,126 @@ function serviceCount(onCallSteps = []) {
...
-        variables={{ id: this.props.userID }}
-        render={({ data }) => this.renderData(data.user)}
-      />
-    )
+  const { data, loading, error } = useQuery(query, {
+    variables: { id: props.userID },
+  })
+
+  if (error) return <GenericError error={error.message} />
+  if (!_.get(data, 'user.id')) {
```
mastercactapus: Can we use a ternary here instead? That avoids nesting `if` statements and makes it a bit more concise.

```suggestion
  if (!_.get(data, 'user.id')) return loading ? <Spinner /> : <ObjectNotFound />
```

PR #160, opened 2019-10-02
services: filter by labels ux enhancement

Auchindoun:

**Description:**
This PR allows individuals to filter services by labels with more ease. When a user selects a label _key_, it will show all services for that key (and populate the search field with the generated search query). From there, they can then select a label _value_ to further refine those results.

**Out of Scope:**
- Allowing to search for service names/descriptions while simultaneously filtering by labels.

**Screenshots:**
image

- Search bar moved from toolbar to just above the list for desktop.
- New filter menu visible when on the service's list page.

- Adds `uniqueValues` option to the `labels` `query`. This acts similarly to the preexisting option `uniqueKeys`.

**Additional Info:**
- Fixes some rendering issues present in `PaginatedList`.


mastercactapus:
Initial thoughts:
- I don't think we should re-use the filter icon for this since it's not a "filter" but rather a helper for configuring/writing the search query for labels.
- Mobile experience feels a bit disjointed, clicking search in the content opens the search above it, and the label form covers the search bar instead of appearing below it. Mobile should probably keep everything at the top so there's less wasted screen space and ensure the existing X works to close/reset (in addition to anything else in the popover)
- The search bar in the content looks nice on widescreen but we need to figure out how to keep things like details pages aligned so it doesn't "bounce" as you navigate.

As far as the new page layout, I'm not 100% sold on the design change yet. Assuming the goal was to solve the "what does the search bar search" issue, we could also update placeholders so they say things like "Search services..." instead of just "Search".

I think it needs a bit more discussion first on the what and whys, as it opens a separate set of issues to solve (i.e. mobile experience, content top padding on details, etc..) that weren't originally part of this work.

---

As for the icon:

Here's a quick rendering of what I was thinking for an in-field icon (ignore the "Name*"):
Screen Shot 2019-10-03 at 7 29 24 AM
Screen Shot 2019-10-03 at 7 29 38 AM

Clicking the icon would toggle the label fields/popover.

The `=` was just a thought because I wanted something a user could associate with a label query, and all label queries need to contain an equal sign, and ID names cannot. Additionally, an aria-label & tooltip that says something like "Search by label", would add some info for the curious.

I like the adornment approach because it makes it more apparent that you're manipulating the search field (and only the search field) since it's not a separate filter like we have on other pages. It also solves the mobile case as we can just have the single search icon, and have the label helper only appear once the field is visible instead of having a list of icons. 

Lastly, if we did have a page-filter (e.g. time-zone, date range, etc..) it would be clear that this is an enhancement for the search field and not something you could use in addition to search, as any other icons or controls would operate on their own, and not change the query.

Auchindoun:
There was no disagreement with the design on Slack @mastercactapus. This is what we decided we could move forward with.

arurao:
Here are my thoughts:
- I personally don't mind the mobile design. Although the optimal way would be to take care of overlaps, it works as of now. The search still has the `X` that ends up closing it. And although the label filter overlaps it, it can be closed by clicking on `done`.
- I do think the filter icon for the label search is a little confusing, considering the other places where we have used the same icon are for other purposes. Having said that, it might be worth considering changing the icon if that isn't too much work and is possible. Otherwise, it definitely is not a blocker and we could keep it as it is right now and change it in the near future.
- Things like adding placeholders could definitely help in achieving a balance of the what this PR is about and what the most optimal solution's design would look like.

Since we did have a brief discussion on Slack before this work was started, my suggestion would be to try and get this PR as an MVP for the feature, making sure to not introduce any logic bugs and we can always come back to this sooner rather than later to improve the overall user experience.

mastercactapus:
@Auchindoun The intent of my review was not to discourage new ideas or design, and was absolutely not meant to be personal. I made an attempt to communicate my own initial thoughts on the look and feel of the proposed design to the application, honestly and clearly. As it was the first time I had a chance to try out or see the new experience, I was admittedly surprised to see changes beyond the scope we had discussed when planning this work. Even so, I apologize if my review came across personally or insulting.

I really appreciate that you made an effort to share your new design and search experience idea early on, and understand your frustration. Unfortunately, most of the team, including myself, was unable to participate in that Slack discussion due to other priorities during this week. Because of that, we did not give feedback earlier. Again, my apologies, for the late-in-cycle feedback.

In hindsight, a thread in chat is great for bouncing ideas, but as a team, I'm sure we can come up with some additional methods offline to ensure everyone gets a fair chance to comment on new proposals _before_ significant work is started. Improving visibility and communication is paramount to preventing this kind of misunderstanding in the future, especially regarding user-facing UI/UX changes.

label/search.go
```diff
@@ -25,7 +26,7 @@ type SearchOptions struct {
 
 	Limit int `json:"-"`
 
-	UniqueKeys bool `json:"u,omitempty"`
+	UniqueKeys bool `json:"k,omitempty"`
```
mastercactapus: This must stay unchanged, otherwise, existing cursors will break.

label/search.go
```diff
@@ -36,8 +37,13 @@ type SearchCursor struct {
 }
 
 var searchTemplate = template.Must(template.New("search").Parse(`
-	SELECT{{if .UniqueKeys}} distinct on (lower(key)){{end}}
-		key, value, tgt_service_id
+	SELECT DISTINCT ON
+		{{if .UniqueKeys}}
+			(lower(key))
```
mastercactapus: We shouldn't use lower() in distinct, because keys (and values) are case-sensitive.

label/search.go
```diff
@@ -52,7 +58,7 @@ var searchTemplate = template.Must(template.New("search").Parse(`
 	{{if .After.Key}}
 		AND (lower(l.key) > lower(:afterKey) AND l.tgt_service_id > :afterServiceID)
 	{{end}}
-	ORDER BY lower(key), tgt_service_id
+	ORDER BY lower(key), lower(value), key, value, tgt_service_id
```
mastercactapus: This should be `lower(key), key, lower(value), value, tgt_service_id` sort by key, then value.

web/src/app/services/ServiceLabelFilterContainer.js
```diff
@@ -0,0 +1,61 @@
...
+          onChange={props.onValueChange}
+          disabled={!props.labelKey}
+        />
+      </Grid>
+    </FilterContainer>
+  )
+}
+
+ServiceLabelFilterContainer.propTypes = {
+  labelKey: p.string,
```
mastercactapus: Instead of breaking out into key/value, can we use a single `onChange` and `value` that uses an object?
```js
value: p.shapeOf({key: p.string, value: p.string}),
onChange: p.func
```

It would clean up the interface a bit and be more consistent I think.

web/src/app/services/ServiceRouter.js
```diff
@@ -31,66 +32,103 @@ const query = gql`
...
-  renderLabels = ({ match }) => (
-    <ServiceLabelList serviceID={match.params.serviceID} />
-  )
+  // grab key and value from the search param, if at all
+  let key = null
+  let value = null
+  if (searchParam.includes('=')) {
+    const searchSplit = searchParam.split(/(!=|=)/)
+    key = searchSplit[0]
+    value = searchSplit[2] // [1] being != or =
```
mastercactapus: The value can contain `=` so we should join the rest of the match so it doesn't get lost:
```js
value = searchSplit.slice(2).join('')
```

web/src/app/services/ServiceRouter.js
```diff
@@ -31,66 +32,103 @@ const query = gql`
...
+            onKeyChange={onKeyChange}
+            onValueChange={onValueChange}
+            onReset={() => setSearchParam()}
+          />
+        }
+      />
     )
   }
+
+  function onKeyChange(newKey) {
```
mastercactapus: As above, if we change to an object this can be simplified like:
```js
function onChange(val) {
    setSearchParam(val.key?`${val.key}=${val.value||''}`:null)
}
```

web/src/app/util/FilterContainer.js
```diff
@@ -37,6 +37,21 @@ export default class FilterContainer extends React.PureComponent {
     anchorEl: null,
   }
 
+  // todo: fill in rest
```
mastercactapus: todo?

web/src/app/util/Search.js
```diff
@@ -51,22 +84,34 @@ export default function Search() {
...
+        className={textClass}
         {...extraProps}
       />
     )
   }
 
   function renderMobile() {
     return (
-      <React.Fragment>
+      <PageActions>
```
mastercactapus: Does `PageActions` support being on the page multiple times/places?

label/search.go
Auchindoun: Yeah, I still need to tweak this per some discussions I had with Aru

web/src/app/util/FilterContainer.js
Auchindoun: yeah, whoever wrote the original didn't add propTypes

web/src/app/util/Search.js
Auchindoun: It does not

web/src/app/services/ServiceRouter.js
Auchindoun: this might reduce the number of lines, but greatly reduces readability

web/src/app/services/ServiceLabelFilterContainer.js
Auchindoun: I'm not really a fan of a `value.value` call, but we can certainly change it to `value.labelValue` and `value.labelKey` if we'd like. I don't think it really makes a difference with the interface/simplification, though, as we would still have to shape the object coming in.

label/search.go
Auchindoun: That's what we had before:
```
SELECT{{if .UniqueKeys}} distinct on (lower(key)){{end}}
		key, value, tgt_service_id
```
I had just broken it up into lines to tweak the rest. What would you suggest we use instead, just `key`?

web/src/app/services/ServiceLabelFilterContainer.js
Auchindoun: On second thought I looked into doing this, but I can't seem to find a clean way to have a single onChange handling different logic depending on key vs value, so I think it would be better to have them separated (as is). It's more clear that way to the reader in my opinion, too. The `onChange` logic can't be in the child (`<ServiceLabelFiltercontianer />`) because depending on the parent rendering it, that `onChange` will change for the `key` and `value` (setting in search params, vs setting in parent state).

mastercactapus: We should be able to do it in the same way we handle forms:

```js
function Container(props) {
  const {onChange, value} = props
  return (
    <KeyField value={value.labelKey} onChange={labelKey => onChange({...value, labelKey}) />
    <ValueField value={value.labelValue} onChange={labelValue => onChange({...value, labelValue}) />
  )
}
```

Auchindoun: As I see it, that would require some form of identifier as well so the parent knows which field to update accordingly in the URL, or storing the value in some object before storing it in the URL, using a useEffect for each type, neither of which I would prefer over my current implementation, which is quick, simple, and easy to read.

Auchindoun:
@dctalbot That's something currently app-wide that I've brought up a few times (stripping out ugly stuff from error messages). It's something I definitely want to to do, but where we do it would affect the errors app-wide, so it would be best to scope that to its own issue. We could certainly hard code it for this particular section to not show "Graphql", but I think it would be best to wait until we can follow through.

web/src/app/services/ServiceLabelFilterContainer.js
mastercactapus: The only field that should be stored in the URL is the search parameter, which will always be `key + '=' + value` and should be updated any time either value is changed. I think we're making this overly complicated, all that's needed (when combined) is:

```js
onChange={({ labelKey, labelValue }) =>
              setSearchParam(labelKey ? labelKey + '=' + labelValue : '')
            }
```

Here's the full change required to simplify this:
```patch
diff --git a/web/src/app/services/ServiceLabelFilterContainer.js b/web/src/app/services/ServiceLabelFilterContainer.js
index 73074c6..06792ff 100644
--- a/web/src/app/services/ServiceLabelFilterContainer.js
+++ b/web/src/app/services/ServiceLabelFilterContainer.js
@@ -9,6 +9,7 @@ import { LabelValueSelect } from '../selection/LabelValueSelect'
 import FilterContainer from '../util/FilterContainer'
 
 export default function ServiceLabelFilterContainer(props) {
+  const { labelKey, labelValue } = props.value
   return (
     <FilterContainer
       icon={<LabelFilterIcon />}
@@ -31,18 +32,20 @@ export default function ServiceLabelFilterContainer(props) {
         <LabelKeySelect
           name='label-key'
           label='Select Label Key'
-          value={props.labelKey}
-          onChange={props.onKeyChange}
+          value={labelKey}
+          onChange={labelKey => props.onChange({ ...props.value, labelKey })}
         />
       </Grid>
       <Grid data-cy='label-value-container' item xs={12}>
         <LabelValueSelect
           name='label-value'
           label='Select Label Value'
-          labelKey={props.labelKey}
-          value={props.labelValue}
-          onChange={props.onValueChange}
-          disabled={!props.labelKey}
+          labelKey={labelKey}
+          value={labelValue}
+          onChange={labelValue =>
+            props.onChange({ ...props.value, labelValue })
+          }
+          disabled={!labelKey}
         />
       </Grid>
     </FilterContainer>
@@ -50,10 +53,8 @@ export default function ServiceLabelFilterContainer(props) {
 }
 
 ServiceLabelFilterContainer.propTypes = {
-  labelKey: p.string,
-  labelValue: p.string,
-  onKeyChange: p.func.isRequired,
-  onValueChange: p.func.isRequired,
+  value: p.shape({ labelKey: p.string, labelValue: p.string }),
+  onChange: p.func,
   onReset: p.func,
 
   // optionally anchors the popover to a specified element's ref
diff --git a/web/src/app/services/ServiceRouter.js b/web/src/app/services/ServiceRouter.js
index cc4ff83..84ee5fc 100644
--- a/web/src/app/services/ServiceRouter.js
+++ b/web/src/app/services/ServiceRouter.js
@@ -38,13 +38,13 @@ export default function ServiceRouter() {
   const setSearchParam = value => dispatch(setURLParam('search', value))
 
   // grab key and value from the search param, if at all
-  let key = null
-  let value = null
+  let labelKey = ''
+  let labelValue = ''
   if (searchParam.includes('=')) {
     const searchSplit = searchParam.split(/(!=|=)/)
-    key = searchSplit[0]
+    labelKey = searchSplit[0]
     // the value can contain "=", so joining the rest of the match such that it doesn't get lost
-    value = searchSplit.slice(2).join('')
+    labelValue = searchSplit.slice(2).join('')
   }
 
   function renderList() {
@@ -62,10 +62,10 @@ export default function ServiceRouter() {
         createLabel='Service'
         searchAdornment={
           <ServiceLabelFilterContainer
-            labelKey={key}
-            labelValue={value}
-            onKeyChange={onKeyChange}
-            onValueChange={onValueChange}
+            value={{ labelKey, labelValue }}
+            onChange={({ labelKey, labelValue }) =>
+              setSearchParam(labelKey ? labelKey + '=' + labelValue : '')
+            }
             onReset={() => setSearchParam()}
           />
         }
@@ -73,25 +73,6 @@ export default function ServiceRouter() {
     )
   }
 
-  function onKeyChange(newKey) {
-    if (newKey) {
-      setSearchParam(newKey + '=')
-    } else {
-      setSearchParam() // clear search if clearing key
-    }
-  }
-
-  function onValueChange(newValue) {
-    // should be disabled if empty, but just in case :)
-    if (!key) return
-
-    if (newValue) {
-      setSearchParam(key + '=' + newValue)
-    } else {
-      setSearchParam(key + '=')
-    }
-  }
-
   function renderDetails({ match }) {
     return <ServiceDetails serviceID={match.params.serviceID} />
   }
```


----

We should always be using `value` and `onChange` for value passing and change callbacks where we can so they work the same as React, MUI, react-select, etc... Sometimes it's going to require a bit of extra work and/or planning but it keeps the component interfaces small, predictable and consistent.

web/src/app/util/FilterContainer.js
mastercactapus: So should we remove the comment or is there work to be done?

web/src/app/services/ServiceLabelFilterContainer.js
Auchindoun: Feel free to commit.. I was happy with my implementation

Auchindoun: I don't think we should force ourselves into an internal API when not needed, though, so I will have to disagree with you there.

web/src/app/util/FilterContainer.js
Auchindoun: There is work to be done, not in scope of this PR.

Auchindoun:
@mastercactapus I wasn't aware of any queries we have having field name errors. I thought this was just in our mutations.

mastercactapus:
@Auchindoun all errors from the validation package will have a field name set, even if it's a query

web/src/app/util/FilterContainer.js
mastercactapus: It's probably not worth the noise then (there's a bunch of places missing propTypes). I suggest we just ignore it or fix it by adding the few missing props.

Auchindoun:
Looks like all errors are rendered through
```
    let noOptionsMessage = 'No options'
    if (optionsError) {
      noOptionsMessage = (
        <ErrorMessage value={optionsError.message || optionsError} />
      )
    }
```
and not using a FormHelper to display under the field. This applies to all of our QuerySelects, not just labels.

Auchindoun:
image
(left: query, right: result)

Searching for something without a `/` in the `labelKeys` query does not return an error, and instead shows no results, which is why we are not seeing an error on the Key query but rather on the Value result set.

mastercactapus:
@Auchindoun Ah, that makes sense. We'll need to look for a key error on the value select and raise it to the key field so it shows up correctly for the user.

Though, alternatively (and maybe a better route), we could consider if we really even need to validate it's a key -- we could validate it as "Search" input inside the SearchValues method. Hard `LabelKey` validation only really needs to be done if we're inserting new data to the DB. For queries, we just need to make sure it's safe to send as a parameter.

That would get rid of the odd error state and if it's an "invalid" key, you would just get no results, which should work fine.

Auchindoun:
> look for a key error on the value select and raise it to the key field so it shows up correctly for the user

I don't know if this should be the correct approach, as the key will have already been in an error state for some time, and a random error message would show in a field _different_ from what the user is currently typing in.

Could you elaborate on your second thought? Not sure I know what you mean. What type of validation do you want to add to the key select query?

mastercactapus:
For the first, we'd have to have some way of passing the error from the value field to the key field -- which feels messy.

For the second, we'd just change this to `valiate.Search` https://github.com/target/goalert/blob/service-search-labels/label/searchvalues.go#L58

Since there's actually no requirement that it be a key (it's read-only and only needs to be safe text). That also makes it use the same validation as the key search so there would never be a discrepancy.

Auchindoun:
@mastercactapus Gotcha. I'll add that validation in there and see how it plays out in the UI

Auchindoun:
So how it would play out right now is that they would just see no results if they were to do `foo=` int the URL or search field, then try to change the label value. All in agreeance? In the URL bar it would show on the main page that the query is already invalid, so they would see right away that there's an issue, and from the search field it would just say no results (potential bug there? do we want to validate `services`' search as LabelKey if `=` is present? @mastercactapus)


From URL:
image


From Search:
image

mastercactapus:
@Auchindoun I think that works. If it's a valid key or not, it's ok. If it's a valid key with no results, or and invalid one with no results, I think we can display it the same.

It might be worth making the same validation change to service search so we don't show an error for invalid keys when searching. Validating the key really only needs to happen when we're putting data in the DB. But if they want to search for the key "foo" they will just get no results since there isn't a key named "foo".

---

It's like if you entered `testing site:invalid_domain` into google, `invalid_domain` can't actually be a domain, but it will still do the search without telling you you're wrong and just return no results

Auchindoun:
Sounds good. I'll update validation on both so the same result set is always expected no matter the source

PR #163, opened 2019-10-09
ui/admin: form scroll fix

KatieMSB:

**Description:**
When admin makes changes in form, confirmation dialog allows for scrolling view with confirm and cancel buttons sticking to the bottom.


web/src/cypress/integration/admin.ts
```diff
@@ -48,7 +48,7 @@ function testAdmin(screen: ScreenFormat) {
       cy.get('button[type="submit"]')
         .contains('Confirm')
         .click()
-      cy.get('div[role="document"]').should('not.exist') // dialog should close
+      cy.get('div[role="document"]').should('exist') // dialog should close
```
Auchindoun: Were these tests failing for you? I believe the div with `role="document"` is the dialog, and shouldn't exist anymore after Cypress clicks on Confirm

KatieMSB: I see the problem, I'll fix that.

web/src/app/dialogs/FormDialog.js
```diff
@@ -193,7 +197,7 @@ export default class FormDialog extends React.PureComponent {
     }
 
     return (
-      <DialogActions>
```
Auchindoun: i think we can leave this as it was in the previous commit, as we still want the header to remain sticky at the top as well

web/src/app/loading/components/LoadingButton.js
```diff
@@ -24,11 +25,13 @@ export default class LoadingButton extends Component {
       noSubmit,
       onClick,
       style,
+      form,
```
mastercactapus: Instead of adding `form` as a new/specific prop, let's just allow all extras to go to the underlying element. That way this component, as a button, will just any/all props a button supports in addition to the ones it uses directly.

```js
const {
  // noSubmit, onClick, etc...
  ...rest
} = this.props

<Button
   // ... existing props
  {...rest}
>
```

web/src/app/dialogs/FormDialog.js
```diff
@@ -192,7 +194,7 @@ export default class FormDialog extends React.PureComponent {
...
+      <DialogActions className={classes.actions}>
```
mastercactapus: missed this

PR #166, opened 2019-10-10
contact methods: suggest country codes by locale

dctalbot:

**Description:**
Suggest default dial code to user on form to create new contact method

Placeholder text for phone number will include default dial code number based on user's IP. Below phone number input will be helper text also suggesting this as an option.

Scenario: a user in Germany
Telephone input placeholder: +491235550123
Helper text below: Please provide your country code e.g. +49 (Germany), +1 (USA), +91 (India)

Defaults to +1, United States

**Additional Info:**
Depends on http://ip-api.com/json but will default to U.S. when not available


web/src/app/users/UserContactMethodForm.js
```diff
@@ -64,7 +64,7 @@ export default class UserContactMethodForm extends React.PureComponent {
           </Grid>
           <Grid item xs={12}>
             <FormField
-              placeholder='+11235550123'
+              placeholder={`${this.props.value.dialCode}1235550123`}
```
Auchindoun: it would be great if we could grab the phone number format + default length per the user's locale as well, if possible with an endpoint similar to above

e.g.
United States placeholder -> `+1 555 555 5555`
India placeholder -> `+91 55555 55555`

(not necessarily including the spaces/dashes/parenthesis, that would be out of scope here i believe)

dctalbot: this is the exact moment I gave up on localization haha. I'll see if there's a library out there that can help

Screen Shot 2019-10-11 at 2 11 46 PM

Auchindoun: Ah I gotcha. Yeah, if we can't find a library that does that for us, I don't think it would be feasible for us to do ourselves

web/src/app/users/UserContactMethodCreateDialog.js
```diff
@@ -22,16 +22,40 @@ export default function UserContactMethodCreateDialog(props) {
     name: '',
     type: 'SMS',
     value: '',
+    countryCode: 'US', // default
+    countryName: 'United States',
   })
 
+  useEffect(() => {
+    async function fetchIpLocale() {
+      const response = await fetch('http://ip-api.com/json')
```
mastercactapus: We should do this once globally rather than every time the dialog opens. Also, is using it this way allowed by their ToS?

web/src/app/util/getExampleNumber.js
```diff
@@ -0,0 +1,251 @@
+const examples = {
```
mastercactapus: We sure none of these are real numbers?

web/src/app/users/UserContactMethodForm.js
```diff
@@ -34,6 +35,24 @@ export default class UserContactMethodForm extends React.PureComponent {
...
+
+    try {
+      exampleNumber = getExampleNumber(locale.countryCode)
+    } catch (e) {
+      exampleNumber = '+1 631 746 3748'
+    }
+
+    const dialCode = exampleNumber.split(' ')[0]
+
+    const targetHQs = ['US', 'IN']
```
mastercactapus: Let's remove this and keep it generic

web/src/app/users/UserContactMethodCreateDialog.js
```diff
@@ -22,16 +22,40 @@ export default function UserContactMethodCreateDialog(props) {
...
+      const countryCode = json.countryCode
+      return { countryCode, countryName }
+    }
+
+    fetchIpLocale().then(({ countryCode, countryName }) => {
+      setCmValue({
+        name: '',
+        type: 'SMS',
+        value: '',
+        countryCode, // NOTE set to mock country code e.g. 'MZ'
```
mastercactapus: mock? not sure what the NOTE means

web/src/app/users/UserContactMethodCreateDialog.js
```diff
@@ -22,16 +22,40 @@ export default function UserContactMethodCreateDialog(props) {
     name: '',
     type: 'SMS',
     value: '',
+    countryCode: 'US', // default
```
mastercactapus: Store these in their own state as they have nothing to do with the contact method

web/src/app/users/UserContactMethodCreateDialog.js
```diff
@@ -22,16 +22,40 @@ export default function UserContactMethodCreateDialog(props) {
...
+
   const [createCM, createCMStatus] = useMutation(createMutation, {
     refetchQueries: ['nrList', 'cmList'],
     onCompleted: result => {
       props.onClose({ contactMethodID: result.createUserContactMethod.id })
     },
     variables: {
       input: {
-        ...cmValue,
+        name: cmValue.name,
```
mastercactapus: as above, keep `cmValue` just as the contact method form's value

web/src/app/users/UserContactMethodForm.js
```diff
@@ -34,6 +35,24 @@ export default class UserContactMethodForm extends React.PureComponent {
...
 
   render() {
+    const locale = this.props.value
+
+    let exampleNumber
+
+    try {
+      exampleNumber = getExampleNumber(locale.countryCode)
+    } catch (e) {
+      exampleNumber = '+1 631 746 3748'
```
mastercactapus: We should use the international format for numbers as we're using that elsewhere

web/src/app/users/UserContactMethodForm.js
```diff
@@ -34,6 +35,24 @@ export default class UserContactMethodForm extends React.PureComponent {
   }
 
   render() {
+    const locale = this.props.value
+
+    let exampleNumber
+
+    try {
```
mastercactapus: I don't think we need the `try` just have getExampleNumber return a default value

web/src/app/util/getExampleNumber.js
dctalbot: Not sure. This list comes from Android (https://github.com/google/libphonenumber)

web/src/app/users/UserContactMethodForm.js
dctalbot: Can you please clarify?

web/src/app/users/UserContactMethodCreateDialog.js
dctalbot: Yes as long as we’re non-commercial (https://members.ip-api.com/legal)

What is the motivation for moving this to global scope? Do you anticipate using the user’s country code in another part of the application? We could cache the value so there is only one fetch per session, but we shouldn’t fetch data for a session that doesn’t need it. I don't imagine this dialog opens very often.

There is also a scenario where e.g. a user in India is on an American VPN and the app thinks they’re in the US. If they switch from VPN to a local network, we want the app to reflect that they’re in India without having a stale cache.

web/src/app/users/UserContactMethodForm.js
dctalbot: Currently the default help text is "Please provide your country code e.g. +1 (USA), +91 (India), +44 (UK)"

Do you want to hard code this for all users?

dctalbot: Every user will now see this: Please provide your country dialing code e.g. +1 (USA), +91 (India)

web/src/app/users/UserContactMethodForm.js
mastercactapus: Here's where we're formatting numbers for contact methods:
https://github.com/target/goalert/blob/master/graphql2/graphqlapp/contactmethod.go#L30

Here's an example of the different formats libphonenumber supports, for reference:
image

dctalbot: Thanks, must be an inconsistency b/w the Go port and the original library.

e.g. in JS:
```
const phoneNumber = parsePhoneNumberFromString('+12133734253')

phoneNumber.format("NATIONAL") === '(213) 373-4253'
phoneNumber.format("INTERNATIONAL") === '+1 213 373 4253'
```

Let's work on leveraging the backend to stay consistent.

web/src/app/users/UserContactMethodCreateDialog.js
```diff
@@ -24,6 +24,23 @@ export default function UserContactMethodCreateDialog(props) {
     value: '',
   })
 
+  const [locale, setLocale] = useState({
+    countryCode: 'US', // default
```
Auchindoun: Are we able to use `luxon` to grab this info by locale?

web/src/app/util/useExamplePhoneNumber.js
```diff
@@ -0,0 +1,18 @@
...
+`
+
+export default function useExamplePhoneNumber(countryCode) {
+  const { data } = useQuery(query, {
+    variables: {
+      input: countryCode,
+    },
+  })
+
+  return data && data.examplePhoneNumber
```
Auchindoun: It may be useful to return data, loading, and error here for the appropriate checks where necessary. That, or just return `queryResult` instead of destructuring it so whichever component is consuming this can use the return value as it pleases

web/src/app/users/UserContactMethodCreateDialog.js
dctalbot: No, as far as I can tell `luxon` will give us timezones and not much else

Auchindoun:
Closing until story is moved into sprint again

Issue #75, opened 2019-08-01
Proxy Path


Haazeel:
Hello guys,

**Is your feature request related to a problem? Please describe:**
When i want put a revers-proxy in front, can't user a context (ex: http://@proxy/goalert/)
I use nginx as reverse_proxy

**Describe the solution you'd like:**
I think, to generate dynamically your path, for take in charge a context root in front.
May be you can reuse --ui-url for that ?

PR #173, opened 2019-10-31
alerts: create alerts for multiple services at once

dctalbot:

**Description:**
The Create Alert dialog now has an option to create an alert on multiple services. This functionality is separated from the single-service alert creation using a MAterial Speed dial component.

**Screenshots:**
Screen Shot 2019-11-04 at 3 15 13 PM

**Additional Info:**
* Blocked by #160 
* Limited to 35 new alerts due to auth checks (2 per new alert). We can increase this number to 50 at the most


Auchindoun:
### Initial Functional Review

1] We'll want to keep the speed dial options synced with what users are currently used to, which is "Create Alert" for the standard flow. We should keep that relatively the same (i.e. "Create Single Alert" or "Alert Single Service"), then the second tooltip for the new feature could read as "Alert Service Group" or something. We'll also want alerting a single service to be the main option for the majority of the time, making it last of the speed dial options (closest to the mouse when hovering over the fab button).

The single service icon could probably be the standard alert icon we use in the sidebar, and for multiple services there is an icon of two keys (we use the key icon for services) that would relay that information well, imo. That is `key-change` and can be found from the `mdi-material-ui` package. Slightly personal preference here, but just my thoughts.
Screen Shot 2019-11-04 at 8 56 00 AM

2] I think we should try and keep the divider line present in the stepper labels at the top of the dialog. This keeps it in-line with standard material practices that you'd see around the web.
Screen Shot 2019-11-04 at 9 01 33 AM
->
krhGl

3] We'll want to reverse the current optional/required boolean in the current form container. Rather than showing all fields as required, we can optionally mark what fields are optional and the asterisks should go away (but will still require the user to fill in necessary to go to the next step).
Screen Shot 2019-11-04 at 9 04 58 AM

4] We should constrain the height of the dialog so it doesn't jump in height when typing in the search text field on step 2.
Screen Shot 2019-11-04 at 9 06 30 AM

4b] Optionally, we can constrain the height of the search results, and have users scroll through that box. Might be easier for everyone to just constrain the dialog height though.

5] Let's add some vertical padding around the `+ Add All` button. It looks a little crammed in there.
Screen Shot 2019-11-04 at 9 08 55 AM

6] We should have text that appears above the search text field on step 2 if nothing has yet been selected, or if the user hasn't searched yet.

e.g. "Use the search box to select your service(s)" and maybe something like "No services currently selected" if they have typed in the field, but haven't clicked on anything yet. This could probably just be a caption with the color set to `textSecondary`.

7] We need to anchor the label filter popover to the search text field such that it opens just below the text field, rather than covering a portion of it:
Screen Shot 2019-11-04 at 9 11 14 AM
->
Screen Shot 2019-11-04 at 9 11 59 AM

8] I think we can simplify the step headers to save on some space.
`Summary and Details` -> `Alert Info`
`Label Selection` -> `Service Selection`
`Confirm and Submit` -> `Summary`

9] I think we can clean this up a bit, too:
Screen Shot 2019-11-04 at 9 15 52 AM
Summary and Details headers should be smaller than the dialog title font size, and the details for those could be on a new line, under the titles, with smaller font size again.

The service chips should also probably have a header above them.

10] It looks like when you submit the form, the dialog drops to a 0 height, then back up to fill its content. This might be resolved if we set a fixed height for the dialog.

11] The `Review Created Alerts` makes it sound like to me that the alerts haven't yet been created. Maybe we could consider changing it to something like `Successfully created alerts!` if all pass, and something else if not?

12] We will need an "Open all" button for the successfully created alerts to open each alert in a separate tab, as per the acceptance criteria.

13] Clicking a alert list item in the last step should open the alert in a new tab, as we don't want that confirmation dialog lost unless the user directly closes the dialog.

13b] Which makes me think, we should disable closing the dialog in the last step (showing created alerts), except for when the user explicitly clicks/taps on the done button.

14] Personal preference, but some form of visual indicator when alerts are created successfully (or not) would be nice:
Screen Shot 2019-11-04 at 9 20 47 AM

15] The `Successfully created x alerts` text should be smaller than the dialog title font size
Screen Shot 2019-11-04 at 9 18 22 AM

web/src/app/alerts/CreateAlertDialog/StepContent/step1.js
```diff
@@ -0,0 +1,196 @@
...
+import { FormField } from '../../../forms'
+import ServiceLabelFilterContainer from '../../../services/ServiceLabelFilterContainer'
+import { Search as SearchIcon } from '@material-ui/icons'
+import FavoriteIcon from '@material-ui/icons/Star'
+import { ServiceChip } from '../../../util/Chips'
+import AddIcon from '@material-ui/icons/Add'
+
+const useStyles = makeStyles(theme => ({
+  addAll: {
+    backgroundColor: theme.palette.grey[100],
```
Auchindoun: The background color contrast ratio of the Add All button although it looks great, doesn't meet WCAG 2.1 requirements, which states that a contrast ratio of at least 3:1 must be met for graphical components.

We could probably get away with making it a `primary['500']` or as a prop, `<Chip color='primary' />` (same as the filter icon), with white text.

Current contrast ratio:
image

---

Ratio of grey button with white text (same ratio for grey button on white background):
image

web/src/app/alerts/CreateAlertDialog/StepContent/step1.js
```diff
@@ -0,0 +1,196 @@
...
+
+  const notice = (
+    <Typography variant='body1' component='p' className={classes.noticeText}>
+      Select services using the search box below
+    </Typography>
+  )
+
+  return (
+    <Grid item xs={12}>
+      <React.Fragment>
```
Auchindoun: This fragment shouldn't be needed since it's contained by the `Grid`

web/src/app/alerts/CreateAlertDialog/StepContent/step1.js
```diff
@@ -0,0 +1,196 @@
...
+              sid => sid !== id,
+            ),
+          })
+        }
+      />
+    )
+  })
+
+  const notice = (
+    <Typography variant='body1' component='p' className={classes.noticeText}>
```
Auchindoun: This should probably be a caption in the top left for consistency, with a text color of `textSecondary`, which is what we use for info text.

e.g.
```
<Typography color='textSecondary' variant='caption'>
  Visit this escalation policy for more information.
</Typography>
```

web/src/app/alerts/CreateAlertDialog/StepContent/step1.js
```diff
@@ -0,0 +1,196 @@
...
+              <ListItemIcon>
+                <FavoriteIcon />
+              </ListItemIcon>
+            )}
+          </ListItem>
+        ))}
+
+        {queriedServices.length === 0 && (
+          <ListItem>
+            <ListItemText secondary={'No services found'} />
```
Auchindoun: nit: remove `{}`

web/src/app/alerts/CreateAlertDialog/StepContent/step2.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react'
```
Auchindoun: File name `step2` -> `Step2`

web/src/app/alerts/CreateAlertDialog/StepContent/step2.js
```diff
@@ -0,0 +1,67 @@
...
+  const Item = props => (
+    <Grid item xs={12}>
+      <Typography
+        variant='subtitle1'
+        component='h3'
+        className={classes.itemTitle}
+      >
+        {props.title}
+      </Typography>
+      <hr className={classes.hrUnderline} />
```
Auchindoun: Material-UI has a `Divider` component you can use that will match theme colors appropriately (will be useful if we generate themes in the future, or have a dark mode)

web/src/app/alerts/CreateAlertDialog/DialogNavigation.js
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  const getBackBtnLabel = () => {
+    switch (activeStep) {
+      case 0:
+        return 'Cancel'
+      default:
+        return 'Back'
+    }
+  }
```
Auchindoun: this can be an if statement or ternary operation since it just has the 1 case:

```
const getBackBtnLabel = () => return activeStep === 0 ? 'Cancel' : 'Back'
```

web/src/app/alerts/CreateAlertDialog/DialogNavigation.js
```diff
@@ -0,0 +1,111 @@
...
+
+  const handleBack = () => {
+    switch (activeStep) {
+      case 0:
+        return onClose()
+      default:
+        return stepBackward()
+    }
+  }
```
Auchindoun: Same as above, this can be converted to a short if statement/ternary operation

web/src/app/alerts/CreateAlertDialog/index.js
```diff
@@ -0,0 +1,106 @@
...
+    }, 1000)
+  }
+
+  return (
+    <Dialog
+      open={props.open}
+      onClose={onLastStep() ? null : onClose} // NOTE only close on last step if user hits Done
+      fullScreen={isWidthDown('md', width)}
+      fullWidth
+      width={'md'}
```
Auchindoun: nit: remove `{}`

web/src/app/alerts/CreateAlertDialog/index.js
```diff
@@ -0,0 +1,106 @@
...
+    >
+      <DialogTitleWrapper
+        fullScreen={isWidthDown('md', width)}
+        title={'Create New Alert'}
+      />
+      <DialogContent style={{ height: '500px' }}>
```
Auchindoun: We should probably set the height on the dialog itself, and use `vh` so it is dynamic based on the user's height, as well as make sure that height isn't set if we're in fullscreen (`isWidthDown('md', width)`)

web/src/app/alerts/CreateAlertDialog/DialogNavigation.js
```diff
@@ -0,0 +1,111 @@
...
+    <DialogActions className={classes.dialogActions}>
+      <Button
+        variant='contained'
+        color='primary'
+        onClick={handleNext}
+        className={classes.button}
+        disabled={nextIsDisabled(activeStep, formFields)}
+      >
+        {getNextBtnLabel()}
+      </Button>
```
Auchindoun: We'll want to set the `type` for the dialog actions if the next button is set to submit, so screen readers/keyboard users can hit submit on enter (if the dialog has focus).

If that doesn't work as intended, we should consider wrapping the step content with a `<Form />`, setting an id on the form, and then additionally setting the prop `form` on the button to that id.

web/src/app/alerts/CreateAlertDialog/index.js
```diff
@@ -0,0 +1,106 @@
...
+                    <StepLabel>{label}</StepLabel>
+                  </Step>
+                ),
+            )}
+          </Stepper>
+        )}
+        <FormContainer
+          onChange={e => onStepContentChange(e)}
+          value={formFields}
+        >
```
Auchindoun: Add `optionalLabels` as a prop to remove the `*` from the alert fields in step 0

web/src/app/alerts/components/CreateAlertSpeedDial.js
```diff
@@ -0,0 +1,110 @@
+import React, { Component } from 'react'
```
Auchindoun: Let's move this file out of the components folder and have it flat under `/alerts/`, so we know that it's not legacy code when we work on the redesign. Perhaps rename `alerts/components` to `alerts/legacyComponents`?

web/src/app/alerts/components/CreateAlertSpeedDial.js
```diff
@@ -0,0 +1,110 @@
...
+      easing: theme.transitions.easing.easeOut,
+    }),
+  },
+  speedDialClose: {
+    transform: 'translate3d(0, 0, 0)',
+    transition: theme.transitions.create('transform', {
+      duration: theme.transitions.duration.leavingScreen,
+      easing: theme.transitions.easing.sharp,
+    }),
+  },
```
Auchindoun: Are these transitions required? We are getting the same transitions on other speed dials without setting any additional CSS

web/src/app/alerts/components/CreateAlertSpeedDial.js
```diff
@@ -0,0 +1,110 @@
...
+})
+
+@withStyles(styles)
+export default class CreateAlertSpeedDial extends Component {
```
Auchindoun: We should convert this to hooks for future maintainability

web/src/app/alerts/CreateAlertDialog/DialogNavigation.js
dctalbot: I like the consistency with its `getNextBtnLabel` counterpart, but open to changing if you want

Auchindoun: That's valid. I'm mostly bringing it up because my IDE was yelling at me :P

web/src/app/alerts/CreateAlertDialog/StepContent/index.js
```diff
@@ -0,0 +1,57 @@
...
+      nodes {
+        id
+        name
+        isFavorite
+      }
+    }
+  }
+`
+
+export default props => {
```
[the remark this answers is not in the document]
dctalbot: Sure, I don't mind if it helps your tooling, but there will still be an `index.js` exporting from the new `./StepContent`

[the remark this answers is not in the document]
dctalbot: Actually, this feels gross, but importing from `./StepContent/StepContent` feels equally gross 🤔 open to suggestions here

web/src/app/alerts/CreateAlertDialog/ServiceListItem.js
```diff
@@ -0,0 +1,76 @@
...
+      id,
+    },
+  })
+
+  let { service } = data || {}
+
+  if (loading) return 'Loading...'
+  if (error) return 'Error fetching data.'
+
+  const serviceUrl = `${window.location.origin}/services/${id}`
```
mastercactapus: Use `absURLSelector` to generate URLs.

web/src/app/alerts/CreateAlertDialog/ServiceListItem.js
```diff
@@ -0,0 +1,76 @@
...
+          </Typography>
+          <Typography color='error' variant='caption'>
+            {err}
+          </Typography>
+        </span>
+
+        <span>
+          <IconButton
+            aria-label='Open service in new tab'
+            onClick={() => window.open(serviceUrl)}
```
mastercactapus: Same as above, use a link & target instead of calling window.open

web/src/app/alerts/CreateAlertDialog/StepContent/Step0.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import { Grid, TextField } from '@material-ui/core'
+import { FormField } from '../../../forms'
+
+export default () => (
```
mastercactapus: Give this and it's file a descriptive name instead of a number like `AlertDetailsStep` or something like that. Also use `function AlertDetailsStep` or assign `.displayName`

web/src/app/alerts/CreateAlertDialog/StepContent/Step1.js
```diff
@@ -0,0 +1,187 @@
...
+    width: '100%',
+    textAlign: 'center',
+    alignSelf: 'center',
+    lineHeight: '9em',
+  },
+}))
+
+const CREATE_ALERT_LIMIT = 35
+
+export default props => {
```
mastercactapus: Same as above, name this something descriptive

web/src/app/alerts/CreateAlertDialog/StepContent/Step1.js
```diff
@@ -0,0 +1,187 @@
...
+  },
+  noticeText: {
+    width: '100%',
+    textAlign: 'center',
+    alignSelf: 'center',
+    lineHeight: '9em',
+  },
+}))
+
+const CREATE_ALERT_LIMIT = 35
```
mastercactapus: put this in `app/config/index.js` or at least higher up. it shouldn't be defined in a specific step (esp. since this step isn't the one that makes the API call)

web/src/app/alerts/CreateAlertDialog/StepContent/Step1.js
```diff
@@ -0,0 +1,187 @@
...
+
+export default props => {
+  const { formFields, queriedServices } = props
+
+  const fieldRef = useRef()
+  const classes = useStyles()
+
+  const labelKey = formFields.searchQuery.split(/(!=|=)/)[0]
+  const labelValue = formFields.searchQuery
+    .split(/(!=|=)/)
```
mastercactapus: This should be `.split('=')` since we only care about values AFTER the first `=`.

Otherwise, it will break with something like `label/key=value!=with=stuff`

web/src/app/alerts/CreateAlertDialog/StepContent/Step1.js
```diff
@@ -0,0 +1,187 @@
...
+    .slice(2)
+    .join('')
```
mastercactapus: We need to re-join with the separator: `.join('=')`

web/src/app/alerts/CreateAlertDialog/StepContent/Step1.js
```diff
@@ -0,0 +1,187 @@
...
+      component='button'
+      label='Add All'
+      size='small'
+      icon={<AddIcon fontSize='small' />}
+      onClick={() => {
+        const toAdd = queriedServices.map(s => s.id)
+
+        // build newState
+        let newState = formFields.selectedServices
+        toAdd.forEach(s => {
```
mastercactapus: If we use `.some` we can jump out early instead of looping through everything once the limit is reached.

Also, we can use the `...` operator like we are below to be consistent.
```js
toAdd.some(s => {
  if (newState.length < CREATE_ALERT_LIMIT) return true
  newState = [...newState, s]
})
```

We should also pass this through `_.uniq` since there will be a delay between adding and the results updating/re-rendering and we don't want to add duplicates (e.g. user double-clicks on a slow machine).

web/src/app/alerts/CreateAlertDialog/StepContent/Step1.js
```diff
@@ -0,0 +1,187 @@
...
+
+        props.onChange({ selectedServices: newState })
+      }}
+    />
+  )
+
+  const serviceChips = formFields.selectedServices.map((id, key) => {
+    return (
+      <ServiceChip
+        key={key}
```
mastercactapus: Use `id` as the key instead of the index. That allows React to delete just the singe DOM element of a removed service chip, instead of updating them all when one is removed.

web/src/app/alerts/CreateAlertDialog/StepContent/Step1.js
```diff
@@ -0,0 +1,187 @@
...
+            </span>
+          ),
+        }}
+      />
+
+      <List aria-label='select service options'>
+        {queriedServices.map((service, key) => (
+          <ListItem
+            button
+            key={key}
```
mastercactapus: Same as above, use the `service.id` instead of index so React can more efficiently update the DOM

web/src/app/alerts/CreateAlertDialog/StepContent/Step2.js
```diff
@@ -0,0 +1,67 @@
...
+  },
+  itemTitle: {
+    paddingBottom: 0,
+  },
+  nudgeRight: {
+    marginLeft: theme.spacing(1),
+  },
+}))
+
+export default props => {
```
mastercactapus: Same as above, let's name this component more descriptively and give the function the same name

web/src/app/alerts/CreateAlertDialog/StepContent/Step2.js
```diff
@@ -0,0 +1,67 @@
...
+
+  return (
+    <Grid container spacing={2}>
+      <Item title='Summary' description={formFields.summary} />
+      <Item title='Details' description={formFields.details} />
+
+      <Item title={`Selected Services (${formFields.selectedServices.length})`}>
+        {formFields.selectedServices.map((id, key) => (
+          <ServiceChip
+            key={key}
```
mastercactapus: Use the `id` as the key, not the index

web/src/app/alerts/CreateAlertDialog/StepContent/Step3.js
```diff
@@ -0,0 +1,86 @@
...
+          )}`}
+        />
+      )}
+
+      {numCreated > 0 && (
+        <Grid item xs={12} className={classes.noPaddingTop}>
+          <List aria-label='Successfully created alerts'>
+            {Object.keys(alertsCreated).map((alias, i) => {
+              const alert = alertsCreated[alias]
+              if (alert) {
```
mastercactapus: Invert this:

`if (!alert) return null`

web/src/app/alerts/CreateAlertDialog/StepContent/Step3.js
```diff
@@ -0,0 +1,86 @@
...
+                return <AlertListItem key={i} id={alertsCreated[alias].id} />
```
mastercactapus: Use a unique id (e.g. `.id`) as the key wherever possible instead of index.

web/src/app/alerts/CreateAlertDialog/StepContent/index.js
mastercactapus: Let's put this in `StepContent.js` and we can export it from `index.js`. That way we keep filenames matching the components they contain so things are easy to search for.

web/src/app/alerts/CreateAlertDialog/StepContent/index.js
```diff
@@ -0,0 +1,57 @@
...
+  const { formFields, mutationStatus, onChange } = props
+
+  // TODO loading error handles
```
mastercactapus: TODO?

web/src/app/alerts/CreateAlertDialog/StepContent/index.js
```diff
@@ -0,0 +1,57 @@
...
+  const { data } = useQuery(query, {
+    variables: {
+      input: {
+        search: formFields.searchQuery,
+        favoritesFirst: true,
+        omit: formFields.selectedServices,
+        favoritesOnly: formFields.searchQuery.length === 0,
```
mastercactapus: We can drop this since we already have `favoritesFirst`

web/src/app/alerts/CreateAlertDialog/index.js
```diff
@@ -0,0 +1,120 @@
...
+  const [
+    createAlerts,
+    { data: alertsCreated, error: alertsFailed, loading: isCreatingAlerts },
+  ] = useCreateAlerts(formFields, setActiveStep)
+
+  const onStepContentChange = e => {
+    setFormFields(prevState => ({ ...prevState, ...e }))
+  }
+
+  const steps = ['Alert Info', 'Service Selection', 'Review', null]
```
mastercactapus: Name this Step titles or something like that?

web/src/app/alerts/CreateAlertDialog/index.js
```diff
@@ -0,0 +1,120 @@
...
+
+  const onLastStep = () => activeStep === steps.length - 1
```
mastercactapus: Can this just be a const?

web/src/app/alerts/CreateAlertDialog/index.js
```diff
@@ -0,0 +1,120 @@
...
+    // NOTE dialog takes time to fade out
+    setTimeout(() => {
+      setActiveStep(0)
+      setFormFields({
+        summary: '',
+        details: '',
+        selectedServices: [],
+        searchQuery: '',
+      })
+    }, 1000)
```
mastercactapus: Let's find a better way than `setTimeout`. This should only be rendered/exist if it's open so we shouldn't need to reset it on close.

web/src/app/alerts/CreateAlertDialog/index.js
```diff
@@ -0,0 +1,120 @@
...
+import { isWidthDown } from '@material-ui/core/withWidth'
+import DialogNavigation from './DialogNavigation'
+import StepContent from './StepContent'
+import { FormContainer, Form } from '../../forms'
+import DialogTitleWrapper from '../../dialogs/components/DialogTitleWrapper'
+import useWidth from '../../util/useWidth'
+import useCreateAlerts from './useCreateAlerts'
+import { fieldErrors } from '../../util/errutil'
+
+export default props => {
```
mastercactapus: Needs a name

web/src/app/alerts/CreateAlertSpeedDial.js
```diff
@@ -0,0 +1,70 @@
...
+const useStyles = makeStyles(theme => ({
+  speedDial: {
+    position: 'fixed',
+    bottom: '2em',
+    right: '2em',
+    zIndex: 9001,
+  },
+}))
+
+export default props => {
```
mastercactapus: Should be named `CreateAlertSpeedDial`

web/src/app/alerts/CreateAlertSpeedDial.js
```diff
@@ -0,0 +1,70 @@
...
+  const classes = useStyles()
+
+  const [showCreateAlertForm, setShowCreateAlertForm] = useState(false)
+  const [showCreateAlertDialog, setShowCreateAlertDialog] = useState(false)
+
+  const showForm = form => {
```
mastercactapus: I don't think we need this, we can just call the proper method from the relevant action below.

e.g. instead of calling `showForm('createAlertByLabel')` let's just call `setShowCreateAlertForm(true)`.

web/src/app/util/Chips.js
```diff
@@ -12,34 +13,60 @@ import {
...
-            <ServiceIcon />
-          </Avatar>
-        }
-        style={style}
-        onDelete={onDelete}
-        onClick={() => history.push(`/services/${id}`)}
-        label={name}
-      />
-    )
+  const history = useHistory()
```
mastercactapus: Don't use `useHistory` directly, instead dispatch push through redux.

web/src/app/util/copyToClipboard-v2.js
```diff
@@ -0,0 +1,16 @@
+export default text => {
+  if (!navigator.clipboard) {
+    window.alert(
```
mastercactapus: Don't use `alert`

web/src/app/util/copyToClipboard-v2.js
```diff
@@ -0,0 +1,16 @@
+export default text => {
```
mastercactapus: Are we replacing `copyToClipboard.js`?

web/src/app/util/copyToClipboard-v2.js
```diff
@@ -0,0 +1,16 @@
...
+  if (!navigator.clipboard) {
+    window.alert(
+      'Copying to clipboard is not supported in this browser. Please use Chrome or Firefox.',
+    )
+    return
+  }
+  navigator.clipboard.writeText(text).then(
+    function() {
+      console.log('Async: Copying to clipboard was successful!')
```
mastercactapus: Don't log unnecessary messages, also some browsers don't define console unless devtools are open anyhow

web/src/app/util/copyToClipboard-v2.js
```diff
@@ -0,0 +1,16 @@
...
+    },
+    function(err) {
+      console.error('Async: Could not copy text: ', err)
```
mastercactapus: Drop the `Async:` prefix

web/src/app/util/useWidth.js
```diff
@@ -0,0 +1,20 @@
+import { useTheme } from '@material-ui/core/styles'
+import useMediaQuery from '@material-ui/core/useMediaQuery'
+
+/**
+ * Only works because the number of with default mui breakpoints.
+ * It will break once you change the number of breakpoints.
+ * See https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level
+ *
+ * https://material-ui.com/components/use-media-query/#migrating-from-withwidth
```
mastercactapus: Move this to the top with a note indicating this was copied directly from the source.

web/src/cypress/integration/alerts.ts
```diff
@@ -288,6 +288,117 @@ function testAlerts(screen: ScreenFormat) {
         cy.get('body').should('contain', svc.name)
       })
     })
+
+    it('should create an alert for mutliple services', () => {
+      cy.createService().then(svc1 => {
+        cy.createService().then(svc2 => {
```
mastercactapus: Let's make these all at once instead of making them in recursive callbacks

web/src/cypress/integration/services.ts
```diff
@@ -547,22 +547,6 @@ function testServices(screen: ScreenFormat) {
       }),
     )
 
-    it('should allow creating alerts', () => {
```
mastercactapus: Creating alerts from the service list page must continue to work.

web/src/app/alerts/CreateAlertDialog/AlertListItem.js
```diff
@@ -0,0 +1,92 @@
...
+      <ListItemText
+        disableTypography
+        style={{
+          display: 'flex',
+          justifyContent: 'space-between',
+          alignItems: 'center',
+        }}
+      >
+        <span>
+          {/* <Typography>
```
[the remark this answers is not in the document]
dctalbot: Wanted to demo it and feedback first. If we aren't displaying the alert's service name or any data besides the alert ID, we can cut out all the mutation stuff in this file

web/src/app/util/useWidth.js
dctalbot: @Auchindoun can you take care of this?

web/src/cypress/integration/services.ts
dctalbot: What is this testing that `alerts.ts` isn't already covering?

Auchindoun: That the service name is auto populated, so only the summary/details needs to be filled out. Likely why we saw that regression earlier without the test catching it

web/src/app/alerts/CreateAlertDialog/StepContent/index.js
dctalbot: At some point I was asked to include this line, but I have no preference. I'll remove for now

web/src/app/alerts/CreateAlertDialog/index.js
dctalbot: Do we want to preserve field data onClose?

web/src/app/util/copyToClipboard-v2.js
dctalbot: Fail silently?

web/src/app/alerts/CreateAlertDialog/index.js
dctalbot: Yes

web/src/app/alerts/CreateAlertDialog/StepContent/Step1.js
dctalbot: May want to revisit `serviceRouter.js` since that's where I copied the logic from. Could be worth extracting into a util function so they're always in sync

Auchindoun: The current logic in `serviceRouter` is:

```
const searchSplit = searchParam.split(/(!=|=)/)
labelKey = searchSplit[0]
// the value can contain "=", so joining the rest of the match such that it doesn't get lost
labelValue = searchSplit.slice(2).join('')
```

it got updated during its PR

Auchindoun: We can probably make it a util function `getServiceLabel` that both use that returns
```
{ key: x, value: y }
```
so the logic remains the same in all locations

web/src/app/alerts/CreateAlertDialog/StepContent/index.js
Auchindoun: If there are no favorites it would default to showing the first services A-Z sorted. I'm kind of in favor of showing only favorites, and searching for the relevant content you want if not applicable

web/src/app/alerts/CreateAlertDialog/index.js
Auchindoun: We do not do this throughout the app. If a form is closed, on opening a new form, it would be empty

web/src/app/util/copyToClipboard-v2.js
Auchindoun: If anything, it should probably be a MUI dialog rather than an alert. Browsers can block alerts, and they are generally frowned upon as they can't really be styled effectively either

web/src/app/alerts/CreateAlertDialog/index.js
dctalbot: Got it, didn't realize there was an onExited callback. Thanks!

web/src/app/alerts/CreateAlertDialog/StepContent/index.js
mastercactapus: Keep in mind not all users of GoAlert will have a large number of services. Also, search-selects currently use `favoritesFirst` as does the search page. Not all users will have favorites either, depending on how they use GoAlert.

Giving some data (A-Z) rather than no data also gives the user a "hint" at what to search for by seeing existing names before entering a query.

Lastly, fav.-only may be a little cleaner in some use-cases, but it would also be different behavior than the rest of the app as well as changing behavior here based on what's in the search field.

web/src/app/alerts/CreateAlertDialog/CreateAlertDialog.js
```diff
@@ -0,0 +1,150 @@
...
+        </Grid>
+        <Grid item className={classes.flexGrow} />
+        <Grid item>
+          <Link
+            href={`/alerts?allServices=1&filter=all&search=${encodeURIComponent(
+              value.summary,
+            )}`}
+            target='_blank'
+            rel='noopener noreferrer'
+          >
```
KatieMSB: Is it necessary to use the Link component here? The Button component has an href prop
https://material-ui.com/api/button/

dctalbot: good catch :) pushing a fix now

web/src/app/dialogs/FormDialog.js
```diff
@@ -183,31 +192,42 @@ export default class FormDialog extends React.PureComponent {
...
+      onBack,
+      onNext,
+    } = this.props
 
     if (alert) {
       return (
         <DialogActions>
           <Button color='primary' onClick={onClose} variant='contained'>
-            Okay
+            Done
```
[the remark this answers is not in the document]
Auchindoun: Ah, I didn't realize this was previous code and not net-new when I made this change. I would argue that this isn't an alert, so we shouldn't have the "Okay" text, in this case. It makes sense in the case where do have it ("Oops, something went wrong! -> Okay"), whereas here it makes more sense/meets our AC to have it say "Done".

We could prob add a prop `buttonTitle` that overrides any of the conditionals. What do you think?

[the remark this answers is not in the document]
dctalbot: We're calling the last step an alert since you can't click out of the dialog. I think "Okay" is more generic and could be used in this case too. Of course we can override, but trying to keep it simple 🤔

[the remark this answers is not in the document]
Auchindoun: Calling it as an alert will also cause the dialog to stop rendering in full screen, which we likely do not want to do since there is a potential large amount of data we can render in the completion step

web/src/app/alerts/CreateAlertDialog/StepContent/CreateAlertServiceSelect.js
```diff
@@ -0,0 +1,244 @@
...
+          className={classes.chipContainer}
+          elevation={0}
+          data-cy='service-chip-container'
+        >
+          {value.length > 0 ? selectedServiceChips : notice}
+        </Paper>
+        {error && (
+          <FormHelperText>
+            {(props.error &&
+              props.error.message.replace(/^./, str => str.toUpperCase())) ||
```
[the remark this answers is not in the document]
dctalbot: Ah good catch, looks like that's leftover from `FormField` where it is also probably just a no-op. I'll cut it out.

web/src/app/dialogs/FormDialog.js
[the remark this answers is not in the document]
Auchindoun: I went ahead and made the button title change since we can't really have the last step be set as an alert without the fullscreen capability on mobile

web/src/app/dialogs/FormDialog.js
```diff
@@ -67,9 +67,17 @@ export default class FormDialog extends React.PureComponent {
     // disables form content padding
     disableGutters: p.bool,
 
+    // overrides any of the main action button titles with this specific text
+    mainActionText: p.string,
```
mastercactapus: Can we call this `primaryActionLabel` to follow MUI's naming conventions?

web/src/app/util/getServiceLabel.js
```diff
@@ -0,0 +1,13 @@
+export default function getServiceLabel(input) {
+  // grab key and value from the input param, if at all
+  let labelKey = ''
+  let labelValue = ''
+  if (input.includes('=')) {
+    const searchSplit = input.split(/(!=|=)/)
```
mastercactapus: Can't split by `!=` or joining wont work

web/src/app/util/getServiceLabel.js
```diff
@@ -0,0 +1,13 @@
...
+    labelKey = searchSplit[0]
+    // the value can contain "=", so joining the rest of the match such that it doesn't get lost
+    labelValue = searchSplit.slice(2).join('')
```
mastercactapus: Need to join on whatever we split on

web/src/cypress/integration/alerts.ts
```diff
@@ -246,47 +246,132 @@ function testAlerts(screen: ScreenFormat) {
...
+      cy.createService().then(s => {
+        svc1 = s
+      })
+
+      cy.createService().then(s => {
+        svc2 = s
+      })
+    })
+
+    it.only('should create an alert for two services', () => {
```
mastercactapus: Remove the `.only` otherwise no other tests will run

web/src/cypress/integration/alerts.ts
```diff
@@ -246,47 +246,132 @@ function testAlerts(screen: ScreenFormat) {
...
-        cy.get('body').should('contain', svc.name)
-      })
+      cy.get('@dialog')
+        .find('[data-cy=service-chip-container]')
+        .contains('[data-cy=service-chip]', svc1.name)
+        .should('be.visible')
+
+      cy.get('@dialog').contains('label', 'Selected Services (2)')
+
+      // TODO test Add All button
```
mastercactapus: Let's add this before merging -- we will not likely come back to it otherwise.

web/src/cypress/integration/alerts.ts
```diff
@@ -246,47 +246,132 @@ function testAlerts(screen: ScreenFormat) {
...
+
+      cy.get('@dialog')
+        .find('li')
+        .should('have.length', 2)
+
+      cy.get('@dialog')
+        .contains('button', 'Done')
+        .click()
+
+      //TODO test errors
```
mastercactapus: If we're going to test this we should do it -- or we can just keep it happy-path for this work.

web/src/cypress/integration/favorites.ts
```diff
@@ -5,18 +5,6 @@ const c = new Chance()
 testScreen('Favorites', testFavorites)
 
 function testFavorites(screen: ScreenFormat) {
-  check(
```
mastercactapus: No longer testing favorite support on services?

web/src/cypress/integration/services.ts
```diff
@@ -547,20 +547,38 @@ function testServices(screen: ScreenFormat) {
...
 
-      cy.get('*[role=dialog]')
-        .contains('button', 'Submit')
+      cy.get('@dialog')
+        .contains('button', 'Next')
         .click()
 
-      cy.location('pathname').should('contain', '/alerts/') // details page
+      // service already selected
+      cy.get('@dialog')
```
mastercactapus: If we're using the new flow/dialog, can we just skip/omit the second step entirely and have it show on the confirm step?

web/src/app/dialogs/FormDialog.js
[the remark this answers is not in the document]
dctalbot: Now that it's not an "alert", we are displaying a back button on the final review step. The only way I can think of getting around this is adding a `hideBackButton` prop since we have three cases: Cancel, Done, and null (no button). Should default to cancel.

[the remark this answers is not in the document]
Auchindoun: Yeahh.. This is why I didn't really see `FormDialog` as a fit for this ticket to begin with. It just keeps making that component more and more messy as these custom features come into play. We can add that prop.

web/src/app/util/getServiceLabel.js
dctalbot: Can you please clarify?

web/src/app/util/getServiceLabel.js
mastercactapus: disregard

web/src/app/util/getServiceLabel.js
mastercactapus: nvm -- missed the group selector :D

web/src/cypress/integration/services.ts
dctalbot: Do we also want to omit the service selection step title?

mastercactapus: Yeah, we can just go from info -> confirm I think -- it would be confusing if there was a step you couldn't get to or was read-only

web/src/app/util/getServiceLabel.test.js
```diff
@@ -0,0 +1,13 @@
+import getServiceLabel from './getServiceLabel'
+
+test('it should split service labels correctly', () => {
+  const check = (search, labelKey, labelValue) =>
+    expect(getServiceLabel(search)).toEqual({ labelKey, labelValue })
+
+  check('wcbn.fm/rfaa=88.3 ann arbor', 'wcbn.fm/rfaa', '88.3 ann arbor')
```
Auchindoun: Gif

dctalbot: Shhh 🤫

web/src/app/alerts/CreateAlertDialog/CreateAlertDialog.js
```diff
@@ -0,0 +1,163 @@
...
+          <Typography>{titleMessage}</Typography>
+        </Grid>
+        <Grid item className={classes.flexGrow} />
+        <Grid item>
+          <Button
+            variant='contained'
+            color='primary'
+            size='small'
+            endIcon={<OpenInNewIcon />}
+            href={`/alerts?allServices=1&filter=all&search=${encodeURIComponent(
```
mastercactapus: Use `absURL` to wrap all GoAlert links, as the UI won't always be mounted directly on the root path (e.g. #75)

web/src/app/util/Chips.js
```diff
@@ -12,34 +15,60 @@ import {
...
+    return data.service.name
   }
+
+  return (
+    <Chip
+      data-cy='service-chip'
+      avatar={
+        <Avatar>
+          <ServiceIcon fontSize='small' />
+        </Avatar>
```
Auchindoun: Let's use `ServiceAvatar` here

web/src/app/util/Chips.js
```diff
@@ -12,34 +15,60 @@ import {
...
+    skip: Boolean(name),
+    fetchPolicy: 'cache-first',
+  })
+
+  const getLabel = () => {
+    if (name) return name
+
+    if (loading) return 'Loading...'
+
+    if (error || !data || !data.service) return 'Error loading name'
```
mastercactapus: We should always get `data.service`, `loading` or `error` so we can probably simplify this. It would also be good to include the error message so it doesn't get swallowed.
```suggestion
    if (error) return `Error: ${error.message}`
```

web/src/cypress/integration/favorites.ts
```diff
@@ -113,5 +105,35 @@ function check(
           .should('contain', name1)
       })
     }
+
+    if (typeName === 'Service') {
```
mastercactapus: I'll take a shot at making this more generic -- otherwise, it probably doesn't belong here, since we're already verifying the list page and this is very specific to the alert flow.

web/src/app/alerts/CreateAlertDialog/StepContent/CreateAlertServiceSelect.js
```diff
@@ -0,0 +1,249 @@
...
+    )
+  }
+
+  const selectedServiceChips = value.map(id => {
+    return (
+      <ServiceChip
+        key={id}
+        clickable={false}
+        id={id}
+        style={{ margin: 3 }}
```
Auchindoun: Let's use a `className` here since we already have it defined at the top.

PR #176, opened 2019-11-05
ui: fix refetch after mutation complete

Auchindoun:


Auchindoun:
Waiting until feature is complete to open

PR #178, opened 2019-11-05
ci: use in-container DB for running tests

mastercactapus:

**Description:**
This PR updates the CI containers and tasks to use Postgres inside the running container. It should lower the possibility of failed builds due to external factors.

### DIFF
--- a/devtools/ci/dockerfiles/build-env/Dockerfile
+++ b/devtools/ci/dockerfiles/build-env/Dockerfile
@@ -2,8 +2,8 @@ FROM golang:1.13.1-alpine
 
 ENV PGDATA=/var/lib/postgresql/data PGUSER=postgres DB_URL=postgresql://postgres@
 RUN apk --no-cache add git nodejs yarn make postgresql postgresql-contrib
-RUN mkdir -p ${PGDATA} /run/postgresql &&\
-    chown postgres ${PGDATA} /run/postgresql &&\
+RUN mkdir -p ${PGDATA} /run/postgresql /var/log/postgresql &&\
+    chown postgres ${PGDATA} /run/postgresql /var/log/postgresql &&\
     su postgres -c "initdb $PGDATA" &&\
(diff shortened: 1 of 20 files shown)


devtools/ci/dockerfiles/build-env/Dockerfile
```diff
@@ -1,2 +1,12 @@
 FROM golang:1.13.1-alpine
-RUN apk --no-cache add git nodejs yarn make postgresql-client
+
+ENV PGDATA=/var/lib/postgresql/data PGUSER=postgres DB_URL=postgresql://postgres@
+RUN apk --no-cache add git nodejs yarn make postgresql postgresql-contrib
```
Auchindoun: how does `postgresql postgresql-contrib` differ from the previous `postgresql-client`?

mastercactapus: `postgresql` depends on/installs the client (as we do still need it). Instead of only adding the client, we're now adding the server & client. The contrib is for the `pgcrypto` extension we need.

web/src/cypress/support/menu.ts
```diff
@@ -24,7 +24,11 @@ function menu(
     const format: 'mobile' | 'wide' = el.data('cy-format')
     expect(format, 'header format').to.be.oneOf(['mobile', 'wide'])
 
-    cy.wrap(sub).click()
+    // Can't just use `.click()` due to
+    // an issue with Cypress 3.6.0
```
Auchindoun: Does this affect `.click()` app wide? This is pretty concerning if this feature is broken on the current version 🙀

mastercactapus: Couldn't nail down exactly what caused it. But was still seeing failures in 3.6.1 (changelog indicated it contained a fix).

Unfortunately 3.6.1 broke `cy.request` anyhow, so reverting back to 3.4.1 for now.

PR #180, opened 2019-11-07
ui/alerts: fix event log order and add ability to load more

KatieMSB:

**Description:**
Load more button appears under alert details logs after more than 35 logs are generated. Load more loads the next 150 logs until all are loaded and then disappears.


web/src/app/alerts/components/AlertDetails.js
```diff
@@ -27,15 +28,153 @@ import Options from '../../util/Options'
...
+        pageInfo {
+          hasNextPage
+          endCursor
+        }
+      }
+    }
+  }
+`
+
+@graphql(query, {
```
Auchindoun: It might be worth bringing this component out into its own file and using the hooks API. It's a lot more straightforward than previous versions of Apollo

https://www.apollographql.com/docs/tutorial/queries/#build-a-paginated-list

web/src/app/alerts/components/AlertDetails.js
```diff
@@ -134,7 +238,10 @@ export default class AlertDetails extends Component {
           className={this.props.classes.tableCardContent}
           style={{ paddingBottom: 0 }}
         >
-          <List>{this.renderAlertLogEvents()}</List>
+          <AlertLogs
+            data={this.props.data.number}
```
mastercactapus: Kind of on @Auchindoun's point, we can move this to its own component that just takes an `alertID` and build it in hooks with the newer stuff in a separate file.

web/src/app/alerts/components/AlertDetailLogs.js
```diff
@@ -0,0 +1,125 @@
...
+      startPolling(2500)
+    } else stopPolling()
+
+    if (props.showExactTimes) {
+      logs = data.alert.data.nodes.map((log, index) => (
+        <div key={index}>
+          <Divider />
+          <ListItem>
+            <ListItemText
+              primary={moment(log.timestamp)
```
mastercactapus: Since we're refactoring this into hooks, let's also switch to `luxon`

web/src/app/alerts/components/AlertDetailLogs.js
```diff
@@ -0,0 +1,125 @@
...
+          </ListItem>
+        </div>
+      ))
+    } else {
+      logs = data.alert.data.nodes.map((log, index) => (
+        <div key={index}>
+          <Divider />
+          <ListItem>
+            <ListItemText
+              primary={moment(log.timestamp)
```
mastercactapus: Same

web/src/app/alerts/components/AlertDetailLogs.js
```diff
@@ -0,0 +1,125 @@
...
+          hasNextPage
+          endCursor
+        }
+      }
+    }
+  }
+`
+export default function AlertDetailLogs(props) {
+  let logs = []
+  const { data, startPolling, stopPolling, fetchMore } = useQuery(query, {
```
mastercactapus: Instead of calling `startPolling` and `stopPolling` explicitly doesn't the default poll interval already handle it for us?

web/src/app/alerts/components/AlertDetailLogs.js
```diff
@@ -0,0 +1,125 @@
...
+    )
+  }
+  return (
+    <List data-cy='alert-logs' style={{ padding: 0 }}>
+      {logs}
+    </List>
+  )
+}
+AlertDetailLogs.propTypes = {
+  alertID: p.number,
```
mastercactapus: I think this is required

web/src/app/alerts/components/AlertDetailLogs.js
```diff
@@ -0,0 +1,125 @@
...
+        <ListItem>
+          <ListItemText primary='No events.' />
+        </ListItem>
+      )
+    }
+    if (data.alert.data.nodes.length <= 35) {
+      startPolling(2500)
+    } else stopPolling()
+
+    if (props.showExactTimes) {
```
mastercactapus: I think we can simplify this all by breaking it up so we don't have so many nested `if` statements. Something like this as an idea:

```js
const doFetchMore = () => { /* fetchMore logic here */}
const renderList = (items, loadMore) => (
    <List data-cy='alert-logs' style={{ padding: 0 }}>
        {items}
        {loadMore && <Button onClick={doFetchMore}>Load More</Button>}
    </List>
)
const renderTime = ts => exactTimes ? luxon_magic1(ts) : luxon_magic2(ts)

if (error) return renderList( <ListItem><ListItemText primary={error.message} /></ListItem>) 
if (loading && !data) return renderList( <ListItem><ListItemText primary='Loading...' /></ListItem>)
if (data.nodes.length===0) return renderList( <ListItem><ListItemText primary='No events.' /></ListItem>)

return renderList(nodes.map((log,index)=>(
  <div key={index}> ... renderTime(log.timestamp) ... </div>
)), data.hasNextPage)
```

web/src/app/alerts/components/AlertDetailLogs.js
Auchindoun: Yep, it does, so we should be able to leave this out. That is, unless we'd like to stop polling if we load more results

web/src/app/alerts/components/AlertDetailLogs.js
```diff
@@ -0,0 +1,144 @@
...
+        }
+      }
+    }
+  }
+`
+export default function AlertDetailLogs(props) {
+  const [poll, setPoll] = useState(POLL_INTERVAL)
+  const { data, error, loading, fetchMore } = useQuery(query, {
+    pollInterval: poll,
+    variables: { id: props.alertID, input: { after: '', limit: 35 } },
```
Auchindoun: We can leave `after: ''` out of the initial query

web/src/app/alerts/components/AlertDetailLogs.js
```diff
@@ -0,0 +1,144 @@
...
+              ),
+            },
+          },
+        }
+      },
+    })
+  }
+  const renderList = (items, loadMore) => {
+    return (
+      <List data-cy='alert-logs' style={{ padding: 0 }}>
```
Auchindoun: You can add the prop `disablePadding` to `List` to remove vertical padding, if that's what you're aiming for with the `List` here. If were looking to strip horizontal padding as well, we can just leave the style as is (or better, convert the inline styles with a `useStyles` hook from `material-ui` at the top of the file, under the `query` const).

web/src/app/alerts/components/AlertDetailLogs.js
```diff
@@ -0,0 +1,144 @@
...
+            'MMM dd yyyy, h:mm:ss a',
+          )}
+          secondary={message}
+        />
+      )
+    return (
+      <ListItemText primary={relativeTime(timestamp)} secondary={message} />
+    )
+  }
+  const relativeTime = timestamp => {
```
Auchindoun: Let's add this to its own file, then we can supplement that util file with `renderItems.test.js` for coverage.

web/src/app/alerts/components/AlertDetailLogs.js
```diff
@@ -0,0 +1,144 @@
...
+    }
+  }
+`
+export default function AlertDetailLogs(props) {
+  const [poll, setPoll] = useState(POLL_INTERVAL)
+  const { data, error, loading, fetchMore } = useQuery(query, {
+    pollInterval: poll,
+    variables: { id: props.alertID, input: { after: '', limit: 35 } },
+  })
+
```
Auchindoun: Let's create two variables that we can use later on to make it easier to have our conditional rendering checks in place:

```
const events = _.get(data, 'alert.data.nodes', [])
const pageInfo = _.get(data, 'alert.data.pageInfo', {})
```

web/src/app/alerts/components/AlertDetailLogs.js
```diff
@@ -0,0 +1,144 @@
...
+    if (Interval.after(from, { days: 1 }).contains(to))
+      return 'Today at ' + to.toFormat('h:mm a')
+    if (Interval.before(from, { days: 1 }).contains(to))
+      return 'Yesterday at ' + to.toFormat('h:mm a')
+    if (Interval.before(from, { weeks: 1 }).contains(to))
+      return 'Last ' + to.weekdayLong + ' at ' + to.toFormat('h:mm a')
+    return to.toFormat('MM/dd/yyyy')
+  }
+
+  if (data && data.alert.data.nodes.length === 0) {
```
Auchindoun: `if (events.length === 0)`

web/src/app/alerts/components/AlertDetailLogs.js
```diff
@@ -0,0 +1,144 @@
...
+    data &&
+      data.alert.data.nodes.map((log, index) => (
+        <div key={index}>
+          <Divider />
+          <ListItem>{renderItems(log.timestamp, log.message)}</ListItem>
+        </div>
+      )),
+    data && data.alert.data.pageInfo.hasNextPage,
+  )
+}
```
Auchindoun:
```
  return renderList(
    events.map((log, index) => (
        <div key={index}>
          <Divider />
          <ListItem>{renderItems(log.timestamp, log.message)}</ListItem>
        </div>
      )),
    pageInfo.hasNextPage,
  )
}
```

web/src/app/alerts/components/AlertDetailLogs.js
```diff
@@ -0,0 +1,144 @@
...
+import { useQuery } from '@apollo/react-hooks'
+import { DateTime, Interval } from 'luxon'
+import { POLL_INTERVAL } from '../../config'
+
+const LIMIT = 149
+
+const query = gql`
+  query getAlert($id: Int!, $input: AlertRecentEventsOptions) {
+    alert(id: $id) {
+      data: recentEvents(input: $input) {
```
Auchindoun: It would probably make more sense to just have this be `recentEvents` without the `data` alias, so that the `_.get` variables are more understandable later on

e.g.
`const events = _.get(data, 'alert.data.nodes', [])`
->
`const events = _.get(data, 'alert.recentEvents.nodes', [])`

web/src/app/alerts/components/AlertDetailLogs.js
```diff
@@ -0,0 +1,137 @@
...
+        }
+      }
+    }
+  }
+`
+export default function AlertDetailLogs(props) {
+  const [poll, setPoll] = useState(POLL_INTERVAL)
+  const { data, error, loading, fetchMore } = useQuery(query, {
+    pollInterval: poll,
+    variables: { id: props.alertID, input: { limit: 35 } },
```
dctalbot: Should probably move magic number into a local const, and rename `LIMIT` to be more specific so we can distinguish between the two e.g. `QUERY_LIMIT` and `FETCH_MORE_LIMIT`

web/src/app/alerts/components/AlertDetailLogs.js
```diff
@@ -0,0 +1,137 @@
...
+        />
+      )
+    return (
+      <ListItemText
+        primary={logTimeFormat(timestamp, DateTime.local().startOf('day'))}
+        secondary={message}
+      />
+    )
+  }
+  if (events === 0) {
```
dctalbot: did you mean `events.length` ? If `events` is an array, this condition will never be true

KatieMSB: yea good catch

PR #182, opened 2019-11-08
ui/overrides: show errors on user conflict

Auchindoun:

**Description:**
There was a bug with showing errors when creating a user override that would conflict with another user on the schedule. This was because we were not checking if the query was still loading before it created the errors array, so the query's data object was `undefined`.

**Additional Info:**
Converted `ScheduleOverrideForm` class component to a functional component for use of the relevant hooks.


web/src/app/schedules/ScheduleOverrideForm.js
```diff
@@ -38,163 +39,159 @@ const query = gql`
...
   },
 })
 
-@connect(state => ({ zone: urlParamSelector(state)('tz', 'local') }))
-@withStyles(styles)
-export default class ScheduleOverrideForm extends React.PureComponent {
-  static propTypes = {
-    scheduleID: p.string.isRequired,
+export default function ScheduleOverrideForm(props) {
+  const { add, remove, errors, scheduleID, value, ...formProps } = props
```
mastercactapus: Since errors isn't required, we should have `errors=[]` so that it defaults to an empty list

web/src/app/schedules/ScheduleOverrideForm.js
```diff
@@ -38,163 +39,159 @@ const query = gql`
...
-      p.shape({
-        field: p.oneOf(['addUserID', 'removeUserID', 'userID', 'start', 'end'])
-          .isRequired,
-        message: p.string.isRequired,
-      }),
-    ),
+  // used to grab conflicting errors from pre-existing overrides
+  const { loading, error, data } = useQuery(query, {
+    variables: {
+      id: userError ? userError.details.CONFLICTING_ID : '',
```
mastercactapus: Simplify to `_.get(userError, 'details.CONFLICTING_ID', '')`?

web/src/app/schedules/ScheduleOverrideForm.js
```diff
@@ -1,24 +1,25 @@
...
 import { urlParamSelector } from '../selectors'
 import { DateRange, ChevronRight, ChevronLeft } from '@material-ui/icons'
 import { DateTimePicker } from '@material-ui/pickers'
 import { DateTime } from 'luxon'
 import { UserSelect } from '../selection'
-import Query from '../util/Query'
 import gql from 'graphql-tag'
 import { mapOverrideUserError } from './util'
 import DialogContentError from '../dialogs/components/DialogContentError'
+import _ from 'lodash-es/lodash.default'
```
dctalbot: Possibly a nit, but is there a reason we're not doing `import _ from lodash-es` ?

Auchindoun: auto imports 🤦‍♂

web/src/app/schedules/ScheduleOverrideForm.js
```diff
@@ -38,163 +39,162 @@ const query = gql`
...
+            fullWidth
+            component={DateTimePicker}
+            mapValue={value => DateTime.fromISO(value, { zone })}
+            mapOnChangeValue={value => value.toISO()}
+            showTodayButton
+            required
+            name='start'
+            leftArrowIcon={<ChevronLeft />}
+            rightArrowIcon={<ChevronRight />}
+            InputProps={{
```
KatieMSB: Also a nit, but looks like leftArrowIcon and rightArrowIcon are defaulted to the left and right chevron unless otherwise set.

Auchindoun: Sounds good. I didn't actually change any of the JSX, just moved it all around. Looks like the package may have been updated with those defaults somewhere along the line. I'll go ahead and remove the icons

web/src/app/schedules/ScheduleOverrideForm.js
```diff
@@ -1,24 +1,25 @@
...
-import { DateRange, ChevronRight, ChevronLeft } from '@material-ui/icons'
+import { DateRange } from '@material-ui/icons'
 import { DateTimePicker } from '@material-ui/pickers'
 import { DateTime } from 'luxon'
 import { UserSelect } from '../selection'
-import Query from '../util/Query'
 import gql from 'graphql-tag'
 import { mapOverrideUserError } from './util'
 import DialogContentError from '../dialogs/components/DialogContentError'
+import _ from 'lodash-es/lodash.default'
```
mastercactapus: We should fix it so they are all the same -- it will make searching and the transition from `lodash-es` to `lodash` easier too.

web/src/app/schedules/ScheduleOverrideForm.js
```diff
@@ -38,163 +39,158 @@ const query = gql`
...
-    value: p.shape({
-      addUserID: p.string.isRequired,
-      removeUserID: p.string.isRequired,
-      start: p.string.isRequired,
-      end: p.string.isRequired,
-    }).isRequired,
+  const classes = useStyles()
+  const params = useSelector(urlParamSelector)
+  const zone = params('tz', 'local')
+  const [userErrors, setUserErrors] = useState([])
```
mastercactapus: Does this need to be stateful or can we just replace `userErrors` with a function that uses `errors`?

It feels like we're duplicating data here and trying to keep it in sync.

web/src/app/schedules/ScheduleOverrideForm.js
Auchindoun: Yep, IDE auto-imports strike again, thanks for catching

web/src/app/schedules/ScheduleOverrideForm.js
Auchindoun: It's an interesting case because we have the standard `errors` variable from the mutation, but have to create the user conflict errors array based off the query if the mutation fails with a `userID` fieldName in question. I'll update the name though so it's more clear what it's being used for

PR #183, opened 2019-11-12
dev: install eslint-plugin-react-hooks

dctalbot:

**Description:**
As we move towards React hooks, we should enforce the "rules of hooks" with `eslint-plugin-react-hooks`


web/src/app/users/UserContactMethodVerificationForm.js
```diff
@@ -52,7 +52,7 @@ export default function UserContactMethodVerificationForm(props) {
   // We only want to display an error in response to a user action.
   useEffect(() => {
     sendCode().catch(() => {})
-  }, [])
+  }, [sendCode])
```
mastercactapus: Won't this cause it to trigger every render (new function from `useMutation`)? We only want it to trigger once on mount

web/src/app/util/Search.js
```diff
@@ -73,12 +72,14 @@ export default function Search(props) {
...
   useEffect(() => {
+    const setSearchParam = value => dispatch(setURLParam('search', value))
+
     const t = setTimeout(() => {
       setSearchParam(search)
     }, DEBOUNCE_DELAY)
 
     return () => clearTimeout(t)
-  }, [search])
+  }, [dispatch, search])
```
mastercactapus: We only want this to trigger on `search` change, not when the dispatch function changes

web/src/app/util/avatar/types.js
```diff
@@ -22,7 +22,7 @@ function useValidImage(srcURL) {
   return valid
 }
 
-function renderAvatar(Fallback, otherProps, imgSrc) {
```
mastercactapus: As this is a render function (and not a Component) shouldn't it stay as `renderAvatar`?

mastercactapus: Future context: the new rule wants this to either be a component or hook since it uses a hook itself; so we changed it to `useAvatar` to be named as a hook.

PR #184, opened 2019-11-12
config/sms: add option to disable link to GoAlert

Auchindoun:

**Description:**
This PR introduces a new config value that allows an admin to disable GoAlert links being rendered in the SMS body of an alert message. This is useful in times where carriers may be blocking messages to end users.


smoketest/twiliosmsurl_test.go
```diff
@@ -0,0 +1,111 @@
...
+		tw := h.Twilio()
+		d1 := tw.Device(h.Phone("1"))
+
+		h.SetConfigValue("General.ShortURL", shortURL)
+		h.SetConfigValue("General.DisableSMSLinks", "true")
+
+		h.CreateAlert(h.UUID("sid"), "test")
+		smsMsg := d1.ExpectSMS("test")
+		tw.WaitAndAssert()
+		if strings.Contains(smsMsg.Body(), shortURL) {
```
mastercactapus: Let's use the assertion library, that way if it fails we'll get more descriptive errors

add import `github.com/stretchr/testify/assert`

And replace with the line: `assert.Contains(t, shortURL, smsMsg.Body())`

It will output something like `expected "foobar baz message body" to contain "http://sho.rt"`

smoketest/twiliosmsurl_test.go
```diff
@@ -0,0 +1,111 @@
...
+		tw := h.Twilio()
+		d1 := tw.Device(h.Phone("1"))
+
+		longURL := h.URL()
+		h.SetConfigValue("General.DisableSMSLinks", "true")
+
+		h.CreateAlert(h.UUID("sid"), "test")
+		smsMsg := d1.ExpectSMS("test")
+		tw.WaitAndAssert()
+		if strings.Contains(smsMsg.Body(), longURL) {
```
mastercactapus: Same as above

smoketest/twiliosmsurl_test.go
Auchindoun: I did try this, but when assert fails it immediately fails the test instead of evaluating true/false for the if statement.

Would something like `assert.NotContainsf(t, smsMsg.Body(), shortURL, "link contains short url")` meet our needs (showing as error msg required to me)?

smoketest/twiliosmsurl_test.go
```diff
@@ -0,0 +1,107 @@
...
+		tw := h.Twilio()
+		d1 := tw.Device(h.Phone("1"))
+
+		h.SetConfigValue("General.ShortURL", shortURL)
+		h.SetConfigValue("General.DisableSMSLinks", "true")
+
+		h.CreateAlert(h.UUID("sid"), "test")
+		smsMsg := d1.ExpectSMS("test")
+		tw.WaitAndAssert()
+		assert.NotContains(t, smsMsg.Body(), shortURL)
```
mastercactapus: On second thought, we should probably just test that it doesn't contain any URL at all when `DisableSMSLinks` is set.

```suggestion
		assert.NotContains(t, smsMsg.Body(), "http")
```

smoketest/twiliosmsurl_test.go
```diff
@@ -0,0 +1,107 @@
...
+		tw := h.Twilio()
+		d1 := tw.Device(h.Phone("1"))
+
+		longURL := h.URL()
+		h.SetConfigValue("General.DisableSMSLinks", "true")
+
+		h.CreateAlert(h.UUID("sid"), "test")
+		smsMsg := d1.ExpectSMS("test")
+		tw.WaitAndAssert()
+		assert.NotContains(t, smsMsg.Body(), longURL)
```
mastercactapus: same

```suggestion
		assert.NotContains(t, smsMsg.Body(), "http")
```

config/config.go
```diff
@@ -25,6 +25,7 @@ type Config struct {
 		DisableLabelCreation   bool   `public:"true" info:"Disables the ability to create new labels for services."`
 		MessageBundles         bool   `public:"true" info:"Enables bundling status updates and alert notifications. Also allows 'ack/close all' responses to bundled alerts."`
 		ShortURL               string `public:"true" info:"If set, messages will contain a shorter URL using this as a prefix (e.g. http://example.com). It should point to GoAlert and can be the same as the PublicURL."`
+		DisableSMSLinks        bool   `public:"true" info:"If set, sms messages will not contain a URL pointing to GoAlert."`
```
dctalbot: Nit: sms -> SMS

Auchindoun:
```suggestion
		DisableSMSLinks        bool   `public:"true" info:"If set, SMS messages will not contain a URL pointing to GoAlert."`
```

PR #190, opened 2019-11-18
dev: add cypress-plugin-retries

mastercactapus:

**Description:**
This PR adds the `cypress-plugin-retries` module configured for 2-retries when running tests.

When a Cypress test fails an additional 2 attempts will be made (3 total) after running `beforeEach`/`afterEach` hooks (but not `before` -- meaning db reset, for instance, will not be performed).

This is in an effort to reduce false-positives for CI test failures.

Failed attempts are still logged/recorded.


web/src/cypress/support/index.ts
```diff
@@ -14,6 +14,8 @@
 // ***********************************************************
 
 // Import commands.js using ES2015 syntax:
+import 'cypress-plugin-retries'
+Cypress.env('RETRIES', 2)
```
Auchindoun: is the `command.js` comment relevant/could be updated?

mastercactapus: Old comment from the original Cypress sample.ts. I'll remove it

PR #199, opened 2019-11-20
go: update mpb and indirect deps

mastercactapus:

**Description:**
This PR updates the mpb package and indirect deps


switchover/dbsync/shell.go
```diff
@@ -169,7 +169,7 @@ func RunShell(oldURL, newURL string) error {
 				mpb.BarClearOnComplete(),
 				mpb.PrependDecorators(
 					decor.OnComplete(
-						decor.StaticName("Truncating tables..."),
+						decor.Name("Truncating tables..."),
```
Auchindoun: I'm not seeing an API change from `StaticName` -> `Name`, is this a breaking change?

I'm looking at https://github.com/vbauerster/mpb/releases

mastercactapus: Yeah, breaking change and they didn't document it from what I could tell. `StaticName` was just calling `Name` though so the behavior is identical.

https://github.com/vbauerster/mpb/commit/8cb26f5221daed4cf19948e04c3450791875300c

PR #200, opened 2019-11-20
graphql: update gqlgen to v0.10.1

mastercactapus:

**Description:**
This PR updates `github.com/99designs/gqlgen` to the latest version.

- Embedded structs are now always pointers
- The `alert/log.Entry` interface is now a struct, since it wasn't being used as an interface, to workaround https://github.com/99designs/gqlgen/issues/920
- Set `Referrer-Policy` header to prevent sending referrer for external assets


alert/log/rawentry.go
```diff
@@ -5,11 +5,12 @@ import (
 	"database/sql"
```
Auchindoun: Worth renaming the file from `rawentry.go` -> `entry.go`?

mastercactapus: yes, missed that

PR #204, opened 2019-11-21
ui: update @material-ui/lab and convert SpeedDial to hooks

dctalbot:

**Description:**
Upgrades `@material-ui/lab` and converts Speed Dial from class component to functional using hooks.

**Screenshots:**
Screen Shot 2019-11-22 at 11 28 43 AM

Speed dial component has new styles as seen in image above (larger font, white background). These are applied when displaying the tooltips persists on touch e.g. on mobile, user will tap speed dial which will remain open until further action is taken.

**Additional Info:**
The tooltip label styles no longer match those of the "fab" buttons.


web/src/app/util/SpeedDial.js
```diff
@@ -1,74 +1,61 @@
...
+      style={{
+        position: 'fixed',
+        bottom: '2em',
+        right: '2em',
+        zIndex: 9001,
+      }}
+    >
+      {props.actions
+        .slice() // TODO why this?
+        .reverse() // TODO why this? If we cut this line, we have to reorder the action arrays everywhere else (3 instances)
```
Auchindoun: Not sure why we are reversing, but I imagine the slice is so we aren't modifying the original array when we reverse. Can probably delete both, but we'll have to update the order where we're using this component

web/src/app/schedules/ScheduleNewOverrideFAB.js
```diff
@@ -9,6 +9,11 @@ export default class ScheduleNewOverrideFAB extends React.PureComponent {
   }
 
   actions = [
+    {
```
mastercactapus: Why was this moved?

web/src/app/schedules/ScheduleRuleList.js
```diff
@@ -119,16 +119,16 @@ export default class ScheduleRuleList extends React.PureComponent {
...
-                label: 'Add Rotation',
-                onClick: () => this.setState({ createType: 'rotation' }),
-                icon: <AccountMultiplePlus />,
-              },
               {
                 label: 'Add User',
                 onClick: () => this.setState({ createType: 'user' }),
                 icon: <AccountPlus />,
               },
+              {
```
mastercactapus: Same, why are we changing menu order?

web/src/app/users/UserDetails.js
```diff
@@ -86,16 +86,16 @@ export default function UserDetails(props) {
...
-              label: 'Add Contact Method',
-              icon: <SettingsPhone />,
-              onClick: () => setCreateCM(true),
-            },
             {
               label: 'Add Notification Rule',
               icon: <AddAlarm />,
               onClick: () => setCreateNR(true),
             },
+            {
```
mastercactapus: Same?

web/src/cypress/support/page-fab.ts
```diff
@@ -18,7 +18,8 @@ function pageFab(dialOption?: string): Cypress.Chainable {
 
   return res
     .parent()
-    .find(`button[role=menuitem][aria-label*=${JSON.stringify(dialOption)}]`)
+    .find(`span[aria-label*=${JSON.stringify(dialOption)}]`)
+    .find('button[role=menuitem]')
```
mastercactapus: Let's join these into a single selector so retry works correctly

web/src/app/util/SpeedDial.js
```diff
@@ -1,74 +1,67 @@
...
+      }}
+      icon={<SpeedDialIcon />}
+      onClick={doToggle}
+      onClose={doClose}
+      onMouseEnter={doOpen}
+      onMouseLeave={doClose}
+      open={open}
+      className={classes.speedDial}
+    >
+      {props.actions.map((action, idx) => (
```
mastercactapus: IIRC the speed dial menu displays things in reverse order, which is why we were switching them here. 

That way the prop-order matches display-order when using the `SpeedDial` component

web/src/app/schedules/ScheduleNewOverrideFAB.js
dctalbot: See conversation above^

We were previously reversing the actions array

web/src/app/util/SpeedDial.js
dctalbot: I'd prefer using the component's default behavior, but can revert 41cd2fe if you want

web/src/app/util/SpeedDial.js
```diff
@@ -1,74 +1,70 @@
...
+  speedDial: {
+    position: 'fixed',
+    bottom: '2em',
+    right: '2em',
+    zIndex: 9001,
+  },
+  staticTooltipLabel: {
+    whiteSpace: 'nowrap',
+  },
+}))
```
Auchindoun: we can leave out `theme` if we aren't using it and just pass the object, the first param supports objects as well as functions

web/src/app/util/SpeedDial.js
```diff
@@ -1,74 +1,70 @@
...
-  state = {
-    open: false,
-  }
+export default function CustomSpeedDial(props) {
+  const [open, setOpen] = useState(false)
+  const classes = useStyles()
 
-  shownState = false
-  _shownTimeout = -1
+  const doToggle = () => setOpen(!open)
```
Auchindoun: let's use the `handle` naming convention set by `jsx-handler-names`, where we name `onEvent` funcs to `handleNounVerb`, i.e. `handleDialOpen` or similar

web/src/app/util/SpeedDial.js
```diff
@@ -1,74 +1,70 @@
...
+  return (
+    <SpeedDial
+      ariaLabel={props.label}
+      FabProps={{
+        'data-cy': 'page-fab',
+      }}
+      icon={<SpeedDialIcon />}
+      onClick={handleToggle}
+      onClose={handleClose}
+      onMouseEnter={handleOpen}
```
mastercactapus: I think we should be naming event handlers after the event, or just inlining the call. E.g. `handleClick` or `handleSpeedDialClick` (if there were multiple).

If we don't need specific event handlers (e.g. it's a simple function) we can just inline wth `() => setOpen(true)` instead.

Otherwise, I feel like the extra step is just adding noise.

mastercactapus: @Auchindoun thoughts?

Auchindoun: Getting around the eslint rule by using `() =>` seems like a way for us to slowly start avoiding the rule over time, which may lead to us just disabling it in the future so I would lean towards the current naming convention.

If we're only calling the set state hook function though, which is what's happening here, then I can see how it'd be better to just inline func it.

If it's using a function we've made that we're referencing though it should probably start with `handle`.

dctalbot: This component is small and simple enough that I'll go ahead and inline the function calls

PR #232, opened 2019-11-26
refetch/eps: convert to hooks

Auchindoun:

**Description:**
This PR converts the escalation policy router and details page to hooks.

Had to re-open on a new branch because I'm bad


web/src/app/escalation-policies/PolicyDetails.js
```diff
@@ -25,88 +27,83 @@ const query = gql`
...
+  const [showDeleteDialog, setShowDeleteDialog] = useState(false)
+  const [showEditDialog, setShowEditDialog] = useState(false)
+
+  const { loading, error, data: _data } = useQuery(query, {
+    variables: {
+      id: props.escalationPolicyID,
+    },
+  })
+
+  const data = _.get(_data, 'escalationPolicy', null)
```
dctalbot: do we get optional chaining with the new babel? we should be able to do 
`const data = _data?.escalationPolicy`

mastercactapus: We still rely on the default value functionality of `_.get` in quite a few places, I'd say stick with it just for consistency for the time being.

If it is an option, and we do switch, we should update everything at once with code shift (or similar tool), but not as part of this work.

PR #243, opened 2019-11-27
admin: provide URLs and copy links from the admin page

mastercactapus:

**Description:**
This PR adds copy-friendly links to config sections on the admin page.

image

The links are generated on the backend using the same logic and config that would be sent to external systems (like Twilio). It was built using the same tools used for the current config so the UI will pickup new "hints" without updates.

Additionally, the Public URL now has it's "public" flag set.


web/src/app/admin/AdminConfig.js
```diff
@@ -127,6 +137,26 @@ export default class AdminConfig extends React.PureComponent {
...
+                          readOnly
+                          label={hintName(h.id)}
+                          value={h.value}
+                          variant='filled'
+                          margin='none'
+                          InputProps={{
+                            margin: 'dense',
+                            endAdornment: (
+                              <InputAdornment position='end'>
+                                <CopyText value={h.value} />
```
dctalbot:
```suggestion
                                <CopyText value={h.value} placement='left' />
```

mastercactapus:
We can't count on them always being URLs, we should also avoid using terms like "integration" here since we already use that for a separate purpose elsewhere.

I used "hints" because it's general and use-cases can vary, meaning you might use/change the value in the external system in some cases. Open to other suggestions though. `#naming-things`

web/src/app/admin/AdminConfig.js
```diff
@@ -127,6 +137,26 @@ export default class AdminConfig extends React.PureComponent {
...
+                      hintGroups[groupID].map(h => (
+                        <TextField
+                          key={h.id}
+                          readOnly
+                          label={hintName(h.id)}
+                          value={h.value}
+                          variant='filled'
+                          margin='none'
+                          InputProps={{
+                            margin: 'dense',
```
Auchindoun: Can we remove dense here? It looks very cluttered with this set, especially when multiple links are displayed.

mastercactapus:
@Auchindoun They are read-only and non-editable. Displaying the link, instead of just a copy label like integration keys, so that it's easy to copy part of a URL and see the full structure based on what they need to achieve.

PR #250, opened 2019-12-03
ui: Route current user's details page to their editable profile

dctalbot:

**Description:**
Users were confused as to why they couldn't edit their profile from their own User Details page. ~~This change redirects them to /profile to avoid this confusion.~~
This change routes them directly to an editable profile view while preserving the `goalert.com/users/:id` url location. This way, we let users share links to their profiles (for tech support or whatever else).

Users will no longer have a read-only view of their User Details page. Viewing other users' User Details pages will remain as-is.


web/src/app/users/UserRouter.js
```diff
@@ -56,34 +52,38 @@ function UserOnCallAssignments() {
...
+
+  return (
+    <Switch>
+      <Route exact path='/users' component={UserList} />
+      <Route
+        exact
+        path='/users/:userID'
+        render={({ match }) =>
+          match.params.userID === userID ? (
+            <Redirect to='/profile' />
```
mastercactapus: Let's see if we can handle this in `routes.js` -- we have all of our other redirects there

mastercactapus:
Please use descriptive branch names for future PRs

web/src/app/users/UserRouter.js
dctalbot: I took a look and I think it's best as-is. This redirect is scoped to `/users/:userID` and having it in the `UserRouter` where we explicitly define `path='/users/:userID'` is most reasonable to me. Feel free to push back, just not yet clear how moving this would improve maintainability

mastercactapus: We put all static redirects in the routes file so that all (non statefull) app routing happens in one place. It also means it doesn't have to render the tree before going back up for a redirect that we already know about. We aren't doing dynamic loading atm/anymore so it's not as noticeable from performance though.

If putting it in routes ends up being too complicated, I'd be ok keeping it in UserRouter for now, but then let's switch it inro a static from/to redirect in the router, instead of mixing it in with the details page render function.

dctalbot: I tried moving into `routes.js` but there's a strange bug. `NewApp` is calling `renderRoutes` as a function, but we'd need to convert it to a functional component is we want to use the `useSessionInfo()` hook. Sounds simple enough, but when we do that, it immediately redirects to the last route component of the array that `renderRoues` returns. The WIP is on `wip-goal-436`

arurao:
Seems like there is something weird with pagination for users query in graphql. I can take a look at it.

web/src/app/users/UserRouter.js
dctalbot: Turns out it's fairly simple to just route directly to the component w/o a `<Redirect />` :
```
<Route
        exact
        path={['/profile', `/users/${userID}`]}
        component={UserProfile}
      />
```


I'm going to move forward with this approach, but the bug below re pagination is still relevant.

PR #251, opened 2019-12-03
ui/profile: prevent adding a notification rule if there are no contact methods

KatieMSB:

**Description:**
'Add Notification Rule' is removed from the speed dial actions on the profile page if the user has no Contact Methods.


web/src/app/users/UserDetails.js
```diff
@@ -79,25 +82,24 @@ export default function UserDetails(props) {
...
   const svcCount = serviceCount(user.onCallSteps)
+  const dialActions = [
+    {
+      label: 'Add Contact Method',
+      icon: <SettingsPhone />,
+      onClick: () => setCreateCM(true),
+    },
+  ]
+  if (user.contactMethods.length)
+    dialActions.push({
```
mastercactapus: Let's disable this instead of hiding, so what the user sees is consistent.

When a new user goes to the profile for the first time, they won't get an indication that the speed dial options changed in the background.

web/src/app/users/UserDetails.js
```diff
@@ -80,6 +83,9 @@ export default function UserDetails(props) {
   const user = _.get(data, 'user')
   const svcCount = serviceCount(user.onCallSteps)
 
+  const iconColor = user.contactMethods.length ? 'primary' : 'disabled'
```
mastercactapus: Instead of changing the icon, let's just disable the individual fab (see below)

web/src/app/util/SpeedDial.js
```diff
@@ -45,8 +45,9 @@ export default function CustomSpeedDial(props) {
             classes={{ staticTooltipLabel: classes.staticTooltipLabel }}
             aria-label={action.label}
             onClick={() => {
-              setOpen(false)
-              action.onClick()
+              if (!action.disabled) {
```
mastercactapus: Let's invert this to get rid of the double negative, and so the primary behavior isn't nested in a block.

We can also make this work a little better:
- If we call `stopPropagation`, it will prevent the SpeedDial from closing when clicking the disabled item
- We can add `TooltipClasses` to "disable" the tooltip text [info here](https://material-ui.com/api/speed-dial-action/#props)
- Add `FabProps={{disabled: action.disabled}}` to use the built-in disable styling for the button [from here](https://material-ui.com/api/fab/#props) instead of manually changing the icon color

// click handler
```js
onClick={e => {
  if (action.disabled) {
    e.stopPropagation()
    return
  }
  action.onClick()
}
```

PR #260, opened 2019-12-09
twilio : make voice verification calls more user-friendly

arurao:

**Description:**
This PR improves the user voice call messages. 
For verification codes, it repeats the code's digits two times, with a delay in between every digit, making it more understandable.
For all voice calls, it introduces a timeout of 10 seconds (versus a default of 5 seconds), which gives the user more time to select their option from the keypad.

### DIFF
--- a/smoketest/twiliovoiceverification_test.go
+++ b/smoketest/twiliovoiceverification_test.go
@@ -77,10 +77,9 @@ func TestTwilioVoiceVerification(t *testing.T) {
 
 	// Since verification code is said twice during one Twilio message
 	c := splitCode(codeStr, 6)
-	c1 := strings.Join(c, "")
 
 	// string to int conversion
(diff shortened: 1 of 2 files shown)


smoketest/twiliovoiceverification_test.go
```diff
@@ -74,7 +75,11 @@ func TestTwilioVoiceVerification(t *testing.T) {
 		return -1
 	}, msg.Body())
 
-	code, _ := strconv.Atoi(codeStr)
+	// Since verification code is said twice during one Twilio message
+	c := splitCode(codeStr, 6)
```
mastercactapus: We can simplify this using slice notation, we should probably assert on the length though.

Something like:
```go
assert.Len(t, codeStr, 12)
code, _ := strconv.Atoi(codeStr[:6])
```

PR #261, opened 2019-12-09
ui: allow typing in date and time fields

### LINKED ISSUE #24: ui: changing a date value within the new override form resets the date after a new poll
**Describe the Bug:**
When creating a new override from the shifts list on a schedule's details page, when either the start or end date is changed it will reset after new data is fetched as a result of the page polling. This bug does not occur from the overrides list, or the calendar.

**Steps to Reproduce:**
1. Go to a schedule
2. Click on Shifts
3. Hover over the page's create fab in the bottom right
4. Click on any create override button
5. Click on start or end date from within the dialog
6. Select a new date
7. Within 3 seconds the selected date will reset to the initial form's value

**Expected Behavior:**
Data that is changed from within the override form should persist.

**Screenshots/Stack Traces:**

<a href="https://gyazo.com/1d1abe53cf3071e0885de3eab73d0fe8">Image from Gyazo</a>

**Application Version:**
`v0.22.0`

**Desktop:**
 - OS: MacOS Mojave
 - Browser: Chrome
 - Version: Chrome v75.0.3770.100

Auchindoun:

**Description:**
This PR enables the use of typing within the date and time fields throughout the app, through the use of our pre-existing package `@material-ui/pickers`. The relevant fields have been updated to be `luxon` `DateTime` objects while editing to avoid parsing errors when converting to an ISO string after each change. This is compensated by the fact that invalid dates will now be validated on the client before formatted as an ISO string and sent to the backend.

The following forms have been updated:
- ScheduleOverrideForm (date and time)
- WizardScheduleForm (date and time)
- ScheduleShiftList (date)
- RotationForm (time)
- ScheduleRuleForm (time)

**Additional Info:**
Inadvertently closes #24


web/src/app/schedules/ScheduleRuleForm.js
```diff
@@ -185,6 +167,15 @@ export default class ScheduleRuleForm extends React.PureComponent {
             <ScheduleTZFilter
               label={tz => `Configure in ${tz}`}
               scheduleID={this.props.scheduleID}
+              onChange={() => {
+                // update value in form container with new TZ
+                onChange({
```
mastercactapus: Same as above, we shouldn't rely on ScheduleTZFilter change to keep things in sync, the `tz` param is already the source of truth.

web/src/app/schedules/ScheduleRuleForm.js
```diff
@@ -244,35 +237,50 @@ export default class ScheduleRuleForm extends React.PureComponent {
...
 
-  renderRuleField(idx) {
-    const parseValue = value => parseClock(value, this.props.zone)
+  renderRuleField(r, idx) {
     const classes = this.props.classes
     return (
       <TableRow key={idx}>
         <TableCell className={classes.startEnd}>
           <FormField
+            data-cy='start-time'
```
mastercactapus: Use the input field

web/src/app/schedules/ScheduleShiftList.js
```diff
@@ -152,9 +161,14 @@ export default class ScheduleShiftList extends React.PureComponent {
...
-  state = {
-    create: null,
-    specifyDuration: false,
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      create: null,
+      specifyDuration: false,
+      startDate: DateTime.fromISO(props.start, { zone: props.zone }) || null,
```
mastercactapus: Why are we duplicating `start` into state?

web/src/app/schedules/ScheduleShiftList.js
```diff
@@ -328,12 +342,21 @@ export default class ScheduleShiftList extends React.PureComponent {
...
                 label='Start Date'
-                value={DateTime.fromISO(this.props.start, { zone })}
-                onChange={e => this.props.handleSetStart(e.toISO())}
+                format='MM/dd/yyyy'
+                mask='__/__/____'
+                placeholder={DateTime.local().toFormat('MM/dd/yyyy')}
+                value={this.state.startDate}
+                onChange={date => {
+                  if (date && date.isValid) {
+                    this.props.handleSetStart(date.toISO())
```
mastercactapus: Should this be DateTime?

web/src/app/schedules/ScheduleTZFilter.js
```diff
@@ -41,7 +41,13 @@ export function ScheduleTZFilter(props) {
       control={
         <Switch
           checked={zone !== 'local'}
-          onChange={e => setZone(e.target.checked ? tz : 'local')}
+          onChange={e => {
+            const newZone = e.target.checked ? tz : 'local'
+            setZone(newZone)
+            if (typeof props.onChange === 'function') {
+              return props.onChange(newZone)
```
mastercactapus: Shouldn't need this, we should always be reading from the `tz` prop

web/src/app/wizard/WizardScheduleForm.js
```diff
@@ -198,6 +199,18 @@ export default class WizardScheduleForm extends React.Component {
                     </InputAdornment>
                   ),
                 }}
+                // keyboard picker props
+                format='MM/dd/yyyy, hh:mm a'
```
mastercactapus: Since we're setting this everywhere, and all the others, should we just be wrapping the picker instead?

Seems like it would be easy for all these pickers to get out-of-sync (e.g. if we change the mask)

Also, is there a way we can make this international friendly? IIRC the US is the only place that does MDY

web/src/app/schedules/ScheduleOverrideForm.js
```diff
@@ -139,14 +147,25 @@ export default function ScheduleOverrideForm(props) {
                 </InputAdornment>
               ),
             }}
+            // keyboard picker props
+            format='MM/dd/yyyy, hh:mm a'
+            mask='__/__/____, __:__ _M'
+            placeholder={DateTime.local()
```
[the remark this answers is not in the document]
Auchindoun: Yes, since the user can backspace and make the field empty

web/src/app/schedules/ScheduleOverrideForm.js
```diff
@@ -97,6 +98,14 @@ export default function ScheduleOverrideForm(props) {
           <ScheduleTZFilter
             label={tz => `Configure in ${tz}`}
             scheduleID={scheduleID}
+            onChange={newZone => {
+              // update value in form container with new TZ
+              onChange({
+                ...value,
+                start: value.start.setZone(newZone),
```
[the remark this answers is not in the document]
Auchindoun: It was originally mapping the value on every change, but we cannot do this as it would change the input every time the user inputs into the field (i.e. CST -> EST toggle, if the user types in '8' in the hour field, '9' would appear instead -- this is unwanted behavior). We need to update the zone on the value only once, when the toggle is flipped

web/src/app/rotations/RotationForm.js
```diff
@@ -17,7 +18,7 @@ export default class RotationForm extends React.PureComponent {
       timeZone: p.string.isRequired,
       type: p.oneOf(rotationTypes).isRequired,
       shiftLength: p.number.isRequired,
-      start: p.string.isRequired,
+      start: p.instanceOf(DateTime),
```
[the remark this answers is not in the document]
Auchindoun: This cannot be required as it can be null if the values are empty (user backspaces). Since all of the fields are required, we won't get into the instance where it breaks, as the required bit is checked before it gets to the validate func. I'll add checks on the validate functions where appropriate though as a sanity check.

web/src/app/schedules/ScheduleOverrideCreateDialog.js
```diff
@@ -41,8 +41,8 @@ export default class ScheduleOverrideCreateDialog extends React.PureComponent {
     defaultValue: p.shape({
       addUserID: p.string,
       removeUserID: p.string,
-      start: p.string,
-      end: p.string,
+      start: p.instanceOf(DateTime),
```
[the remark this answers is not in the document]
Auchindoun: see above

web/src/app/schedules/ScheduleRuleForm.js
Auchindoun: see above

web/src/app/schedules/ScheduleShiftList.js
Auchindoun: Since the date can be invalid while they are typing, we don't want to store an invalid timestamp into the URL. It will only update the URL with an ISO timestamp once a valid date is entered into the form field, thus, an invalid date can be stored in state, but only valid dates will be in the URL.

web/src/app/schedules/ScheduleShiftList.js
Auchindoun: This is the ISO timestamp stored in the URL, if the date typed in is valid.

web/src/app/wizard/WizardScheduleForm.js
Auchindoun: We could probably put effort into making it dynamically change the format depending on where they live, but with the placeholder and defaults stored in the field I don't think it's necessary/worth the time here to try and do something like this. Could be a nice improvement later on down the road though.

What do you mean by "wrapping the picker" with a format string, though? I'm not sure I fully understand. I can use an exported const for the date-time format, similar to our poll intervals.

web/src/app/schedules/ScheduleTZFilter.js
Auchindoun: This change will be dependent on the [above conversation](https://github.com/target/goalert/pull/261#discussion_r357352326) ^

web/src/app/wizard/WizardScheduleForm.js
mastercactapus: Currently we are displaying dates as something like `December 12` for pickers, which side-steps the issue since the month is spelled out.

Changing the display format to one that isn't correct for the current locale will cause confusion -- `01/02/2020` does not mean the same thing in all countries, so hard-coding it is problematic (Feb 1st or Jan 2nd) because we need to support multiple time-zones/countries.

Wrapping it is because we're passing the same mask, placeholder, validation, etc.. in a lot of places -- if the defaults of the keyboard picker aren't usable, we should wrap it so we don't need to copy/paste and keep-in-sync all those options.

web/src/app/schedules/ScheduleShiftList.js
mastercactapus: Would it be possible to store this in a wrapper component instead? Said component would only trigger `onChange` when it changes to a new _valid_ value, instead of storing mixed state everywhere else.

web/src/app/forms/FormField.js
```diff
@@ -175,6 +175,11 @@ export class FormField extends React.PureComponent {
     if (render) return render(props)
     const Component = component
 
+    let helperCy = null
+    if (this.props['data-cy']) {
+      helperCy = this.props['data-cy'] + '-form-helper'
```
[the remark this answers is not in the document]
Auchindoun: How do you see this working? The input is several children below the form helper, and it's not a direct descendant.

we have:
```
div
> div
>> input - form field
p - form helper
```

which would mean we would need to do something like `cy.get('input[name="start"]').parent().parent().siblings()` to grab this, which seems way less convenient and more fragile than the current solution

web/src/app/schedules/ScheduleShiftList.js
Auchindoun: Do we need to do that though? The state is isolated to this component, and will update the URL as soon as it gets a valid date

Auchindoun:
Closed as @mastercactapus is reworking the solution.

PR #263, opened 2019-12-10
ui/profile: added '+' as input adornment to phone field for contact method

KatieMSB:

**Description:**
Added input adornment for phone number field when adding a new contact method.

### DIFF
--- a/web/src/app/users/UserContactMethodForm.js
+++ b/web/src/app/users/UserContactMethodForm.js
@@ -34,19 +34,13 @@ export default class UserContactMethodForm extends React.PureComponent {
     onChange: () => {},
   }
 
-  state = {
-    cmAdornment: true,
-  }
-
   render() {
     const cleanValue = val => {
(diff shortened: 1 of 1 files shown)


web/src/app/users/UserContactMethodForm.js
```diff
@@ -33,14 +34,19 @@ export default class UserContactMethodForm extends React.PureComponent {
...
+  state = {
+    cmAdornment: true,
+  }
+
   render() {
     const cleanValue = val => {
       val = val.replace(/[^0-9]/g, '')
 
       if (!val) {
+        this.setState({ cmAdornment: true })
```
mastercactapus: We should avoid using state to store calculated values, though in this case, I think we can just always render the adornment.

In the `FormField` we can use the `mapValue` prop to strip the `+` from the value since we are rendering our own.

Something like:
```js
mapValue={v => v.replace(/^\+/, '')}
```

PR #277, opened 2019-12-16
cypress: update Cypress to 3.8.1 and refactor where needed

dctalbot:

**Description:**
Updates to Cypress 3.8.1

Content Scrolls below AppBar:
- See `AlertsList`, `NewApp`, `CheckedAlertsFormControl`

Adds form test helpers:
- See `support/dialog.ts`, `support/forms.ts`

Remove legacy `ConfirmationDialog` component:
- See `ServiceAlerts`

All other changes are general test hardening, future-proofing etc.

**Additional Info:**
Making the `<main />` element scrollable broke the `CalendarToolbar` and the `CheckedAlertsFormControl` due to positioning issues. These are addressed in this PR and also #292


web/src/cypress/integration/schedules.ts
```diff
@@ -202,6 +202,16 @@ function testSchedules(screen: ScreenFormat) {
...
 
+    it('should add a user as an assignment', () => {
+      cy.pageFab('User')
+
+      // select create rotation
+      cy.get('input[name=targetID').selectByLabel(rot.users[0].name)
+      cy.get('button')
+        .contains('Submit')
+        .click()
+    })
```
mastercactapus: We should assert that the user assignment was added

web/src/cypress/integration/schedules.ts
```diff
@@ -215,6 +225,34 @@ function testSchedules(screen: ScreenFormat) {
...
+        .click()
+
+      cy.get('tr').should('have.length', 3)
+
+      // select create rotation
+      cy.get('input[name=targetID]').selectByLabel(rot.name)
+      cy.get('button')
+        .contains('Submit')
+        .click()
+    })
```
mastercactapus: Same as above, we should assert that the submit actually takes effect.

web/src/cypress/integration/schedules.ts
```diff
@@ -215,6 +225,34 @@ function testSchedules(screen: ScreenFormat) {
...
+        .contains('Today')
+        .click()
+      cy.get('button')
+        .contains('AM')
+        .click()
+      cy.get('body')
+        .contains('OK')
+        .click()
+
+      cy.get('tr').should('have.length', 3)
```
mastercactapus: This is a bit ambiguous, we should scope it, at least to the dialog.

web/src/cypress/integration/schedules.ts
```diff
@@ -215,6 +225,34 @@ function testSchedules(screen: ScreenFormat) {
       cy.get('body').should('not.contain', rot.name)
     })
 
+    it('should create multiple rules on an assignment', () => {
+      cy.pageFab('Rotation')
+
+      cy.get('input[name=Sunday]').click()
+      cy.get('button[aria-label="Add rule"]').click()
+
+      cy.get('td')
```
mastercactapus: Is there a better/more specific selector we can use? `td` is a bit generic, also this is page-wide and not scoped to the dialog.

web/src/cypress/integration/schedules.ts
```diff
@@ -328,10 +366,10 @@ function testSchedules(screen: ScreenFormat) {
         cy.get('button')
           .contains('Submit')
           .click()
+        cy.get('[role=dialog]').should('not.exist')
```
mastercactapus: We should add this after the submits for the other actions in this test

web/src/cypress/integration/schedules.ts
```diff
@@ -343,14 +381,6 @@ function testSchedules(screen: ScreenFormat) {
         cy.get('body')
           .should('not.contain', users[0].name)
           .should('contain', users[1].name)
-
-        cy.get('button[data-cy=other-actions]').menu('Delete')
```
mastercactapus: How are we handling testing that delete works?

web/src/cypress/integration/schedules.ts
dctalbot: Is `cy.get('body').contains('li', rot.users[0].name)` not sufficient?

web/src/cypress/support/dialog.ts
```diff
@@ -0,0 +1,126 @@
+declare namespace Cypress {
+  interface Chainable<Subject> {
+    /** Click a dialog button with the given text and wait for it to dissapear. */
```
m17ch: dissapear -> disappear

web/src/cypress/integration/escalationPolicySteps.ts
```diff
@@ -37,34 +37,21 @@ function testSteps(screen: ScreenFormat) {
       cy.fixture('users').then(users => {
         const u1 = users[0]
         const u2 = users[1]
+        const del = c.integer({ min: 1, max: 9000 })
```
dctalbot:
```suggestion
        const delay = c.integer({ min: 1, max: 9000 })
```

Confused this with delete at first. I'd rahter be clear and consistent since we call this var `delay` in another test

m17ch: I had the same issue -- appreciate the rename.

web/src/cypress/integration/escalationPolicySteps.ts
```diff
@@ -107,38 +91,28 @@ function testSteps(screen: ScreenFormat) {
           cy.reload()
         })
         .then(() => {
+          const del = c.integer({ min: 1, max: 9000 })
```
dctalbot: same as above

PR #281, opened 2019-12-16
ui: fix refetch behavior on mutations

mastercactapus:
**Description:**
This is a feature PR -- code has been reviewed, only functional validation necessary.

It resolves issues where data would be stale after mutations, as well as issues with deleting items (e.g. 404 instead of redirect to list page). As part of that work details pages have been converted to hooks.

**Additional Info**
- All visible pages/components should update when any change (mutation) is made across the app
- Details pages have been converted to hooks and should load quicker when coming from the list page
- List pages will now poll for updates until/unless the user starts paginating

**Out of Scope**
- Rotation users: no spinner when loading
- EP: steps: no spinner when loading


web/src/app/util/promiseBatch.js
```diff
@@ -0,0 +1,54 @@
...
+    this._active++
+  }
+  this.done = () => {
+    this._active--
+    if (this._active === 0) this._resolve()
+  }
+}
+
+function start() {
+  if (!current) current = new Batch(1)
```
dctalbot:
```suggestion
  if (!current) current = new Batch(0)
```

Looks like an off-by-one bug is causing the crash since you would have unresolved promises building up

PR #283, opened 2019-12-16
ui/QueryList: always display new data

mastercactapus:

**Description:**
This PR updates the `QueryList` component to use the `network-only` fetch policy and enable polling when not paginating through results.

It fixes issues when coming from a details page (e.g. favorite or delete) and having the list be out-of-date.

### DIFF
--- a/web/src/app/lists/QueryList.js
+++ b/web/src/app/lists/QueryList.js
@@ -5,7 +5,7 @@ import { useQuery } from '@apollo/react-hooks'
 import { Grid, makeStyles } from '@material-ui/core'
 import { once } from 'lodash-es'
 import { PaginatedList } from './PaginatedList'
-import { ITEMS_PER_PAGE } from '../config'
+import { ITEMS_PER_PAGE, POLL_INTERVAL } from '../config'
 import { searchSelector, urlKeySelector } from '../selectors'
 import { fieldAlias } from '../util/graphql'
 import Search from '../util/Search'
@@ -18,9 +18,10 @@ const useStyles = makeStyles({
   },
 })
 
-const buildFetchMore = (fetchMore, after) => {
-  return once(num =>
-    fetchMore({
+const buildFetchMore = (fetchMore, after, stopPolling) => {
+  return once(num => {
+    stopPolling()
+    return fetchMore({
       variables: {
         input: {
           first: num,
@@ -38,8 +39,8 @@ const buildFetchMore = (fetchMore, after) => {
           },
         }
       },
-    }),
-  )
+    })
+  })
 }
 
 export default function QueryList(props) {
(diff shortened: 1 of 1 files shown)


web/src/app/lists/QueryList.js
```diff
@@ -75,7 +78,11 @@ export default function QueryList(props) {
   if (data && data.data && data.data.nodes) {
     items = data.data.nodes.map(props.mapDataNode)
     if (data.data.pageInfo.hasNextPage) {
-      loadMore = buildFetchMore(fetchMore, data.data.pageInfo.endCursor)
+      loadMore = buildFetchMore(
+        fetchMore,
+        data.data.pageInfo.endCursor,
+        stopPolling,
```
Auchindoun: Are we covering enabling polling if the user goes back to the first page?

mastercactapus: No, we can't poll once we start paginating since it joins multiple queries together. We also don't have the page info from QueryList (it just keeps adding).

So once they start paginating it stops until they leave/come-back or change their search.

I'm sure we could make it work though if we add a "reset" method of some sort and give it to PaginatedList -- it would have to dump the loaded pages though

Auchindoun: I imagine it's fine and probably preferred to reset previously loaded pages in favor of polling if they go back to the first page. I thought we had implemented this before

mastercactapus: Maybe in the Relay days, though I think we did the same thing, once they hit next polling stops -- it's been so long I can't remember. I'll look at adding it in, if it's simple enough it would make explaining the behavior easier -- polls if on the first page.

mastercactapus: TL;DR; After messing around with this for a while, I don't think it's going to be worth the complexity to make this work.

---

Calling `startPolling` when going back to the first page is easy enough, but it introduces a race condition where if the user goes to any page but the first after that, it will end up in a broken state once the poll interval hits.

This can be mitigated by calling `refetch` at the same time, but the same race still exists based on the completion of that query, rather than the interval. It's just harder to detect.

We also can't just call `stopPolling` again because there could be an in-flight request which would cause the same issue.

To actually fix it, we need to synchronously dump the data beyond the first page before calling `startPolling`, making the data set to match what we had before pagination started.

Unfortunately, even trying to use `readQuery` and `writeQuery` to do that falls short because we need the original `endCursor` and up-to-date `hasNextPage` information.

We would need to ensure that the original `endCursor` never gets out-of-sync, so it would have to be stored by something like `fetchMore` which feels messy, though it may be possible.

Given that the user has to scroll to the bottom of the page to paginate, and it's reset on changing the query at all, I'm thinking it would just be better to leave as-is. If they start paginating, the result set is frozen. Anything else, by default, will auto-update.

Maybe in the future, we can have the paginated data itself refresh rather than switching modes at all or something like that. (e.g. using a before & after query, poll each page instead of aggregating results) which would be a more solid solution.

PR #293, opened 2019-12-18
ui: fix input label shrinking bug

KatieMSB:

**Description:**
Added fix for input labels that would overlap in text fields that had autofilled values.

**Screenshots:**
Screen Shot 2019-11-18 at 2 27 49 PM

**Additional Info**
Link to documentation in material ui for this fix:
https://material-ui.com/components/text-fields/#shrink

### DIFF
--- a/web/src/app/forms/FormField.js
+++ b/web/src/app/forms/FormField.js
@@ -109,10 +109,11 @@ export class FormField extends React.PureComponent {
       ...otherFieldProps
     } = this.props
 
-    const InputLabelProps = {
+    let InputLabelProps = {
       required: required && !optionalLabels,
       ..._inputProps,
     }
+    if (this.props.value) InputLabelProps = { ...InputLabelProps, shrink: true }
 
     const baseLabel = typeof _label === 'string' ? _label : startCase(name)
     const label =
@@ -174,7 +175,6 @@ export class FormField extends React.PureComponent {
 
     if (render) return render(props)
     const Component = component
-    if (props.value) props.InputLabelProps = { shrink: true }
 
     return (
       <FormControl fullWidth={props.fullWidth} error={Boolean(props.error)}>
--- a/web/src/app/forms/FormField.js
+++ b/web/src/app/forms/FormField.js
@@ -109,11 +109,11 @@ export class FormField extends React.PureComponent {
       ...otherFieldProps
     } = this.props
 
-    let InputLabelProps = {
+    const InputLabelProps = {
       required: required && !optionalLabels,
+      shrink: Boolean(this.props.value),
       ..._inputProps,
     }
-    if (this.props.value) InputLabelProps = { ...InputLabelProps, shrink: true }
 
     const baseLabel = typeof _label === 'string' ? _label : startCase(name)
     const label =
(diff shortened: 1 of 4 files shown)


m17ch:
Would you mind linking the material-ui issue that recommends this workaround (in PR message)?

web/src/app/forms/FormField.js
```diff
@@ -109,10 +109,11 @@ export class FormField extends React.PureComponent {
       ...otherFieldProps
     } = this.props
 
-    const InputLabelProps = {
+    let InputLabelProps = {
       required: required && !optionalLabels,
       ..._inputProps,
     }
+    if (this.props.value) InputLabelProps = { ...InputLabelProps, shrink: true }
```
Auchindoun: Instead of rebuilding the object again, we can evaluate in the initial creation:
```
const InputLabelProps = {
  required: required && !optionalLabels,
  shrink: Boolean(this.props.value)
  ..._inputProps,
}
```

web/src/app/selection/MaterialSelect.js
```diff
@@ -47,11 +47,16 @@ export default class MaterialSelect extends Component {
...
       name,
       placeholder,
-      InputLabelProps,
+      InputLabelProps: _InputLabelProps,
 
       ...props
     } = this.props
 
+    const InputLabelProps = {
+      shrink: Boolean(this.props.value),
```
dctalbot:
```suggestion
      shrink: Boolean(this.props.value.length > 0),
```
This will always evaluate to `true` since this.props.value is and Array (`Boolean([]) === true`)

KatieMSB: This was the issue in FormField but looks like MaterialSelect is returning an object for each instance.

web/src/app/forms/FormField.js
```diff
@@ -124,9 +125,10 @@ export class FormField extends React.PureComponent {
       hint,
       value: mapValue(get(value, fieldName)),
     }
+
     const InputLabelProps = {
       required: required && !optionalLabels,
-      shrink: Boolean(props.value),
+      shrink: hasValue(props.value),
```
mastercactapus: Can we use `_.isEmpty` instead of our own?

Also `shrink: false` the not same as not specifying shrink at all: we still want it to handle shrink on focus and whatever other behavior material would follow. We may need to stick with the original implementation and only specify the shrink prop *if* the value is non-empty.

mastercactapus: @KatieMSB @dctalbot What do you think of having an explicit helper? That way it's easy to remove/delete once fixed in material

Something like this maybe:
```js
// shrinkWorkaround fixes a bug in material where non-empty values
// fail to trigger the label text to shrink.
//
// Usage:
// const InputProps = {
//  // ...
//  ...shrinkWorkaround(value)
// }
function shrinkWorkaround(value) {
  if (_.isEmpty(value)) return {}
  return {shrink: true}
}
```

dctalbot: Sounds good. Looks like _.isEmpty returns `true` when it's passed a `Number`. I think we will only be passing in `Strings` but we may need to account for that too.  e.g. `if (typeof value is Number) return false.

mastercactapus: I think in that case we'd want shrink to be true -- since we'd be displaying the number. But yes, we'll need to handle that. We can just add it to the conditional `_.isEmpty && !_.isNumber` or something like that. Good catch

PR #315, opened 2019-12-31
feature/cal-subscriptions: graphql

Auchindoun:

**Description:**
This PR introduces the concept of a calendar subscription within GraphQL, as well as a migration to create a `calendar_subscriptions` table to store the data in.

**Out of Scope:**
Generating the URL as well as other feature work.


calendarsubscription/store.go
```diff
@@ -0,0 +1,152 @@
...
+	"database/sql"
+	uuid "github.com/satori/go.uuid"
+	"github.com/target/goalert/permission"
+	"github.com/target/goalert/util"
+	"github.com/target/goalert/util/sqlutil"
+	"github.com/target/goalert/validation/validate"
+	"time"
+)
+
+type Store interface {
```
mastercactapus: I think we were going to stop using the giant interfaces for these and just use the store as a struct going forward like in `auth/basic`, `config` and `timezone`.

graphql2/graphqlapp/app.go
```diff
@@ -3,6 +3,7 @@ package graphqlapp
 import (
 	context "context"
 	"database/sql"
+	"github.com/target/goalert/calendarsubscription"
```
mastercactapus: nit: should be sorted

calendarsubscription/store.go
```diff
@@ -0,0 +1,152 @@
...
+
+	var cs CalendarSubscription
+	err = db.findOne.QueryRowContext(ctx, id).Scan(&cs.ID, &cs.Name, &cs.UserID, &cs.LastAccess, &cs.Disabled)
+	if err != nil {
+		return nil, err
+	}
+	return &cs, nil
+}
+
+func (db *DB) CreateSubscriptionTx(ctx context.Context, tx *sql.Tx, cs *CalendarSubscription) (res bool, err error) {
```
mastercactapus: This should return the created subscription as it's a Store method -- bool is just for the GraphQL interfaces.

calendarsubscription/store.go
```diff
@@ -0,0 +1,152 @@
...
+		return false, err
+	}
+
+	stmt := db.create
+	if tx != nil {
+		stmt = tx.StmtContext(ctx, stmt)
+	}
+
+	n.ID = uuid.NewV4().String()
+	n.UserID = permission.UserID(ctx)
```
mastercactapus: User ID should be passed to the store and then permission check should be `permission.Admin, permission.MatchUser(cs.UserID)`

Explicit in the store, implicit in GraphQL.

calendarsubscription/store.go
```diff
@@ -0,0 +1,152 @@
...
+	n.LastAccess = time.Now()
```
mastercactapus: Never rely on system time, DB time is always our source of truth. This should be part of the create query using `now()` as the `last_access` value instead of passing it in.

calendarsubscription/store.go
```diff
@@ -0,0 +1,152 @@
...
+	n.Disabled = false
```
mastercactapus: False is the default value in Go, no need to set it explicitly

calendarsubscription/store.go
```diff
@@ -0,0 +1,152 @@
...
+type DB struct {
+	db *sql.DB
+
+	findOne *sql.Stmt
+	create  *sql.Stmt
+	update  *sql.Stmt
+	delete  *sql.Stmt
+}
+
+func NewDB(ctx context.Context, db *sql.DB) (*DB, error) {
```
mastercactapus: nit: Let's try to add comments for public/exported methods as part of the dev life work

migrate/migrations/20191230125304-calendar-subscriptions.sql
```diff
@@ -0,0 +1,11 @@
+-- +migrate Up
+CREATE TABLE calendar_subscriptions (
+    id uuid NOT NULL UNIQUE,
```
dctalbot:
```suggestion
    id uuid PRIMARY KEY,
```
what's the primary key on this table?

calendarsubscription/calendarsubscription.go
```diff
@@ -0,0 +1,25 @@
...
+	ID         string    `json:"id"`
+	Name       string    `json:"name"`
+	UserID     string    `json:"user_id"`
+	LastAccess time.Time `json:"last_access"`
+	Disabled   bool      `json:"disabled"`
+}
+
+func (cs CalendarSubscription) Normalize() (*CalendarSubscription, error) {
+	err := validate.Many(
+		validate.IDName("Name", cs.Name),
```
mastercactapus: We should validate the ID and UserID as well.

calendarsubscription/store.go
```diff
@@ -0,0 +1,152 @@
...
+	return true, nil
+}
+
+func (db *DB) UpdateSubscriptionTx(ctx context.Context, tx *sql.Tx, cs *CalendarSubscription) (err error) {
+	err = permission.LimitCheckAny(ctx, permission.Admin, permission.User)
+	if err != nil {
+		return err
+	}
+
+	err = validate.UUID("CalendarSubscriptionID", cs.ID)
```
mastercactapus: Should be handled by `Normalize`

calendarsubscription/store.go
```diff
@@ -0,0 +1,152 @@
...
+
+	_, err = stmt.ExecContext(ctx, n.ID, n.Name, n.UserID, n.LastAccess, n.Disabled)
+	if err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+func (db *DB) UpdateSubscriptionTx(ctx context.Context, tx *sql.Tx, cs *CalendarSubscription) (err error) {
+	err = permission.LimitCheckAny(ctx, permission.Admin, permission.User)
```
mastercactapus: This should replace `permission.User` with `permission.MatchUser(cs.UserID)`. Additionally, the update query will need to add `user_id = ` to the `WHERE` clause otherwise any user could update any other user's subscription.

calendarsubscription/store.go
```diff
@@ -0,0 +1,152 @@
...
+	_, err = stmt.ExecContext(ctx, n.ID, n.Name, n.Disabled)
+	if err != nil {
+		return err
+	}
+
+	return err
+}
+
+func (db *DB) DeleteSubscriptionsTx(ctx context.Context, tx *sql.Tx, ids []string) error {
+	err := permission.LimitCheckAny(ctx, permission.Admin, permission.User)
```
mastercactapus: Need to check the user ID like in update

graphql2/graphqlapp/calendarsubscription.go
```diff
@@ -0,0 +1,100 @@
...
+
+		return err
+	})
+
+	return res, err
+}
+
+func (m *Mutation) UpdateCalendarSubscription(ctx context.Context, input graphql2.UpdateCalendarSubscriptionInput) (bool, error) {
+	err := withContextTx(ctx, m.DB, func(ctx context.Context, tx *sql.Tx) error {
+		cs := &calendarsubscription.CalendarSubscription{
```
mastercactapus: This needs some sort of `FindOneForUpdateTx` and only update the provided fields.

For example, you should be able to update the name without providing the `disabled` value.

See `UpdateUserContactMethod` for an example: https://github.com/target/goalert/blob/master/graphql2/graphqlapp/contactmethod.go#L76

graphql2/schema.graphql
```diff
@@ -38,6 +38,12 @@ type Query {
   # Returns a single schedule with the given ID.
   schedule(id: ID!): Schedule
 
+  # Returns the public information of a calendar subscription
+  calendarSubscription(id: ID!): CalendarSubscription
```
mastercactapus: This was missed during the proposal, but we'll need it for the edit dialog to work. However, it should match contact method query for naming i.e. `userCalendarSubscription`

graphql2/schema.graphql
```diff
@@ -38,6 +38,12 @@ type Query {
...
+
+  # Returns a paginated list of calendar subscriptions
+  calendarSubscriptions(input: CalendarSubscriptionSearchOptions): CalendarSubscriptionConnection!
```
mastercactapus: This should exist on the `User` struct as previously discussed in the proposal: https://github.com/target/goalert/compare/calendar-subscription-api#diff-04e3af9ffc876a383b6f192a8ff8450dR841

graphql2/schema.graphql
```diff
@@ -293,6 +303,16 @@ input CreateScheduleInput {
   targets: [ScheduleTargetInput!]
 }
 
+input CreateCalendarSubscriptionInput {
+  name: String!
```
mastercactapus: This type is missing the `ScheduleID` and the `NotificationMinutes` array.
https://github.com/target/goalert/compare/calendar-subscription-api#diff-04e3af9ffc876a383b6f192a8ff8450dR269

We should also probably allow passing in `disabled` to keep the API matching between create and update (i.e. update is a subset of create)

graphql2/schema.graphql
```diff
@@ -293,6 +303,16 @@ input CreateScheduleInput {
...
+}
+
+input UpdateCalendarSubscriptionInput {
+  id: ID!
+  name: String
+  disabled: Boolean
```
mastercactapus: Missing `notificationMinutes`

graphql2/schema.graphql
```diff
@@ -435,6 +460,14 @@ type Schedule {
   isFavorite: Boolean!
 }
 
+type CalendarSubscription {
```
mastercactapus: Missing `notificationMinutes`, `scheduleID` and `schedule`
https://github.com/target/goalert/compare/calendar-subscription-api#diff-04e3af9ffc876a383b6f192a8ff8450dR282

migrate/migrations/20191230125304-calendar-subscriptions.sql
```diff
@@ -0,0 +1,11 @@
...
+    name TEXT NOT NULL,
+    user_id uuid NOT NULL REFERENCES users(id),
+    last_access TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
+    disabled BOOLEAN NOT NULL DEFAULT false
+);
```
mastercactapus: We will need to store the `schedule_id` and `config` here as well:
https://github.com/target/goalert/compare/calendar-subscription-api#diff-04e3af9ffc876a383b6f192a8ff8450dR282

smoketest/graphql2calendarsubscriptions_test.go
```diff
@@ -0,0 +1,124 @@
...
+			return
+		}
+		err := json.Unmarshal(g.Data, &res)
+		if err != nil {
+			t.Fatal("failed to parse response:", err)
+		}
+	}
+
+	// find one query
+	var cs struct {
```
mastercactapus: Can we use the existing struct?

graphql2/graphqlapp/calendarsubscription.go
```diff
@@ -0,0 +1,100 @@
...
+		if err != nil {
+			return err
+		}
+
+		return err
+	})
+
+	return err == nil, err
+}
+func (q *Query) CalendarSubscriptions(ctx context.Context, opts *graphql2.CalendarSubscriptionSearchOptions) (conn *graphql2.CalendarSubscriptionConnection, err error) {
```
mastercactapus: This really should just be an array using a `FindAllByUser` approach -- the result set will always be scoped to a specific user like contact methods (or like integration keys on a service).

It's a lot of overhead to add/maintain otherwise, search would generally be for things created by other users and/or >100 results.

Here's what we had settled on previously: https://github.com/target/goalert/compare/calendar-subscription-api#diff-04e3af9ffc876a383b6f192a8ff8450dR841

calendarsubscription/store.go
Auchindoun: We won't have an ID to normalize on Create, would we need to make 2 normalize functions for both mutation types?

mastercactapus: We still have to generate the ID in the store method, so we can just generate it before calling Normalize instead of after.

app/initstores.go
```diff
@@ -117,6 +118,14 @@ func (app *App) initStores(ctx context.Context) error {
 	if err != nil {
 		return errors.Wrap(err, "init schedule store")
 	}
+
+	if app.CalendarSubscriptionStore == nil {
+		app.CalendarSubscriptionStore, err = calendarsubscription.NewStore(ctx, app.db)
```
dctalbot:
```suggestion
		app.CalendarSubscriptionStore, err = calendarsubscription.NewDB(ctx, app.db)
```

Even though I prefer `NewStore` since it's more idiomatic, probably better to be consistent and stick with `NewDB`. Future work could be renaming all of them at once

app/initstores.go
```diff
@@ -117,6 +118,14 @@ func (app *App) initStores(ctx context.Context) error {
...
+		//app.CalendarSubscriptionStore, err = calendarsubscription.NewStore(ctx, app.db)
```
dctalbot:
```suggestion
```

calendarsubscription/calendarsubscription.go
```diff
@@ -0,0 +1,31 @@
...
+type CalendarSubscription struct {
+	ID         string    `json:"id"`
+	Name       string    `json:"name"`
+	UserID     string    `json:"user_id"`
+	LastAccess time.Time `json:"last_access"`
+	Disabled   bool      `json:"disabled"`
+	Config     []byte    `json:"config"`
+	ScheduleID string    `json:"schedule_id"`
+
+	NMinutes []int `json:"notification_minutes"`
```
dctalbot: I'd rename `NMinutes` to be something more semantic, like `NotificationMinutes`

Auchindoun: I believe the plan is to use `reminderMinutes` ultimately, from a Slack conversation. I can make this change if you didn't get a chance to see it @KatieMSB

KatieMSB: Sure sounds good

calendarsubscription/store.go
```diff
@@ -0,0 +1,189 @@
...
+	row := wrapTx(ctx, tx, b.findOneUpd).QueryRowContext(ctx, id)
+	err = row.Scan(&cs.ID, &cs.Name, &cs.UserID, &cs.Disabled, &cs.Config)
+	if err != nil {
+		return nil, err
+	}
+	return &cs, nil
+}
+
+// UpdateSubscriptionTx updates a calendar subscription with given information.
+func (b *Store) UpdateSubscriptionTx(ctx context.Context, tx *sql.Tx, cs *CalendarSubscription) error {
```
mastercactapus: This can be just `UpdateTx` since we're already `calendarsubscription.Store` and only have one type

calendarsubscription/store.go
```diff
@@ -0,0 +1,189 @@
...
+			return nil, err
+		}
+		calendarsubscriptions = append(calendarsubscriptions, cs)
+	}
+
+	return calendarsubscriptions, nil
+}
+
+// DeleteSubscriptionsTx removes calendar subscriptions with the given ids.
+func (b *Store) DeleteSubscriptionsTx(ctx context.Context, tx *sql.Tx, ids []string) error {
```
mastercactapus: This can be simplified to `DeleteTx` for the same reason. Additionally, let's make this variadic: `ids ...string` like in `heartbeat/store.go`

calendarsubscription/store.go
```diff
@@ -0,0 +1,189 @@
...
+	err := permission.LimitCheckAny(ctx, permission.Admin, permission.MatchUser(permission.UserID(ctx)))
```
mastercactapus: we can use just `permission.User` as matching against the current user is redundant.

graphql2/graphqlapp/calendarsubscription.go
```diff
@@ -0,0 +1,98 @@
...
+	return (*CalendarSubscription)(a)
+}
+
+func (a *CalendarSubscription) ReminderMinutes(ctx context.Context, obj *calendarsubscription.CalendarSubscription) ([]int, error) {
+	var config calendarsubscription.Config
+	err := json.Unmarshal(obj.Config, &config)
+	obj.ReminderMinutes = config.ReminderMinutes
+	return obj.ReminderMinutes, err
+}
+func (a *CalendarSubscription) ScheduleID(ctx context.Context, obj *calendarsubscription.CalendarSubscription) (string, error) {
```
mastercactapus: Is this required? it should map to the `ScheduleID` field automatically (json definitions may be changing that)

graphql2/graphqlapp/calendarsubscription.go
```diff
@@ -0,0 +1,98 @@
...
+	e := *obj
+	return e.ScheduleID, nil
+}
+func (a *CalendarSubscription) Schedule(ctx context.Context, obj *calendarsubscription.CalendarSubscription) (*schedule.Schedule, error) {
+	return a.ScheduleStore.FindOne(ctx, obj.ScheduleID)
+}
+func (a *CalendarSubscription) URL(ctx context.Context, obj *calendarsubscription.CalendarSubscription) (*string, error) {
+	var err error
+	return nil, err
```
mastercactapus: We should return an error, stub URL, or omit this field entirely from this PR. Probably the latter.

graphql2/schema.graphql
```diff
@@ -231,6 +234,12 @@ type Mutation {
   setLabel(input: SetLabelInput!): Boolean!
 
   createSchedule(input: CreateScheduleInput!): Schedule
+
+  userCreateCalendarSubscription(
```
mastercactapus: Should be `createUserCalendarSubscription` to follow the naming conventions

graphql2/schema.graphql
```diff
@@ -231,6 +234,12 @@ type Mutation {
...
+    input: UserCreateCalendarSubscriptionInput!
```
mastercactapus: Same

graphql2/schema.graphql
```diff
@@ -231,6 +234,12 @@ type Mutation {
...
+  ): CalendarSubscription!
+  userUpdateCalendarSubscription(input: UserUpdateCalendarSubscriptionInput!): Boolean!
```
mastercactapus: Same

migrate/migrations/20191216145826-calendar-subscriptions.sql
```diff
@@ -0,0 +1,15 @@
+-- +migrate Up
+
+CREATE TABLE user_calendar_subscriptions (
+    id UUID PRIMARY KEY,
+    name TEXT NOT NULL,
+    user_id UUID NOT NULL REFERENCES users (id) ON DELETE CASCADE,
+    last_access TIMESTAMPTZ NOT NULL DEFAULT now(),
```
mastercactapus: As per the UI discussion, I think this will be nullable (and default to null) and we're going to add a `last_updated` field that is NOT NULL defaulting to `now()` and set when updated.

calendarsubscription/calendarsubscription.go
```diff
@@ -0,0 +1,31 @@
+package calendarsubscription
+
+import (
+	"github.com/target/goalert/validation/validate"
+	"time"
+)
+
+type CalendarSubscription struct {
+	ID         string    `json:"id"`
```
mastercactapus: I think we can remove the `json:""` definitions, we only had those in older stuff because we needed it for the legacy gql API. New stuff should work by default.

graphql2/graphqlapp/calendarsubscription.go
KatieMSB: No not required, removed now. Was generated previously before ScheduleID was added to the CalendarSubscription struct.

calendarsubscription/calendarsubscription.go
```diff
@@ -0,0 +1,39 @@
...
+
+func (cs CalendarSubscription) Normalize() (*CalendarSubscription, error) {
+	if cs.ID == "" {
+		cs.ID = uuid.NewV4().String()
+	}
+
+	err := validate.Many(
+		validate.IDName("Name", cs.Name),
+		validate.UUID("ID", cs.ID),
+		validate.UUID("UserID", cs.UserID),
```
arurao: We should probably add validation here for returning error if reminderMinutes is negative(example: -1). Currently it accepts all int values.

util/errutil/maperror.go
```diff
@@ -21,6 +21,8 @@ func MapDBError(err error) error {
 	switch dbErr.Code {
 	case "23503": // fkey constraint
 		switch dbErr.ConstraintName {
+		case "user_calendar_subscriptions_schedule_id_fkey":
```
arurao: Should we also add `user_calendar_subscriptions_user_id_fkey`?

migrate/migrations/20191216145826-calendar-subscriptions.sql
```diff
@@ -0,0 +1,18 @@
...
+    id UUID PRIMARY KEY,
+    name TEXT NOT NULL,
+    user_id UUID NOT NULL REFERENCES users (id) ON DELETE CASCADE,
+    last_access TIMESTAMPTZ,
+    last_update TIMESTAMPTZ NOT NULL DEFAULT now(),
+    disabled BOOLEAN NOT NULL DEFAULT FALSE,
+    schedule_id UUID NOT NULL REFERENCES schedules (id) ON DELETE CASCADE,
+    config JSONB NOT NULL,
+
+    UNIQUE(name, schedule_id)
```
arurao: Let's also modify this to be `UNIQUE(name, schedule_id, user_id)`

calendarsubscription/calendarsubscription.go
mastercactapus: I think negative is OK, it would mean a user wants an extra notification say 5 minutes after their shift starts. As long as the input doesn't cause us to output something wrong (i.e. as long as we can generate a valid iCal/VALARM entry) it's probably okay.

mastercactapus: Having said that, we def. should validate the length of the array at least before serializing and sending it to the DB.

arurao: That sounds good to me. Should length of array be restricted to a specific limit?

mastercactapus: Yeah I think so. Maybe 15? I think on the UI we were thinking of making sure 3 looks good, and max of 5. API could be a little more flexible though -- the goal is for nobody to ever know there is a limit during normal usage, but keep it bounded in the case of mistakes or "curiosity"

calendarsubscription/store.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	err = validate.UUID("ID", id)
+	if err != nil {
+		return nil, err
+	}
+
+	var cs CalendarSubscription
+	err = cs.scanFrom(b.findOne.QueryRowContext(ctx, id).Scan)
+	if err != nil {
```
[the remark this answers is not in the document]
arurao: Simply adding this should do it
`if err == sql.ErrNoRows { 
  log.Debugf(ctx, "query calendar subscription: no rows matched")
  err = nil
}`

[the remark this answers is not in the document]
mastercactapus: We could do `return nil, validation.NewFieldError("ID", "not found")` or just `return nil, nil`.

I don't think it's something we've been super consistent on yet. But I think `return nil, nil` is most common.

The debug line is probably overkill -- IIRC we're only doing debug in the case of alert escalation failures returning `ErrNoRows`

calendarsubscription/calendarsubscription.go
```diff
@@ -29,6 +29,7 @@ func (cs CalendarSubscription) Normalize() (*CalendarSubscription, error) {
 	}
 
 	err := validate.Many(
+		validate.ArrayLength("ReminderMinutes", cs.Config.ReminderMinutes, 15),
```
arurao: Rather than adding a separate function, and since we are using this at only once place as of now, could we use something like : 
`err = validate.Range("ReminderMinutes", len(cs.Config.ReminderMinutes), 0, 15)`

validation/validate/array.go
```diff
@@ -0,0 +1,13 @@
+package validate
```
arurao: If we use `validate.Range` as mentioned above, we don't need this.

calendarsubscription/store.go
```diff
@@ -97,11 +98,11 @@ func (b *Store) FindOne(ctx context.Context, id string) (*CalendarSubscription,
 
 	var cs CalendarSubscription
 	err = cs.scanFrom(b.findOne.QueryRowContext(ctx, id).Scan)
-	if err != nil {
-		return nil, err
+	if err == sql.ErrNoRows {
+		return nil, validation.NewFieldError("ID", "Not found")
 	}
```
arurao: I think we should keep the part 
`if err != nil { return nil, err }`
And then
`return &cs, nil`
as the last line.

PR #330, opened 2020-01-07
ui: Convert ScheduleOverrideCreateDialog to hooks

dctalbot:

**Additional Info:**
Calendar schedule pages are lagging, including dialogs. Converted the dialogs to hooks, but didn't gain any significant improvement in perf

### DIFF
--- a/web/src/app/schedules/ScheduleOverrideCreateDialog.js
+++ b/web/src/app/schedules/ScheduleOverrideCreateDialog.js
@@ -72,7 +72,7 @@ export default function ScheduleOverrideCreateDialog(props) {
           disabled={loading}
           errors={fieldErrors(error)}
           value={value}
-          onChange={newValue => setValue({ ...value, ...newValue })}
+          onChange={newValue => setValue(newValue)}
         />
       }
     />
--- a/web/src/app/schedules/ScheduleOverrideCreateDialog.js
+++ b/web/src/app/schedules/ScheduleOverrideCreateDialog.js
@@ -1,6 +1,6 @@
-import React from 'react'
+import React, { useState } from 'react'
 import p from 'prop-types'
-import { Mutation } from 'react-apollo'
+import { useMutation } from 'react-apollo'
 import FormDialog from '../dialogs/FormDialog'
 import { DateTime } from 'luxon'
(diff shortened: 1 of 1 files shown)


web/src/app/schedules/ScheduleOverrideCreateDialog.js
```diff
@@ -33,79 +33,64 @@ const mutation = gql`
...
+      onSubmit={mutate}
+      form={
+        <ScheduleOverrideForm
+          add={props.variant !== 'remove'}
+          remove={props.variant !== 'add'}
+          scheduleID={props.scheduleID}
+          disabled={loading}
+          errors={fieldErrors(error)}
+          value={value}
+          onChange={newValue => setValue({ ...value, ...newValue })}
```
Auchindoun: We should be fine to keep using `value => this.setState({ value })` here; were you running into issues?

dctalbot: When converting to hooks, I extracted the keys of the state object as the destructured array values so instead of 
```js
{
value: { x: 1, y: 2, z: 'abc' }
}
```
you'll have
```js
const [value, setValue] = { x: 1, y: 2, z: 'abc' }
```

Let me know if this needs further clarification

Auchindoun: Gotcha!

mastercactapus: The form container handle's aggregating all the props, so `setValue(newValue)` should work rather than merging them since `newValue` will always have all props `value` has with changes already applied.

web/src/app/schedules/ScheduleOverrideCreateDialog.js
```diff
@@ -33,79 +33,64 @@ const mutation = gql`
...
-      </Mutation>
-    )
-  }
+  return (
+    <FormDialog
+      onClose={props.onClose}
+      title={copyText[props.variant].title}
+      subTitle={copyText[props.variant].desc}
+      errors={nonFieldErrors(error)}
+      onSubmit={mutate}
```
mastercactapus: Wrap this so we don't pass the event args to `mutate`
```suggestion
      onSubmit={() => mutate()}
```

PR #341, opened 2020-01-09
ui: Always allow clearing of select input fields

dctalbot:

Users will now be able to clear select fields, even if they are required by the form. (The form still won't submit when the fields are empty).


Auchindoun:
@dctalbot Can we add tests for these cases that you've mentioned? We can make a separate Cypress file dedicated to testing all the paths we can think of for Material Select fields. I don't think we have to test every select in the app, since they all pass through the same wrappers iirc any regressions _should_ be found testing on just one or two.

Auchindoun:
I think if we do set this to the default, we should fix the relevant warnings that aren't used to having empty values in state. Otherwise, we should do as you mentioned and set the default to false to start

Auchindoun:
Closing as strategy is changing slightly to solve this issue.

PR #349, opened 2020-01-13
ui: use native time-pickers when available

mastercactapus:

**Description:**
This PR replaces usages of date, time, and date-time pickers with new wrapper components.

The new ISO- components:
- use an ISO time string directly for value and on-change
- handle the page tz parameter directly/automatically
- use native controls when available (falling-back to material) which support locale formatting and keyboard input
- are fully supported (incl. fallback) with `dialogForm` and `form` test helpers
- no longer need manual mapValue functions and adornments

Native support is determined by feature-detection using Modernizr.
`ScheduleRuleCreateDialog` was converted to hooks.


web/src/app/schedules/ScheduleRuleCreateDialog.js
```diff
@@ -1,79 +1,96 @@
...
-                rules: this.state.value.rules,
+  return (
+    <FormDialog
+      onClose={onClose}
+      title={`Add ${startCase(targetType)} to Schedule`}
+      errors={nonFieldErrors(mutationStatus.error)}
+      maxWidth='md'
+      onSubmit={() => {
+        mutate({
+          variables: {
```
dctalbot: Move the `variables` key/value pair into the `useMutation` block so it matches our call to `useQuery`

web/src/package.json
```diff
@@ -87,6 +87,8 @@
     "lodash-es": "4.17.15",
     "luxon": "1.21.3",
     "mdi-material-ui": "6.10.0",
+    "modernizr": "^3.8.0",
```
dctalbot: Specify explicit semvers

web/src/app/schedules/ScheduleRuleEditDialog.js
```diff
@@ -57,30 +59,34 @@ export default class ScheduleRuleEditDialog extends React.Component {
...
 
-  renderDialog(data, commit, status) {
+  renderDialog(data, commit, status, zone) {
     const defaults = {
       targetID: this.props.target.id,
-      rules: data.rules.map(r =>
-        pick(r, ['id', 'start', 'end', 'weekdayFilter']),
-      ),
+      rules: data.rules.map(r => ({
+        ..._.pick(r, ['id', 'weekdayFilter']),
```
Auchindoun: Could you explain how the spread operator works before a lodash func (`..._.pick`)? I haven't seen it used in this context before so it's a bit confusing to me

web/src/app/schedules/util.js
```diff
@@ -30,6 +30,15 @@ export function mapRuleTZ(fromTZ, toTZ, rule) {
   }
 }
 
+export function clockToISO(clockTime, zone) {
+  return DateTime.fromFormat(clockTime, 'HH:mm', { zone })
+    .toUTC()
+    .toISO()
+}
+export function isoToClock(isoTime, zone) {
```
Auchindoun: Could this perhaps be renamed to `isoToMilitary` or s/t? The `Clock` keyword confused me at first and I had to look at the def. here to understand

web/src/app/schedules/ScheduleRuleEditDialog.js
dctalbot:
```js
id: r.id,
weekdayFilter: r.weekdayfilter
```
is how I'd recommend writing this. Easy to read, and doesn't depend on `lodash`

web/src/app/util/ISOPickers.js
```diff
@@ -0,0 +1,139 @@
...
+  useEffect(() => {
+    setInputValue(dtValue.toFormat(format))
+  }, [value, zone])
+
+  const handleChange = e => {
+    setInputValue(e.target.value)
+
+    const newVal = isoInput(e.target.value)
+    if (newVal && newVal !== value) {
+      onChange(newVal)
```
Auchindoun: Can you add a note here as to what `onChange`'s intended use is? Hard to tell with it being passed through as a param, especially with `handleChange` presumably being that

web/src/app/util/ISOPickers.js
```diff
@@ -0,0 +1,139 @@
...
+  { value, onChange, timeZone, ...otherProps },
+  { format, truncateTo, type, Fallback },
+) {
+  const native = hasInputSupport(type)
+  const params = useSelector(urlParamSelector)
+  const zone = timeZone || params('tz', 'local')
+  const dtValue = DateTime.fromISO(value, { zone })
+  const [inputValue, setInputValue] = useState(dtValue.toFormat(format))
+
+  const parseInput = input => {
```
Auchindoun: Would `isoToDateTime` be a more accurate name?

web/src/app/util/ISOPickers.js
```diff
@@ -0,0 +1,139 @@
...
+      }
+      return dt
+    }
+
+    const iso = DateTime.fromISO(input)
+    if (iso.isValid) return iso
+
+    return null
+  }
+  const isoInput = input => {
```
Auchindoun: Same here, maybe `inputToISO`?

web/src/app/util/ISOPickers.js
```diff
@@ -0,0 +1,139 @@
...
+    extraProps.leftArrowButtonProps = { 'data-cy': 'month-back' }
+    extraProps.rightArrowButtonProps = { 'data-cy': 'month-next' }
+  }
+
+  return (
+    <Fallback
+      value={dtValue}
+      onChange={v => handleChange({ target: { value: v } })}
+      showTodayButton
+      autoOk
```
Auchindoun: I don't think should always be on– For instance, in the override form after you select the time this will automatically close the form which doesn't give the user a chance to review their details, or toggle between AM/PM. I believe the only place where we had this set to `true` originally was in `ScheduleShiftList`

web/src/cypress/integration/timepickers.ts
```diff
@@ -0,0 +1,171 @@
...
+
+import { testScreen } from '../support'
+const c = new Chance()
+
+testScreen('Time Pickers', testTimePickers)
+
+function testTimePickers(screen: ScreenFormat) {
+  describe('Time', () => {
+    const check = (name: string, params: string, display: string) =>
+      it(name, () => {
```
Auchindoun: Can we name these with references to where they are testing these pickers? The hardcoded commands for specific dialogs but the general titles threw me off for a sec

web/src/cypress/support/form.ts
```diff
@@ -1,14 +1,111 @@
...
+  const isAM = hour < 12
+  if (!isAM) hour -= 12
+
+  return cy
+    .contains('button', isAM ? 'AM' : 'PM')
+    .click() // select AM or PM
+
+    .get('[role=dialog][data-cy=picker-fallback] [role=menu]')
+    .parent()
+    .then(clickArc(hour / 12)) // select the hour
```
Auchindoun: It looks like we should be able to select the correct hour without doing any math to determine where the button is on the page. From what I'm seeing we can reference the picker once it's open and search for the span for each hour on the clock: `<span class="MuiTypography-root MuiPickersClockNumber-clockNumber MuiTypography-body1" style="transform: translate(55px, 19.6px);">1</span>`

E.g. maybe something like `cy.get('dialog picker-fallback menu span').contains(1).click()`

web/src/cypress/support/form.ts
```diff
@@ -21,7 +118,15 @@ function fillFormField(
   return cy.get(selector).then(el => {
     const isSelect =
       el.parents('[data-cy=material-select]').data('cy') === 'material-select'
+    const pickerFallback = el
+      .parents('[data-cy-fallback-type]')
+      .data('cyFallbackType')
     if (isSelect) {
+      if (DateTime.isDateTime(value)) {
```
Auchindoun: nit: Can we add spacing to break all this apart a bit for readability? Not a big deal if you'd rather not but just a personal nitpick :)

web/src/cypress/support/schedule.ts
```diff
@@ -124,9 +124,10 @@ function setScheduleTarget(
     }
   }`
   const { scheduleID, target } = tgt
+  const { schedule, ...params } = tgt
```
Auchindoun:
```suggestion
  const { schedule, scheduleID, target, ...params } = tgt
```

web/src/app/schedules/ScheduleRuleEditDialog.js
Auchindoun: I was also thinking this; it took me a minute to digest this call and remember the return type of `pick` to make the spread op possible 🙂

web/src/cypress/support/form.ts
Auchindoun: It looks like you put a significant amount of thought into this though, so correct me if I'm wrong

web/src/app/schedules/ScheduleRuleEditDialog.js
mastercactapus: Makes sense -- it used to have 4, but now that there's only 2 it's just overhead

web/src/app/schedules/util.js
mastercactapus: Hmm. We are using `Clock` as the type on the backend to represent `HH:mm` values, so I didn't think twice about it. But I can see how it'd be confusing (I think that's actually why Go calls it "Kitchen" time when AM/PM is involved).

I'll try thinking of a different way of naming it since all the types are similar. A literal name for this would be `ISOTimestampToISOTimeWithoutOffset` so maybe `timeFromISO` and `isoFromTime`?

Using "from" instead of "to" might be better as well now that I think about it because it loses the date information (e.g. not a 1-to-1 conversion).

mastercactapus: Might be worth noting, we also use `ClockTime` as the type in the GraphQL schema too... maybe we should stick with clock and just make sure to add comments (should have been there anyway)?

web/src/app/util/ISOPickers.js
mastercactapus: I think we want to keep `parse` as it's taking in raw input, in one of many possible formats, and we've been mostly consistent in naming things like that. But the DateTime output is important, I'll try coming up with something for that.

web/src/cypress/support/form.ts
mastercactapus: That could work for hours, but not minutes. Rather than having 2 code paths I just opted to use the same one so there's less to go wrong/debug.

web/src/cypress/support/schedule.ts
mastercactapus: lol oops

mastercactapus: oh wait. This was intentional, because params needs to include both `scheduleID` and `target`.

I'll add a comment.

web/src/app/util/ISOPickers.js
mastercactapus: Ah, yes I had it backward. I think in the interest of simplicity, and consistency, I'll just leave it always off. It is a fall-back after all, but better the user can set it how they want than needing to remember to do AM/PM first, for example.

web/src/cypress/integration/timepickers.ts
```diff
@@ -0,0 +1,171 @@
...
+        })
+
+        cy.dialogFinish('Submit')
+
+        cy.get('body').contains('Sun from 2:05 PM to 5:56 PM')
+      })
+
+    describe('Native', () => {
+      check(
+        'should allow managing time values in the same zone',
```
dctalbot:
```suggestion
        'should manage time values in local timezone',
```

web/src/cypress/integration/timepickers.ts
```diff
@@ -0,0 +1,171 @@
...
+        '?tz=America/New_York',
+        'Sun from 3:04 PM to 4:23 AM',
+      )
+      check(
+        'should allow managing time values in the alt. zones',
```
dctalbot:
```suggestion
        'should manage time values in alt. timezone',
```

web/src/cypress/integration/timepickers.ts
dctalbot: and all that follow... etc

web/src/cypress/integration/timepickers.ts
dctalbot: and all that follow... etc

web/src/app/util/ISOPickers.js
Auchindoun: Sounds good to me!

web/src/app/schedules/util.js
Auchindoun: I see. The only thing that concerns me is it's a data type that we have defined on the schema/backend, so context is inherently required to understand what this function might do to newcomers, as opposed to a more generally understood term. I understand where you're coming from though

web/src/app/schedules/ScheduleRuleCreateDialog.js
```diff
@@ -1,79 +1,96 @@
...
+  return (
+    <FormDialog
+      onClose={onClose}
+      title={`Add ${startCase(targetType)} to Schedule`}
+      errors={nonFieldErrors(mutationStatus.error)}
+      maxWidth='md'
+      loading={queryStatus.loading || mutationStatus.loading}
+      onSubmit={() => {
+        mutate()
+      }}
```
dctalbot: `onSubmit={mutate}` should suffice

web/src/app/schedules/util.js
dctalbot: I'd also advocate for a different term here. These functions don't touch the backend.
When I think of "clock time" I think of an analog 🕑that only does 12 hrs. I'd suggest `24HourTime`

web/src/app/schedules/ScheduleRuleCreateDialog.js
mastercactapus: We can't do that as `onSubmit` is an event handler -- it will pass in arguments not meant for `mutate` and we don't want it to interpret the return value of `mutate`

web/src/app/schedules/util.js
mastercactapus: Actually, they do -- these functions are *only* used for converting to/from the GraphQL data type to and from the backend into ISO timestamps behind the scenes of the UI. I think that's where some of the confusion is stemming from.

Maybe `isoToGQLClockTime` and `gqlClockTimeToISO`? That way it's clear it's intent is for the backend API and not any type of formatting/use.

PR #353, opened 2020-01-14
devtools/resetdb: use pgtype.Time instead of custom encoder

mastercactapus:

**Description:**
This PR removes the use of a custom encoder in the `resetdb` tool now that `pgtype` added support for the Postgres `Time` type.


devtools/resetdb/main.go
```diff
@@ -98,6 +100,11 @@ func fillDB(ctx context.Context, url string) error {
 			log.Fatal(err)
 		}
 	}
+	asTime := func(c rule.Clock) (t pgtype.Time) {
+		t.Status = pgtype.Present
+		t.Microseconds = time.Duration(c).Microseconds()
```
Auchindoun: Do we need this level of granularity? I believe on the UI we go to the minute

mastercactapus: The `pgtype.Time` (and Postgres itself) uses microseconds as the underlying unit for the `Time` type. So anything we send to it will need to be converted at some point. We handle truncation to minutes in other layers, but it all still boils down to microseconds in the end.

Here we're just casting `rule.Clock` to it's underlying `time.Duration` type and using the built-in conversion method to get the `Microseconds` value we need.

mastercactapus: image

PR #357, opened 2020-01-15
engine: implement APIKeyExpireDays

KatieMSB:

**Description:**
Added Maintenance.APIKeyExpireDays to config that disables calendar API keys unused after set number of days


engine/cleanupmanager/db.go
```diff
@@ -38,7 +39,8 @@ func NewDB(ctx context.Context, db *sql.DB) (*DB, error) {
 
 		// Abort any cleanup operation that takes longer than 3 seconds
 		// error will be logged.
-		setTimeout:    p.P(`SET LOCAL statement_timeout = 3000`),
-		cleanupAlerts: p.P(`delete from alerts where id = any(select id from alerts where status = 'closed' AND created_at < (now() - $1::interval) order by id limit 100 for update)`),
+		setTimeout:     p.P(`SET LOCAL statement_timeout = 3000`),
+		cleanupAlerts:  p.P(`delete from alerts where id = any(select id from alerts where status = 'closed' AND created_at < (now() - $1::interval) order by id limit 100 for update)`),
+		cleanupAPIKeys: p.P(`update user_calendar_subscriptions set disabled = true where id = any(select id from user_calendar_subscriptions where last_update < (now() - $1::interval))`),
```
mastercactapus: Add a limit to this (100 like above should be fine) so it's always bounded, as well as `for update` .

We should probably add `skip locked` to both of these as well (the inner query). Since it's cleanup and measured in days, we don't need to add any contention -- it will run often enough.

PR #361, opened 2020-01-16
calendarsubscription: implement ical generation

arurao:

**Description:**
This PR implements the actual iCal generation logic. Given a list of shift & using the minutes for reminders, the method outputs the iCal.


calendarsubscription/calendarsubscription_test.go
```diff
@@ -0,0 +1,37 @@
...
+			cs := CalendarSubscription{}
+
+			t1, _ := time.Parse(time.RFC3339, "2020-01-01T22:08:41+00:00")
+			t2, _ := time.Parse(time.RFC3339, "2020-01-07T22:08:41+00:00")
+
+			s1 := oncall.Shift{UserID: "cb75f78a-0f7c-42fa-99f8-6b30e92a9518", Start: t1, End: t2}
+			shifts := []oncall.Shift{}
+			shifts = append(shifts, s1)
+
+			reminderMinutes := make([]int, 2)
```
dctalbot:
```suggestion
			reminderMinutes := [2]int{5, 10}
```

calendarsubscription/calendarsubscription_test.go
```diff
@@ -0,0 +1,37 @@
...
+			reminderMinutes[0] = 5
```
dctalbot:
```suggestion
```

calendarsubscription/calendarsubscription_test.go
```diff
@@ -0,0 +1,37 @@
...
+			reminderMinutes[1] = 10
```
dctalbot:
```suggestion
```

calendarsubscription/calendarsubscription.go
```diff
@@ -40,3 +44,45 @@ func (cs CalendarSubscription) Normalize() (*CalendarSubscription, error) {
 
 	return &cs, nil
 }
+
+func (cs CalendarSubscription) RenderICalFromShifts(shifts []oncall.Shift, reminderMinutes []int) ([]byte, error) {
+	type iCalOptions struct {
+		Shifts          []oncall.Shift `json:"s,omitempty"`
```
dctalbot:
```suggestion
		Shifts          []oncall.Shift `json:"s, omitempty"`
```

calendarsubscription/calendarsubscription.go
```diff
@@ -40,3 +44,45 @@ func (cs CalendarSubscription) Normalize() (*CalendarSubscription, error) {
...
+		ReminderMinutes []int          `json:"r, omitempty"`
+	}
+
+	var iCalTemplate = `
+		BEGIN:VCALENDAR
+		VERSION:2.0
+		PRODID:-//ZContent.net//Zap Calendar 1.0//EN
```
dctalbot:
```suggestion
		PRODID:-//GoAlert//Calendar Subscriptions//EN
```

calendarsubscription/calendarsubscription.go
```diff
@@ -40,3 +44,45 @@ func (cs CalendarSubscription) Normalize() (*CalendarSubscription, error) {
...
+		CALSCALE:GREGORIAN
+		METHOD:PUBLISH
+		{{range .Shifts}}
+		BEGIN:VEVENT
+		SUMMARY:On-Call
```
Auchindoun: Should we include the schedule name here, since users will be able to subscribe to multiple schedules? e.g. `On-call for GoAlert Primary Schedule`

Or perhaps we could set the calendar name under the `VCALENDAR` section (a `NAME` property is supported for the calendar title: https://icalendar.org/New-Properties-for-iCalendar-RFC-7986/5-1-name-property.html) and keep the `SUMMARY` simple for each `VEVENT` as something like `On-call shift`

arurao: That sounds good! Would it also be worth including the calendar subscription name too?

calendarsubscription/calendarsubscription_test.go
mastercactapus: The suggested change would make this a fixed array instead of a slice of length 2.

mastercactapus: If you omit the 2, and use `[]int{5,10}` though you'd get what you want, and could remove the subsequent lines

calendarsubscription/calendarsubscription.go
mastercactapus: No space is correct:
https://golang.org/pkg/encoding/json/#Marshal

calendarsubscription/calendarsubscription.go
mastercactapus: Let's wait until the final pass when we test everything together on actual devices to add names -- I don't want us to spin until we can actually try it.

Then we can try different things, see what is the best UX for diff. calendar providers.

calendarsubscription/calendarsubscription.go
```diff
@@ -46,3 +50,45 @@ func (cs CalendarSubscription) Normalize() (*CalendarSubscription, error) {
 
 	return &cs, nil
 }
+
+func (cs CalendarSubscription) RenderICalFromShifts(shifts []oncall.Shift, reminderMinutes []int) ([]byte, error) {
+	type iCalOptions struct {
+		Shifts          []oncall.Shift `json:"s,omitempty"`
+		ReminderMinutes []int          `json:"r, omitempty"`
```
mastercactapus: Remove space

calendarsubscription/calendarsubscription.go
arurao: Sure! We can add more UX then after some testing, I'll revert my changes for the name field for now.

calendarsubscription/calendarsubscription.go
```diff
@@ -46,3 +50,47 @@ func (cs CalendarSubscription) Normalize() (*CalendarSubscription, error) {
...
+	type iCalOptions struct {
+		Shifts          []oncall.Shift `json:"s,omitempty"`
+		ReminderMinutes []int          `json:"r,omitempty"`
+		ScheduleName    string         `json:"n,omitempty"`
+	}
+
+	var iCalTemplate = `
+		BEGIN:VCALENDAR
+		VERSION:2.0
+		PRODID:-//GoAlert//Calendar Subscriptions//EN
```
mastercactapus: Hmm, should we include the version? Also I think the cal-sub bit is redundant (since it's already a calendar).
>       ;Any text that describes the product and version
>       ;and that is generally assured of being unique.

Maybe this should be `//GoAlert//{{.Version}}//EN` (the current version of GoAlert, like in page source).

It would help if we generated bogus iCal data with a specific version, for example.

calendarsubscription/calendarsubscription.go
```diff
@@ -46,3 +50,45 @@ func (cs CalendarSubscription) Normalize() (*CalendarSubscription, error) {
...
+		BEGIN:VALARM
+		ACTION:DISPLAY
+		DESCRIPTION:REMINDER
+		TRIGGER:-PT{{.}}M
+		END:VALARM
+		{{end}}
+
+		END:VCALENDAR`
+
+	iCal := template.Must(template.New("iCal").Parse(iCalTemplate))
```
mastercactapus: Let's do this outside the function as we do in `search.go` and `playground.go`, no need to recompile. Also `Must` will cause a panic at runtime if used like this (and the template was invalid).

calendarsubscription/calendarsubscription.go
```diff
@@ -46,3 +50,45 @@ func (cs CalendarSubscription) Normalize() (*CalendarSubscription, error) {
...
 }
+
+func (cs CalendarSubscription) renderICalFromShifts(shifts []oncall.Shift, reminderMinutes []int) ([]byte, error) {
+	type iCalOptions struct {
+		Shifts          []oncall.Shift `json:"s,omitempty"`
+		ReminderMinutes []int          `json:"r,omitempty"`
+	}
+
+	var iCalTemplate = `
+		BEGIN:VCALENDAR
```
mastercactapus: We should add a comment above this template to the RFC it's built against, I'm assuming this one: https://tools.ietf.org/html/rfc5545

calendarsubscription/calendarsubscription_test.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/target/goalert/oncall"
+	"github.com/target/goalert/permission"
+)
+
+func TestRenderICalFromShifts(t *testing.T) {
+	check := func(ctx context.Context, schedID string, userID string) {
+		t.Run("ical", func(t *testing.T) {
+			cs := CalendarSubscription{}
+
+			t1, _ := time.Parse(time.RFC3339, "2020-01-01T22:08:41+00:00")
```
mastercactapus: Use `time.Date` instead of parsing for tests

calendarsubscription/calendarsubscription_test.go
```diff
@@ -0,0 +1,35 @@
...
+	"time"
+
+	"github.com/target/goalert/oncall"
+	"github.com/target/goalert/permission"
+)
+
+func TestRenderICalFromShifts(t *testing.T) {
+	check := func(ctx context.Context, schedID string, userID string) {
+		t.Run("ical", func(t *testing.T) {
+			cs := CalendarSubscription{}
```
mastercactapus: Use `var cs CalendarSubscription` for an empty declaration

calendarsubscription/calendarsubscription_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+			t1, _ := time.Parse(time.RFC3339, "2020-01-01T22:08:41+00:00")
+			t2, _ := time.Parse(time.RFC3339, "2020-01-07T22:08:41+00:00")
+
+			s1 := oncall.Shift{UserID: "cb75f78a-0f7c-42fa-99f8-6b30e92a9518", Start: t1, End: t2}
+			shifts := []oncall.Shift{}
```
mastercactapus: We can move the line above this inside:

```go
shifts := []oncall.Shift{{UserID: ""...}}
```

calendarsubscription/calendarsubscription.go
```diff
@@ -46,3 +50,45 @@ func (cs CalendarSubscription) Normalize() (*CalendarSubscription, error) {
 
 	return &cs, nil
 }
+
+func (cs CalendarSubscription) renderICalFromShifts(shifts []oncall.Shift, reminderMinutes []int) ([]byte, error) {
```
mastercactapus: `reminderMinutes` is already available in `cs.Config.ReminderMinutes`, no need for the param

oncall/store.go
```diff
@@ -3,14 +3,16 @@ package oncall
 import (
 	"context"
 	"database/sql"
+
```
mastercactapus: nit: extra whitespace

smoketest/calendarsubscription_test.go
```diff
@@ -0,0 +1,72 @@
...
+				reminderMinutes: [%d]
+				scheduleID: "%s",
+			}) {
+				url
+			}
+		}
+	`, "foobar", 5, h.UUID("schedId")), &cs)
+
+	// get token
+	s := strings.Split(cs.CreateUserCalendarSubscription.URL, "token=")
```
mastercactapus: Doing this could double-encode the token value, and will break if any other params are present as `url.Values` is re-encoded before sending and splitting does not decode.

We can simplify it a bit though:
- parse as URL, assert no err, assert the path is `/api/v2/calendar`
- make the GET request using the original URL string

We just want to know, whatever string we get, works, and said string points to the path we expect. We should do some assertion on the contents and content-type of the response as well (e.g. make sure we actually get iCal data)

smoketest/calendarsubscription_test.go
```diff
@@ -0,0 +1,72 @@
...
+	v := make(url.Values)
+	v.Set("token", s[1])
+
+	resp, err := http.PostForm(h.URL()+"/api/v2/calendar", v)
```
mastercactapus: Should be GET

calendarsubscription/http.go
```diff
@@ -43,8 +43,11 @@ func (s *Store) ServeICalData(w http.ResponseWriter, req *http.Request) {
 	}
 
 	_ = filtered
```
dctalbot: Sorry if this is another go-ism that I'm not familiar with, but is this line extraneous?

arurao: Good catch! `_` is used whenever there is a value that is not being used. This was there with the previous PR where we were stubbing out the call to the endpoint. But since that is no longer the case, we no longer need this line, since we will be passing `filtered` as an argument to the call.

mastercactapus: It was there in my PR because filtered wasn't being used, but should be removed now

mastercactapus: It's a compile error to have an unused var, so `_ = thing` is the convention to work around it for example code.

calendarsubscription/calendarsubscription.go
```diff
@@ -24,6 +29,36 @@ type CalendarSubscription struct {
...
+	{{range .ReminderMinutes}}
+	BEGIN:VALARM
+	ACTION:DISPLAY
+	DESCRIPTION:REMINDER
+	TRIGGER:-PT{{.}}M
+	END:VALARM
+	{{end}}
+
+	END:VCALENDAR
+`))
```
mastercactapus: Based on my understanding of the RFC, I think we need to adjust a few things to follow the spec:
- Move `BEGIN:VCALENDAR` to the first line
- Remove the tabs/indentation before each line
- `VALARM` should happen inside the `VEVENT`
- Replace LF with CRLF (e.g. `strings.ReplaceAll(`BEGIN....`, "\n", "\r\n")`)

calendarsubscription/calendarsubscription.go
```diff
@@ -46,3 +81,16 @@ func (cs CalendarSubscription) Normalize() (*CalendarSubscription, error) {
 
 	return &cs, nil
 }
+
+// RFC can be found at https://tools.ietf.org/html/rfc5545
```
mastercactapus: Let's put this next to the iCal template

calendarsubscription/calendarsubscription.go
```diff
@@ -35,28 +35,26 @@ type iCalOptions struct {
...
+var iCalTemplate = template.Must(template.New("ical").Parse(`BEGIN:VCALENDAR
+PRODID:-//GoAlert//{{.Version}}//EN
+CALSCALE:GREGORIAN
+METHOD:PUBLISH
+{{range .Shifts}}
+BEGIN:VEVENT
+SUMMARY:On-Call Shift
+DTSTART:{{.Start}}
+DTEND:{{.End}}
+{{range .ReminderMinutes}}
```
mastercactapus: Since we're inside the range of `.Shifts` we won't have access to `.ReminderMinutes` without putting it in a variable first.

```
--- FAIL: TestRenderICalFromShifts (0.00s)
    --- FAIL: TestRenderICalFromShifts/ical (0.00s)
        calendarsubscription_test.go:22: err = render ical template:: template: ical:10:8: executing "ical" at <.ReminderMinutes>: can't evaluate field ReminderMinutes in type oncall.Shift; want nil
```

Something like this:
```go
{{- $mins := .ReminderMinutes }}
{{range .Shifts}}
{{range $mins}}
```

calendarsubscription/calendarsubscription_test.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/target/goalert/oncall"
+)
+
+func TestRenderICalFromShifts(t *testing.T) {
```
mastercactapus: Since this is a method test, it should be named in the style `TestTypeName_MethodName`
```suggestion
func TestCalendarSubscription_RenderICalFromShifts(t *testing.T) {
```

smoketest/calendarsubscription_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"fmt"
+	"net/http"
+	"net/url"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/target/goalert/smoketest/harness"
+)
+
+func TestICal(t *testing.T) {
```
mastercactapus: Follow the filename:
```suggestion
func TestCalendarSubscription(t *testing.T) {
```

PR #371, opened 2020-01-22
calendarsubscription: restrict token based on timestamp

KatieMSB:

**Description:**
Adds created_at field to calendar subscriptions and sets the issued_at timestamp for tokens that must match the created_at field.

### DIFF
--- a/calendarsubscription/calendarsubscription.go
+++ b/calendarsubscription/calendarsubscription.go
@@ -14,6 +14,7 @@ type CalendarSubscription struct {
 	UserID     string
 	ScheduleID string
 	LastAccess time.Time
+	CreatedAt  time.Time
 	Disabled   bool
 
 	// Config provides necessary parameters CalendarSubscription Config (i.e. ReminderMinutes)
--- a/calendarsubscription/calendarsubscription.go
+++ b/calendarsubscription/calendarsubscription.go
@@ -14,7 +14,6 @@ type CalendarSubscription struct {
 	UserID     string
 	ScheduleID string
 	LastAccess time.Time
-	CreatedAt  time.Time
 	Disabled   bool
 
 	// Config provides necessary parameters CalendarSubscription Config (i.e. ReminderMinutes)
--- a/calendarsubscription/store.go
+++ b/calendarsubscription/store.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"encoding/json"
+	"time"
 
 	"github.com/dgrijalva/jwt-go"
 	"github.com/target/goalert/keyring"
@@ -54,7 +55,7 @@ func NewStore(ctx context.Context, db *sql.DB, apiKeyring keyring.Keyring, oc on
 
 		findOne: p.P(`
 			SELECT
-				id, name, user_id, disabled, schedule_id, config, last_access
+				id, name, user_id, disabled, schedule_id, config, last_access, created_at
 			FROM user_calendar_subscriptions
 			WHERE id = $1
 		`),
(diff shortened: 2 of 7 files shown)


calendarsubscription/calendarsubscription.go
```diff
@@ -14,6 +14,7 @@ type CalendarSubscription struct {
 	UserID     string
 	ScheduleID string
 	LastAccess time.Time
+	CreatedAt  time.Time
```
mastercactapus: We can leave this out of the struct, we only need it during Authorize as part of the query

calendarsubscription/store.go
```diff
@@ -188,9 +189,16 @@ func (s *Store) CreateTx(ctx context.Context, tx *sql.Tx, cs *CalendarSubscripti
 		return nil, err
 	}
 
+	var now time.Time
+	err = s.now.QueryRowContext(ctx).Scan(&now)
```
mastercactapus: We can't issue a new query while scanning; if we're in a transaction, we need this query to happen within the transaction, and since we're scanning we can't issue a new one till it's done reading.

Instead, we can pass IssuedAt as a parameter in the auth query so we don't need to pull anything out.

mastercactapus:
If we take a slightly different approach we can sidestep needing multiple queries.
- update `authUser` to pass in the IssuedAt value `WHERE .... AND created_at = $2`
- update `create` to add `RETURNING created_at` to used as the IssuedAt value

calendarsubscription/store.go
```diff
@@ -48,13 +49,13 @@ func NewStore(ctx context.Context, db *sql.DB, apiKeyring keyring.Keyring, oc on
 		authUser: p.P(`
 			UPDATE user_calendar_subscriptions
 			SET last_access = now()
-			WHERE NOT disabled AND id = $1
+			WHERE NOT disabled AND id = $1 AND created_at = $2
```
mastercactapus: I think `created_at` needs to be truncated before comparing -- the `IssuedAt` field only has a precision of seconds, but `TIMESTAMPTZ` is precise to microseconds. If I'm correct, it will be unlikely that the two will match during auth.
```suggestion
			WHERE NOT disabled AND id = $1 AND date_trunc('second', created_at) = $2
```

calendarsubscription/store.go
```diff
@@ -128,7 +130,7 @@ func (s *Store) Authorize(ctx context.Context, token string) (context.Context, e
 	}
 
 	var userID string
-	err = s.authUser.QueryRowContext(ctx, c.Subject).Scan(&userID)
+	err = s.authUser.QueryRowContext(ctx, c.Subject, c.IssuedAt).Scan(&userID)
```
mastercactapus: Do we need to cast this to `time.Time` first?
```suggestion
	err = s.authUser.QueryRowContext(ctx, c.Subject, time.Unix(c.IssuedAt, 0)).Scan(&userID)
```

calendarsubscription/store.go
```diff
@@ -48,13 +49,13 @@ func NewStore(ctx context.Context, db *sql.DB, apiKeyring keyring.Keyring, oc on
...
 			SET last_access = now()
-			WHERE NOT disabled AND id = $1
+			WHERE NOT disabled AND id = $1 AND date_trunc('second', created_at) = $2
 			RETURNING user_id
 		`),
 
 		findOne: p.P(`
 			SELECT
-				id, name, user_id, disabled, schedule_id, config, last_access
+				id, name, user_id, disabled, schedule_id, config, last_access, created_at
```
mastercactapus: I think this was leftover: `sql: expected 8 destination arguments in Scan, not 7`
```suggestion
				id, name, user_id, disabled, schedule_id, config, last_access
```

PR #372, opened 2020-01-22
ui: migrate from moment to luxon

dctalbot:

**Description:**
Migrates any use of the datetime formatter Moment to use Luxon which is much smaller in bundle size. (Although Moment is a dependency of cypress; thus `yarn.lock` remains unchanged).

**User-facing Changes:**
`CalendarEventWrapper.js` now includes timezone

Before:
Screen Shot 2020-01-22 at 3 08 40 PM
After:
Screen Shot 2020-01-22 at 3 08 57 PM

### DIFF
--- a/web/src/app/schedules/CalendarEventWrapper.js
+++ b/web/src/app/schedules/CalendarEventWrapper.js
@@ -99,7 +99,7 @@ export default class CalendarEventWrapper extends Component {
     const { classes, event } = this.props
 
     let overrideCtrls = null
-    if (DateTime.fromJSDate(event.end).toUTC() > DateTime.utc()) {
+    if (DateTime.fromJSDate(event.end) > DateTime.utc()) {
       overrideCtrls = (
         <React.Fragment>
           <Grid item className={classes.buttonContainer}>
--- a/web/src/app/schedules/CalendarEventWrapper.js
+++ b/web/src/app/schedules/CalendarEventWrapper.js
@@ -7,7 +7,6 @@ import RemoveIcon from '@material-ui/icons/Delete'
 import Tooltip from '@material-ui/core/Tooltip'
 import withStyles from '@material-ui/core/styles/withStyles'
 import { connect } from 'react-redux'
-import moment from 'moment'
 import { urlParamSelector } from '../selectors'
 import { DateTime, Duration } from 'luxon'
 
@@ -100,7 +99,7 @@ export default class CalendarEventWrapper extends Component {
     const { classes, event } = this.props
 
     let overrideCtrls = null
-    if (moment(event.end).isAfter(moment())) {
+    if (DateTime.fromJSDate(event.end).toUTC() > DateTime.utc()) {
       overrideCtrls = (
         <React.Fragment>
           <Grid item className={classes.buttonContainer}>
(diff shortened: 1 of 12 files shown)


web/src/app/util/luxon-helpers.test.js
```diff
@@ -0,0 +1,73 @@
...
+  return new Date(new Date(saturday).setHours(23, 59, 59, 999))
+}
+
+test('it should yield 12am on Sunday of the week of the date given', () => {
+  // get beginning of today's week
+  const result1 = getLuxonStartOfWeek()
+  const result2 = getNativeStartOfWeek()
+  expect(result1.toMillis()).toBe(result2.getTime())
+
+  const davidsBirthday = new Date('November 26, 1997 03:24:00') // wednesday
```
arurao: How about renaming these to something generic like `d1` or `date1` and so on?

web/src/app/util/luxon-helpers.test.js
```diff
@@ -0,0 +1,73 @@
...
+  expect(getLuxonStartOfWeek(midnight).toMillis()).toBe(midnight.getTime())
+})
+
+test('it should yield almost midnight on Saturday of the week of the date given', () => {
+  // get end of today's week
+  const result1 = getLuxonEndOfWeek()
+  const result2 = getNativeEndOfWeek()
+  expect(result1.toMillis()).toBe(result2.getTime())
+
+  const davidsBirthday = new Date('November 26, 1997 03:24:00') // wednesday
```
arurao: same as above

web/src/app/schedules/CalendarEventWrapper.js
```diff
@@ -100,7 +99,7 @@ export default class CalendarEventWrapper extends Component {
     const { classes, event } = this.props
 
     let overrideCtrls = null
-    if (moment(event.end).isAfter(moment())) {
+    if (DateTime.fromJSDate(event.end).toUTC() > DateTime.utc()) {
```
mastercactapus: `toUTC` shouldn't be necessary, `#valueOf` returns the epoch timestamp which would already account for timezone differences.

web/src/app/util/luxon-helpers.js
```diff
@@ -0,0 +1,33 @@
+import { DateTime } from 'luxon'
+
+// getLuxonStartOfWeek
+// Given a JS Date object, this yields a DateTime object set to the previous Sunday at 12am.
+// If no date is provided, default to now
+// This is excluded from Luxon core since certain countries start the week on Saturday or Sunday
+// Conventionally, weeks start on a Monday
+export function getLuxonStartOfWeek(JSDate = new Date()) {
```
mastercactapus: I think it would be clearer to just call this `getStartOfWeek` or `startOfWeek` and operate on `DateTime` directly, doing the to/from jsDate stuff only in the specific places we need to (i.e. calendar) instead of in the helper.

web/src/app/schedules/CalendarEventWrapper.js
dctalbot: Right, but I feel like this is easier to read than 
`DateTime.fromJSDate(event.end) > DateTime.local()`

mastercactapus: These are all identical:
- `DateTime.fromJSDate(event.end) > DateTime.local()`
- `DateTime.fromJSDate(event.end) > DateTime.utc()`
- `DateTime.fromJSDate(event.end).toUTC() > DateTime.local()`
- `DateTime.fromJSDate(event.end).toUTC() > DateTime.utc()`

But the `.toUTC()` implies it's necessary before comparing (i.e. for future code) and is just extra noise. Anyone looking at it is going to wonder why it's there, or assume they need it.

mastercactapus: My preference would be `DateTime.fromJSDate(event.end) > DateTime.utc()`

PR #389, opened 2020-01-28
smoketests: Migrate system limit tests to graphQL2

dctalbot:

**Description:**
Migrates system limit tests over to new graphql2 endpoint


smoketest/harness/graphql.go
```diff
@@ -63,6 +64,20 @@ func (h *Harness) SetConfigValue(id, value string) {
 	h.Trigger()
 }
 
+// SetSystemLimit will update the value of a system limit given an id (e.g. `RulesPerSchedule`).
+func (h *Harness) SetSystemLimit(id limit.ID, value int) {
```
Auchindoun: I'm not sure if a func to directly manipulate the DB is appropriate in the harness' `graphql.go` file, especially if it's not in the current schema

smoketest/systemlimits_test.go
```diff
@@ -59,7 +60,11 @@ func TestSystemLimits(t *testing.T) {
 		insert into escalation_policy_steps (id, escalation_policy_id, delay)
 		values
 			({{uuid "act_ep_step"}}, {{uuid "act_ep"}}, 15),
-			({{uuid "act_ep_step2"}}, {{uuid "act_ep2"}}, 15);
+			({{uuid "act_ep_step2"}}, {{uuid "act_ep2"}}, 15),
+			({{uuid "ep_step1"}}, {{uuid "step_ep"}}, 15),
```
Auchindoun: How does `act_ep_step` differ from `ep_step`?

smoketest/systemlimits_test.go
```diff
@@ -149,8 +154,21 @@ func TestSystemLimits(t *testing.T) {
...
+			if isMultiInsert {
+				noErr(doQLErr(t, addQuery(3), parseID))
+				setLimit(2)
+				mustErr(doQLErr(t, addQuery(4), parseID))
+				noErr(doQLErr(t, addQuery(1), parseID))
+				noErr(doQLErr(t, addQuery(2), parseID))
+				mustErr(doQLErr(t, addQuery(3), parseID))
+				setLimit(-1)
+				noErr(doQLErr(t, addQuery(4), parseID))
+				return
```
Auchindoun: Could you add a little info here for this bit? The random numbers/func calls seem a bit magical without context.

smoketest/harness/graphql.go
dctalbot: We didn't want to update the api for test purposes only, but yeah we will want to transition this once the admin config page is implemented. No strong opinion on this though. @mastercactapus what do you think?

mastercactapus: The test harness should be okay for this, for now. All the limits are implemented as DB triggers, so we don't need to sync with the engine or anything and will work as intended. The harness accesses the DB directly for things like time-shifting, so it's not a new connection.

But it is intended to be temporary until we build a new mutation & query. We decided not to add any new GraphQL mutations in this scope, and instead leave that for when we add limits config it to the Admin page so we can account for any "hidden" requirements. At that time we will update this to use the mutation instead of direct-to-DB.

However, it might be worth adding a comment inside the method to indicate that the query is only temporary, and should be replaced with a GQL call once system limits have been added to the schema. 

@dctalbot would you mind adding that?

Auchindoun: Sounds good, yeah a `todo` that makes this known would be preferable.

PR #393, opened 2020-01-29
deps: ignore @date-io/luxon <= 2.2.0

dctalbot:

**Description:**
Addresses #391 
and #402

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -6,4 +6,4 @@ update_configs:
     ignored_updates:
       - match:
           dependency_name: "@date-io/luxon"
-          version_requirement: "<=2.0.1"
+          version_requirement: "<=2.2.0"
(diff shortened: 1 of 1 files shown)


Auchindoun:
Is there an issue with the latest Luxon release(s) and our project?

.dependabot/config.yml
```diff
@@ -5,8 +5,5 @@ update_configs:
     update_schedule: "daily"
     ignored_updates:
       - match:
-          dependency_name: "cypress"
-          version_requirement: "<=3.7.0"
-      - match:
-          dependency_name: "react-infinite-scroll-component"
-          version_requirement: "<=5.0.3"
```
Auchindoun: Have you verified that we are good to start upgrading this package again?

dctalbot: We are already on `5.0.4` so this is stale as is

dctalbot: Can't recall why it was ignored originally, but it seems to be working as intended now

Auchindoun: Sounds good

PR #396, opened 2020-01-30
feature/cal-subscriptions: final draft

arurao:

**Description:**
This PR wraps up some loose ends for the calendar subscriptions work. It makes sure that the generated url shows up on the UI and is accessible. It also fixes the start and end times in the ical format.  It adds `VERSION` and `DTSTAMP` to the iCalendar template.


web/src/app/schedules/calendar-subscribe/CalendarSuccessForm.js
```diff
@@ -25,13 +25,14 @@ const useStyles = makeStyles(theme => ({
 
 export default function CalenderSuccessForm(props) {
   const classes = useStyles()
+  const url = props.url.replace('https://', '')
```
dctalbot:
```suggestion
  const url = props.url.replace('https://', 'webcal://')
```

We can combine the `.replace` and concat (line 35) methods into one line here. That way, the `url` var is already formatted correctly wherever it's used

web/src/app/schedules/calendar-subscribe/CalendarSuccessForm.js
```diff
@@ -25,13 +25,14 @@ const useStyles = makeStyles(theme => ({
...
   return (
     <Grid container spacing={2}>
       <Grid item xs={12} className={classes.subscribeButtonContainer}>
         <Button
           color='primary'
           variant='contained'
-          href={'webcal://' + props.url}
+          href={'webcal://' + url}
```
dctalbot:
```suggestion
          href={url}
```
See comment above

web/src/app/schedules/calendar-subscribe/CalendarSuccessForm.js
```diff
@@ -25,13 +25,14 @@ const useStyles = makeStyles(theme => ({
 
 export default function CalenderSuccessForm(props) {
   const classes = useStyles()
+  const url = props.url.replace('https://', 'webcal://')
```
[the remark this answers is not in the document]
m17ch: Handled in https://github.com/target/goalert/pull/396/commits/fbb40fead76f6bc74509e95be7a994c9d1d0529f

calendarsubscription/calendarsubscription.go
```diff
@@ -34,19 +34,23 @@ type iCalOptions struct {
 	Shifts          []oncall.Shift `json:"s,omitempty"`
 	ReminderMinutes []int          `json:"r,omitempty"`
 	Version         string         `json:"v,omitempty"`
+	Generated       time.Time      `json:"g,omitempty"`
```
mastercactapus: Can this be `GeneratedAt`? Also, instead of `iCalOptions` maybe this should be `iCalRenderData` or something like that as it's specific and only used by the template.

PR #400, opened 2020-01-30
feature/cal-subscriptions: admin kill switch to disable endpoint access and create actions

Auchindoun:

**Description:**
This PR creates an admin config option to disable the ability to access the calendar endpoint, as well as when attempting to create a new calendar subscription via GraphQL.

**Screenshots:**
Screen Shot 2020-01-30 at 3 59 20 PM
Screen Shot 2020-01-30 at 4 34 54 PM


calendarsubscription/store.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"encoding/json"
+	"github.com/target/goalert/config"
```
mastercactapus: nit: sort with `goimports`

smoketest/calendarsubscription_test.go
```diff
@@ -3,24 +3,23 @@ package smoketest
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/stretchr/testify/assert"
+	"github.com/target/goalert/smoketest/harness"
```
mastercactapus: goimports

calendarsubscription/http.go
```diff
@@ -12,7 +12,7 @@ import (
 func (s *Store) ServeICalData(w http.ResponseWriter, req *http.Request) {
 	ctx := req.Context()
 	src := permission.Source(ctx)
-	if src.Type != permission.SourceTypeCalendarSubscription {
+	if src.Type != permission.SourceTypeCalendarSubscription || isCreationDisabled(ctx) {
```
mastercactapus: Instead of checking this here, do it in the `Authorize` method so that the context is never given the authorization to start.

It also might be clearer to just do the cfg check in both places without a helper since it's a bool check

Auchindoun: Sounds good

Auchindoun: Yeah, I made the helper as part of the AC for the story but since it's small enough it would make sense to do it inline where needed

Auchindoun:
More renderer crashes on CI. Will run full suite locally and report back with results.

PR #410, opened 2020-02-04
smoketests: migrate queries to use GraphQL 2

dctalbot:

**Description:**
This removes the `GraphlQLQuery` method in favor of `GraphlQLQuery2` without introducing any new functionality or features.

### DIFF
--- a/smoketest/graphqlalert_test.go
+++ b/smoketest/graphqlalert_test.go
@@ -194,12 +194,13 @@ func TestGraphQLAlert(t *testing.T) {
 		mutation {
 			createEscalationPolicyStep(input:{
 				delayMinutes: 60,
-				escalationPolicyID: "%s"
+				escalationPolicyID: "%s",
+				targets: [{id: "%s", type: user}, {id: "%s", type: schedule}]
 			}){
 				id
 			}
 		}
-	`, esc.CreateEscalationPolicy.ID), &step)
+	`, esc.CreateEscalationPolicy.ID, uid2, sched.CreateSchedule.ID), &step)
 	var svc struct{ CreateService struct{ ID string } }
 	doQL(fmt.Sprintf(`
 		mutation {
--- a/smoketest/graphqlalert_test.go
+++ b/smoketest/graphqlalert_test.go
@@ -67,7 +67,10 @@ func TestGraphQLAlert(t *testing.T) {
 			ID string
 		}
 	}
-	doQL(fmt.Sprintf(`
+
+	createCM := func(userID, phone string, cm interface{}) harness.TwilioExpectedMessage {
+		t.Helper()
+		doQL(fmt.Sprintf(`
 		mutation {
 			createUserContactMethod(input:{
 				userID: "%s",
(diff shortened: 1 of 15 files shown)


smoketest/harness/harness.go
```diff
@@ -659,26 +659,26 @@ func (h *Harness) WaitAndAssertOnCallUsers(serviceID string, userIDs ...string)
 	getUsers := func() []string {
 		var result struct {
 			Service struct {
-				OnCall []struct {
-					UserID   string `json:"user_id"`
-					UserName string `json:"user_name"`
-				} `json:"on_call_users"`
+				OnCallUsers []struct {
+					UserID   string `json:"userID"`
```
mastercactapus: We can drop all these JSON tags, since they now match the field names (unmarshaling is already case-insensitive).

smoketest/graphqlalert_test.go
```diff
@@ -59,11 +62,15 @@ func TestGraphQLAlert(t *testing.T) {
 	uid1, uid2 := h.UUID("u1"), h.UUID("u2")
 	phone1, phone2 := h.Phone("u1"), h.Phone("u2")
 
-	var cm1, cm2 struct{ CreateContactMethod struct{ ID string } }
+	var cm1, cm2 struct {
+		CreateUserContactMethod struct {
+			ID string `json:"id"`
+		} `json:"createUserContactMethod"`
```
mastercactapus: We can drop the JSON tags

smoketest/graphqlalert_test.go
```diff
@@ -59,11 +62,15 @@ func TestGraphQLAlert(t *testing.T) {
...
+	}
 	doQL(fmt.Sprintf(`
 		mutation {
-			createContactMethod(input:{
-				user_id: "%s",
+			createUserContactMethod(input:{
```
mastercactapus: Hmm. at this point is it worth a helper func?

Something like:

```go
createCM := func(userID, phone string, cm interface{}) TwilioExpectedMessage {
    t.Helper()
	doQL(fmt.Sprintf(`
		mutation {
			createUserContactMethod(input:{
				userID: "%s",
				name: "default",
				type: SMS,
				value: "%s"
			}) {
				id
			}
		}
    `, userID, phone), cm)
    return h.Twilio().Device(phone).ExpectSMS("verification")
}

msg1 := createCM(uid1, phone1, &cm1)
msg2 := createCM(uid2, phone2, &cm2)
h.Twilio().WaitAndAssert()
```

smoketest/graphqlalert_test.go
```diff
@@ -87,114 +94,190 @@ func TestGraphQLAlert(t *testing.T) {
...
+
+	tw := h.Twilio()
+	d1 := tw.Device(phone1)
+	d2 := tw.Device(phone2)
+
+	msg1 := d1.ExpectSMS("verification")
+	msg2 := d2.ExpectSMS("verification")
+	tw.WaitAndAssert() // wait for code, and ensure no notifications went out
+
+	codeStr1 := strings.Map(func(r rune) rune {
```
mastercactapus: Probably just break out the map func instead of copying it, since they are used right here

```go
digits := func(r rune) rune {
    // ...
}
codeStr1 := strings.Map(digits, msg1.Body())
codeStr2 := strings.Map(digits, msg2.Body())
```

smoketest/graphqlupdaterotation_test.go
```diff
@@ -42,82 +42,100 @@ func TestGraphQLUpdateRotation(t *testing.T) {
...
-				shift_length: 2,
-				schedule_id: "%s"
-			}){
-				rotation {
-					id
-					name
-				}
-			}
+				name: "new name",
+			})
```
mastercactapus: We should test updating all the fields, though previously we were only asserting on shift length.

Let's update it to check all the fields after updating, and can use the `assert` library to make it cleaner instead of the manual `if` checks.

```go
assert.Equal(t, "new name", .Name)
assert.Equal(t, 2, .ShiftLength)
```

smoketest/graphqluserfavorites_test.go
```diff
@@ -46,77 +47,83 @@ func TestGraphQLUserFavorites(t *testing.T) {
...
+				}
+				favorite: true
+			})
 		}
 	`, h.UUID("sid1")), nil)
 
 	var s struct {
 		Service struct {
-			IsUserFav bool `json:"is_user_favorite"`
+			IsFavoite bool `json:"isFavorite"`
```
mastercactapus: Fix spelling and remove JSON tags

smoketest/graphqlusers_test.go
```diff
@@ -40,19 +40,24 @@ func TestGraphQLUsers(t *testing.T) {
 	}
 
 	var res struct {
-		Users []user.User
+		Users struct {
+			Nodes []struct {
+				ID string `json:"id"`
```
mastercactapus: Drop the tags

smoketest/listalerts_test.go
```diff
@@ -40,9 +45,14 @@ func TestListAlerts(t *testing.T) {
 	}
 
 	var res struct {
-		Alerts []struct {
-			ID      int
-			Service struct{ Name string }
+		Alerts struct {
+			Nodes []struct {
+				ID      int `json:"id,string"`
```
mastercactapus: Should be `json:"alertID"`

smoketest/graphqlalert_test.go
dctalbot: I wondered the same, I'll try a refactor once I address other comments and see how it feels

smoketest/listalerts_test.go
dctalbot: Not sure that's right

This is the resp body:
```json
{
  "data": {
    "alerts": {
      "nodes": [
        {
          "id": "79766",
          "service": {
            "id": "7f2e3509-2806-4148-8059-59a17e5495ad",
            "name": "Chief content-based Mobility Service"
          }
        },
...
}
```

PR #416, opened 2020-02-05
ui/accessibility: make escalation policy form fields focusable

dctalbot:

**Description:**
Previously, keyboard users did not receive visual feedback when tabbing through the escalation policy form. This resolves that issue by adding the `tabindex` attribute where needed.

**Screenshots:**

Screen Shot 2020-02-05 at 12 08 46 PM

Keyboard users are now able to receive visual feedback via the browser's default outline when tabbing through focusable form fields

**Other:**
Adds `aria-expanded` attribute to buttons of active accordion fields

### DIFF
--- a/web/src/app/escalation-policies/PolicyStepForm.js
+++ b/web/src/app/escalation-policies/PolicyStepForm.js
@@ -136,6 +136,7 @@ export default class PolicyStepForm extends React.Component {
                   <Step>
                     <StepButton
                       aria-label='Show Rotations Select'
+                      aria-expanded={(step === 0).toString()}
                       data-cy='rotations-step'
                       icon={<RotationsIcon />}
                       optional={optionalText}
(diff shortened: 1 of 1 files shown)


web/src/app/escalation-policies/PolicyStepForm.js
```diff
@@ -103,6 +103,7 @@ export default class PolicyStepForm extends React.Component {
         classes={{
           badge: classes.badge,
         }}
+        tabIndex='0'
       >
```
mastercactapus: We can add an aria label here, something like "toggle {txt}"

web/src/app/escalation-policies/PolicyStepForm.js
```diff
@@ -103,6 +103,8 @@ export default class PolicyStepForm extends React.Component {
         classes={{
           badge: classes.badge,
         }}
+        tabIndex='0'
+        aria-label={`toggle ${txt}`}
```
Auchindoun: nit:
```suggestion
        aria-label={`Toggle ${txt}`}
```

since output would be `toggle Add Users`

PR #417, opened 2020-02-05
cypress: update support functions to utilize gql2

Auchindoun:

**Description:**
This PR updates several support commands found under `web/src/cypress/support` using the old `graphql` api to the newer, more supported `graphql` api.

**Out of Scope:**
- Alert tests, to be completed in tandem with updating the alert UI components to the newer api.

### DIFF
--- a/web/src/cypress/support/profile.ts
+++ b/web/src/cypress/support/profile.ts
@@ -217,7 +217,7 @@ function clearContactMethods(id: string): Cypress.Chainable {
   })
 }
 
-function resetProfile(prof?: Profile): Cypress.Chainable<boolean> {
+function resetProfile(prof?: Profile): Cypress.Chainable {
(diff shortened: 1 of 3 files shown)


web/src/cypress/support/profile.ts
```diff
@@ -135,75 +142,94 @@ function addNotificationRule(
...
+            id: cm.id,
+          },
+        ],
+      })
     })
   })
 }
-function resetProfile(prof?: Profile): Cypress.Chainable<Profile> {
+
+function resetProfile(prof?: Profile): Cypress.Chainable<boolean> {
```
mastercactapus: Since this should always be true, or it would error/fail, maybe just return `Cypress.Chainable` like we do with `clearContactMethods` so it's not tied to the mutation at all.

PR #421, opened 2020-02-05
ui/material-select: set isClearable to always true

KatieMSB:

**Description:**
Sets isClearable in material-select component to always be true for all fields. If required, field will be reset to last set value (or default) after losing focus.

### DIFF
--- a/web/src/app/selection/MaterialSelect.js
+++ b/web/src/app/selection/MaterialSelect.js
@@ -111,7 +111,6 @@ export default class MaterialSelect extends Component {
             ) {
               this.setState({ isCleared: false })
             }
-            onChange(value)
           }}
           onChange={val => {
             if (required && val === null) {
--- a/web/src/app/selection/MaterialSelect.js
+++ b/web/src/app/selection/MaterialSelect.js
@@ -27,6 +27,7 @@ export default class MaterialSelect extends Component {
     this.state = {
       isCleared: false,
     }
+    this.buttonClicked = false
   }
 
   static propTypes = {
@@ -83,11 +84,26 @@ export default class MaterialSelect extends Component {
     }
     const { isCleared } = this.state
 
+    const isDescendent = (parent, child) => {
+      if (!child) return false
+      if (child.isSameNode(parent)) return true
+      return isDescendent(parent, child.parentNode)
+    }
+
     return (
+      // going to look into replacing component with material-ui autocomplete component
+      // eslint-disable-next-line jsx-a11y/click-events-have-key-events,jsx-a11y/no-static-element-interactions
       <div
         data-cy='material-select'
         data-cy-ready={!props.isLoading}
         className={classes.root}
+        onMouseDownCapture={e => {
+          if (isDescendent(this.clearButtonRef.current, e.target))
+            this.buttonClicked = true
+        }}
+        onClick={() => {
+          this.buttonClicked = false
+        }}
       >
         <Select
           menuPortalTarget={document.body}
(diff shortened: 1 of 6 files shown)


web/src/app/selection/QuerySelect.js
```diff
@@ -96,7 +96,8 @@ function makeUseValues(query, mapNode) {
 
     const result = value.map((v, i) => {
       const name = 'data' + i
-      if (!data || !data[name]) return { value: v, label: 'Loading...' }
+      if (!data || !data[name] || _.isEmpty(data[name]))
```
dctalbot:
```suggestion
      if (!data || _.isEmpty(data[name]))
```

```
_.isEmpty("") === _.isEmpty(null) === _.isEmpty(undefined)
```

web/src/cypress/support/dialog.ts
```diff
@@ -62,8 +65,23 @@ function dialogFinish(s: string): Cypress.Chainable {
...
+      return cy.get(selector).clear()
+    })
+}
+
 Cypress.Commands.add('dialogFinish', dialogFinish)
 Cypress.Commands.add('dialogTitle', dialogTitle)
 Cypress.Commands.add('dialogForm', dialogForm)
 Cypress.Commands.add('dialogContains', dialogContains)
 Cypress.Commands.add('dialogClick', dialogClick)
+Cypress.Commands.add('dialogClearField', dialogClearField)
```
mastercactapus: This should be handled by setting a form field to an empty string (with `form()` or `dialogForm()`)

web/src/app/selection/MaterialSelect.js
```diff
@@ -44,6 +52,7 @@ export default class MaterialSelect extends Component {
       classes,
       disabled,
       required,
+      onBlur,
```
mastercactapus: Leftover?

web/src/app/selection/MaterialSelect.js
```diff
@@ -88,12 +98,27 @@ export default class MaterialSelect extends Component {
...
           components={components}
-          onChange={onChange}
+          onBlur={e => {
+            if (
+              isCleared &&
+              e.relatedTarget !== this.clearButtonRef.current &&
+              required
+            ) {
+              this.setState({ isCleared: false })
+              onChange(value)
```
mastercactapus: No need to fire `onChange` since the value doesn't change.

web/src/app/selection/MaterialSelect.js
```diff
@@ -88,12 +98,27 @@ export default class MaterialSelect extends Component {
...
+            }
+          }}
+          onChange={val => {
+            if (!required) onChange(val)
```
mastercactapus: It looks like this could fire `onChange` twice, let's invert it so that calling `onChange` is the primary action.

Somethine like:

```js
if (required && val === null) {
    // internally track clear state, without calling onChange if required is true
    this.setState({isCleared: true})
    return
}

onChange(val)
```

web/src/app/selection/MaterialSelect.js
KatieMSB: Yes, removed now.

web/src/app/selection/MaterialSelect.js
```diff
@@ -88,12 +97,30 @@ export default class MaterialSelect extends Component {
...
-          onChange={onChange}
+          onBlur={e => {
+            if (
+              isCleared &&
+              e.relatedTarget !== this.clearButtonRef.current &&
+              required
+            ) {
+              this.setState({ isCleared: false })
+            }
+            onChange(value)
```
mastercactapus: I don't think we need this in `onBlur` at all, do we?

web/src/app/selection/MaterialSelect.js
```diff
@@ -81,6 +82,10 @@ export default class MaterialSelect extends Component {
       value: value ? (multiple ? value.join(',') : value.value) : '',
     }
     const { isCleared } = this.state
+    const getValue = () => {
+      if (multiple) return
```
mastercactapus: Doesn't this need to return the current value?
```suggestion
      if (multiple) return value
```

KatieMSB: Returning the value won't clear the field.

mastercactapus: As-is this will result in `undefined` being passed as the `value` prop to the `Select` component in **all** cases (for multi select), clear or not. Previously we were passing the value directly, which will be an empty array if cleared.

Locally it seems to work either way without warnings, but if there is a reason `value` must always be `undefined` for multi-select, let's add a comment for future reference.

KatieMSB: Moving the condition to onChange works and won't return undefined.

PR #422, opened 2020-02-05
ui: fix alerts list pagination

Auchindoun:

**Description:**
Fixes pagination on alerts lists and adds a test to ensure a regression isn't caused in the future.

### DIFF
--- a/web/src/app/alerts/components/AlertsList.js
+++ b/web/src/app/alerts/components/AlertsList.js
@@ -333,6 +333,7 @@ export default class AlertsList extends Component {
             </Hidden>
             <List id='alerts-list' style={{ padding: 0 }} data-cy='alerts-list'>
               <InfiniteScroll
+                scrollableTarget='content'
                 next={() => loadMore(this.getQueryData(offset))}
                 dataLength={len}
                 hasMore={hasMore}
--- a/web/src/app/alerts/components/AlertsList.js
+++ b/web/src/app/alerts/components/AlertsList.js
@@ -28,6 +28,7 @@ import {
 } from '../../selectors/url'
 import AlertsListControls from '../components/AlertsListControls'
 import { LegacyGraphQLClient } from '../../apollo'
+import Spinner from '../../loading/components/Spinner'
 
 const LIMIT = 25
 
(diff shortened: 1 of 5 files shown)


web/src/app/alerts/components/AlertsList.js
```diff
@@ -331,18 +331,34 @@ export default class AlertsList extends Component {
...
                 {content}
-              </InfiniteScroll>
-            </List>
+              </List>
+            )}
+            {!isLoading && (
+              <List
+                id='alerts-list'
+                style={{ padding: 0 }}
+                data-cy='alerts-list'
```
dctalbot: Instead of two code blocks for the loading/loaded states, we could do
`data-cy={isLoading ? 'alerts-list-loading' : alerts-list }`
and use the `loader` prop

```
<InfiniteScroll
  loader={content}
```

We can also use a custom loading component, like a spinner or 'Loading...' text.

web/src/cypress/support/alert.ts
```diff
@@ -1,11 +1,14 @@
 import { Chance } from 'chance'
 import { DateTime } from 'luxon'
+import { sha512 } from 'js-sha512'
```
mastercactapus: I think we can just put generic text in there, it's generated as a hash, but just compared as a string. Something like `1:manual:createManyAlerts_#` where `#` is just a counter.

Then we don't need to worry about the overhead of calculating the hash, or the extra dep

Auchindoun: Interesting, I was messing around with a lot of strings and kept getting back responses that the value was invalid when inserting to the DB. I'll try this

web/src/cypress/support/alert.ts
```diff
@@ -172,6 +177,36 @@ function createAlert(a?: AlertOptions): Cypress.Chainable<Alert> {
...
+  }
+
+  // build query
+  let query =
+    'insert into alerts (service_id, summary, details, dedup_key) values '
+  let rows: Array<string> = []
+  for (let i = 0; i < count; i++) {
+    const summary = alertOptions.summary || c.word()
+    const details = alertOptions.details || c.sentence()
+    const dedupKey = 'manual:1:createManyAlerts_' + i
```
mastercactapus: Declare a variable outside this function, and increment it and use that instead of the `for` var. Otherwise if you call this twice on the same DB you'll get identical dedup keys.

PR #452, opened 2020-02-17
typescript: configuration and basic examples

dctalbot:

**Description:**
This introduces the minimal configuration required to compile typescript alongside our existing javascript.

**Additional Info:**
As part of our "developer life" initiative, we want to take advantage of the great tooling in the typescript ecosystem


web/src/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+  "compilerOptions": {
+    "strict": true,
+    "lib": ["es6", "dom"],
+    "moduleResolution": "node",
```
m17ch: I don't believe `moduleResolution` is needed per https://www.typescriptlang.org/v2/en/tsconfig#moduleResolution

I removed and ran `yarn fmt` and tsc did not yell at me.

web/src/webpack.config.js
```diff
@@ -64,15 +64,6 @@ module.exports = {
         test: /\.md$/,
         use: 'raw-loader',
       },
-      {
-        test: /\.ts$/,
-        exclude: [/node_modules/],
-        use: [
-          {
-            loader: 'ts-loader',
```
Forfold: Typescript loader no longer needed with these updates? 🤔

dctalbot: On line 40 you'll see we're now resolving typescript with the `babel-loader`, in conjunction with the `@babel/preset-typescript` preset. I don't think this loader config was being used anyway, since previously we only had typescript in /cypress and those files have their own bundle config i.e. `webpack.cy`press.js

web/src/app/util/SetFavoriteButton.tsx
```diff
@@ -1,11 +1,22 @@
 import React from 'react'
-import p from 'prop-types'
 import IconButton from '@material-ui/core/IconButton'
 import FavoriteFilledIcon from '@material-ui/icons/Star'
 import FavoriteBorderIcon from '@material-ui/icons/StarBorder'
 import Spinner from '../loading/components/Spinner'
 
-export function SetFavoriteButton({ typeName, isFavorite, loading, onClick }) {
+interface SetFavoriteButtonInput {
```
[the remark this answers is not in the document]
dctalbot: Good call, we should decide on a convention. How about
`xyzProps` for React components and `xyzInput` for misc. javascript functions?

PR #460, opened 2020-02-19
ui: replace react-select with @material-ui/lab/autocomplete

dctalbot:

**Description:**
This PR introduces no new functionality. It reimplements `react-select` with `@material-ui/lab/autocomplete` which is a more natural fit with our component library. 

There are stylistic changes that a user may notice, like no clearing X, a different drop-down icon, and changes in padding/margin

**Additional Info:**
Reference: https://material-ui.com/api/autocomplete/


web/src/app/selection/MaterialSelect.js
```diff
@@ -1,10 +1,152 @@
...
+  makeStyles,
+  MenuItem,
+  ListItemIcon,
+  Typography,
+  Paper,
+  Chip,
+} from '@material-ui/core'
+import { Autocomplete } from '@material-ui/lab'
+
+export const styles = () => ({
```
Forfold: I don't see this being imported anywhere, so we should be fine leaving the `export` out and removing the constant in favor of defining the styles within the func's parameter

web/src/app/selection/MaterialSelect.js
```diff
@@ -1,10 +1,152 @@
...
+    required,
+    value: propsValue,
+
+    // classes
+    // error: boolean
+    // fullWidth,
+    // hint,
+    // noOptionsMessage,
+    // placeholder: string
+    // theme,
```
Forfold: Still need?

dctalbot: Wasn't sure initially, but made some decisions and narrowed them down. We should chat briefly about placeholder and decide if we need it going forward

Forfold: Sounds good!

web/src/app/selection/MaterialSelect.js
```diff
@@ -1,10 +1,148 @@
...
+          let newInputVal = ''
+          if (!multiple) {
+            if (canCreate && options.length === 1) {
+              newInputVal = options[0].value
+            } else {
+              newInputVal = valueObj.label
+            }
+          }
+          setInputValue(newInputVal)
+        }
```
mastercactapus: - fire `onChange` handlers after input has been handled rather than before
- let's flatten this somehow
- what happens with 1 valid search result (not created) and `canCreate` is true? Can we set a flag on created options instead?

web/src/app/selection/QuerySelect.js
```diff
@@ -250,19 +250,20 @@ export function makeQuerySelect(displayName, options) {
       else onChange((newVal && newVal.value) || null)
     }
 
-    let noOptionsMessage = 'No options'
+    let noOptionsText = 'No options'
     if (optionsError) {
-      noOptionsMessage = (
+      noOptionsText = optionsError.message || optionsError
```
mastercactapus: leftover?

web/src/cypress/support/select-by-label.ts
```diff
@@ -22,12 +22,9 @@ type findByLabelFn = (label: string) => Cypress.Chainable
...
 function selectByLabel(sub: any, label: string): Cypress.Chainable {
-  return findByLabel(sub, label)
-    .click()
-    .get('[data-cy=select-dropdown]')
-    .should('not.exist')
-    .get('ul[role=listbox]')
-    .should('not.exist')
+  findByLabel(sub, label).click()
+
+  return cy.get('[data-cy=select-dropdown]').should('not.exist')
```
mastercactapus: These look equivalent, other than we no longer wait for the listbox to disappear, why the change?

web/src/app/selection/MaterialSelect.js
dctalbot: Removed the canCreate prop in favor of an option flag

web/src/cypress/support/select-by-label.ts
dctalbot: Yes you're correct, sorry it got a little mixed up (was seeing if I could avoid using the data-cy tag). Will revert and revisit since tests are flaking at the moment

web/src/app/selection/MaterialSelect.tsx
```diff
@@ -0,0 +1,170 @@
...
+interface MaterialSelectProps {
+  disabled?: boolean
+  error?: boolean
+  isLoading?: boolean
+  label?: string
+  multiple?: boolean // allow selecting multiple values
+  noOptionsText?: ReactNode
+  name?: string
+  required?: boolean
+  onChange: Function
```
mastercactapus: Let's define the signatures here instead of just using `Function`

Something like:
```ts
onChange: (value: string|string[]) => void
onInputChange: (value: string) => void
```

web/src/app/selection/MaterialSelect.tsx
```diff
@@ -0,0 +1,170 @@
...
+  makeStyles,
+  MenuItem,
+  ListItemIcon,
+  Typography,
+  Paper,
+  Chip,
+} from '@material-ui/core'
+import { Autocomplete } from '@material-ui/lab'
+
+function asArray(value: any) {
```
mastercactapus: Don't use any unless we have to, as it disables type checking.

web/src/app/selection/MaterialSelect.tsx
mastercactapus: Think we could swap props using `multiple`?

```ts
interface MaterialSelectCommonProps {
    disabled?: boolean
    // shared stuff
}
interface MaterialSelectSingleProps extends MaterialSelectCommonProps {
    multiple?: false
    value?: SelectOption
    onChange: (SelectOption) => void
}
interface MaterialSelectMultiProps extends MaterialSelectCommonProps {
    multiple: true
    value?: SelectOption[]
    onChange: (SelectOption[]) => void
}
```

web/src/cypress/support/select-by-label.ts
```diff
@@ -84,6 +84,10 @@ function findByLabel(sub: any, label: string): Cypress.Chainable {
 
       cy.get('[data-cy=select-dropdown]').should('not.contain', 'Loading')
 
+      cy.get('[data-cy=select-dropdown]')
```
mastercactapus: Isn't this doing the same assertion as the selector below?

web/src/app/selection/MaterialSelect.tsx
dctalbot: This was intentional. What do you have in mind?

web/src/cypress/support/select-by-label.ts
dctalbot: This resolves a flaky test. `.should()` will continually retry until the predicate is met whereas `.contains()` will not. However, we need to return a `Chainable` object which is why `.contains()` is still used here

web/src/app/selection/MaterialSelect.tsx
mastercactapus: Something like this to keep the types intact:

```ts
function asArray<T>(value?: T | T[]): T[] {
  if (!value) return []

  return Array.isArray(value) ? value : [value]
}
```

PR #462, opened 2020-02-20
ui: alerts list to graphql2

Forfold:

**Description:**
This PR updates the alerts list page and relevant components pertaining to the alerts list to use the `graphql2` endpoint, as well as converts majorly refactored components to hooks. Removes custom functionality of checkboxes and infinite scrolling and utilizes the new functionality of these features from within QueryList.

**Out of Scope:**
Converting files to TypeScript and some class components to hooks

- Search and filtering moved from toolbar to above list to match other list pages using `QueryList`
- Checkbox alignment adjusted so it does not move around page
- Loading spinners app-wide when rendered with text properly display MUI Typography
- Adds loading spinner footer on alerts list when loading more items
- Adds empty message footer on alerts list "Displaying all results." when `hasNextPage` returns false from a query
- Closed alerts will show as `ok` status (resolved)


alert/search.go
```diff
@@ -60,6 +60,7 @@ var searchTemplate = template.Must(template.New("search").Parse(`
 	{{end}}
 	{{ if .Search }}
 		AND (
+			CAST(a.id AS TEXT) ilike :search OR
```
mastercactapus: For ID searching, we should probably do an exact match instead, wyt?

graphql2/graphqlapp/alert.go
```diff
@@ -111,9 +115,15 @@ func (q *Query) Alerts(ctx context.Context, opts *graphql2.AlertSearchOptions) (
 			if err != nil {
 				return nil, err
 			}
+			if len(s.Services) == 0 {
+				return &graphql2.AlertConnection{
```
mastercactapus: Let's add a comment explaining fav only & no services will always return empty result set.

web/src/app/lists/PaginatedList.js
```diff
@@ -170,27 +184,38 @@ export class PaginatedList extends React.PureComponent {
 
     // provide a message to display if there are no results
     emptyMessage: p.string,
+
+    // renders as an infinite list instead of the default
+    // page by page behavior
+    infiniteScroll: p.bool,
+
+    // to be used if using the infinite scroller
+    hasMore: p.bool,
```
mastercactapus: We can determine if there are more to load by checking if `loadMore` was provided, e.g. `Boolean(props.loadMore)`

web/src/app/lists/PaginatedList.js
```diff
@@ -148,8 +156,11 @@ const mapStateToProps = state => {
 @connect(mapStateToProps)
 export class PaginatedList extends React.PureComponent {
   static propTypes = {
-    // headerNote will be displayed at the top of the list.
-    headerNote: p.oneOfType([p.string, p.element]),
+    // cardHeader will be displayed at the top of the card
+    cardHeader: p.node,
```
mastercactapus: Does this replace `listHeader` if used?

web/src/app/lists/PaginatedList.js
```diff
@@ -170,27 +184,38 @@ export class PaginatedList extends React.PureComponent {
 
     // provide a message to display if there are no results
     emptyMessage: p.string,
+
+    // renders as an infinite list instead of the default
```
mastercactapus: Maybe we can describe this in terms of specific behavior. E.g. "if set, `loadMore` will be called when the user scrolls towards the bottom of the list"

web/src/app/lists/PaginatedList.js
```diff
@@ -278,15 +305,17 @@ export class PaginatedList extends React.PureComponent {
...
+    let items = this.props.items
+    if (!this.props.infiniteScroll) {
+      items = items.slice(page * itemsPerPage, (page + 1) * itemsPerPage)
+    }
+    items = items.map(this.renderItem)
 
     // Display full list when loading
     if (!noPlaceholder) {
-      while (this.isLoading() && items.length < ITEMS_PER_PAGE) {
+      while (this.isLoading() && items.length < itemsPerPage) {
```
mastercactapus: For placeholders to work the way they do otherwise, I think we need pageCount here
```suggestion
      while (this.isLoading() && items.length < itemsPerPage * this.pageCount()) {
```

web/src/app/lists/QueryList.js
```diff
@@ -86,18 +96,29 @@ export default function QueryList(props) {
...
+        <Grid item className={classes.search}>
+          <Search endAdornment={searchAdornment} />
+        </Grid>
       </Grid>
 
       <Grid item xs={12}>
         <PaginatedList
           {...listProps}
           key={urlKey}
+          hasMore={hasNextPage}
```
mastercactapus: `loadMore` is already conditional

web/src/app/lists/QueryList.js
```diff
@@ -133,8 +154,19 @@ QueryList.propTypes = {
   // If set, the search string param is ignored.
   noSearch: p.bool,
 
-  // filters additional to search, set in the search text field.
+  // controls unrelated to search, but still modify results, rendered to
+  // the left of the search text field
+  controls: p.node,
```
mastercactapus: Do we have more cases than check all/some + actions, or can we make the check all/none a feature of the list itself?

If we do that, we can stop tracking things like `onDataChange` (which duplicates data + double renders) and the indirection around it (incl. the redux action I think) can go away and instead just pass in something like menu options. Let's chat on this.

web/src/app/lists/QueryList.js
```diff
@@ -72,12 +78,16 @@ export default function QueryList(props) {
...
   })
 
+  let nodes = []
   let items = []
   let loadMore
+  let hasNextPage = false
 
   if (data && data.data && data.data.nodes) {
+    nodes = data.data.nodes
     items = data.data.nodes.map(props.mapDataNode)
```
mastercactapus: Can we simplify everything accessing this by just doing something like `const nodes = _.get(data, 'data.nodes', [])` on line 81?

alert/search.go
Forfold: Yeah I think that makes sense, iirc we were doing that on the old endpoint anyways

web/src/app/lists/PaginatedList.js
Forfold: You could have both if you wanted to, 1 would be rendered at the top of the card, and the other would be rendered as the first list item with text

web/src/app/lists/PaginatedList.js
Forfold: Works for me

web/src/app/lists/QueryList.js
Forfold: We don't (check all, check none, ack, close, and escalate are all actions), but I'm not sure what this would look like without changing more of how it looks which we did want to avoid for now if we could help it

web/src/app/lists/QueryList.js
Forfold: We can, but I was avoiding refactoring more code to instead use the if statement already in place without pulling more stuff in

I can make the adjustments though if preferred

web/src/app/lists/PaginatedList.js
Forfold: I'll look into this :+1:

web/src/app/alerts/AlertsCheckboxControls.js
```diff
@@ -0,0 +1,287 @@
...
+  function setAll() {
+    setCheckedAlerts(alerts.map(a => a.id))
+  }
+
+  function setNone() {
+    return setCheckedAlerts([])
+  }
+
+  function handleToggleSelectAll() {
+    if (areNoneChecked()) {
```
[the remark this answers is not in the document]
Forfold: Since this is a boolean getter method, I would prefer to opt to use the is/are naming convention (e.g. isChecked, areNoneChecked, etc)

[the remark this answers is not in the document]
Forfold: However with the refactor it looks like we're only using this function in one spot now, so I'll remove and do the check inline.

web/src/app/lists/PaginatedList.js
Forfold: This is working as intended, this is controlling how many items to render, so we wouldn't want to render `itemsPerPage * n`, we only want to render placeholders that would match what a full list would look like for that page.

web/src/app/lists/QueryList.js
mastercactapus: It seems odd to be accessing the same data in two different ways. I think it would be more clear if we simplify and clean it up

web/src/app/lists/PaginatedList.js
mastercactapus: That would mean when you get to the bottom, you would not get placeholders for the next (loading) page of results, correct?

Though that would be different from the alerts list page is today, it would be breaking from the behavior of PaginatedList (which does use placeholders when paginating) which we are migrating to.

Since it would be a simple change, let's just have it as a quick standup item and decide.

web/src/app/lists/QueryList.js
mastercactapus: I'd rather put the work into extending our re-usable components with features we use, than to have arbitrary "hook" points that we re-use in one-off places.

In this case, it would be useful to be able to do multi select on lists (e.g. remove multiple people from rotation, disable all integration keys temporarily, etc...) in the future, and get's us away from needing to do the bouncing of data around just to get it to work.

web/src/app/lists/PaginatedList.js
Forfold: Sure, however I don't think it's relevant for infinite scrolling lists since we are displaying a "🔄Loading..." indicator at the bottom of the list, in favor of placeholder list items (since we are appending to the current list and not rendering net-new content)

web/src/app/lists/QueryList.js
Forfold: I can play around with some concepts of the checkbox controls inside the card/list so the data is available without having to use redux, though it will more than likely delay getting this out and working on the next steps of decommissioning gql2

user/favorite/store.go
```diff
@@ -3,7 +3,6 @@ package favorite
 import (
 	"context"
 	"database/sql"
```
m17ch: Think we want the space separator here.

alert/search.go
```diff
@@ -60,6 +60,7 @@ var searchTemplate = template.Must(template.New("search").Parse(`
 	{{end}}
 	{{ if .Search }}
 		AND (
+			CAST(a.id AS TEXT) = substring(:search FROM '\%(.*?)\%') OR
```
mastercactapus: Maybe we can pass this as a `NamedArg` as a `NullInt64` rather than doing regex at query-time -- that way the DB can skip the check altogether if it's not valid and not need to cast either.

Query would become:
```sql
a.id = :searchID
```

And in `QueryArgs` we could do something like:
```go
var searchID sql.NullInt64
if i, err := strconv.ParseInt(opts.Search, 10, 64); err == nil {
  searchID.Valid = true
  searchID.Int64 = i
}
```

web/src/app/alerts/AlertsList.js
```diff
@@ -0,0 +1,369 @@
...
+      nodes {
+        id
+        alertID
+        status
+        summary
+        details
+        createdAt
+        serviceID
+        service {
+          name
```
mastercactapus: We should ask for the `id` with `service {}` so that it joins/updates the correct document in cache, since we're fetching the name.
```suggestion
        service {
          id
          name
```

web/src/app/lists/ControlledPaginatedList.tsx
```diff
@@ -67,6 +67,7 @@ export interface ControlledPaginatedListAction {
 
   // label to display (e.g. "Close alerts")
   label: string
+  ariaLabel?: string
```
mastercactapus: We already have label text, `ariaLabel` is redundant -- we should re-use the existing OR if there is already a `title` attribute in the markup, we don't need to override it with `aria-label` as a screen reader will already read the title.

I'd rather make the label more descriptive if possible than always defining two descriptions for the same action -- where possible. (e.g. The tooltip/label could just say "Close Alerts" instead of just "Close" which would work for visual and keyboard users)

m17ch:
From functional testing, I'm noticing extremely high CPU usage  (in chrome task manager) / sluggish responsiveness on the alerts list if I scroll down a few times on the closed alerts tab (filter set to include all services) -- are you able to reproduce?

web/src/app/alerts/AlertsList.js
```diff
@@ -0,0 +1,369 @@
...
+function getStatusFilter(s) {
+  switch (s) {
+    case 'acknowledged':
+      return ['StatusAcknowledged']
+    case 'unacknowledged':
+      return ['StatusUnacknowledged']
+    case 'closed':
+      return ['StatusClosed']
+    case 'all':
+      return ['StatusAcknowledged', 'StatusUnacknowledged', 'StatusClosed']
```
mastercactapus: For all the filter can be omitted/empty (`[]`)

web/src/app/alerts/AlertsList.js
```diff
@@ -0,0 +1,369 @@
...
+
+export default function AlertsList(props) {
+  const classes = useStyles()
+  const width = useWidth()
+  const isFullScreen = isWidthDown('md', width)
+
+  // state initialization
+  const [errorMessage, setErrorMessage] = useState('')
+  const [updateMessage, setUpdateMessage] = useState('')
+  const [updateComplete, setUpdateComplete] = useState(false)
```
mastercactapus: Instead of trying to track/keep in sync with these, let's switch to using the `useMutation` API/status/errors directly.

web/src/app/alerts/AlertsList.js
```diff
@@ -0,0 +1,369 @@
...
+
+  // get redux url vars
+  const params = useSelector(urlParamSelector)
+  const allServices = params('allServices')
+  const filter = params('filter', 'active')
+  const isFirstLogin = params('isFirstLogin')
+
+  // always open unless clicked away from or there are favorite services present
+  const [_showNoFavoritesWarning, setShowNoFavoritesWarning] = useState(true)
```
mastercactapus: Would this be clearer if we just track it as "dismissed"?
```suggestion
  const [noFavDismissed, SetNoFavDismissed] = useState(false)
```

web/src/app/alerts/AlertsList.js
```diff
@@ -0,0 +1,369 @@
...
+              },
+            },
+          })
+            .then(res =>
+              onCompleted(res?.data?.updateAlerts?.length ?? 0, checkedItems),
+            )
+            .catch(err => {
+              setUpdateComplete(true)
+              setErrorMessage(err.message)
+            })
```
mastercactapus: As above, let's move away from relying on the Promise API for these, since we are provided state by the `useMutation` hook

m17ch:
I'm getting the "It looks like you have no favorited services..." popup on alerts list even though, I have services favorited:
image

web/src/app/alerts/AlertsList.js
```diff
@@ -0,0 +1,369 @@
...
+                },
+              },
+            })
+              .then(res =>
+                onCompleted(res?.data?.updateAlerts?.length ?? 0, checkedItems),
+              )
+              .catch(err => {
+                setUpdateComplete(true)
+                setErrorMessage(err.message)
+              })
```
mastercactapus: Same

web/src/app/alerts/AlertsList.js
```diff
@@ -0,0 +1,369 @@
...
+              .then(res =>
+                onCompleted(
+                  res?.data?.escalateAlerts?.length ?? 0,
+                  checkedItems,
+                ),
+              )
+              .catch(err => {
+                setUpdateComplete(true)
+                setErrorMessage(err.message)
+              })
```
mastercactapus: Same

web/src/app/lists/ControlledPaginatedList.tsx
Forfold: I think that `aria-label` can be justified here as the tooltip should be short and concise for the majority of users (e.g. the current Ack, Close, Escalate), as they can see the relation next to the checkbox element. Though for screen readers I think it's acceptable they we add additional context to help them out, e.g. Acknowledge Selected Alerts, since we aren't limited on screen space.

web/src/app/alerts/AlertsList.js
Forfold: We can do that, but I found it simpler to read as opposed to showing the update based off of three variables for the three mutations, as opposed to the one variable in state

web/src/app/alerts/AlertsList.js
Forfold: `onCompleted` is not properly called from within the `useMutation` options object when calling the `mutate` func from the hook, so we are forced to use the Promise API in this situation

web/src/app/alerts/AlertsList.js
mastercactapus: I think we should. With state, we are duplicating data that we already have from `useMutation`, and proper behavior depends on us keeping it in sync from inside the Promise callbacks.

As we are switching to hooks, we should be consistent in following the new API and removing usages of Promises with Apollo.

web/src/app/alerts/AlertsList.js
mastercactapus: I'm not sure we need to, `data.updatedAlerts.length` should be provided by `useMutation` as well as that the mutation was complete. Both of which we are tracking in state, but should be relying on the hooks API instead.

The only value we need to persist is the total number of alerts, which we have when the button is clicked, before we call the mutation. (and that's assuming we don't have access to the mutations input vars from the hooks API)

Forfold: That's correct, the main reason I moved the onComplete/onError calls to then/catch, respectively, is because `checkedItems` is not persisted and we don't have access to it outside of the `onClick()` provided to us by `QueryList` and children.

web/src/app/lists/ControlledPaginatedList.tsx
mastercactapus: Instead, can we indicate it's a region (e.g. of actions) which will give context to "Close" "Ack" etc..?

I'd like to move away from having two textual descriptions of things, esp. in an API, and think it would be worthwhile to find a way to make it clearer without having to rely on aria-label  overriding the text already on the page.

web/src/app/alerts/AlertsList.js
Forfold: So turns out we do need it in state as we require a manual closing of the snackbar– I can update the variable name so it's more clear that this is more than just duplicate data in state.

web/src/app/alerts/AlertsList.js
mastercactapus: In that case we can store that single value in state, instead of the others and still drop the usage of the Promise API.

Maybe call it `selectedAlertCount` or `lastSelectedAlertCount`?

We already have `result.error` for the error state and completed & count based on `data` (or `.called && !.loading`).

Forfold: That works, though I'm not sure how far in which direction we want to go since we're at a win/lose in either way: data in state that could be avoided as it is available elsewhere, or, using a promise to workaround an apollo bug

web/src/app/alerts/AlertsList.js
```diff
@@ -54,7 +54,7 @@ export const alertsListQuery = gql`
 
 const updateMutation = gql`
   mutation UpdateAlertsMutation($input: UpdateAlertsInput!) {
-    updateAlerts(input: $input) {
+    updateAlerts(inputa: $input) {
```
dctalbot: typo?

Forfold: yeah, leftover from testing error messages. still working on a couple of fixes @mastercactapus has requested

web/src/app/alerts/AlertsList.js
```diff
@@ -0,0 +1,348 @@
...
+          title: `${a.alertID}: ${a.status
+            .toUpperCase()
+            .replace('STATUS', '')}`,
+          subText: (props.serviceID ? '' : a.service.name + ': ') + a.summary,
+          action: (
+            <Typography variant='caption'>
+              {formatTimeSince(a.createdAt)}
+            </Typography>
+          ),
+          url: `/alerts/${a.id}`,
```
dctalbot: Should use `absURLSelector` which handles the new prefix work
e.g.
`const absURL = useSelector(absURLSelector)`
```absURL(`/alerts/${a.id}`)```

File: web/src/app/lists/PaginatedList.tsx [hunk 23]
```diff
@@ -8,7 +8,7 @@ import CircularProgress from '@material-ui/core/CircularProgress'
 import Grid from '@material-ui/core/Grid'
 import IconButton from '@material-ui/core/IconButton'
 import List from '@material-ui/core/List'
-import ListItem from '@material-ui/core/ListItem'
+import ListItem, { ListItemProps } from '@material-ui/core/ListItem'
```
dctalbot: `11:20  error  'ListItemProps' is defined but never used  @typescript-eslint/no-unused-vars`

Forfold:
> Seeing some odd behavior when it comes to checkbox actions. 🤔 Sometimes nothing is updated until refresh. Other times it needs to wait until the next poll. Ideally, it would update instantly.

Yeah, current behavior on alerts list performs the same way, that is, changes from checkbox actions will be displayed on the next poll. It would be a nice feature to have, but I believe that should remain a separate body of work

web/src/app/lists/PaginatedList.tsx
Forfold: Interesting this wasn’t found with `make check` or `yarn fmt` on my end. Thanks for catching.

dctalbot: `make check`, `yarn run check`, and `yarn lint` should find it. `yarn fmt` no longer does linting, just formatting w/ Prettier. Lmk if you still can't catch it. Maybe there is a missing config option

web/src/app/alerts/AlertsList.js
Forfold: Isn't the opposite true, that we should no longer be using `absURLSelector`, in favor of `AppLink`, which is being used in `PaginatedList`? @mastercactapus

dctalbot: Ah I see, didn't realize this went straight to `AppLink`. Should be good to go then!

mastercactapus:
@Forfold we should spend a little time looking into the update issue; using the new endpoint & hooks should have "fixed" that automatically (compared with old alerts list) as there's nothing special about the alerts list (just a query).

Primarily, we need to make sure the recent changes to QueryList didn't break updates (e.g. for other pages as well)

Forfold:
resolution after talking with @mastercactapus: fix client refetching while the page is still actively polling (easy enough fix pushed up)

out of scope: updates/changes on pages after loading more data

PR #464, opened 2020-02-20
admin/systemlimits: system limits ui

KatieMSB:

**Description:**
Defines system limit schema in graphql and creates ui page in admin submenu to read and set limits.


devtools/limitapigen/main.go
```diff
@@ -0,0 +1,118 @@
...
+	f, err := parser.ParseFile(fset, "../limit/id.go", nil, parser.ParseComments)
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+
+	s := SystemLimitScanner{}
+	ast.Walk(&s, f)
+	sort.Slice(s.data, func(i, j int) bool { return s.data[i].ID < s.data[j].ID })
+	tmpl.Execute(w, s.data)
```
mastercactapus: Check for and log errors for this

graphql2/graphqlapp/limit.go
```diff
@@ -0,0 +1,38 @@
...
+	"context"
+	"database/sql"
+
+	"github.com/target/goalert/graphql2"
+	"github.com/target/goalert/limit"
+	"github.com/target/goalert/permission"
+)
+
+func (q *Query) SystemLimits(ctx context.Context) ([]graphql2.SystemLimit, error) {
+	err := permission.LimitCheckAny(ctx, permission.System, permission.Admin)
```
mastercactapus: Do this check in the store instead

graphql2/graphqlapp/limit.go
```diff
@@ -0,0 +1,38 @@
...
+		return nil, err
+	}
+	limits, err := q.LimitStore.All(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return graphql2.MapLimitValues(limits), nil
+}
+func (m *Mutation) SetSystemLimits(ctx context.Context, input []graphql2.SystemLimitInput) (bool, error) {
+	err := permission.LimitCheckAny(ctx, permission.System, permission.Admin)
```
mastercactapus: Same as above, do the permissions check in the Store

graphql2/graphqlapp/limit.go
```diff
@@ -0,0 +1,38 @@
...
+	if err != nil {
+		return false, err
+	}
+	err = withContextTx(ctx, m.DB, func(ctx context.Context, tx *sql.Tx) error {
+		l := limit.Limits{}
+		l, err := graphql2.ApplyLimitValues(l, input)
```
mastercactapus: Err needs to be checked

graphql2/graphqlapp/limit.go
```diff
@@ -0,0 +1,38 @@
...
+
+		for id, max := range l {
+			err = m.LimitStore.UpdateLimitsTx(ctx, tx, string(id), max)
```
mastercactapus: Err needs to be checked (inside the loop)

limit/id.go
```diff
@@ -7,9 +7,11 @@ type ID string
...
-	ContactMethodsPerUser       ID = "contact_methods_per_user"
-	EPStepsPerPolicy            ID = "ep_steps_per_policy"
+	NotificationRulesPerUser ID = "notification_rules_per_user"
+	ContactMethodsPerUser    ID = "contact_methods_per_user"
+	// Maximum steps on a single escalation policy step
+	EPStepsPerPolicy ID = "ep_steps_per_policy"
+	// Maximum actions on a single escalation policy step
 	EPActionsPerStep            ID = "ep_actions_per_step"
 	ParticipantsPerRotation     ID = "participants_per_rotation"
 	RulesPerSchedule            ID = "rules_per_schedule"
```
mastercactapus: We should add descriptions for some of these others:
- `RulesPerSchedule` pertains to all rules for all assignments/targets or per assignment/target, can't remember -- but we should add a description indicating exactly how it works
- `UnackedAlertsPerService`, only affects newly created alerts, but not re-escalated ones
- `UserOverridesPerSchedule` only limits future overrides (e.g. end in the future)

We could probably describe each for consistency, but those 3 come to mind as being not totally clear

web/src/app/admin/AdminLimits.js
```diff
@@ -0,0 +1,149 @@
...
+  },
+  groupTitle: {
+    fontSize: '1.1rem',
+  },
+  saveDisabled: {
+    color: 'rgba(255, 255, 255, 0.5)',
+  },
+})
+
+@withStyles(styles)
```
mastercactapus: Since this is new, let's write it in Hooks (preferably typescript)

web/src/app/admin/AdminLimits.js
```diff
@@ -0,0 +1,149 @@
...
+export default class AdminLimits extends React.PureComponent {
+  state = {
+    tab: 0,
```
mastercactapus: What is `tab` used for?

web/src/cypress/integration/admin.ts
```diff
@@ -1,12 +1,67 @@
...
+
+    it('should allow updating system limits values', () => {
+      const newContactMethods = c.integer({ min: 0, max: 1000 })
+      const newEPActions = c.integer({ min: 0, max: 1000 })
+
+      cy.form({
+        ContactMethodsPerUser: newContactMethods.toString(),
+        EPActionsPerStep: newEPActions.toString(),
+      })
+      cy.get('button[data-cy=save]').click()
```
mastercactapus:
```suggestion
      cy.dialogClick('Save')
```

web/src/cypress/integration/admin.ts
```diff
@@ -1,12 +1,67 @@
...
+      )
+    })
+
+    it('should reset pending system limit value changes', () => {
+      cy.form({
+        ContactMethodsPerUser: c.integer({ min: 0, max: 1000 }).toString(),
+        EPActionsPerStep: c.integer({ min: 0, max: 1000 }).toString(),
+      })
+
+      cy.get('button[data-cy="reset"]').click()
```
mastercactapus: Use `dialogClick`

web/src/cypress/support/limits.ts
```diff
@@ -0,0 +1,52 @@
+declare namespace Cypress {
+  interface Chainable {
+    getLimits: typeof getLimits
+    updateLimits: typeof updateLimits
+  }
+}
+
+interface Limits {
+  id: ID
```
mastercactapus: Would this be easier as an object in the helper?

If you need to look up a particular limit for a test you'd have to scan through the array each time otherwise.

We could define it as a struct or a map with an enum of possible keys.

web/src/cypress/integration/admin.ts
```diff
@@ -1,12 +1,67 @@
...
-  let cfg: Config
-  describe('Admin Page', () => {
+  describe('Admin System Limits Page', () => {
+    let limits: Limits[]
+    beforeEach(() => {
+      cy.getLimits().then(l => {
+        limits = l
+        return cy.visit('/admin/limits')
+      })
+      return cy.visit('/admin/limits')
```
mastercactapus: Why are we visiting/loading the page twice?

web/src/cypress/support/limits.ts
```diff
@@ -0,0 +1,52 @@
...
+  description: string
+  value: number
+}
+interface SystemLimitInput {
+  id: ID
+  value: number
+}
+enum ID {
+  NotificationRulesPerUser = 'notification_rules_per_user',
```
mastercactapus: How are these values used? The schema/API AFAIK uses `NotificationRulesPerUser`

web/src/cypress/integration/admin.ts
KatieMSB: These buttons are in the top bar, not in a dialog.

web/src/cypress/integration/admin.ts
KatieMSB: Same as above, not in a dialog.

mastercactapus:
Routing to the first page in the list (Config) was intentional when that functionality was built out in order to point to the most important/common page of available options.

Another example was with some of the proposed alerts list changes (e.g. bundle list of active, separate page for searching closed/past alerts): if we click on "Alerts" we'd want to default to the active list, rather than a blank page or having the user click twice for the most common action.

Likewise here, the Config would be the "default" most common until/unless we have something else that would be more commonly used in the future. It also follows that you get to config by clicking on Admin today, we're just adding another option here. If we are to change that, we will also need to ensure we update documentation to say "Click Admin, then Config" instead of just "Admin".

Since Config is more-or-less mandatory though, and System Limits already has defaults set, I think it still makes sense to default to Config instead of having to click twice.

Forfold:
While I agree that there can be a default route (i.e. `/admin` will default to our config, and `/alerts` would default to the list, I disagree that we should force the user to load a page they do not desire to see.

web/src/app/admin/AdminLimits.tsx
```diff
@@ -0,0 +1,171 @@
...
+          }}
+        >
+          Save
+        </Button>
+      </PageActions>
+    )
+  }
+
+  const render = () => {
+    return (
```
Forfold: We can return here flat without creating the function `render`

web/src/app/admin/AdminLimits.tsx
```diff
@@ -0,0 +1,171 @@
...
+export default function AdminLimits() {
+  const classes = useStyles()
+  const [confirm, setConfirm] = useState(false)
+  const [values, setValues] = useState({})
+
+  const { data, loading } = useQuery(query)
+
+  if (loading && !data) {
+    return <Spinner />
+  }
```
Forfold: we should display an error if it fails to query: `const { data, error, loading } = useQuery(query)` and `if (error) return <GenericError error={error.message} />`

web/src/app/admin/AdminDialog.js
```diff
@@ -4,58 +4,58 @@ import List from '@material-ui/core/List'
...
 
     return (
       <FormDialog
         disableGutters
         title={`Apply Configuration Change${changes.length > 1 ? 's' : ''}?`}
         onClose={this.props.onClose}
-        onSubmit={() =>
-          commit({
+        onSubmit={() => {
+          return commit({
```
Forfold: we can return here without making a block/explicit return call:
```
onSubmit={() => commit({ ... })}
```

devtools/limitapigen/main.go
```diff
@@ -0,0 +1,121 @@
...
+	return s
+}
+
+var tmpl = template.Must(
+	template.
+		New("maplimit.go").
+		Funcs(template.FuncMap{
+			"quote": strconv.Quote,
+		}).Parse(`
+// Code generate by devtools/limitapigen DO NOT EDIT.
```
mastercactapus: Should appear as the first line, and generate -> generated

```suggestion
		}).Parse(`// Code generated by devtools/limitapigen DO NOT EDIT.
```

limit/id.go
```diff
@@ -7,17 +7,22 @@ type ID string
...
+	ParticipantsPerRotation ID = "participants_per_rotation"
+	// Pertains to all rules for all assignments/targets.
+	RulesPerSchedule          ID = "rules_per_schedule"
+	IntegrationKeysPerService ID = "integration_keys_per_service"
+	// Only affects newly created alerts, but not re-escalated ones.
 	UnackedAlertsPerService     ID = "unacked_alerts_per_service"
 	TargetsPerSchedule          ID = "targets_per_schedule"
 	HeartbeatMonitorsPerService ID = "heartbeat_monitors_per_service"
-	UserOverridesPerSchedule    ID = "user_overrides_per_schedule"
+	// Only limits future overrides (e.g. end in the future).
```
mastercactapus:
```suggestion
	// Only limits future overrides (i.e. end in the future).
```

limit/id.go
```diff
@@ -7,16 +7,21 @@ type ID string
...
-	ContactMethodsPerUser        ID = "contact_methods_per_user"
-	EPStepsPerPolicy             ID = "ep_steps_per_policy"
-	EPActionsPerStep             ID = "ep_actions_per_step"
-	ParticipantsPerRotation      ID = "participants_per_rotation"
-	RulesPerSchedule             ID = "rules_per_schedule"
-	IntegrationKeysPerService    ID = "integration_keys_per_service"
-	UnackedAlertsPerService      ID = "unacked_alerts_per_service"
-	TargetsPerSchedule           ID = "targets_per_schedule"
-	HeartbeatMonitorsPerService  ID = "heartbeat_monitors_per_service"
+	NotificationRulesPerUser ID = "notification_rules_per_user"
```
dctalbot: Personal preference maybe, but I don't like having comments for a seemingly random set of these constants. Some are listed as "Maximum ..." Are the others max or min? Do we need comments here at all?

limit/store.go
```diff
@@ -47,18 +35,35 @@ func NewDB(ctx context.Context, db *sql.DB) (*DB, error) {
 	}, p.Err
 }
 
-// ResetAll implements the Store interface.
-func (db *DB) ResetAll(ctx context.Context) error {
+// Update will update all configurable limits.
```
dctalbot:
```suggestion
// UpdateLimitsTx updates all configurable limits.
```

limit/id.go
KatieMSB: I agree, and we should add comments for each of them, the ones marked stood out as the most confusing.

PR #470, opened 2020-02-24
smoketests: rename graphql2 tests

dctalbot:

**Description:**
- Removes all test files named with a graphql2 prefix. 
- Merge `graphql2users_test` and `graphqlusers_test` 
- Delete `graphql2alerts_test` which is covered by `graphqlalerts_test`


mastercactapus:
We do want the end-to-end test (using graphql to create contact methods, etc..) as it's the only smoketest that verifies the entire flow, meaning you can get an SMS from a *mostly* empty DB using **only** the API. The system limits test already uses the `createAlert` mutation so the other alert-only gql test is already redundant.

 As for renaming, make sure to update test names to match the new file names.

PR #471, opened 2020-02-24
ui/avatars: convert avatar util to typescript

dctalbot:

**Description:**
Part of our "developer life" initiative is converting all util functions to ts

- Renamed file: `util/avatar/types.js` to `util/avatars.tsx`


web/src/app/util/avatars.tsx
```diff
@@ -0,0 +1,74 @@
+import React, { useEffect, useState } from 'react'
+import { Layers, RotateRight, Today, VpnKey, Person } from '@material-ui/icons'
+import { useSessionInfo } from './RequireConfig'
+import { Avatar, AvatarTypeMap, SvgIconProps } from '@material-ui/core'
+import { OverridableComponent } from '@material-ui/core/OverridableComponent'
+
+type IconProps = (props: SvgIconProps) => JSX.Element //type alias
+interface AvatarProps extends OverridableComponent<AvatarTypeMap> {
+  userID?: string
```
mastercactapus: This is only used for the `UserAvatar` and is required. We should define it separately from the others.

web/src/app/util/avatars.tsx
```diff
@@ -0,0 +1,76 @@
...
+      {...DOMProps}
+    >
+      {validImage ? null : <Fallback fontSize='small' />}
+    </Avatar>
+  )
+}
+
+export function UserAvatar(props: UserAvatarProps) {
+  const { userID } = props
+  return useAvatar(Person, props, `/api/v2/user-avatar/${userID}`)
```
mastercactapus: We don't need UserAvatarProps inside `useAvatar` let's clean it up in `UserAvatar` instead.
```suggestion
  otherProps: AvatarProps,
  imgSrc?: string,
) {
  const validImage = useValidImage(imgSrc)

  return (
    <Avatar
      alt=''
      src={validImage ? imgSrc : undefined}
      data-cy={validImage ? null : 'avatar-fallback'}
      {...otherProps}
    >
      {validImage ? null : <Fallback fontSize='small' />}
    </Avatar>
  )
}

export function UserAvatar(props: UserAvatarProps) {
  const { userID, ...otherProps } = props
  return useAvatar(Person, otherProps, `/api/v2/user-avatar/${userID}`)
```

dctalbot: I wanted to do this originally, but it isn't valid since ts apparently isn't smart enough to derive the inheritance of otherProps. I may me missing something, but I was really surprised to find this limitation

mastercactapus: We can cast it manually if the inferencing isn't working automatically.

PR #472, opened 2020-02-24
http: add proper support for subpaths

### LINKED ISSUE #75: Proxy Path

mastercactapus:

**Description:**
This PR addresses issues preventing GoAlert from being served from a subpath of a domain.

You can start GoAlert with a prefix by setting `GOALERT_HTTP_PREFIX`, for example:
- `make start GOALERT_HTTP_PREFIX=/goalert`
- `make start-prod GOALERT_HTTP_PREFIX=/goalert`
- `make cy-wide GOALERT_HTTP_PREFIX=/goalert`

Where `/goalert` is the desired prefix, and the UI can be accessed at `http://localhost:3030/goalert`

Fixes #75 

**Additional Info:**
- A new command-line flag `--http-prefix` has been added allowing per-instance prefix setting.
- The index/HTML page is now generated via Go template
- Version information is now embedded in `app.js` as well as meta tags (instead of HTML comments, as Go templates strip them out and are harder to access)
- A new `make start-prod` command has been added allowing production mode local dev
- Public URL paths are no longer automatically stripped
- Rewrite helpers no longer loop through the serve mux twice
- `inliner` devtool can now calculate `sha256` hash of file data
- `simpleproxy` now supports the `-trim` flag for testing
- Twilio validation now uses the original `RequestURI` instead of mapped paths solving issues with validation when transitioning callback URLs
- Cache headers & ETags are now set for UI assets
- tests updated to better support prefixed paths


web/src/app/util/AppLink.tsx
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export function AppLink(props: AppLinkProps) {
+  const { to: _to, ...other } = props
+  const path = useSelector(urlPathSelector)
+
+  if (/^(mailto:|https?:\/\/)/.test(_to)) {
+    return <a href={_to} {...other} />
+  }
+
```
dctalbot:
```suggestion

  if (/^(mailto:|https?:\/\/)/.test(_to)) {
    return <a href={_to} {...other} />
  }

const path = useSelector(urlPathSelector)
```

dctalbot: Are there other protocols we should look for e.g. `tel:`, `sms:` or should we keep this explicit? Could also go generic like e.g. `/.*:\/\/)/` ?

web/src/package.json
```diff
@@ -136,6 +136,7 @@
     "babel-plugin-transform-class-properties": "6.24.1",
     "babel-plugin-typescript-to-proptypes": "1.2.1",
     "chance": "1.1.4",
+    "copy-webpack-plugin": "5.1.1",
```
Forfold: I'm not seeing any other references to `html-webpack-plugin`, so we should be good to remove it here

mastercactapus:
@dctalbot Give it another try, if your Public URL was empty, the "fallback" would not include the prefix, but now does.

If you have Public URL set, make sure it has the prefix included

PR #473, opened 2020-02-24
ui: convert copy to clipboard util to typescript

dctalbot:

### DIFF
--- a/web/src/app/util/copyToClipboard.ts
+++ b/web/src/app/util/copyToClipboard.ts
@@ -14,7 +14,7 @@ export default function copyToClipboard(text: string) {
  * that state is saved upon completion of copying.
  */
 function fallback(str: string) {
-  const textArea = <HTMLTextAreaElement>document.createElement('textArea')
+  const textArea = document.createElement('textArea') as HTMLTextAreaElement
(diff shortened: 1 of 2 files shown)


web/src/app/util/copyToClipboard.ts
```diff
@@ -13,25 +13,27 @@ export default function copyToClipboard(text) {
...
   if (isOS()) {
     const range = document.createRange()
     range.selectNodeContents(textArea)
     const selection = window.getSelection()
-    selection.removeAllRanges()
-    selection.addRange(range)
+    if (selection) {
+      selection.removeAllRanges()
+      selection.addRange(range)
+    }
```
Forfold: If it fails to get the selection, does the copy still succeed? If not, let's return an error here if selection is falsey

web/src/app/util/copyToClipboard.ts
```diff
@@ -14,26 +14,32 @@ export default function copyToClipboard(text: string) {
  * that state is saved upon completion of copying.
  */
 function fallback(str: string) {
+  console.log('herere')
```
Forfold: rm

PR #477, opened 2020-02-25
ui: convert paginated list to typescript

mastercactapus:

**Description:**
This PR converts `PaginatedList` to typescript and hooks as well as the URL selectors.

A `useWidth` helper has been added for components that were using the `@useWidth()` decorator.


Forfold:
Screen Shot 2020-02-25 at 1 11 11 PM

Looks like "No results" didn't translate over correctly. I can take a look at this

web/src/app/lists/PaginatedList.tsx
```diff
@@ -0,0 +1,299 @@
...
+            />
+          )}
+          <RightIcon />
+        </IconButton>
+      </Grid>
+    </Grid>
+  )
+}
+
+const loadingStyle = {
```
dctalbot: If we can, let's move this into `useStyles` so we know where to look

PR #478, opened 2020-02-25
ui: convert Diff and debug util to typescript

dctalbot:


web/src/app/util/Diff.tsx
```diff
@@ -0,0 +1,89 @@
...
+        {part.value}
+      </span>
+    )
+  })
+
+  const hideRemoved = diff.length === 1 && diff[0].added // net new string
+  const hideAdded = diff.length === 1 && diff[0].removed // deleted whole string
+
+  return (
+    <>
```
Forfold: Can we use `React.Fragment` here?

PR #480, opened 2020-02-25
ui: remove reminderMinutes

m17ch:

**Description:**
This PR removes `reminderMinutes` fields from the UI when creating or editing calendar subscriptions and updates cypress tests.  It also sets a default reminder on new calendar subscriptions created in the UI to be the on-call shift start.  No GraphQL or backend adjustments were made (still support specifying alternate reminderMinutes).

### DIFF
--- a/web/src/app/schedules/calendar-subscribe/CalendarSubscribeCreateDialog.js
+++ b/web/src/app/schedules/calendar-subscribe/CalendarSubscribeCreateDialog.js
@@ -62,7 +62,7 @@ export default function CalendarSubscribeCreateDialog(props) {
       input: {
         scheduleID: value.scheduleID,
         name: value.name,
-        reminderMinutes: [0],
+        reminderMinutes: [0], // default reminder at shift start time
         disabled: false,
       },
     },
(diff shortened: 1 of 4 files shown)


web/src/app/schedules/calendar-subscribe/CalendarSubscribeCreateDialog.js
```diff
@@ -62,7 +62,7 @@ export default function CalendarSubscribeCreateDialog(props) {
       input: {
         scheduleID: value.scheduleID,
         name: value.name,
-        reminderMinutes: value.reminderMinutes.map(r => r && r.value),
+        reminderMinutes: [0],
```
Forfold: Let's leave a quick comment that we are defaulting to notify at time of shift on creation

web/src/app/schedules/calendar-subscribe/CalendarSubscribeForm.js
```diff
@@ -49,7 +20,7 @@ export default function CalendarSubscribeForm(props) {
             fullWidth
             component={TextField}
             name='name'
-            label='Subscription Name'
+            label='Subscription Source Name'
```
Forfold: 👍 This should a good way to distinguish that this is for GoAlert organization, and not what they will see on their devices, I think 🎉

PR #491, opened 2020-03-02
tcpa: log notification opt-in/opt-out events

dctalbot:

**Description:**
When a user opts in or out from notifications this PR will log a user's contact method ID.

**User facing changes**
Users will be required to unsubscribe/disable contact methods independently. (Currently disabling notification from VOICE would also disable SMS notifications).

**Additional Info:**
We want to maintain a record of consent and keep a current record of consumers who have consented to receiving communications and the type of communications to which they have consented


engine/engine.go
```diff
@@ -357,11 +358,20 @@ func (p *Engine) Stop(ctx context.Context, d notification.Dest) error {
...
 	permission.SudoContext(ctx, func(ctx context.Context) {
-		err = p.cfg.ContactMethodStore.DisableByValue(ctx, contactmethod.TypeFromDestType(d.Type), d.Value)
+		err = p.cfg.ContactMethodStore.DisableByValue(ctx, contactMethodType, d.Value)
 	})
 
+	// NOTE TCPA compliance: maintain a record of consent/dissent
+	if err == nil {
+		userID := permission.UserID(ctx)
+		time := time.Now().String()
+		logStr := fmt.Sprintf("user %s stopped %s notifications at %s", userID, contactMethodType, time)
```
mastercactapus: `log.Logf` takes its first argument as a format string. But use `WithFields` for the ID and keep the message fixed in this case.

- We'll need to return the CM id from the query above, UserID can probably be omitted, as the CM id already links to the user ID
- The context is likely not associated with the user, since we're using `SudoContext` above (meaning `permission.UserID` will most likely return an empty string)
- The log text should just be something fixed like "Contact method STOP received." or something of that nature. 
- This should probably be in the store method instead, in case it's ever called from elsewhere. Anything that triggers Enable/Disable should log. That way we can also avoid adding/changing the API when returning the contact method ID

notification/store.go
```diff
@@ -270,5 +271,12 @@ func (db *DB) VerifyContactMethod(ctx context.Context, cmID string, code int) er
 		return validation.NewFieldError("code", "invalid code")
 	}
 
+	// NOTE: maintain a record of consent/dissent
+	logctx := log.WithFields(ctx, log.Fields{
```
mastercactapus: nit: capital `C`

user/contactmethod/store.go
```diff
@@ -102,20 +97,27 @@ func NewDB(ctx context.Context, db *sql.DB) (*DB, error) {
...
-		_, err = db.disablePhone.ExecContext(ctx, n.Value)
-	default:
-		_, err = db.disable.ExecContext(ctx, n.Type, n.Value)
+
+	var id string
+	err = db.disable.QueryRowContext(ctx, n.Type, n.Value).Scan(&id)
+
+	if err == nil {
+		// NOTE: maintain a record of consent/dissent
+		logctx := log.WithFields(ctx, log.Fields{
```
mastercactapus: nit: capital `C`

PR #493, opened 2020-03-02
ui: Link -> AppLink

mastercactapus:

**Description:**
This PR replaces usage of `<a>`,  `<Link>` and . `absURL` with a helper component `<AppLink>`.

This is in preparation for changes that allow/fix handling for URL prefixes.


web/src/app/services/IntegrationKeyForm.js
```diff
@@ -62,9 +63,9 @@ export default class IntegrationKeyForm extends React.PureComponent {
                     endAdornment: (
                       <InputAdornment position='end'>
                         <Tooltip title='API Documentation' placement='right'>
-                          <a href='/docs' target='_blank'>
+                          <AppLink to='/docs' target='_blank'>
```
dctalbot: This is missing `rel='noopener noreferrer'` to prevent tabnabbing. We have it almost everywhere else, but maybe it's worth adding in the `AppLink` component a check like 
```
if (props.target === '_blank') props.rel='noopener noreferrer'
```

mastercactapus: Updated with `newTab` prop as discussed in chat

PR #508, opened 2020-03-05
ui: use graphql2 api in alert details

Forfold:

**Description:**
This PR converts the api calls from alert details pages to use the Graphql2 client in Apollo.

**Out of Scope:**
Converting alert details files to hooks and TypeScript

Rotations will now appear in the escalation policy countdown table

### DIFF
--- a/web/src/app/alerts/components/AlertDetails.js
+++ b/web/src/app/alerts/components/AlertDetails.js
@@ -90,7 +90,7 @@ export default class AlertDetails extends Component {
           style={{ paddingBottom: 0 }}
         >
           <AlertDetailLogs
-            alertID={this.props.data.number}
+            alertID={this.props.data.alertID}
             showExactTimes={this.state.showExactTimes}
           />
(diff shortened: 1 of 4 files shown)


web/src/app/alerts/components/AlertDetails.js
```diff
@@ -122,26 +122,40 @@ export default class AlertDetails extends Component {
...
+    const state = this.props.data.state
+
     return {
-      repeat: eps.repeat,
-      numSteps: eps.steps.length,
-      steps: eps.steps,
+      repeat: state?.repeatCount,
+      numSteps: ep.steps.length,
+      steps: ep.steps,
       status: alert.status.toLowerCase(),
```
mastercactapus: Now that we're using the new API, maybe we should get rid of the use of `.toLowerCase()` everywhere for status comparisons.

That way every usage has identical values (e.g. `StatusUnacknowledged`).

PR #513, opened 2020-03-09
dev: fix typescript file checking on commit and make check

dctalbot:
- [x] Identified the issue which this PR solves.
- [x] Read the [**CONTRIBUTING**](https://github.com/target/goalert/blob/master/CONTRIBUTING.md) document.
- [x] Code builds clean without any errors or warnings.
- [x] Added appropriate tests for any new functionality.
- [x] All new and existing tests passed.
- [x] Added comments in the code, where necessary.
- [x] Ran `make check` to catch common errors. Fixed any that came up.

**Description:**
This PR combines our cypress/ and app/ `tsconfig.json` files and updates the `package.json` to lint the project with a unified config.

- husky will **not** run typechecking pre-commit
- `make check` should fail if there are any typescript errors
- uses `babel-loader` to build cypress bundles

**Out of Scope**
- eslint will not check typescript files (current state)

PR #520, opened 2020-03-11
feature/eslint-typescript: add typescript-eslint

dctalbot:

**Description:**
Continuation of #513 to add typescript-eslint and trigger in 'yarn fmt' and husky pre-commit hook.


web/src/.eslintrc.js
```diff
@@ -23,6 +32,15 @@ module.exports = {
     // handled by prettier
     'react/jsx-curly-newline': 'off',
     'react/jsx-indent': 'off',
+
+    // typescript-eslint rules
+    '@typescript-eslint/explicit-function-return-type': 'off',
+    '@typescript-eslint/no-namespace': 'off',
+    '@typescript-eslint/no-use-before-define': 'off',
+    '@typescript-eslint/no-explicit-any': 'off',
```
Forfold: how much effort would it take to have this one stay on? we probably shouldn't use `any` anywhere we can help it

web/src/cypress/integration/escalationPolicySteps.ts
```diff
@@ -83,11 +83,9 @@ function testSteps(screen: ScreenFormat) {
       })
     })
 
-    it('should edit a step', () => {
-      let s1: EPStep
+    it.only('should edit a step', () => {
```
Forfold:
```suggestion
    it('should edit a step', () => {
```

web/src/cypress/integration/pagination.ts
```diff
@@ -39,7 +39,7 @@ function testPagination(screen: ScreenFormat) {
 
 function testPaginating(label: string, url: string, create: createManyFunc) {
   let names: Array<string> = []
-  let nameSubstr: string = ''
+  let nameSubstr = ''
```
Forfold: linter doesn't want type def?

web/src/cypress/support/config.ts
```diff
@@ -1,83 +1,85 @@
...
   }
   Feedback?: {
-    Enable?: Boolean
-    OverrideURL?: String
+    Enable?: boolean
+    OverrideURL?: string
   }
 }
-interface Config {
+export interface Config {
```
Forfold: im not seeing where this is being referenced, still need export?

web/src/cypress/support/graphql.ts
```diff
@@ -10,16 +12,12 @@ interface GraphQLResponse {
   errors: [any]
 }
 
-function graphql2(query: string, variables?: any): Cypress.Chainable<any> {
```
Forfold: i'm curious as to why it's removing function return types, is this a rule we can enable?

web/src/cypress/support/index.ts
```diff
@@ -35,12 +31,18 @@ import './page-search'
...
+export * from './limits'
 export * from './util'
 
 import './fail-fast'
+Cypress.env('RETRIES', 2)
+Cypress.Cookies.defaults({
+  whitelist: 'goalert_session.2',
+})
+
+export * from './util'
```
Forfold: duplicate of [line 40](https://github.com/target/goalert/pull/520/files#diff-e433f8c67e8c459348b83b0482bee448R40)?

web/src/.eslintrc.js
dctalbot: This would be very difficult in certain cases like the `Autocomplete` abstraction we chatted about this week. I think having explicit `any` turned on but implicit `any` turned off strikes the right balance of developer freedom

web/src/cypress/integration/pagination.ts
dctalbot: inferred types are preferable

web/src/cypress/support/config.ts
dctalbot: Yes, `cypress/support` has an entrypoint at `cypress/support/index.ts` that makes this available in the namespace. It's used in `admin.ts`

web/src/cypress/support/graphql.ts
dctalbot: Inferred types are preferable. Less code, less chance for developer error

PR #531, opened 2020-03-16
dev: fix typescript file checking on commit and make check

KatieMSB:

**Description:**
This PR moves src/tsconfig.json to app/ and keeps the cypress/ and app/tsconfig.json files separate. Also updates the package.json to lint the project with a unified config and creates separate yarn scripts to format and type check app/ and cypress/.

- husky will **not** run typechecking pre-commit
- make check should fail if there are any typescript errors
- uses babel-loader to build cypress bundles

**Out of Scope:**
eslint will not check typescript files (current state)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -169,7 +169,7 @@ check: generate web/src/node_modules
 	go vet ./...
 	go run github.com/gordonklaus/ineffassign .
 	CGO_ENABLED=0 go run honnef.co/go/tools/cmd/staticcheck ./...
-	(cd web/src && yarn fmt)
+	(cd web/src && yarn fmt && yarn tsc-app && yarn tsc-cypress)
 	./devtools/ci/tasks/scripts/codecheck.sh
 
 check-all: check test smoketest cy-wide-prod-run cy-mobile-prod-run
--- a/Makefile
+++ b/Makefile
@@ -159,7 +159,7 @@ start-prod: bin/waitfor web/inline_data_gen.go bin/runjson
 	bin/runjson <devtools/runjson/localdev-prod.json
 
 jest: web/src/node_modules
-	cd web/src && yarn tsc-app && node_modules/.bin/jest $(JEST_ARGS)
+	cd web/src && node_modules/.bin/jest $(JEST_ARGS)
 
 test: web/src/node_modules jest
 	go test -short ./...
--- a/Makefile
+++ b/Makefile
@@ -159,7 +159,7 @@ start-prod: bin/waitfor web/inline_data_gen.go bin/runjson
 	bin/runjson <devtools/runjson/localdev-prod.json
 
 jest: web/src/node_modules
-	cd web/src && node_modules/.bin/jest $(JEST_ARGS)
+	cd web/src && yarn tsc-app && node_modules/.bin/jest $(JEST_ARGS)
 
 test: web/src/node_modules jest
 	go test -short ./...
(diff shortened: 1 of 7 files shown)


web/src/cypress/tsconfig.json
```diff
@@ -4,7 +4,19 @@
...
     "target": "es5",
     "lib": ["es2015", "dom"],
-    "types": ["cypress"]
+    "types": ["cypress"],
+    "jsx": "react",
+    "allowJs": true,
+    "allowSyntheticDefaultImports": true,
+    "esModuleInterop": true,
+    // TODO: set to noImplicitAny to true
+    "noImplicitAny": true,
```
Forfold: rm comment?

dctalbot: Will resolve this in a follow-up PR

Forfold: It looks like it's already set to `true` though?

KatieMSB: that was a typo, meant to be `isolatedModules` not `noImplicitAny`. PR here for that fix: #532

Makefile
```diff
@@ -159,7 +159,7 @@ start-prod: bin/waitfor web/inline_data_gen.go bin/runjson
 	bin/runjson <devtools/runjson/localdev-prod.json
 
 jest: web/src/node_modules
-	cd web/src && node_modules/.bin/jest $(JEST_ARGS)
+	cd web/src && yarn tsc-app && node_modules/.bin/jest $(JEST_ARGS)
```
mastercactapus: We can skip this for running tests

web/src/package.json
```diff
@@ -5,9 +5,11 @@
   "main": "index.js",
   "scripts": {
     "slint": "stylelint ./app/**/*.css",
-    "lint": "eslint . && tsc -p ./tsconfig.json",
+    "lint": "eslint . && tsc -p app/tsconfig.json",
```
mastercactapus: Should check both app & cypress

web/src/cypress/tsconfig.json
```diff
@@ -4,7 +4,19 @@
     "baseUrl": "../node_modules",
     "target": "es5",
     "lib": ["es2015", "dom"],
-    "types": ["cypress"]
+    "types": ["cypress"],
+    "jsx": "react",
+    "allowJs": true,
+    "allowSyntheticDefaultImports": true,
+    "esModuleInterop": true,
+    // TODO: set isolatedModules to true
```
mastercactapus: js comments are not valid JSON

web/src/cypress/tsconfig.json
```diff
@@ -4,7 +4,19 @@
     "baseUrl": "../node_modules",
     "target": "es5",
```
mastercactapus: I think we can drop this now

web/src/cypress/tsconfig.json
```diff
@@ -4,7 +4,19 @@
...
     "lib": ["es2015", "dom"],
```
mastercactapus: Can we use `es6` to match what's in app?

web/src/app/tsconfig.json
```diff
@@ -14,6 +14,6 @@
     "skipLibCheck": true,
     "forceConsistentCasingInFileNames": true
   },
-  "include": ["./app/"],
+  "include": ["**/*"],
   "exclude": ["./node_modules/", "./build/"]
```
mastercactapus: I think we can drop this now that the file is in `app`

PR #535, opened 2020-03-16
devtool: sendit proxy

mastercactapus:

**Description:**
This PR adds a new devtool, `sendit` that can be used similarly to ngrok or serveo, but in environments where a self-hosted option is preferable.

**Out-of-scope**
- Only supports prefixed routing (e.g. URL path, not subdomain)
- Does not support TLS connections (should be behind a load balancer that does TLS termination)
- No direct `make start` integration -- should be run separately, if/when needed

If any of the out-of-scope items should be part of this PR, let me know and I can work on something. But wanted to get the bare tool for now, and see what is really needed after people have gotten a chance to try/test it.


devtools/sendit/server.go
```diff
@@ -0,0 +1,236 @@
...
+
+	return sess.OpenContext(ctx)
+}
+
+// serveOpen will perform initial authentication and esablish a tunnel session.
+func (s *Server) serveOpen(w http.ResponseWriter, req *http.Request) {
+	if len(s.authSecret) > 0 {
+		_, err := TokenSubject(s.authSecret, TokenAudienceAuth, req.FormValue("token"))
+		if err != nil {
+			if err != nil {
```
dctalbot: Looks like a redundant branch condition

mastercactapus: Lol nice catch. copy pasta

devtools/sendit/session.go
```diff
@@ -0,0 +1,211 @@
...
+	case <-sess.doneCh:
+		return
+	default:
+	}
+
+	select {
+	case <-sess.stream.Ready():
+	case <-sess.doneCh:
+		return
+	}
```
Forfold: do we need to be repeating this logic waiting on the done channel to return?

devtools/sendit/session.go
```diff
@@ -0,0 +1,211 @@
...
+	delete(sess.s.sessionsByPrefix, sess.Prefix)
+}
+
+type ioContext struct {
+	fn     func([]byte) (int, error)
+	cancel func()
+}
+
+func (ctx *ioContext) Read(p []byte) (int, error)  { return ctx.fn(p) }
+func (ctx *ioContext) Write(p []byte) (int, error) { return ctx.fn(p) }
```
Forfold: is the error return type needed here, or does `fn` return two vars? e.g. `return ctx.fn(p), nil` or removing the `error` return type?

devtools/sendit/stream.go
```diff
@@ -0,0 +1,189 @@
...
+
+func (s *Stream) Write(p []byte) (int, error) {
+	var w io.WriteCloser
+	select {
+	case w = <-s.writeCh:
+	case <-s.closeCh:
+		return 0, io.ErrClosedPipe
+	}
+
+	n, err := w.Write(p)
```
Forfold: missing error check here and in `Read()`? could you explain a little why we would want to return n and error if both evaluate as truthy?

devtools/sendit/stream.go
```diff
@@ -0,0 +1,189 @@
...
+		closeCh:   make(chan struct{}),
+		readyCh:   make(chan struct{}),
+	}
+	return s
+}
+
+// Ready will return a channel indicating when the stream is ready for reading and writing.
+func (s *Stream) Ready() <-chan struct{} { return s.readyCh }
+
+func (s *Stream) Write(p []byte) (int, error) {
```
Forfold: should we check if the stream is closed before we write?
```
if s.isClosed {
	return io.ErrClosedPipe
}
```

devtools/sendit/client.go
```diff
@@ -0,0 +1,170 @@
...
+			select {
+			case <-t.C:
+				err := setPipe()
+				if err != nil {
+					log.Println("ERROR: set pipe:", err)
+					return
+				}
+			case <-ctx.Done():
+				return
+			}
```
Forfold: same here

devtools/sendit/session.go
```diff
@@ -0,0 +1,211 @@
...
+	default:
+	}
+
+	select {
+	case <-sess.doneCh:
+		return nil, io.ErrClosedPipe
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case <-sess.readyCh:
+	}
```
dctalbot: I think we can remove this redundant switch statement

devtools/sendit/README.md
```diff
@@ -0,0 +1,40 @@
+# SendIt
+
+SendIt is a tool for setting up a reverse proxy to localhost that's accesible externally. It's similar to tools like `ngrok` and `serveo`.
```
dctalbot:
```suggestion
SendIt is a tool for setting up a reverse proxy to localhost that's accessible externally. It's similar to tools like `ngrok` and `serveo`.
```

devtools/sendit/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+Usage: `sendit-server -secret <SECRET_STRING>`
+
+If exposing publicly, it's recommended to start in secure mode with the `-secret` flag set. This will require clients to provide a token generated with the `sendit-token` command.
+
+## Client
+
+Usage: `sendit -token <TOKEN> <SERVER_URL>/<DESIRED_PREFIX> <LOCAL_URL>`
+
+Example: `sendit https://example.com/foobar http://localhost:3030`
```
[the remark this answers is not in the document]
mastercactapus: There are no domain requirements, the subdomain we used was just for our own purposes.

devtools/sendit/session.go
mastercactapus: Yes, `select` makes no guarantees on order and we want to exit if `doneCh` is set, even if the stream is ready. If we only had one, it would be effectively random which one would get picked each time, even if closed.

devtools/sendit/session.go
mastercactapus: These implement the `io.Reader` and `io.Writer` interfaces, respectively. So the func signatures are what they need to be. And yes `fn` returns `(int, error)` so the values are just a passthrough.

devtools/sendit/stream.go
mastercactapus: This is just a wrapper function that handles the synchronization of the underlying readers/writers so they can be swapped in a thread-safe way. We don't actually care about the return values, only that we provide them to the caller.

The exception is `Read` which needs to handle EOF a little differently so that instead of declaring we got to the end, we start reading from the next one. Similar to how `io.MultiReader` works -- but in a dynamic way.

devtools/sendit/session.go
Forfold: Can we document that here and other relevant places for readability and maintenance?

devtools/sendit/stream.go
mastercactapus: We are already doing so by way of `<-s.closeCh`. It uses the channel to make the check to keep it thread-safe as we don't want to access `s.isClosed` while `s.isClosed` is being set by another thread.

devtools/sendit/client.go
mastercactapus: Same as above, it chooses one at random if multiple are ready, so we need to break it out to give context priority.

There will never be a guarantee that `setPipe` will not be called if the context is cancelled at exactly the right time, but that's okay. But by breaking it up, we can guarantee `setPipe` will not be called if the context is cancelled before calling the function.

devtools/sendit/session.go
mastercactapus: Breaking it apart is the only way to prioritize the close/cancellation cases above the `readyCh` check. When multiple cases are ready, the one fulfilled is chosen at random, thus the need to check the short-circuit cases separately/first.

PR #539, opened 2020-03-17
dev: add typescript-eslint

KatieMSB:

**Description:**
Adds typescript-eslint to project and sets base rules. Also configures .ts, .tsx files to be linted in pre-commit hook. 

**Out of Scope:**
TODOs listed in eslint rules to be handled in future prs individually.


web/src/cypress/integration/pagination.ts
```diff
@@ -39,7 +39,7 @@ function testPagination(screen: ScreenFormat) {
 
 function testPaginating(label: string, url: string, create: createManyFunc) {
   let names: Array<string> = []
-  let nameSubstr: string = ''
```
Forfold: Is there a reason we are removing the type definition for the string but not for the array of strings

web/src/package.json
```diff
@@ -130,6 +130,8 @@
     "@types/react-redux": "7.1.7",
     "@types/react-router": "5.1.4",
     "@types/react-router-dom": "5.1.3",
+    "@typescript-eslint/eslint-plugin": "^2.24.0",
+    "@typescript-eslint/parser": "^2.24.0",
```
Forfold:
```suggestion
    "@typescript-eslint/eslint-plugin": "2.24.0",
    "@typescript-eslint/parser": "2.24.0",
```

plus a rerun of `yarn install` after to update `yarn.lock`

web/src/cypress/integration/pagination.ts
KatieMSB: This was removed by the `eslint --fix` since it's being set when declared it infers the type.

Forfold: Yeah, I suppose I'm curious why it's inferring some types but not others. Are only primitives inferred?

KatieMSB: Yea it's only inferring the primitives.

web/src/cypress/support/graphql.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint no-unused-expressions: 0 */
```
mastercactapus: Let's rewrite those two assertions so we don't need to disable the rule.

The first is just to display the message as a failure. A better way may be to do `expect.fail(data.errors[0].message)`

Second can be `expect(data).to.not.have.property('errors')` I think.

Do a negative test locally to make sure the test/failure output is useful

web/src/cypress/support/page-fab.ts
```diff
@@ -12,19 +12,30 @@ declare global {
...
-
   // standard page fab
-  if (!dialOption) return res.click()
+  if (!dialOption)
+    return cy
+      .get('button[data-cy=page-fab]')
+      .should('be.visible')
+      .then(res => {
+        res.click()
+      })
```
mastercactapus: No need for `.then` it will already be chained/queued.
```suggestion
      .click()
```

web/src/cypress/support/page-fab.ts
```diff
@@ -12,19 +12,30 @@ declare global {
...
+      res
+        .trigger('mouseover')
+        .parent()
+        .find(
+          `span[aria-label*=${JSON.stringify(
+            dialOption,
+          )}] button[role=menuitem]`,
+        )
+        .click()
+    })
```
mastercactapus: No need for `.then` Cypress handles queuing/chaining
```suggestion
    .trigger('mouseover')
    .parent()
    .find(
      `span[aria-label*=${JSON.stringify(
        dialOption,
      )}] button[role=menuitem]`,
    )
    .click()
```

PR #555, opened 2020-03-30
dev: eslint enforce @typescript-eslint/no-explicit-any

dctalbot:

**Description:**
This PR provides compatibility for the default eslint rule of @typescript-eslint/no-explicit-any


web/src/app/util/debug.ts
```diff
@@ -1,12 +1,14 @@
-export function debug(args: any): void {
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export function debug(message?: any, ...optionalParams: any[]): void {
   if (!window.console) return // some browsers don't define console if the devtools are closed
-  if (console.debug) console.debug(args)
+  if (console.debug) console.debug(message, optionalParams)
   // not all browsers have `.debug` defined
-  else console.log(args)
+  else console.log(message, optionalParams)
```
mastercactapus: missing spread

web/src/app/util/debug.ts
```diff
@@ -1,12 +1,14 @@
...
 }
-export function warn(...args: any[]): void {
+
+export function warn(message?: any, ...optionalParams: any[]): void {
   if (!window.console) return // some browsers don't define console if the devtools are closed
-  if (console.warn) console.warn(...args)
+  if (console.warn) console.warn(...message, optionalParams)
   // not all browsers have `.warn` defined
-  else console.log(...args)
+  else console.log(...message, optionalParams)
```
mastercactapus: Spread is on wrong argument

web/src/cypress/support/fail-fast.ts
```diff
@@ -1,3 +1,5 @@
+// NOTE the runner mthod is not inherent to Cypress
```
mastercactapus: spelling

web/src/app/lists/PaginatedList.tsx
```diff
@@ -192,6 +192,8 @@ export function PaginatedList(props: PaginatedListProps): JSX.Element {
     // must be explicitly set when using, in accordance with TS definitions
     const urlProps = item.url && {
       component: AppLink,
+      // NOTE button: false? not assignable to true
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
Forfold: you can change the comment here to say that `button` must be set to true when using `to`. `button={false}` was displaying as invalid code with all the URL props set (but off) IIRC. you may want to verify with your IDE though.

web/src/app/lists/PaginatedList.tsx
```diff
@@ -277,12 +279,9 @@ export function PaginatedList(props: PaginatedListProps): JSX.Element {
...
     // explicitly set props to load more, if loader function present
-    const loadProps: any = {}
-    if (loadMore) {
-      loadProps.hasMore = true
-      loadProps.next = loadMore
-    } else {
-      loadProps.hasMore = false
+    const loadProps = {
+      hasMore: Boolean(loadMore),
+      next: loadMore || (() => {}),
```
Forfold: you may want to double check that this does not have any unintended consequences, since the comment indicates that `next` was explicitly _not_ set (`undefined` as opposed to `() => {}`) if `loadMore` is false

web/src/app/util/debug.ts
```diff
@@ -1,12 +1,14 @@
-export function debug(args: any): void {
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export function debug(message?: any, ...optionalParams: any[]): void {
```
Forfold: `message` should always be a `string` here, shouldn't it? if `error` as an object is passed here instead of `error.message`, it probably shouldn't be

web/src/cypress/support/config.ts
```diff
@@ -14,71 +14,49 @@ declare global {
...
-    OverrideURL?: string
-  }
+interface General {
+  PublicURL: string
+  DisableLabelCreation: boolean
+  NotificationDisclaimer: string
+  DisableCalendarSubscriptions: boolean
+}
+interface Auth {
+  RefererURLs: [string]
```
Forfold: nit: can we add spacing between all these interfaces?

web/src/app/lists/PaginatedList.tsx
dctalbot: `button={false}` is fine, it's when we set it to `true` that the assignment is invalid, so we cast it as `any`

Basically, the component wants us to use it like `<ListItem button />` as opposed to `<ListItem button={true} />` but there's no way to validate _presence_ of a prop AFAIK

web/src/app/util/debug.ts
dctalbot: Not necessarily e.g. we can console log objects. You can inspect the `console.log` method to view the types of its parameter list since we just copied those names/types

web/src/app/lists/PaginatedList.tsx
dctalbot: Thanks, I just validated it works as intended, and the docs indicate we have it right:
> hasMore:	boolean tells the InfiniteScroll component on whether to call next function on reaching the bottom ...

I'll add a warning in the function stub (should never be seen)

PR #562, opened 2020-03-31
dev: eslint enforce @typescript-eslint/no-namespace

dctalbot:

**Description:**
This PR provides compatibility for the default eslint rule of @typescript-eslint/no-namespace

- Moves extensions of the Cypress namespace into a `types/` directory
- `tsconfig.json` is made aware of this directory
- removes `@typescript-eslint/no-namespace` config from `eslintrc.js`
- resolve implicit `any`s


web/src/cypress/types/alert.d.ts
```diff
@@ -0,0 +1,40 @@
...
+    createAlert: typeof createAlert
+    createManyAlerts: typeof createManyAlerts
+    closeAlert: typeof closeAlert
+    createAlertLogs: typeof createAlertLogs
+  }
+}
+
+interface Alert {
+  number: number
+  id: number
```
Forfold, line 12: Are we still using the old API for alert calls? New schema should be `id` and `alertID`

dctalbot: Good catch, this is also an issue in `master`, but will be solved when we start generating static response types for graphql requests. I'll update this here in any case.

Forfold: Sounds good

PR #563, opened 2020-03-31
dev: eslint enforce @typescript-eslint/no-use-before-define

KatieMSB:

**Description:**
All function definitions are hoisted above their use to make compatible with the eslint rule @typescript-eslint/no-use-before-define.


web/src/app/apollo.js
```diff
@@ -100,6 +100,14 @@ const simpleCacheTypes = [
...
+const cache = new InMemoryCache({
+  cacheRedirects: {
+    Query: {
+      ...queryCache,
+    },
+  },
+})
+
 simpleCacheTypes.forEach(name => {
   queryCache[camelCase(name)] = (_, args) =>
```
Forfold: We'll have to be careful here, iirc it was a little awkward and order does matter here because we are defining `queryCache = {}`, then iterating over `simpleCacheTypes` to set the keys in `queryCache`, then creating the `cache` from that.

We may need to ignore the rule in this file if moving the `cache` def above causes issues

web/src/cypress/support/alert.ts
```diff
@@ -47,6 +47,59 @@ declare global {
...
+            name
+            description
+            repeat
+        }
+      }
+    }
+  }`
+
+  return cy.graphql(query, { id }).then(res => res.alert)
+}
```
Forfold, line 101: Why does this differ from the previous code before re-ordering?

web/src/app/lists/ControlledPaginatedList.tsx
```diff
@@ -125,6 +118,12 @@ export default function ControlledPaginatedList(
     }
     return []
   }
+  const [_checkedItems, setCheckedItems] = useState<Array<string | number>>([])
```
Forfold, line 121: Let's retain spacing between definitions in js/ts files.

web/src/cypress/integration/alerts.ts
```diff
@@ -378,7 +376,7 @@ function testAlerts(screen: ScreenFormat): void {
     beforeEach(() => {
       cy.createAlertLogs({ count: 200 }).then(_logs => {
         logs = _logs
-        return cy.visit(`/alerts/${logs.alert.id}`)
+        return cy.visit(`/alerts/${logs.alert.number}`)
```
Forfold, line 379: Was this causing a test failure due to the eslint changes?

KatieMSB: Currently looking into it, timepickers tests are also failing for some reason.

Forfold: The current schema doesn't include `alert.number`, probably related to the comment I left above:
```
type Alert {
  id: ID!
  alertID: Int!
  status: AlertStatus!
  summary: String!
  details: String!
  createdAt: ISOTimestamp!
  serviceID: ID!
  service: Service

  # Escalation Policy State for the alert.
  state: AlertState

  # Recent log entries for the alert.
  recentEvents(input: AlertRecentEventsOptions): AlertLogEntryConnection!
}
```

web/src/cypress/support/alert.ts
KatieMSB: should be fixed now

PR #564, opened 2020-03-31
gql2: migrate service ack/close all mutations

Forfold:

**Description:**
This PR drops and replaces the old `updateAlertStatusByService` mutation into the new schema, updating all relevant references to use the newer code.


graphql2/schema.graphql
```diff
@@ -289,10 +289,17 @@ type Mutation {
   updateUserOverride(input: UpdateUserOverrideInput!): Boolean!
   updateHeartbeatMonitor(input: UpdateHeartbeatMonitorInput!): Boolean!
 
+  updateAlertStatusByService(input: UpdateAlertStatusByServiceInput!): Boolean!
```
mastercactapus, line 292: I'm thinking we could call this `updateAlertsByService` to match `updateAlerts` which also only changes status.

Forfold: That sounds good. I couldn’t think of a good name so I went with the old API nomenclature 🙂

PR #578, opened 2020-04-02
ui: fix search focus lost after typing on EP list

Forfold:

**Description:**
This PR fixes an issue on the escalation policy list where focus was lost after typing in the search field. This was due to the fact that we were using the `component` prop where we should have used `render` prop while utilizing `react-router-dom`.

### DIFF
--- a/web/src/cypress/support/page-search.ts
+++ b/web/src/cypress/support/page-search.ts
@@ -12,6 +12,7 @@ export function pageSearch(s: string): Cypress.Chainable {
     cy.get('[data-cy=search-field] input')
       .type(`{selectall}${s}`)
       .should('have.value', s)
+      .should('have.focus')
   })
 }
 
--- a/web/src/app/escalation-policies/PolicyRouter.js
+++ b/web/src/app/escalation-policies/PolicyRouter.js
@@ -47,16 +47,16 @@ export default function PolicyRouter() {
 
   return (
     <Switch>
-      <Route exact path='/escalation-policies' component={renderList} />
+      <Route exact path='/escalation-policies' render={renderList} />
       <Route
         exact
         path='/escalation-policies/:escalationPolicyID'
-        component={renderDetails}
+        render={renderDetails}
       />
       <Route
         exact
         path='/escalation-policies/:escalationPolicyID/services'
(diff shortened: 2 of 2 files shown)


web/src/cypress/support/page-search.ts
```diff
@@ -9,9 +9,12 @@ export function pageSearch(s: string): Cypress.Chainable {
       }) // since we're running tests, it's ok if it is already open
     }
 
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
     cy.get('[data-cy=search-field] input')
       .type(`{selectall}${s}`)
+      .wait(250) // wait a necessary amount of time for debounce delay
```
mastercactapus, line 15: The delay can change, and we can't really count on timing for CI vs local -- what's happening under the covers that causes it to break?

Forfold: This check is here to ensure that no regressions come up in the future similar to what I found yesterday. We could remove it/not include, but then it could happen again without us knowing and resort in us finding a bug like this in prod. We can find a way to import the DELAY const being used in the config if you would prefer. Regardless of timing it should pass (after debounce and processing/logic field should always remain in focus), so an arbitrary amount of 1s would suffice even, but that would slow tests down quite a bit so I wanted to keep it minimal.

mastercactapus: By under the covers, I mean the root issue of what was failing before. I don't think a delay is the correct solution, as there is no guarantee on debounce timing. As per `setTimeout`, the only guarantee of execution is a minimum time, but a slow enough CI job or test environment could have debounce process & render after say 2 seconds and should still pass.

If we need to wait for something, there is never a guarantee on timing with tests, every environment will have edge cases, slow runs, etc.. and result in a flaky test unless we get to the bottom of the root issue.

If there is no possible solution than to wait for debounce to complete, then we should implement a way to wait for it directly, rather than arbitrary waiting. But to know that we'd have to consider what the core problem that debounce is causing.

---

From the PR description I got the impression it was a focus lost issue, and not a debounce/issue (e.g. re-render caused the components to be swapped out) -- which would have no benefit from an added delay here.

PR #582, opened 2020-04-06
dev: generate typescript types from graphql schema

dctalbot:

**Description:**
This PR introduces a new devtool, gqltsgen, which parses `schema.graphql` and ouputs `schema.d.ts`, a types definition file of mapped graphql models

- `make check` will verify that `schema.graphql` and `schema.d.ts` are in sync with the git diff check we already have in place
- a follow up PR will provide examples for how to leverage these types with the intention of removing much of the manually-defined type definitions in `cypress/types`


devtools/gqltsgen/main.go
```diff
@@ -0,0 +1,84 @@
...
+	doc, err := parser.ParseSchemas(src...)
+	if err != nil {
+		log.Fatal("ERORR:", err)
+	}
+
+	w := os.Stdout
+	if *out != "" {
+		fd, err := os.Create(*out)
+		if err != nil {
+			panic(err)
```
mastercactapus, line 39: Use `log.Fatal` as elsewhere

mastercactapus:
@dctalbot Yes, we need a task so that dependencies are properly tracked and built. Developers should never need to run `make generate` as it should happen as part of the normal build as-required.

`make generate` is a `.PHONY` task (always executes) that works as a catch-all, primarily so that `make check` can detect out-of-date and/or mismatched output files.

You can reference other generate tasks from the Makefile for examples:
- `migrate/inline_data_gen.go:`
- `graphql2/mapconfig.go:`
- `graphql2/generated.go:`
- `web/inline_data_gen.go`

One thing I missed -- it should also be a dependency of `cypress:`

Makefile
```diff
@@ -158,7 +158,10 @@ cy-wide-prod-run: web/inline_data_gen.go cypress
 cy-mobile-prod-run: web/inline_data_gen.go cypress
 	make cy-mobile-prod CY_ACTION=run
 
-start: bin/waitfor web/src/node_modules web/src/build/vendorPackages.dll.js bin/runjson
+web/src/app/schema.d.ts: graphql2/schema.graphql
```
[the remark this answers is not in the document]
dctalbot: good catch, thanks!

devtools/gqltsgen/main.go
```diff
@@ -0,0 +1,93 @@
...
+
+func main() {
+	out := flag.String("out", "", "Output file.")
+	flag.Parse()
+	log.SetFlags(log.Lshortfile)
+	var src []*ast.Source
+	for _, file := range flag.Args() {
+		data, err := ioutil.ReadFile(file)
+		if err != nil {
+			log.Fatal("ERORR:", err)
```
Forfold, line 22:
```suggestion
			log.Fatal("ERROR:", err)
```

devtools/gqltsgen/main.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+		src = append(src, &ast.Source{
+			Name:  file,
+			Input: string(data),
+		})
+	}
+
+	doc, err := parser.ParseSchemas(src...)
+	if err != nil {
+		log.Fatal("ERORR:", err)
```
Forfold, line 32:
```suggestion
		log.Fatal("ERROR:", err)
```

devtools/gqltsgen/main.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	w := os.Stdout
+	if *out != "" {
+		fd, err := os.Create(*out)
+		if err != nil {
+			log.Fatal("ERORR:", err)
```
Forfold, line 39:
```suggestion
		log.Fatal("ERROR:", err)
```

PR #590, opened 2020-04-08
wds: re-enable hot module reloading

Forfold:

**Description:**
This PR re-enables hot module reloading with the Webpack Dev Server by serving the hot reload hash files to the build directory for WDS to pick up and trigger a reload.

### DIFF
--- a/web/src/webpack.config.js
+++ b/web/src/webpack.config.js
@@ -87,8 +87,6 @@ module.exports = (env = { GOALERT_VERSION: 'dev' }) => ({
 
     // host: HOST,
     port: PORT,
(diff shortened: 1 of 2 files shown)


web/src/webpack.config.js
```diff
@@ -85,6 +87,8 @@ module.exports = (env = { GOALERT_VERSION: 'dev' }) => ({
 
     // host: HOST,
     port: PORT,
+
+    writeToDisk: true,
```
mastercactapus, line 91: I think we can drop this now and just keep the Filename bits above

PR #600, opened 2020-04-13
dev: generate string union types instead of const enums

dctalbot:

**Description:**
Ambient const enums are not allowed when the '--isolatedModules' flag is provided with typescript, effectively making the exported enums unusable

- uses string union types instead of enums
- run `prettier -l --write schema.d.ts` after generating
- reintroduce double newlines for readability

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -160,6 +160,7 @@ cy-mobile-prod-run: web/inline_data_gen.go cypress
 
 web/src/schema.d.ts: graphql2/schema.graphql
 	go generate ./web/src
+	(cd web/src && yarn prettier -l --write schema.d.ts)
 
 start: bin/waitfor web/src/node_modules web/src/build/vendorPackages.dll.js bin/runjson web/src/schema.d.ts
 	# force rebuild to ensure build-flags are set
--- a/Makefile
+++ b/Makefile
@@ -158,9 +158,8 @@ cy-wide-prod-run: web/inline_data_gen.go cypress
 cy-mobile-prod-run: web/inline_data_gen.go cypress
 	make cy-mobile-prod CY_ACTION=run
 
-web/src/schema.d.ts: graphql2/schema.graphql
+web/src/schema.d.ts: graphql2/schema.graphql web/src/node_modules
 	go generate ./web/src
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -160,6 +160,7 @@ cy-mobile-prod-run: web/inline_data_gen.go cypress
 
 web/src/schema.d.ts: graphql2/schema.graphql
 	go generate ./web/src
+	(cd web/src && yarn prettier -l --write schema.d.ts)
```
mastercactapus, line 163: Do this in the generate file rather than relying on it here. `go generate ./web/src` should result in the final file & formatting

dctalbot: Do you know if a way to run an executable from `main.go` that's aware of our $PATH and config options?

mastercactapus: You can use `exec.LookupPath` (I think that was the one) but it would be easier to just add it to the generate file instead.

We are doing something similar in `graphql2/gen.go` where we are running `goimports` (our go formatter) after we generate the files. You can just add a second `go:generate` line that will call `yarn prettier` and should be good.

You could also call it directly instead of relying on yarn with `./node_modules/.bin/prettier` and that might be a bit safer/better.

Oh, and I forgot -- since we rely on prettier, add `web/src/node_modules` to as a dependency of `web/src/schema.d.ts` on line 161. That will make sure things are handled e.g. if you run `make clean` first (or just cloned the repo).

web/src/schema.d.ts
```diff
@@ -30,60 +30,64 @@ export interface Query {
...
-  IntegrationKeysPerService = 'IntegrationKeysPerService',
-  UnackedAlertsPerService = 'UnackedAlertsPerService',
-  TargetsPerSchedule = 'TargetsPerSchedule',
-  HeartbeatMonitorsPerService = 'HeartbeatMonitorsPerService',
-  UserOverridesPerSchedule = 'UserOverridesPerSchedule',
-}
+
+export type ConfigType = 'string' | 'stringList' | 'integer' | 'boolean'
+
+export type SystemLimitID =
```
m17ch, line 77: 😍

PR #603, opened 2020-04-14
alert-log: added notification not sent

KatieMSB:

**Description:**
Indicates when no notification is sent because on-call users have no immediate notification rule.


engine/npcyclemanager/update.go
```diff
@@ -27,6 +29,41 @@ func (db *DB) update(ctx context.Context, all bool, alertID *int) error {
...
+		if err != nil {
+			return err
+		}
+		data = append(data, rec)
+	}
+
+	for _, rec := range data {
+		err = db.log.LogTx(ctx, tx, rec.alertID, alertlog.TypeNoNotificationSent, rec.meta)
+		if err != nil {
+			return errors.Wrap(err, "queue outgoing messages")
```
mastercactapus, line 64: Wrap message should be something like "log no notification sent", queuing happens above

engine/npcyclemanager/update.go
```diff
@@ -27,6 +29,41 @@ func (db *DB) update(ctx context.Context, all bool, alertID *int) error {
 	}
 	log.Debugf(ctx, "Updating notification cycles.")
 
-	_, err = db.lock.Exec(ctx, db.queueMessages)
-	return errors.Wrap(err, "queue outgoing messages")
+	tx, err := db.lock.BeginTx(ctx, nil)
+	if err != nil {
+		return err
```
mastercactapus, line 34: let's wrap these since it's in an engine module

messages should just be concise as to what was attempted, e.g. here can be "begin tx"

engine/npcyclemanager/update.go
```diff
@@ -27,6 +29,41 @@ func (db *DB) update(ctx context.Context, all bool, alertID *int) error {
...
+	rows, err := tx.StmtContext(ctx, db.queueMessages).QueryContext(ctx)
+	if err != nil {
+		return err
+	}
+	defer rows.Close()
+
+	type record struct {
+		alertID int
+		userID  string
+		meta    *alertlog.NotificationMetaData
```
mastercactapus, line 47: Since this isn't used we can leave it out, just pass nil as the final arg to `LogTx`

migrate/migrations/20200413113132-add-no-notification-alert-log.sql
```diff
@@ -1,6 +1,7 @@
 -- +migrate Up notransaction
 
 ALTER TYPE enum_alert_log_event ADD VALUE IF NOT EXISTS 'no_notification_sent';
+ALTER TYPE enum_alert_log_subject_type ADD VALUE IF NOT EXISTS 'no_notification';
```
mastercactapus, line 4: `no_notitifcation` is the event type, the subject type should be `user` rather than adding a new one.

PR #608, opened 2020-04-15
graphql: deprecate v1 endpoint

Forfold:

**Description:**
This PR introduces the first stage in deprecating the GraphQL V1 endpoint. This can be disabled from within the Admin config section of the UI. Upon disabling, all requests to `/v1/graphql` will return a `404`. This option is currently opt-in, though in the future it will transition to be opt-out, and then finally to a complete deletion of the endpoint. We recommend using `/api/graphql` moving forward.


Forfold:
> Noticed that `make smoketest GOALERT_HTTP_PREFIX=/foobar` is failing. I wonder if it's worth adding this to our pipeline so we don't forget to test this in the future

Hmm, it looks like it's failing in master too when testing with a prefix @dctalbot. If anything I'd say we create an issue to represent that separately.

app/inithttp.go
```diff
@@ -122,6 +122,9 @@ func (app *App) initHTTP(ctx context.Context) error {
 		// limit max request size
 		maxBodySizeMiddleware(app.cfg.MaxReqBodyBytes),
 
+		// return a 404 if the v1 graphql api is disabled
+		handleServingV1GraphQLRequests(),
```
mastercactapus, line 126: This should be something like `graphQLV1DeprecationMiddleware` or `graphQLV1DisableMiddleware`. Handle implies it's a handler rather than middleware.

Though, it could probably just be inlined too, there's not much logic to it.

config/config.go
```diff
@@ -27,6 +27,7 @@ type Config struct {
 		DisableSMSLinks              bool   `public:"true" info:"If set, SMS messages will not contain a URL pointing to GoAlert."`
 		DisableLabelCreation         bool   `public:"true" info:"Disables the ability to create new labels for services."`
 		DisableCalendarSubscriptions bool   `public:"true" info:"If set, disables all active calendar subscriptions as well as the ability to create new calendar subscriptions."`
+		DisableV1GraphQL             bool   `public:"true" info:"Deprecated API. Disables the ability to use the /v1/graphql api. Please use /api/graphql instead."`
```
mastercactapus, line 30: Let's rephrase this a bit, maybe "Disables the deprecated /v1/graphql endpoint (replaced by /api/graphql)."

Still communicates the same info, but in more of a message-to-admin style than documentation style. Also, this doesn't need to be public, as there is no user-facing code that uses or references it anymore.

web/src/cypress/support/config.ts
```diff
@@ -86,6 +89,9 @@ function merge(dst: Config, src: ConfigInput): Config {
   return dst
 }
 
+/*
+ * Updates the current config with newCfg
```
mastercactapus, line 93: Can we format these a little differently?

```
// updateConfig updates the current config with newConfig
```

That would match the Go style, and the majority of our js/ts function descriptions.
For this one it might also be helpful to indicate it merges it with the current though so it's obvious what "update" means (without looking at "set")

```
// updateConfig updates the current config by merging the provided newCfg values
```

Something like that

web/src/cypress/support/config.ts
```diff
@@ -65,6 +65,9 @@ function getConfig(): Cypress.Chainable<Config> {
   return cy.adminLogin(true).then((tok: string) => getConfigDirect(tok))
 }
 
+/*
+ * Replaces the current config completely with cfg
```
mastercactapus, line 69: Same as below, let's format this a little differently to match

web/src/cypress/support/util.ts
```diff
@@ -77,6 +77,15 @@ export function testScreen(
...
       it(adminLogin ? 'admin login' : 'login', () => {}) // required due to mocha skip bug
     }
+
+    before(() =>
+      cy.updateConfig({
+        General: {
+          DisableV1GraphQL: true,
+        },
+      }),
+    )
```
mastercactapus, line 87: Let's just add this field to the default config in `resetConfig` instead of adding a step

web/src/cypress/support/config.ts
Forfold: That sounds good. I'm used to the convention I learned in college where `// comments` are for single line items, and
```
/*
 * block comments
 */
```
are for function definitions.

dctalbot: I'd follow the respective language conventions. If this were Go, inline is used basically everywhere besides package comments (See [here](https://golang.org/doc/effective_go.html#commentary))

Typescript/javascript is less formal, but there are tools like jsdoc/typedoc that have well-defined rules

Forfold: That definitely makes sense. For instance, above the test def [here](https://github.com/target/goalert/pull/608/files#diff-199d37c4de5ee0e358ded986a41688bbR11) I'm using `// inline` comments

PR #609, opened 2020-04-15
cypress: remove api endpoint option

dctalbot:

**Description:**
In the spirit of #608 this PR removes the option to specify a graphql API endpoint other than `'/api/graphql'`

### DIFF
--- a/web/src/cypress/support/graphql.ts
+++ b/web/src/cypress/support/graphql.ts
@@ -5,7 +5,7 @@ interface GraphQLResponse {
 }
 
 // runs a graphql query returning the data response (after asserting no errors)
-function graphql2(query: string, variables?: any): Cypress.Chainable<any> {
+function graphql(query: string, variables?: any): Cypress.Chainable<any> {
   const url = '/api/graphql'
   if (!variables) variables = {}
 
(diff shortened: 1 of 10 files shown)


web/src/cypress/support/graphql.ts
```diff
@@ -5,11 +5,8 @@ interface GraphQLResponse {
 }
 
 // runs a graphql query returning the data response (after asserting no errors)
-function graphql(
-  query: string,
-  variables?: any,
-  url = '/v1/graphql',
-): Cypress.Chainable<any> {
+function graphql2(query: string, variables?: any): Cypress.Chainable<any> {
```
Forfold, line 8: What do you think of keeping the name as `graphql`? I think in the very near future we will be able to start stripping away code like this and have a single source of truth rather than `graphql` and `graphql2` constantly.

That would cause a bit of bloat in this PR since we would have to update references like `cy.graphql2()` to `cy.graphql()`, but I think it's worth it. I'd be happy to help, lemme know your thoughts

dctalbot: I agree, let's get it done while we're thinking about it

PR #616, opened 2020-04-16
sms: support all twilio stop codes

Forfold:

**Description:**
This PR enables all Twilio stop codes via SMS: STOP, STOPALL, UNSUBSCRIBE, CANCEL, END, and QUIT.

**Additional Info:**
Stop codes are now processed on exact matches, to fall inline with [Twilio's processing logic](https://support.twilio.com/hc/en-us/articles/223134027-Twilio-support-for-opt-out-keywords-SMS-STOP-filtering-): 
> Only single-word messages will trigger the block. So, for example, replying STOP will stop the customer from receiving messages from that particular Twilio number, but replying "STOP PLEASE" or "PLEASE CANCEL" will not.

### DIFF
--- a/notification/twilio/sms.go
+++ b/notification/twilio/sms.go
@@ -266,7 +266,10 @@ func (s *SMS) ServeMessage(w http.ResponseWriter, req *http.Request) {
 	}
 
 	body := req.FormValue("Body")
-	if strings.Contains(strings.ToLower(body), "stop") {
+	// twilio unsubscribes on single-word matches
+	// i.e. "please stop" will not result in unsubscribing on their end.
+	checkBody := func(str string) bool { return strings.ToLower(body) == str }
+	if checkBody("stop") || checkBody("stopall") || checkBody("unsubscribe") || checkBody("cancel") || checkBody("end") || checkBody("quit") {
 		err := retry.DoTemporaryError(func(int) error {
 			errCh := make(chan error, 1)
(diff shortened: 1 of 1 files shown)


notification/twilio/sms.go
```diff
@@ -266,7 +266,10 @@ func (s *SMS) ServeMessage(w http.ResponseWriter, req *http.Request) {
 	}
 
 	body := req.FormValue("Body")
-	if strings.Contains(strings.ToLower(body), "stop") {
+	// twilio unsubscribes on single-word matches
+	// i.e. "please stop" will not result in unsubscribing on their end.
+	checkBody := func(str string) bool { return strings.ToLower(body) == str }
+	if checkBody("stop") || checkBody("stopall") || checkBody("unsubscribe") || checkBody("cancel") || checkBody("end") || checkBody("quit") {
```
mastercactapus, line 272: Maybe we can just break this out into it's own func -- should make things easier when we add support for the start codes as well. We can also get away with calling `ToLower` once for all options and avoid all the `||` conditions.

```go
func isStopMessage(body string) bool {
  switch strings.ToLower(body) {
  case "stop", "stopall",  "...":
    return true
  }

  return false
}

func ServeMessage() {
  // ...
  body := req.FormValue("Body")
  if isStopMessage(body) {
    // ...
  }
  // ...
}
```

PR #627, opened 2020-04-20
devtool: restart backend task for cypress tests

arurao:

**Description:**
This PR adds a new devtool, `procwrap`, to make sure that there are no deadlocks when executing a query during cypress tests, by ensuring the order of events to always be as: 
Stop backend - Execute query - Start backend

Procwrap handles signals like stop, start, signal and shutdown.
Tasks have been added to the cypress plugins that take care of handling plugins like trigger, start and stop.

**Additional Info:**
Developers can test this by running commands like `make cy-wide` and the others, and observe in the tests that wherever a query needs to be executed, the order of stop backend - execute query - start backend is preserved. 
For example: In profile.ts the order can be seen as
Screen Shot 2020-04-15 at 2 02 29 PM
Developers should be able to run the cypress test commands as before, without observing any deadlocks.


devtools/procwrap/main.go
```diff
@@ -0,0 +1,188 @@
...
+		mx.Lock()
+		defer mx.Unlock()
+	}
+	if cmd == nil {
+		return lastExitCode
+	}
+	log.Println("stopping", flag.Arg(0))
+	close(stopping)
+
+	cmd.Process.Signal(os.Interrupt)
```
mastercactapus, line 172: I think this was intended to be passed `sig`

PR #630, opened 2020-04-21
sms: support all twilio start codes

Forfold:

**Description:**
This PR enables [all Twilio start codes via SMS](https://support.twilio.com/hc/en-us/articles/223134027-Twilio-support-for-opt-out-keywords-SMS-STOP-filtering-): START, YES, and UNSTOP.

**Additional Info:**
Fixes polling on the contact method list by using apollo's `useQuery` hook instead of our util component `<Query />`.


user/contactmethod/store.go
```diff
@@ -96,6 +105,33 @@ func NewDB(ctx context.Context, db *sql.DB) (*DB, error) {
...
+	var id string
+	err = db.enable.QueryRowContext(ctx, n.Type, n.Value).Scan(&id)
+
+	if err == nil {
+		// NOTE: maintain a record of consent/dissent
+		logCtx := log.WithFields(ctx, log.Fields{
+			"contactMethodID": id,
+		})
+
+		log.Logf(logCtx, "Contact method resubscribe code received.")
```
mastercactapus, line 129: As with below, this can be `Contact method START received.` to match `STOP`.

We could use "subscribe/resubscribe/unsubscribe", "opt-in/opt-out", etc.. but since STOP is pretty common verbiage, I think START and STOP might be the most predictable/guessable words if one needed to search logs/records for a timestamp.

In this case, I think an explicit predictable keyword is best (even if less readable) since that's the primary reason/use for adding these two log lines.

notification/manager.go
```diff
@@ -226,7 +226,10 @@ func (m *Manager) receive(ctx context.Context, providerID string, resp *MessageR
 	}),
 		"response received",
 	)
-	if resp.Result == ResultStop {
+
+	if resp.Result == ResultStart {
+		return m.r.Start(ctx, resp.From)
+	} else if resp.Result == ResultStop {
 		return m.r.Stop(ctx, resp.From)
 	}
```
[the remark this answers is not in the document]
Forfold: There is another flat return below it (`return m.r.Receive(ctx, resp.ID, resp.Result)`) which makes the `else if` necessary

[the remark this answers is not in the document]
Forfold: Can use a switch if preferred though

user/contactmethod/store.go
Forfold: That works, though we will have to be extra cognizant in the future if we are debugging cases using the non-recommended codes as it could lead to confusion while searching the logs. Future/new developers might not know this either. Opt-out/in might be the most appropriate words on second though, since re/un-subscribe could also be confused for keywords.

Forfold: We did go with `opt-out` instead of `STOP` in the previous PR that we merged, so I think it may be best to stick with that path as we move forward.

PR #634, opened 2020-04-22
deps: revert updates to material-ui/core introduced in #624

Forfold:

**Description:**
This PR patches a failing test in CI due to a patch update with `@material-ui/core`. Updates to this package will be ignored by dependabot until at least the next minor version increase is introduced.

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -7,3 +7,6 @@ update_configs:
       - match:
           dependency_name: "@date-io/luxon"
           version_requirement: "<=2.4.0"
+      - match:
+        dependency_name: "@material-ui/core"
+        version_requirement: "<=4.9.x"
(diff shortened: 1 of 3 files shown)


.dependabot/config.yml
```diff
@@ -7,3 +7,6 @@ update_configs:
       - match:
           dependency_name: "@date-io/luxon"
           version_requirement: "<=2.4.0"
+      - match:
```
m17ch, line 10: I wonder for each ignored_update, if we should get in the practice of adding a comment with a link to the PR that describes why the entry exists (for future reference).

i.e. Maybe simply include above each match line:
`# see https://github.com/target/goalert/pull/634 `

What do you think?

m17ch:
nit: since you're not really patching any cypress tests, I'd adjust the PR name & description: 

i.e 

> revert updates to `material-ui/core` introduced in https://github.com/target/goalert/pull/624

PR #651, opened 2020-04-28
ui: Convert Admin Section to TypeScript

KatieMSB:

**Description:**
Converts AdminSection to use hooks and typescript.


web/src/app/admin/AdminSection.tsx
```diff
@@ -0,0 +1,105 @@
...
+}
+
+interface FieldProps {
+  id: string
+  label: string
+  description?: string
+  value: string
+  type: ConfigType
+  password?: boolean
+}
```
dctalbot, line 31: Sorry I didn't notice this last time I reviewed, but we should probably 
```ts
interface FieldProps extends ConfigValue {
  label: string
}
```

PR #656, opened 2020-04-30
alert logs: display status of outgoing messages

Forfold:

**Description:**
This PR adds to the alert logs the last status of an outgoing message, as well as logs when a message fails to send. This is visible to a user by viewing the alert details page of an alert (i.e. `mygoalert.com/alerts/1234`).

**Out of Scope:**
- Failure logging when a contact method is disabled.
- More flavorful error messages/styling.
- Messages appear to be `queued` far longer than they actually are. For instance, it will remain `queued` although the text has already been delivered. I've noticed this lasts for about 60 seconds, give or take the time to poll on the UI (3s).

image

image

image


alert/log/entry.go
```diff
@@ -152,6 +154,22 @@ func (e Entry) String() string {
...
+	if e.lastStatus.String == "failed" {
+		return "ERROR"
+	}
+
+	if e.lastStatus.String == "delivered" {
+		return "OK"
+	}
+
+	return ""
+}
```
mastercactapus, line 171: Since these are specific for only one type of Entry, I wonder if there's a better way we can do this. What would the `Status` of an escalation be for instance?

We could just have one method `OutgoingMessageState` that returns a struct or nil with `StatusDetails` and `LastStatus` (since it only has value when there is a message id). I'd also wait until GraphQL to map to `ERROR` or `OK` and keep this part plain as possible.

Alternatively, we could also leave Entry untouched, and do a separate query for the state (e.g. with a datafetcher using the ID from meta). That would have the benefit that we wouldn't always be joining to the messages table. We would only access it if there was a log entry with a message ID in the results.

The former would probably be the most straightforward and resolves the ambiguity. Though the latter would probably cleanest from an API standpoint (not adding arbitrary fields/methods and keeping everything general on Entry).

Forfold: > Alternatively, we could also leave Entry untouched, and do a separate query

Would there be any limitations to this? After the first page we load the next 150 logs, and if many of those have their state set, that's a lot of extra trips to the DB.

mastercactapus: The datafetcher would batch them together, so it would only be one query for all 150 worst case (though not all 150 would be `notification_sent`).

Forfold: Let's try leaving Entry untouched and grab the data as a separate query, then

engine/sendmessage.go
```diff
@@ -114,15 +114,21 @@ func (p *Engine) sendMessage(ctx context.Context, msg *message.Message) (*notifi
 		return &notification.MessageStatus{State: notification.MessageStateFailedPerm}, nil
 	}
 
+	meta := alertlog.NotificationMetaData{
+		MessageID: msg.ID,
+	}
+
 	status, err := p.cfg.NotificationSender.Send(ctx, notifMsg)
 	if err != nil {
+		p.cfg.AlertLogStore.MustLog(ctx, msg.AlertID, alertlog.TypeNotificationSent, meta)
```
mastercactapus, line 123: In this case, something went wrong before making the attempt so we wouldn't want to log anything. This also could be a test or verification message with no AlertID.

Any case that should be recorded as a sent-but-retry (temp.) or sent-but-failed (perm.) will indicate in the status but with a `nil` err and get logged below.

graphql2/graphqlapp/alert.go
```diff
@@ -39,6 +41,36 @@ func (a *AlertLogEntry) Message(ctx context.Context, obj *alertlog.Entry) (strin
 	return e.String(), nil
 }
 
+func (a *AlertLogEntry) State(ctx context.Context, obj *alertlog.Entry) (*graphql2.AlertLogEntryState, error) {
+	e := *obj
+	meta, ok := e.Meta().(*alertlog.NotificationMetaData)
+	if !ok {
```
mastercactapus, line 47: Also, check that `meta` isn't nil -- as will be the case for invalid or missing metadata (e.g. old log entries made before this PR). Otherwise `meta.MessageID` will panic below.
```suggestion
	if !ok || meta == nil {
```

graphql2/graphqlapp/alert.go
```diff
@@ -39,6 +41,36 @@ func (a *AlertLogEntry) Message(ctx context.Context, obj *alertlog.Entry) (strin
...
+	s, err := (*App)(a).FindOneAlertLogMessageStatus(ctx, meta.MessageID)
+	if err != nil {
+		return nil, errors.Wrap(err, "find alert log state")
+	}
+
+	status := ""
+	switch s.State {
+	case notification.MessageStateFailedPerm:
+		status = "ERROR"
+	case notification.MessageStateDelivered:
```
mastercactapus, line 60: Sent should probably be OK as well -- we don't always have delivery confirmation.
```suggestion
	case MessageStateSent, notification.MessageStateDelivered:
```

graphql2/graphqlapp/alert.go
```diff
@@ -39,6 +41,36 @@ func (a *AlertLogEntry) Message(ctx context.Context, obj *alertlog.Entry) (strin
...
+		status = "OK"
+	}
+
+	if s.Details == "" {
+		return nil, nil
+	}
```
mastercactapus, line 66: We can probably drop this -- we could check details & status and return nil if empty, but I don't think we should make the API depend on that.

UI can just do `state?.details` if we want to do this check. I don't think we should have a case where we have a failure but no message, but it wouldn't change the fact that it had still failed from an API standpoint

graphql2/graphqlapp/dataloaders.go
```diff
@@ -37,9 +39,23 @@ func (a *App) registerLoaders(ctx context.Context) context.Context {
 	ctx = context.WithValue(ctx, dataLoaderKeyService, dataloader.NewServiceLoader(ctx, a.ServiceStore))
 	ctx = context.WithValue(ctx, dataLoaderKeyUser, dataloader.NewUserLoader(ctx, a.UserStore))
 	ctx = context.WithValue(ctx, dataLoaderKeyCM, dataloader.NewCMLoader(ctx, a.CMStore))
+	ctx = context.WithValue(ctx, dataLoaderKeyAlertLogMessageStatus, dataloader.NewAlertLogMessageStatusLoader(ctx, a.NotificationStore))
 	return ctx
 }
 
+func (app *App) FindOneAlertLogMessageStatus(ctx context.Context, id string) (*notification.MessageStatus, error) {
```
mastercactapus, line 46: We should probably update the names to be `NotificationMessageStatus` for consistency

graphql2/schema.graphql
```diff
@@ -705,6 +705,12 @@ type AlertLogEntry {
   id: Int!
   timestamp: ISOTimestamp!
   message: String!
+  state: AlertLogEntryState
+}
+
+type AlertLogEntryState {
+  details: String!
+  status: String
```
mastercactapus, line 713: Instead of a string, maybe `'OK' | 'ERR'` as an enum. Will make it a lot easier for consumers of the API to know/understand what to expect

notification/store.go
```diff
@@ -280,3 +295,45 @@ func (db *DB) VerifyContactMethod(ctx context.Context, cmID string, code int) er
 
 	return nil
 }
+
+func (db *DB) FindManyMessageStatuses(ctx context.Context, ids ...string) ([]MessageStatus, error) {
+	err := permission.LimitCheckAny(ctx, permission.User, permission.Admin)
```
mastercactapus, line 300: Can just be a `User` check, as `Admin` will automatically qualify (some old code references both, but it's unnecessary with the current permissions package).

notification/store.go
```diff
@@ -280,3 +295,45 @@ func (db *DB) VerifyContactMethod(ctx context.Context, cmID string, code int) er
...
+		if err != nil {
+			return nil, err
+		}
+
+		switch lastStatus {
+		case "delivered":
+			s.State = MessageStateDelivered
+		case "failed":
+			s.State = MessageStateFailedPerm
+		}
```
mastercactapus, line 333:
```suggestion
		switch lastStatus {
		case "pending", "sending", "queued_remotely":
			s.State = MessageStateActive
		case "sent":
			s.State = MessageStateSent
		case "delivered":
			s.State = MessageStateDelivered
		case "failed", "bundled": // bundled message was not sent (replaced) and should never be re-sent
			// temporary if retry
			if hasNextRetry {
				s.State = MessageStateFailedTemp
			} else {
				s.State = MessageStateFailedPerm
			}
		default:
			return nil, fmt.Errorf("unknown last_status %s", lastStatus)
		}
```

Here are all the possible types for reference:
```
    'pending',
    'sending',
    'queued_remotely',
    'sent',
    'delivered',
    'failed',
    'bundled'
```

As for the `hasNextRetry` we can add a `next_retry_at notnull` to the SELECT query

graphql2/graphqlapp/alert.go
```diff
@@ -39,6 +41,36 @@ func (a *AlertLogEntry) Message(ctx context.Context, obj *alertlog.Entry) (strin
...
+
+	s, err := (*App)(a).FindOneAlertLogMessageStatus(ctx, meta.MessageID)
+	if err != nil {
+		return nil, errors.Wrap(err, "find alert log state")
+	}
+
+	status := ""
+	switch s.State {
+	case notification.MessageStateFailedPerm:
```
mastercactapus, line 58: Should we add Temp to this too? Or since it will retry leave it out? It will have an error message until it tries again..

Forfold: Temp works. If it's listed as a "temp" failure we can list it as `Failed. Retrying...` on the UI.

graphql2/graphqlapp/alert.go
Forfold: Should we then change the UI to say "Sent" when sent? I don't think it would make sense to display "Delivered" on the UI when it might not have

Forfold: Right now, the UI displays "Delivered" if the status is "OK"

graphql2/graphqlapp/alert.go
Forfold: This is a remnant from before I refactored, without this check it was returning

```
state: {
  details: "",
  status: "",
}
```
Looks like it's good to remove now :+1:

notification/store.go
Forfold: Should we pass this back to graphql transparently instead of mapping these to `ERROR` and `OK`? Then we can make more granular hints for our users (i.e. `Sent` vs `Delivered`, `Failed` vs `Failed. Retrying`, etc)

graphql2/graphqlapp/alert.go
Forfold: Edit: See below, this will have to remain out of scope for more granular details wording

PR #665, opened 2020-05-06
dev: parallelize backend smoketests

arurao:

**Description:**
This PR handles parallelizing backend smoketests, wherever possible. Some backend smoketests were using `t.Run` for many different test cases but were not calling `t.Parallel` inside the new test scope. Running them in parallel speeds up these tests significantly.


Forfold:
Are you able to provide benchmarks to the description (notably the overall time to run before vs. after these changes) for easy reference later on?

smoketest/graphqloncall_test.go
```diff
@@ -17,54 +17,55 @@ import (
...
-		idCounter++
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			h := harness.NewHarness(t, "", "escalation-policy-step-reorder")
+			defer h.Close()
 
-		u1, u2 := h.CreateUser(), h.CreateUser()
-		data.User1 = u1
-		data.User2 = u2
+			doQL := func(t *testing.T, query string, res interface{}) {
```
mastercactapus, line 29: Since this has `t` passed in, it can be lifted out of the `t.Run` closure

smoketest/migrations_test.go
```diff
@@ -667,6 +667,7 @@ func TestMigrations(t *testing.T) {
 			if checkDiff(t, "UP", migrationName, upSnap, s) {
 				t.Fatalf("UP migration did not restore previous schema")
 			}
+			t.Parallel()
```
mastercactapus, line 670: Migration tests cannot be run in parallel without each having an independent harness/DB. Which should probably wait for the refactor.

smoketest/systemlimits_test.go
```diff
@@ -22,11 +25,11 @@ func TestSystemLimits(t *testing.T) {
...
-			({{uuid "generic_user1"}}, 'User 1'),
-			({{uuid "generic_user2"}}, 'User 2'),
-			({{uuid "generic_user3"}}, 'User 3'),
-			({{uuid "generic_user4"}}, 'User 4'),
-			({{uuid "generic_user5"}}, 'User 5');
+			('50322144-1e88-43dc-b638-b16a5be7bad6', 'User 1'),
+			('dfcc0684-f045-4a9f-8931-56da8a014a44', 'User 2'),
+			('016d5895-b20f-42fd-ad6c-7f1e4c11354d', 'User 3'),
+			('dc8416e1-bf15-4248-b09f-f9294adcb962', 'User 4'),
+			('c1dadc8b-b0fc-41e3-a015-5a14c5c19433', 'User 5');
```
mastercactapus, line 32: We should stick to using the UUID generators for consistency. They should be thread-safe. Additionally, the SQL query will be evaluated for each (e.g. used as template) so `generic_user1` will match `generic_user1` within each `t.Run`

smoketest/systemlimits_test.go
```diff
@@ -135,18 +111,87 @@ func TestSystemLimits(t *testing.T) {
...
+
+				return "", g.Errors[0].Message
+			}
+
+			doQuery := func(t *testing.T, query string) string {
+				id, errMsg := doQL(t, query)
+				assert.Empty(t, errMsg, "error message")
+				return id
+			}
+			doQueryExpectError := func(t *testing.T, query string, expErr string) {
```
mastercactapus, line 146:
```suggestion
			doQueryExpectError := func(t *testing.T, query, expErr string) {
```

smoketest/systemlimits_test.go
```diff
@@ -159,31 +204,111 @@ func TestSystemLimits(t *testing.T) {
 				7. set limit to -1
 				8. create (should work)
 			*/
+			t.Parallel()
+			h := harness.NewHarness(t, sql, "limit-configuration")
+			defer h.Close()
+
+			doQL := func(t *testing.T, query string) (string, string) {
```
mastercactapus, line 211: I think we can move these helpers out of the t.Run scope and flatten things a bit, for this one we'll just need to add `h` as an argument.

smoketest/systemlimits_test.go
```diff
@@ -159,31 +204,111 @@ func TestSystemLimits(t *testing.T) {
...
+						return p
+					}
+					if uuid, ok := uuids[id]; ok {
+						return uuid
+					}
+					uuid := h.UUID(id)
+					uuids[id] = uuid
+					return uuid
+				},
+			})
```
mastercactapus, line 260: I think `h.UUID()` would be easier to read than going through empty template rendering, or am I missing something?

arurao: I agree. `h.UUID()` would be much easier, but with the tests running in parallel the harness is now only accessible within either the `t.Run()` or the helper functions (`doQL()`) I am not sure how to send `h.UUID()` from the test cases. Or is there a way to incorporate the harness within `check()`?

smoketest/systemlimits_test.go
arurao: Yes, the reason why we can not use the UUID generators is because of the refactoring, the harness is only accessible within `t.Run()` and helper functions. More details in the next comment below.

smoketest/systemlimits_test.go
mastercactapus: The harness can be passed as a parameter, much like we're doing with `t` where needed

arurao:
> Are you able to provide benchmarks to the description (notably the overall time to run before vs. after these changes) for easy reference later on?

Previously
- graphqloncall_test.go : 97s
- graphqloncallassignments_test.go : 97s

Now
- graphqloncall_test.go : 26s
- graphqloncallassignments_test.go : 45s

smoketest/graphqloncallassignments_test.go
```diff
@@ -29,33 +29,11 @@ func TestGraphQLOnCallAssignments(t *testing.T) {
...
-		if res == nil {
-			return
-		}
-		err := json.Unmarshal(g.Data, &res)
-		if err != nil {
-			t.Fatal(err)
-		}
-	}
-
 	var idCounter int
```
mastercactapus, line 37: Since this is used for names per-harness, it can/should live within `t.Run`. Otherwise we need to use something like `atomic.AddInt32` to handle the concurrent access.

PR #686, opened 2020-05-18
alert logs: display when contact method is disabled

Forfold:

**Description:**
This PR logs notification_sent (but failed) when the applicable user's contact method is disabled.

### DIFF
--- a/engine/message/db.go
+++ b/engine/message/db.go
@@ -264,7 +264,7 @@ func NewDB(ctx context.Context, db *sql.DB, c *Config, a alertlog.Store) (*DB, e
 				set
 					last_status = 'failed',
 					last_status_at = now(),
-					status_details = 'contact method disabled',
+					status_details = 'failed: contact method disabled',
 					cycle_id = null,
 					next_retry_at = null
 				from user_contact_methods cm
--- a/engine/message/db.go
+++ b/engine/message/db.go
@@ -264,7 +264,7 @@ func NewDB(ctx context.Context, db *sql.DB, c *Config, a alertlog.Store) (*DB, e
 				set
 					last_status = 'failed',
 					last_status_at = now(),
-					status_details = 'failed: contact method disabled',
+					status_details = 'contact method disabled',
 					cycle_id = null,
 					next_retry_at = null
 				from user_contact_methods cm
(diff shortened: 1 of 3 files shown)


engine/message/db.go
```diff
@@ -569,11 +575,37 @@ func (db *DB) _SendMessages(ctx context.Context, send SendFunc, status StatusFun
 		return errors.Wrap(err, "clear disabled status updates")
 	}
 
-	_, err = tx.Stmt(db.failDisabledCM).ExecContext(execCtx)
+	// processes disabled CMs and writes to alert log if disabled
+	cms, err := tx.Stmt(db.failDisabledCM).QueryContext(execCtx)
```
mastercactapus, line 579: name this `rows` instead of `cms` as a covention, then add a `defer rows.Close()` after the err check.

engine/message/db.go
```diff
@@ -256,19 +259,22 @@ func NewDB(ctx context.Context, db *sql.DB, c *Config) (*DB, error) {
...
+					cycle_id = null,
+					next_retry_at = null
+				from user_contact_methods cm
+				where
+					msg.last_status = 'pending' and
+					msg.message_type != 'verification_message' and
+					cm.id = msg.contact_method_id and
+					cm.disabled
+				returning msg.id, alert_id
+			) select distinct id, alert_id from disabled
```
mastercactapus, line 277: We'll need to add the `user_id` here so we can use it for the log context.

engine/message/db.go
```diff
@@ -569,11 +575,37 @@ func (db *DB) _SendMessages(ctx context.Context, send SendFunc, status StatusFun
...
+	}
+	cms.Close()
+
+	for _, m := range msgs {
+		meta := alertlog.NotificationMetaData{
+			MessageID: m.MessageID,
+		}
+
+		// log failures
+		db.alertlogstore.MustLogTx(ctx, tx, m.AlertID, alertlog.TypeNotificationSent, meta)
```
mastercactapus, line 606: Add the user id to the context here -> `MustLogTx(permission.UserContext(ctx, userID, permission.RoleUser), tx, ...)`

engine/message/db.go
mastercactapus: ^ the defer is so that we always close the rows, even if we `return err` before we finish reading.

PR #690, opened 2020-05-20
alerts list: always show alerts the current user is notified for

Forfold:

**Description:**
This PR enables functionality such that any user visiting the alerts list homepage will see any alerts they have been notified for, on top of their current filters (show all services, favorite services, ack only, etc).

Navigating to an alerts list from a specific service will remain the same (aside from an added header note), showing all alerts specific to that service (regardless of being favorited/notified).

**Screenshots:**

Legend
- NOTIFIED summaries: alerts from a service that **has not** been favorited, but are alerts that the user has been notified of
- FAVE summaries: alerts from a service that **has** been favorited

Home page
All services

---

Set
Unset

---

- Removes snackbar for favorited services (“Looks like you have no favorited services…“)
- Adds description of what alerts are being shown above alerts list (`headerNote` in `PaginatedList`)
  - "Showing [alert type] alerts you have been notified of and from any services you have favorited."
  - "Showing [alert type] alerts for all services."
  - "Showing [alert type] alerts for the service [service name]"
- Adds tooltip to favorite button on services page
  - "Favorite this service to always see its alerts on your homepage"
  - "Unfavorite this service to stop seeing its alerts on your homepage"

- Adds `includeNotified` variable to the `alerts` query schema in GraphQL, which defaults to `true` if left null

### DIFF
--- a/alert/search.go
+++ b/alert/search.go
@@ -25,6 +25,9 @@ type SearchOptions struct {
 	// Services, if specified, will restrict alerts to those with a matching ServiceID.
 	Services []string `json:"v,omitempty"`
 
+	// AllServices reflects if all services are to return their alerts
+	AllServices bool `json:"l,omitempty"`
+
 	After SearchCursor `json:"a,omitempty"`
 
 	// Omit specifies a list of alert IDs to exclude from the results.
@@ -45,23 +48,7 @@ type SearchCursor struct {
 }
 
 var searchTemplate = template.Must(template.New("search").Parse(`
-	{{ if .NotifiedUserID }}
-		SELECT
-			a.id,
-				a.summary,
-				a.details,
-				a.service_id,
-				a.source,
-				a.status,
-				created_at,
-				a.dedup_key
-		FROM alerts a
-		JOIN alert_logs al ON al.alert_id = a.id AND al.event = 'notification_sent'
-		WHERE al.sub_user_id = :currentUserID
-		UNION
-	{{ end }}
-
-	SELECT
+	SELECT DISTINCT
 		a.id,
 		a.summary,
 		a.details,
@@ -73,6 +60,9 @@ var searchTemplate = template.Must(template.New("search").Parse(`
 	FROM alerts a
 	{{ if .Search }}
 		JOIN services svc ON svc.id = a.service_id
+	{{ end }}
+	{{ if and .NotifiedUserID (not .AllServices) }}
+		JOIN alert_logs al ON al.alert_id = a.id
 	{{ end }}
 	WHERE true
 	{{ if .Omit }}
@@ -88,8 +78,12 @@ var searchTemplate = template.Must(template.New("search").Parse(`
 	{{ if .Status }}
 		AND a.status = any(:status::enum_alert_status[])
 	{{ end }}
-	{{ if .Services }}
+	{{ if and .Services (not .NotifiedUserID)}}
 		AND a.service_id = any(:services)
+	{{ else if and .NotifiedUserID (not .AllServices) (not .Services)}}
+		AND (al.sub_user_id = :currentUserID AND al.event = 'notification_sent')
+	{{ else if and .Services .NotifiedUserID}}
+		AND (a.service_id = any(:services) OR (al.sub_user_id = :currentUserID AND al.event = 'notification_sent'))
 	{{ end }}
 	{{ if .After.ID }}
 		AND (
--- a/alert/search.go
+++ b/alert/search.go
@@ -40,8 +40,8 @@ type SearchOptions struct {
 }
 
 type IDFilter struct {
-	Valid bool
-	IDs   []string
+	Valid bool     `json:"v,omitempty"`
+	IDs   []string `json:"i,omitempty"`
 }
 
 type SearchCursor struct {
--- a/alert/search.go
+++ b/alert/search.go
@@ -48,7 +48,7 @@ type SearchCursor struct {
 }
 
 var searchTemplate = template.Must(template.New("search").Parse(`
-	SELECT DISTINCT
+	SELECT
 		a.id,
 		a.summary,
 		a.details,
@@ -60,9 +60,6 @@ var searchTemplate = template.Must(template.New("search").Parse(`
 	FROM alerts a
 	{{ if .Search }}
 		JOIN services svc ON svc.id = a.service_id
-	{{ end }}
-	{{ if and .NotifiedUserID (not .AllServices) }}
-		JOIN alert_logs al ON al.alert_id = a.id
 	{{ end }}
 	WHERE true
 	{{ if .Omit }}
@@ -78,12 +75,12 @@ var searchTemplate = template.Must(template.New("search").Parse(`
 	{{ if .Status }}
 		AND a.status = any(:status::enum_alert_status[])
 	{{ end }}
-	{{ if and .Services (not .NotifiedUserID)}}
-		AND a.service_id = any(:services)
-	{{ else if and .NotifiedUserID (not .AllServices) (not .Services)}}
-		AND (al.sub_user_id = :currentUserID AND al.event = 'notification_sent')
-	{{ else if and .Services .NotifiedUserID}}
-		AND (a.service_id = any(:services) OR (al.sub_user_id = :currentUserID AND al.event = 'notification_sent'))
+	{{ if .Services }}
+		AND (a.service_id = any(:services)
+			{{ if .NotifiedUserID }}
+				OR a.id = any(select alert_id from alert_logs where event = 'notification_sent' and sub_user_id = :currentUserID)
+			{{ end }}
+		)
 	{{ end }}
 	{{ if .After.ID }}
 		AND (
--- a/alert/search.go
+++ b/alert/search.go
@@ -75,7 +75,7 @@ var searchTemplate = template.Must(template.New("search").Parse(`
 	{{ if .Services }}
 		AND (a.service_id = any(:services)
 			{{ if .NotifiedUserID }}
-				OR a.id = any(select alert_id from alert_logs where event = 'notification_sent' and sub_user_id = :notifiedUserID)
+				OR a.id = any(select alert_id from alert_logs where event in ('notification_sent', 'no_notification_sent') and sub_user_id = :notifiedUserID)
 			{{ end }}
 		)
 	{{ end }}
--- a/alert/search.go
+++ b/alert/search.go
@@ -34,6 +34,10 @@ type SearchOptions struct {
 	// will also be added to the results of the query
 	IncludeNotifiedUser string `json:"e,includenotifieduser"`
 
+	// FavoriteServicesOnly is used as a conditional for this template. Merging of
+	// alerts from favorited services is handled separately.
+	FavoriteServicesOnly bool
+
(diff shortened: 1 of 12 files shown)


alert/search.go
```diff
@@ -41,6 +45,22 @@ type SearchCursor struct {
...
+				a.summary,
+				a.details,
+				a.service_id,
+				a.source,
+				a.status,
+				created_at,
+				a.dedup_key
+		FROM alerts a
+		JOIN alert_logs al ON al.alert_id = a.id AND al.event = 'notification_sent'
+		WHERE al.sub_user_id = :currentUserID
```
mastercactapus, line 60: This should be mixed into the where clause -- It needs to respect Omit, Search, Status, pagination etc.. The only filter it should override is the service id restriction

```
	{{ if .Services }}
		AND (a.service_id = any(:services) OR sub_user_id)
	{{ end }}
```

alert/search.go
```diff
@@ -30,6 +30,10 @@ type SearchOptions struct {
 	// Omit specifies a list of alert IDs to exclude from the results.
 	Omit []int `json:"o,omitempty"`
 
+	// IncludeNotifiedUser will ensure all alerts the specified user has been notified for
+	// will also be added to the results of the query
+	IncludeNotifiedUser string `json:"e,omitempty"`
```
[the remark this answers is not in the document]
Forfold: I was thinking along the same line, but couldn't think of a good name, so I'll go with this 👍

alert/search.go
Forfold: I originally had this in the where clause but was running into some issues with aggregating the notified alerts on top of the logic currently in place. I'll keep brainstorming

Forfold: Just updated without using the ternary, let me know your thoughts once you get a chance @mastercactapus!

alert/search.go
```diff
@@ -25,11 +25,18 @@ type SearchOptions struct {
 	// Services, if specified, will restrict alerts to those with a matching ServiceID.
 	Services []string `json:"v,omitempty"`
 
+	// AllServices reflects if all services are to return their alerts
+	AllServices bool `json:"l,omitempty"`
+
```
mastercactapus, line 30: This is already accomplished/implied by omitting the `Services` filter.

graphql2/graphqlapp/alert.go
```diff
@@ -121,6 +121,9 @@ func (q *Query) Alerts(ctx context.Context, opts *graphql2.AlertSearchOptions) (
 		s.Search = *opts.Search
 	}
 	s.Omit = opts.Omit
+	if opts.IncludeNotified != nil && *opts.IncludeNotified == true {
```
mastercactapus, line 124:
```suggestion
	if opts.IncludeNotified != nil && *opts.IncludeNotified {
```

web/src/app/alerts/AlertsList.js
```diff
@@ -233,6 +214,25 @@ export default function AlertsList(props) {
...
+
+    if (includeNotified && favoritesOnly) {
+      return 'Showing alerts you have been notified of and from any services you have favorited.'
+    }
+
+    if (allServices) {
+      return 'Showing alerts for all services.'
+    }
+
+    return ''
```
mastercactapus, line 233: What case would this be? If it's not expected, maybe throw an error here, or at least log one.

web/src/app/util/SetFavoriteButton.tsx
```diff
@@ -2,26 +2,35 @@ import React from 'react'
...
 interface SetFavoriteButtonProps {
   typeName: 'rotation' | 'service' | 'schedule'
   isFavorite?: boolean
   loading: boolean
   onClick: Function
+
+  tooltipOptions?: {
+    enabled: boolean
+    setMessage?: string // The tooltip displayed when the favorite has already been set
+    unsetMessage?: string // The tooltip displayed when the favorite has not yet been set
```
mastercactapus, line 17: Can we hard-code these and use `typeName`? I feel like any/everywhere we have a `service` fav. button we probably want the same message rather than having the ability to change it.

Forfold: We can, but we made the favorite button generic and this could be useful for the other pages as well (rotations, schedules, etc)

web/src/app/alerts/AlertsList.js
Forfold: Currently just a default, which is used on the services alerts pages. If an unknown filter comes in (or one that we are not covering on purpose), we don't display anything. In the case of the services list, we only care about displaying the alerts for that service (no "all services" or "my alerts" to worry about distinguishing between), so we can leave the note blank

alert/search.go
Forfold: I needed to use this to distinguish between `AllServices` and `FavoritesOnly`, because both trickle down to `.Services` in the query but are handled differently. It's been about a week so I can't remember the exact behavior I was seeing though. I can try to recreate my issue, if you'd like

web/src/app/util/SetFavoriteButton.tsx
mastercactapus: It is generic but not type-agnostic. In the case of a rotation or schedule, we'd add a new type and associated messages. We're currently using `typeName` to generate the aria-labels rather than having them passed in which fills a similar role to the tooltip -- both sources of truth should live in the same place I think.

If we rendered a `service` favorite button in two places, we'd have to copy and paste both tooltip messages to both places to get them to match, which makes it seem like it belongs inside the component.

web/src/app/alerts/AlertsList.js
mastercactapus: On the service alert page, it wouldn't be a bad idea to display "Displaying alerts for Foo Bar." or something like that. When we did UX testing, it came up that it was hard to tell the difference between the two alert pages -- always having a filter string could help with that.

Also it might be good to include the status filter in that string as well, especially on mobile where the headers are not rendered. Again following the newer convention of always describing the current filter set. (e.g. schedule shifts).

alert/search.go
mastercactapus: We should work through that a bit more -- they should be equivalent at this level. Either we are filtering by service ID or we're not regardless of why. If we are, then the notified user ID is relevant, otherwise it can be ignored as the service ID filter is the only criteria overridden by the notification ID.

Forfold: Gotcha, I'll circle back to this when I have some time and get back to you

alert/search.go
```diff
@@ -69,7 +73,11 @@ var searchTemplate = template.Must(template.New("search").Parse(`
 		AND a.status = any(:status::enum_alert_status[])
 	{{ end }}
 	{{ if .Services }}
-		AND a.service_id = any(:services)
+		AND (a.service_id = any(:services)
+			{{ if .NotifiedUserID }}
+				OR a.id = any(select alert_id from alert_logs where event = 'notification_sent' and sub_user_id = :currentUserID)
```
mastercactapus, line 78: One additional thought for this, we can add `no_notification_sent` for users who were on-call but misconfigured.
```suggestion
				OR a.id = any(select alert_id from alert_logs where event in ('notification_sent', 'no_notification_sent') and sub_user_id = :currentUserID)
```

graphql2/graphqlapp/alert.go
```diff
@@ -143,14 +146,6 @@ func (q *Query) Alerts(ctx context.Context, opts *graphql2.AlertSearchOptions) (
...
 				return nil, err
 			}
-
-			// favorites only with no returned services will
-			// return an empty result set
-			if len(s.Services) == 0 {
-				return &graphql2.AlertConnection{
-					PageInfo: &graphql2.PageInfo{},
-				}, nil
-			}
```
mastercactapus, line 153: This was a workaround for the same reason you added `AllServices`. The consequence is if we have no favorites the response will contain alerts from all services.

Some options:
- Leave as is
- differentiate between `nil` and `len() == 0` (generally not a Go way)
- Pass a single impossible service ID (e.g. all zero UUID), probably easiest
- if `len() == 0` and `IncludeNotifiedUserID`, only include notified alerts (seems a little "magic", needs care to ensure graphql API behaves)
- add a `NotifiedUserOnly` bool to `SearchOptions` that will override the service filter. (will need to be added to cursor) maybe cleanest

I think most importantly, is the GraphQL API behavior:
- favoritesOnly = true, no favorites, and includeNotified = false should return nothing
- favoritesOnly = true, no favorites, and includeNotified = true should return only notified
- favoritesOnly = false, return everything

If we don't then you'll get all alerts, use the fav. button, and have things disappear from the alerts list which might be confusing.

alert/search.go
Forfold: True, I'll go ahead and update the alerts list text to cover this as well 👍

m17ch:
I am seeing the following backend errors (are you able to reproduce)?
`error="find alert log state: sql: Scan error on column index 3, name \"provider_msg_id\": converting NULL to string is unsupported"`

Forfold:
@m17ch I didn't see that originally, I'll do some investigating now though!

web/src/app/alerts/AlertsList.js
```diff
@@ -106,7 +107,7 @@ function getStatusFilter(s) {
 export default function AlertsList(props) {
   const classes = useStyles()
   const width = useWidth()
-  const isFullScreen = isWidthDown('md', width)
+  const isMobileScreenSize = isWidthDown('md', width)
```
m17ch, line 110: 💯 better var name (I **always** thought this was opposite)

web/src/app/services/ServiceDetails.js
```diff
@@ -4,6 +4,7 @@ import gql from 'graphql-tag'
 import { useQuery } from 'react-apollo'
 import { Redirect } from 'react-router-dom'
 import _ from 'lodash-es'
+import Tooltip from '@material-ui/core/Tooltip'
```
m17ch, line 7: is this import still needed (I don't see it referenced)?  If so, not sure why `es-lint` didn't catch this?

web/src/app/alerts/AlertsList.js
Forfold: Yeah, it's 100% always confusing to me, too! Thankfully with things changing to `width: md` in hooks it will eventually go away all together

web/src/app/util/SetFavoriteButton.tsx
```diff
@@ -42,4 +44,22 @@ export function SetFavoriteButton({
       </IconButton>
     </form>
   )
+
+  // todo: add other types as needed
```
m17ch, line 48: Did we create a story for this or will we just spot in future (by searching todos)?

Forfold: I didn't, I left it in as a todo when it was an API change, but now that we are using `typename` and can hardcode the strings as needed, I think we can delete this todo

m17ch:
> I am seeing the following backend errors (are you able to reproduce)?
> `error="find alert log state: sql: Scan error on column index 3, name \"provider_msg_id\": converting NULL to string is unsupported"`

This was likely due to some db contents from my work on another branch -- I've regen'ed the db and have been unable to reproduce (you can disregard unless I holler again).

Forfold:
@m17ch @mastercactapus needed to fix a few things that weren't caught in a pre-commit, ready for a final approval now

PR #726, opened 2020-06-12
hooks: converted AppBarSearchContainer.js to hooks

jakegleason:

**Description:**

converted AppBarSearchContainer.js to hooks

### DIFF
--- a/web/src/app/util/AppBarSearchContainer.js
+++ b/web/src/app/util/AppBarSearchContainer.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState, useEffect } from 'react'
 import p from 'prop-types'
 import { debounce } from 'lodash-es'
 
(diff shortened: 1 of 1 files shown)


web/src/app/util/AppBarSearchContainer.js
```diff
@@ -125,3 +116,11 @@ export default function AppBarSearchContainer(props) {
     </SearchContext.Consumer>
   )
 }
+
+SearchProvider.propTypes = {
+  actions: null,
+}
```
Forfold, line 122: I don't think this is needed, as `actions` is used in `state` as opposed to coming from a `prop`

web/src/app/util/AppBarSearchContainer.js
```diff
@@ -14,94 +14,85 @@ export function SearchContainer() {
...
-      >
-        {this.props.children}
-      </SearchContext.Provider>
-    )
-  }
+  return (
+    <SearchContext.Provider
+      value={{
+        actions: actions,
+        setActions: props.setActions,
```
Forfold, line 67:
```suggestion
        setActions: handleSetActions,
```

web/src/app/util/AppBarSearchContainer.js
jakegleason: Removed actions.

web/src/app/util/AppBarSearchContainer.js
jakegleason: Changed `props.setActions` to `handleSetActions`. Tested by dragging browser to a smaller width.

Forfold:
Closing per our conversation with @mastercactapus that we may need to restructure our render tree a bit to achieve the desired results using functional components in the way we want.

PR #730, opened 2020-06-15
hooks: Converted PageActions.js to hooks

jakegleason:

**Description:**
This PR converts `PageActions.js` to hooks.

### DIFF
--- a/web/src/app/util/PageActions.js
+++ b/web/src/app/util/PageActions.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState, useEffect } from 'react'
 import p from 'prop-types'
 import { debounce } from 'lodash-es'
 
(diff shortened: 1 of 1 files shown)


web/src/app/util/PageActions.js
```diff
@@ -8,104 +8,96 @@ const PageActionsContext = React.createContext({
...
+  // state = {
+  //   actions: null,
+  // }
+  const [actions, setActions] = useState(null)
 
-  _mountCount = 0
-  _mounted = false
+  let _mountCount = 0
+  let _mounted = false
+  let _pending = null
```
mastercactapus, line 27: I'm not sure that these will work the same in hooks -- each render will re-run this function, re-declaring them, rather than persisting. I'll try to spend some time this afternoon investigating.

web/src/app/util/PageActions.js
```diff
@@ -8,104 +8,96 @@ const PageActionsContext = React.createContext({
...
 }
 
-class PageActionUpdater extends React.PureComponent {
-  static propTypes = {
-    setActions: p.func.isRequired,
-  }
-
-  _mounted = false
+function PageActionUpdater(props) {
+  let _mounted = false
```
mastercactapus, line 82: Same as above, each render will potentially have this reset -- we'd only have it shared during the useEffect and its subsequent cleanup call, but future renders will lose the value.

Forfold:
Closing per our conversation with @mastercactapus that we may need to restructure our render tree a bit to achieve the desired results using functional components in the way we want.

PR #732, opened 2020-06-17
alert log: add smoketests

Forfold:

**Description:**
This PR adds tests for the following scenarios in the alert log:
- successful sent notification gets logged
- disabled contact method gets logged
- sms failure gets logged
- voice failure gets logged
- "no immediate rule" gets logged
- "no on-call users" gets logged
- "no escalation policy steps" gets logged

### DIFF
--- a/smoketest/alertlog_test.go
+++ b/smoketest/alertlog_test.go
@@ -43,7 +43,7 @@ func TestVoiceFailure(t *testing.T) {
 func TestNoImmediateNR(t *testing.T) {
 	t.Parallel()
 
-	var sql = makeSQL(true, false, true, true)
+	var sql = makeSQL(false, false, true, true)
 	h := harness.NewHarness(t, sql, "add-no-notification-alert-log")
 	defer h.Close()
 
@@ -57,12 +57,40 @@ func TestNoImmediateNR(t *testing.T) {
 	assert.Contains(t, msg, "no immediate rule")
 }
 
+// DONE !!!
 func TestNoOnCallUsers(t *testing.T) {
 	t.Parallel()
+
+	var sql = makeSQL(false, true, true, false)
+	h := harness.NewHarness(t, sql, "add-no-notification-alert-log")
+	defer h.Close()
+
+	// create alert
+	doQL(h, t, makeCreateAlertMut(h), nil)
+	h.Trigger()
+	logs := getLogs(h, t)
+
+	// most recent entry
+	var details = logs.Alert.RecentEvents.Nodes[0].State.Details
+	assert.Equal(t, "No one was on-call", details)
 }
 
+// DONE !!!
 func TestNoEPSteps(t *testing.T) {
 	t.Parallel()
+
+	var sql = makeSQL(false, true, false, false)
+	h := harness.NewHarness(t, sql, "add-no-notification-alert-log")
+	defer h.Close()
+
+	// create alert
+	doQL(h, t, makeCreateAlertMut(h), nil)
+	h.Trigger()
+	logs := getLogs(h, t)
+
+	// most recent entry
+	var details = logs.Alert.RecentEvents.Nodes[0].State.Details
+	assert.Equal(t, "No escalation policy steps", details)
 }
 
 func makeSQL(disabledCM bool, nr bool, epStep bool, epStepUser bool) string {
--- a/smoketest/alertlog_test.go
+++ b/smoketest/alertlog_test.go
@@ -9,29 +9,63 @@ import (
 	"github.com/target/goalert/smoketest/harness"
 )
 
-type alertLogTestCfg struct {
-	CMDisabled bool
-	CMType     string
-	NR         bool
-	EPStep     bool
-	EPStepUser bool
-}
+func TestAlertLog(t *testing.T) {
+	t.Parallel()
 
-type alertLogs struct {
-	Alert struct {
-		RecentEvents struct {
-			Nodes []struct {
-				Message string
-				State   struct {
-					Details string
+	type alertLogs struct {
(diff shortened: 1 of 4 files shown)


smoketest/alertlog_test.go
```diff
@@ -0,0 +1,252 @@
...
+	doQL(h, t, makeCreateAlertMut(h), nil)
+	h.Trigger()
+	logs := getLogs(h, t)
+
+	// most recent entry
+	var details = logs.Alert.RecentEvents.Nodes[0].State.Details
+	assert.Equal(t, "No escalation policy steps", details)
+}
+
+func makeSQL(disabledCM bool, cmType string, nr bool, epStep bool, epStepUser bool) string {
```
mastercactapus, line 138: Let's make this a struct so it's easier to read where it's used.

```go
type alertLogTestCfg struct {
  CMDisabled bool
  CMType string
  NR bool
  EPStep bool
  EPStepUser bool
}
```

Lastly, what do you think of allowing a data parameter for the initSQL?

```go
const alertLogSQLTmpl = `
		insert into users (id, name, email) 
		values ({{uuid "user"}}, 'bob', 'joe');
		insert into user_contact_methods (id, user_id, name, type, value, disabled) 
		values ({{uuid "cm1"}}, {{uuid "user"}}, 'personal', '{{.CMType}}', {{phone "1"}}, {{.CMDisabled}});

		{{- if .NR}}
		insert into user_notification_rules (user_id, contact_method_id, delay_minutes) 
		values ({{uuid "user"}}, {{uuid "cm1"}}, 0);
		{{- end}}
`
h := harness.NewHarnessWithData(t, alertLogSQLTmpl, "add-no-notification-alert-log", alertLogTestCfg{
  CMType: "SMS",
  NR: true,
  EPStep: true,
  EPStepUser: true,
})
```


I think it would be a lot cleaner to have the if logic as part of the template. The `Harness.execQuery` method can just add a `data interface{}` argument and pass that to `Execute` instead of the hard-coded `nil` value and it should be good.

smoketest/alertlog_test.go
```diff
@@ -0,0 +1,252 @@
...
+	Alert struct {
+		RecentEvents struct {
+			Nodes []struct {
+				Message string `json:"message"`
+				State   struct {
+					Details string `json:"details"`
+				} `json:"state"`
+			} `json:"nodes"`
+		} `json:"recentEvents"`
+	} `json:"alert"`
```
mastercactapus, line 197: Tags are redundant and can be removed since we're only decoding -- it's already case-insensitive.

smoketest/alertlog_test.go
```diff
@@ -0,0 +1,252 @@
...
+	// add service
+	setupSQL = setupSQL + `
+		insert into services (id, escalation_policy_id, name) 
+		values ({{uuid "sid"}}, {{uuid "eid"}}, 'service');
+	`
+
+	return setupSQL
+}
+
+type logs struct {
```
mastercactapus, line 187: Name this `alertLogs` or `alertLogResp` since it shares a namespace with all other smoketests.

smoketest/alertlog_test.go
```diff
@@ -0,0 +1,252 @@
...
+	// create alert
+	doQL(h, t, makeCreateAlertMut(h), nil)
+	h.Trigger()
+	logs := getLogs(h, t)
+
+	// most recent entry
+	var msg = logs.Alert.RecentEvents.Nodes[0].Message
+	assert.Contains(t, msg, "Notification sent")
+	h.Twilio(t).Device(h.Phone("1")).ExpectSMS("Alert #1: foo")
+}
```
mastercactapus, line 28: On second thought, we can probably make one high-level test, and define our types and methods inside it (so we don't need to worry about namespace pollution)

```go
func TestAlertLog(t *testing.T) {
  t.Parallel()
  const sql = `...`
  type cfg struct {
    CMDisabled bool
    // ...
  }
  type logs struct {
    Alert struct {
      // ...
    }
  }

  check := func(desc string, c cfg, assert func(*testing.T, logs)) {
    t.Run(desc, func(t *testing.T) {
      t.Parallel()
      h := harness.NewHarnessWithData(t, sql, c, "add-no-notification-alert-log")
      defer h.Close()

      doQL(h, t, makeCreateAlertMut(h), nil)
      h.Trigger()
      var l logs
      doQL(h, t, logsQuery, &l)
      assert(t, l)
    })
  }


  check("DisabledContactMethod", cfg{CMDisabled: true, CMType: "SMS", ...}, func(t *testing.T, l logs) {
    assert.Equal(t, "contact method disabled", l.Alert.RecentEvents.Nodes[0].State.Details)
  })
}
```

smoketest/alertlog_test.go
```diff
@@ -0,0 +1,227 @@
...
+
+	type config struct {
+		CMDisabled bool
+		CMType     string
+		NR         bool
+		EPStep     bool
+		EPStepUser bool
+	}
+
+	doQL := func(h *harness.Harness, query string, res interface{}) {
```
mastercactapus, line 36: This will need `t *Testing.T` as an argument so that it uses the `t` from `t.Run` and not the top-level test instance.

```suggestion
	doQL := func(t *testing.T, h *harness.Harness, query string, res interface{}) {
```

smoketest/alertlog_test.go
```diff
@@ -0,0 +1,227 @@
...
+		g := h.GraphQLQuery2(query)
```
mastercactapus, line 37: Mark this as a helper so any errors get reported from the place that called the query

```suggestion
		t.Helper()
		g := h.GraphQLQuery2(query)
```

PR #741, opened 2020-06-24
ui: Converted ServiceForm to Typescript

jakegleason:

**Description:**

- Delete goalert/web/src/app/services/ServiceForm.js
- Convert ServiceForm to ts

### DIFF
--- a/web/src/app/services/ServiceForm.tsx
+++ b/web/src/app/services/ServiceForm.tsx
@@ -59,24 +59,3 @@ export default function ServiceForm(props: ServiceFormProps): JSX.Element {
     </FormContainer>
   )
 }
-
-// ServiceForm.propTypes = {
-//   value: p.shape({
-//     name: p.string,
-//     description: p.string,
-//     escalationPolicyID: p.string,
-//   }).isRequired,
-
-// indicates that the escalation policy is a required field
-//   epRequired: p.bool,
-
-//   errors: p.arrayOf(
-//     p.shape({
-//       field: p.oneOf(['name', 'description', 'escalationPolicyID']).isRequired,
-//       message: p.string.isRequired,
-//     }),
-//   ),
-
-//   onChange: p.func,
-
-//   disabled: p.bool,
--- a/web/src/app/services/ServiceForm.tsx
+++ b/web/src/app/services/ServiceForm.tsx
@@ -1,13 +1,29 @@
 import React from 'react'
-import p from 'prop-types'
 import Grid from '@material-ui/core/Grid'
 import TextField from '@material-ui/core/TextField'
 import { EscalationPolicySelect } from '../selection/EscalationPolicySelect'
 import { FormContainer, FormField } from '../forms'
 
-export default function ServiceForm(props) {
-  const { epRequired, ...containerProps } = props
+interface ServiceFormProps {
+  value: {
+    name: string
+    description: string
+    escalationPolicyID: string
 
+    errors: {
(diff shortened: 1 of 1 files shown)


web/src/app/services/ServiceForm.tsx
```diff
@@ -1,13 +1,29 @@
...
+interface ServiceFormProps {
+  value: {
+    name: string
+    description: string
+    escalationPolicyID: string
+
+    errors: {
+      field: ['name', 'description', 'escalationPolicyID']
+      message: string
+    }
```
mastercactapus, line 16: Errors should be an array

web/src/app/services/ServiceForm.tsx
```diff
@@ -1,13 +1,29 @@
...
+
+    onChange: Function
 
+    disabled: boolean
+  }
```
mastercactapus, line 21: `value` should only contain `name` `description` and `escalationPolicyID`

web/src/app/services/ServiceForm.tsx
```diff
@@ -1,13 +1,29 @@
...
+    name: string
+    description: string
+    escalationPolicyID: string
+
+    errors: {
+      field: ['name', 'description', 'escalationPolicyID']
+      message: string
+    }
+
+    onChange: Function
```
mastercactapus, line 18: Define the signature rather than just the Function type

web/src/app/services/ServiceForm.tsx
```diff
@@ -1,13 +1,29 @@
...
 
+    disabled: boolean
```
Forfold, line 20: These will need to be out of the value object as only `name`, `description`, and `escalationPolicyID` are nested under `value` in the original `ServiceForm.js`

web/src/app/services/ServiceForm.tsx
jakegleason: Changed errors to an array. Also, changed `field` to a union between the values:
```
errors: {
    field: 'name' | 'description' | 'escalationPolicyID'
    message: string
  }[]
```

web/src/app/services/ServiceForm.tsx
jakegleason: `value` now only contains `name` `description` and `escalationPolicyID`

web/src/app/services/ServiceForm.tsx
jakegleason: Defined signature instead of type: 
`onChange: (val: Value) => void`

web/src/app/services/ServiceForm.tsx
jakegleason: `value` now only contains `name` `description` and `escalationPolicyID`

web/src/app/services/ServiceForm.tsx
```diff
@@ -1,13 +1,32 @@
...
 import TextField from '@material-ui/core/TextField'
 import { EscalationPolicySelect } from '../selection/EscalationPolicySelect'
 import { FormContainer, FormField } from '../forms'
 
-export default function ServiceForm(props) {
-  const { epRequired, ...containerProps } = props
+interface Value {
+  name: string
+  description: string
+  escalationPolicyID: string
```
Forfold, line 10: while `Value` should be required, these three keys inside of `Value` should be optional

jakegleason: Made requested changes and re-commited.

PR #742, opened 2020-06-24
ui: Changed HeartbeatMonitorForm to Typescript

jakegleason:

**Description:**

- Delete goalert/web/src/app/services/HeartbeatMonitorForm.js
- Convert HeartbeatMonitorForm to ts


web/src/app/services/HeartbeatMonitorForm.tsx
```diff
@@ -1,19 +1,32 @@
...
+  value: {
+    name: string
+    timeoutMinutes: [number, string]
+  }
 
-export default function HeartbeatMonitorForm(props) {
+  errors: {
+    field: ['name', 'timeoutMinutes']
+    message: string
+  }
```
mastercactapus, line 23: `errors` is an array

web/src/app/services/HeartbeatMonitorForm.tsx
```diff
@@ -1,19 +1,32 @@
...
+
+  onChange: Function
```
mastercactapus, line 25: Define signature rather than `Function` type

web/src/app/services/HeartbeatMonitorForm.tsx
jakegleason: Changed `errors` to an array:
```
errors: {
    field: 'name' | 'timeoutMinutes'
    message: string
  }[]
```

web/src/app/services/HeartbeatMonitorForm.tsx
jakegleason: Changed to `onChange: (val: Value) => void`

PR #755, opened 2020-06-30
admin: add toolbox for SMS testing and number lookup

mastercactapus:

**Description:**
This PR adds a `Toolbox` in the Admin section that can be used to lookup number & carrier information, as well as send SMS messages for testing (e.g. to test for carrier filtering).

Additionally, gqlgen was updated to the latest version to resolve a schema issue. 
https://gqlgen.com/recipes/migration-0.11/

image


graphql2/graphqlapp/mutation.go
```diff
@@ -29,6 +32,39 @@ func (a *Mutation) SetFavorite(ctx context.Context, input graphql2.SetFavoriteIn
...
+func (a *Mutation) DebugSendSms(ctx context.Context, input graphql2.DebugSendSMSInput) (*graphql2.DebugSendSMSInfo, error) {
+	err := permission.LimitCheckAny(ctx, permission.Admin)
+	if err != nil {
+		return nil, err
+	}
+
+	err = validate.Many(
+		validate.Phone("To", input.To),
+		validate.Phone("From", input.From),
+		validate.Text("Body", input.Body, 1, 1000),
```
Forfold, line 49: Should we limit this to 255?

graphql2/graphqlapp/mutation.go
```diff
@@ -29,6 +32,39 @@ func (a *Mutation) SetFavorite(ctx context.Context, input graphql2.SetFavoriteIn
 	}
 	return true, nil
 }
+
+type safeErr struct{ error }
+
+func (safeErr) ClientError() bool { return true }
+
+func (a *Mutation) DebugSendSms(ctx context.Context, input graphql2.DebugSendSMSInput) (*graphql2.DebugSendSMSInfo, error) {
```
Forfold, line 40: should this mutation be under `/graphqlapp/toolbox`?

graphql2/graphqlapp/toolbox.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var result struct {
+		Carrier struct {
+			CC   string `json:"mobile_country_code"`
+			NC   string `json:"mobile_network_code"`
+			Name string
+			Type string
```
Forfold, line 53: do we need json tags here?

web/src/app/admin/AdminNumberLookup.tsx
```diff
@@ -0,0 +1,187 @@
...
+  const [lookup, { data: carrData, loading: carrLoading }] = useMutation(
+    carrierInfoMut,
+    {
+      variables: { number: '+' + number },
+      onError: (err) => setLastError(err),
+    },
+  )
+
+  function renderListItem(label: string, _text: string): JSX.Element {
+    const text = (_text === undefined ? '' : _text).toString()
```
Forfold, line 71: is there a reason we are explicitly checking for undefined? might be best just to check if falsey, i.e. `_text ? _text : ''`

and also we shouldn't need to have a `.toString()` since we have type safety on `_text` coming in.

web/src/app/util/CopyText.tsx
```diff
@@ -22,19 +22,37 @@ interface CopyTextProps {
...
-    <Tooltip
-      onClose={() => setShowTooltip(false)}
-      open={showTooltip}
-      title='Copied!'
-      placement={props.placement || 'right'}
-    >
+  let content
+  if (props.textOnly) {
+    content = (
+      <span
```
Forfold, line 35: `Typography`? It should be able to handle root level html tags for `span`

web/src/app/admin/AdminNumberLookup.tsx
Forfold: I'm wondering if we need this check at all as since the parameter is not optional, we should always get a string to use. 🤔

graphql2/graphqlapp/mutation.go
mastercactapus: Since it's an admin-only tool and not being displayed in a list or stored anywhere I left it a bit long so there's a bit more flexibility in what you want to test (e.g. multi-segment SMS). The 255 elsewhere was just for DB & title size reasons.

If we wanted to limit it I'd say we put the logic to limit it to a single segment, but if anything we could extend it in the future to show a calculated segment counter.

graphql2/graphqlapp/toolbox.go
mastercactapus: No, the values that come in are `type` and `name` which will already parse correctly from the response, the others only needed it because of the `snake_case` is not handled by `UnmarshalJSON` (but upper/lower case is).

web/src/app/admin/AdminNumberLookup.tsx
mastercactapus: The issue is the values come from the GraphQL response which is the `any` type -- they can be undefined but the type checking will miss it.

But now that you mention it -- I'll update it to use the schema type and fix the signature accordingly.

Forfold: Sounds good

web/src/app/util/CopyText.tsx
mastercactapus: It already is rendered in/wrapped in Typography through the `ListItemText` component.

As far as here in `CopyText` -- `<span>` was used so it just inherits the styles and can have event handlers attached -- trying to follow the same behavior as `<AppLink>`, which uses a base `<a>` tag when `textOnly` is unset. That way it relies on it's container for any styling and with or without a link can be put into whatever container is needed.

PR #759, opened 2020-06-30
ui: Converted IntegrationKeyDeleteDialog to Typescript

jakegleason:

**Description:**
Delete goalert/web/src/app/services/IntegrationKeyDeleteDialog.js
Convert IntegrationKeyDeleteDialog to ts

### DIFF
--- a/web/src/app/services/IntegrationKeyDeleteDialog.tsx
+++ b/web/src/app/services/IntegrationKeyDeleteDialog.tsx
@@ -1,5 +1,4 @@
 import React from 'react'
-import p from 'prop-types'
 
 import gql from 'graphql-tag'
 import { useQuery, useMutation } from 'react-apollo'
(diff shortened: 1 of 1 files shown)


web/src/app/services/IntegrationKeyDeleteDialog.tsx
```diff
@@ -45,7 +50,7 @@ export default function IntegrationKeyDeleteDialog(props) {
   const [deleteKey, deleteKeyStatus] = useMutation(mutation, {
     onCompleted: props.onClose,
     update: (cache) => {
-      const { service } = cache.readQuery({
+      const { ...service } = cache.readQuery({
```
Forfold, line 53:
```suggestion
      const { service } = cache.readQuery({
```

We'll want to continue explicitly calling for the service prop, as destructuring it here under the name `service` will grab everything from the result of `readQuery`, which we don't want (we only want the `service` object from the cache.

With the destructuring the following warning/errors show in the console after deleting:
image

PR #771, opened 2020-07-07
Active text fix

alejandra-martinez:

**Description:**
Updated secondary text in alert log to show the last status and details of an outgoing alert message
Updated schema to access last status of outgoing message


mastercactapus:
Instead of extending the schema/API, let's handle this in the backend entirely. We can return the `details` field with the last status already mixed in.

PR #778, opened 2020-07-08
ui: Changed IntegrationKeyForm to Ts 

jakegleason:

**Description:**

- Delete goalert/web/src/app/services/IntegrationKeyForm.js
- Convert IntegrationKeyForm to ts


web/src/app/services/IntegrationKeyForm.tsx
```diff
@@ -13,11 +12,29 @@ import { AppLink } from '../util/AppLink'
...
   infoIcon: {
-    color: theme.palette.primary['500'],
+    color: theme.palette.primary.main,
   },
 }))
 
-export default function IntegrationKeyForm(props) {
+interface Value {
+  name: string
+  type: ['generic' | 'grafana' | 'site24x7' | 'email']
```
Forfold, line 21: Since these are optional in `propTypes`, we should make them optional here via `name?:` and `type?:`

mastercactapus: I think they were meant to be required but got missed.

For `type` I think it needs to be without the square brackets since it shouldn't be an array

web/src/app/services/IntegrationKeyForm.tsx
```diff
@@ -13,11 +12,29 @@ import { AppLink } from '../util/AppLink'
...
+interface Value {
+  name: string
+  type: 'generic' | 'grafana' | 'site24x7' | 'email'
+}
+
+interface IntegrationKeyFormProps {
+  value: Value
+
+  errors: {
+    field: ['name' | 'type']
```
mastercactapus, line 28: Shouldn't be an array:

```suggestion
    field: 'name' | 'type'
```

jakegleason: Made requested changes.

PR #783, opened 2020-07-09
ui: override enhancements

Quynhd07:

**Description:**
This PR removes the use of the word "Override" from the UI and replaces the buttons to "temp. add", "replace", and "remove". It also makes the user being removed/replaced from the calendar read-only.


web/src/app/schedules/ScheduleOverrideCreateDialog.js
```diff
@@ -82,6 +83,7 @@ ScheduleOverrideCreateDialog.propTypes = {
   scheduleID: p.string.isRequired,
   variant: p.oneOf(['add', 'remove', 'replace']).isRequired,
   onClose: p.func,
+  disableRemoveField: p.bool,
```
mastercactapus, line 86: I'm thinking this should be `removeUserReadOnly` based on function (even if we end up passing disabled to MUI)

That also follows the pattern in `CalendarSubscribeForm` which uses `scheduleReadOnly` to achieve the same purpose.

Forfold:
Adding a title to the buttons makes sense, and also helps out with our a11y friends as well

web/src/app/schedules/CalendarEventWrapper.js
```diff
@@ -106,6 +106,7 @@ export default class CalendarEventWrapper extends Component {
               onClick={() => this.handleShowOverrideForm('replace')}
               variant='contained'
               color='primary'
+              title='Temporarily replace a user from this schedule'
```
mastercactapus, line 109: Since this will be locked for one specific user, can we put their name here instead of "a user"? I believe `event.title` will be the user's name.

web/src/cypress/integration/alerts.ts
```diff
@@ -182,7 +182,10 @@ function testAlerts(screen: ScreenFormat): void {
         'UNACKNOWLEDGED',
       )
 
-      cy.get(`[data-cy=select-all] input`).check()
+      cy.reload()
+      cy.get(`span[data-cy=item-${alert1.id}] input`).check()
+      cy.get(`span[data-cy=item-${alert2.id}] input`).check()
+      cy.get(`span[data-cy=item-${alert3.id}] input`).check()
```
mastercactapus, line 188: This test should have been unaffected by this PR/changes -- is this necessary?

web/src/package.json
```diff
@@ -66,7 +66,7 @@
     "@apollo/react-hooks": "3.1.5",
     "@babel/runtime": "7.9.6",
     "@date-io/luxon": "1.3.13",
-    "@material-ui/core": "4.9.8",
+    "@material-ui/core": "4.10.2",
```
mastercactapus, line 69: It looks like there is a dependabot PR to update to 4.11 already -- if this PR needs this update, let's get the MUI PR in first so this one doesn't need to do any of it's own package updates.

If it doesn't then let's update `package.json` and `yarn.lock` to use the versions from master rather than introducing the update on this PR.

That way it will be easier to identify if any test regressions are from the UI update or the changes specific to this PR.

Forfold: Fixed with PR #786

PR #784, opened 2020-07-09
alert log: last status and details 

alejandra-martinez:

**Description:**
message details gives full context on what is going on with the outgoing message

### DIFF
--- a/graphql2/graphqlapp/alert.go
+++ b/graphql2/graphqlapp/alert.go
@@ -74,9 +74,54 @@ func (a *AlertLogEntry) notificationSentState(ctx context.Context, obj *alertlog
 	case notification.MessageStateSent, notification.MessageStateDelivered:
 		status = "OK"
 	}
+	var details string
+	switch s.LastStatus {
+	case notification.MessageLastStatusDelivered:
+		if s.Details == "delivered" {
+			details = s.Details
+		} else {
+			details = ":delivered" + s.Details
+		}
+	case notification.MessageLastStatusPending:
+		if s.Details == "pending" {
+			details = s.Details
+		} else {
+			details = ":pending" + s.Details
+		}
+	case notification.MessageLastStatusSending:
+		if s.Details == "sending" {
+			details = s.Details
+		} else {
+			details = "sending:" + s.Details
+		}
+	case notification.MessageLastStatusSent:
+		if s.Details == "sent" {
+			details = s.Details
+		} else {
+			details = "sent:" + s.Details
+		}
+	case notification.MessageLastStatusQueuedRemotely:
+		if s.Details == "queued_remotely" {
+			details = s.Details
+		} else {
+			details = "queued_remotely:" + s.Details
+		}
+	case notification.MessageLastStatusFailed:
+		if s.Details == "failed" {
+			details = s.Details
+		} else {
+			details = "failed:" + s.Details
+		}
+	case notification.MessageLastStatusBundled:
+		if s.Details == "bundled" {
+			details = s.Details
+		} else {
+			details = "bundled:" + s.Details
+		}
+	}
 
 	return &graphql2.AlertLogEntryState{
-		Details: s.Details,
+		Details: details,
 		Status:  &status,
 	}, nil
 }
--- a/graphql2/graphqlapp/alert.go
+++ b/graphql2/graphqlapp/alert.go
@@ -74,48 +74,34 @@ func (a *AlertLogEntry) notificationSentState(ctx context.Context, obj *alertlog
 	case notification.MessageStateSent, notification.MessageStateDelivered:
 		status = "OK"
 	}
-	var details string
+	var details = s.Details
 	switch s.LastStatus {
 	case notification.MessageLastStatusDelivered:
-		if s.Details == "delivered" {
-			details = s.Details
-		} else {
-			details = ":delivered" + s.Details
+		if s.Details != "delivered" {
+			details = "delivered:" + s.Details
 		}
 	case notification.MessageLastStatusPending:
-		if s.Details == "pending" {
-			details = s.Details
-		} else {
-			details = ":pending" + s.Details
+		if s.Details != "pending" {
+			details = "pending:" + s.Details
 		}
 	case notification.MessageLastStatusSending:
-		if s.Details == "sending" {
-			details = s.Details
-		} else {
+		if s.Details != "sending" {
 			details = "sending:" + s.Details
 		}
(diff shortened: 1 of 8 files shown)


notification/messagestatus.go
```diff
@@ -19,6 +19,9 @@ type MessageStatus struct {
 	// Details can contain any additional information about the State (e.g. "ringing", "no-answer" etc..).
 	Details string
 
+	//LastStatus labels information given in details for quick overall (failed, delivered, etc..).
+	LastStatus MessageLastStatus
+
```
mastercactapus, line 24: This is already represented by `State` (if not as granular).

Rather than adding a new field, let's update `MessageState`

We can add `MessageStatePending` and change `MessageStateActive` to `MessageStateSending` to make it a bit more clear. Pending probably shouldn't result in `Active` as it does today anyhow. Don't forget to update the switch statement

The DB/Store isn't the only place generating `MessageStatus` so we'd have to update all sites to be consistent, however since it's only used for display purposes we can just make `State` do what we want and then add the text in the GraphQL layer.

graphql2/graphqlapp/alert.go
```diff
@@ -74,9 +74,40 @@ func (a *AlertLogEntry) notificationSentState(ctx context.Context, obj *alertlog
...
+		}
+	case notification.MessageLastStatusFailed:
+		if s.Details != "failed" {
+			details = "failed:" + s.Details
+		}
+	case notification.MessageLastStatusBundled:
+		if s.Details != "bundled" {
+			details = "bundled:" + s.Details
+		}
+	}
```
mastercactapus, line 107: We can simplify this a bit (and using `State` as in another comment) so that it reads better.


If we structure the code so that the shared/common logic is only written once, it makes the switch easier to scan for just the string/prefix differences.
```
	var prefix string
	switch s.State {
	case notification.MessageStatePending:
		prefix = "pending"
	case notification.MessageStateSending:
		prefix = "sending"
	case notification.MessageStateSent:
		prefix = "sent"
	// ...
	}

	details := s.Details
	if prefix != details {
		details = prefix + ": " + details
	}
```

Also for reference, use `var` for empty declarations and `:=` for assigning declarations.
```
// either
var details string
// or
details := s.Details
```

Using `var foo = ` should only be done (by convention) for package-level variables where `:=` is invalid syntax.

graphql2/graphqlapp/alert.go
```diff
@@ -74,9 +74,40 @@ func (a *AlertLogEntry) notificationSentState(ctx context.Context, obj *alertlog
 	case notification.MessageStateSent, notification.MessageStateDelivered:
 		status = "OK"
 	}
+	var details = s.Details
+	switch s.LastStatus {
```
Forfold, line 78: would it work to make this more this dynamic? with something like:
```
var details = s.Details
if s.LastStatus !== s.Details {
  details = s.LastStatus + ": " + s.Details
}
```

graphql2/graphqlapp/alert.go
```diff
@@ -75,8 +76,29 @@ func (a *AlertLogEntry) notificationSentState(ctx context.Context, obj *alertlog
...
+		prefix = "Delivered"
+	case notification.MessageStateFailedTemp:
+		prefix = "Failed"
+	case notification.MessageStateFailedPerm:
+		prefix = "Failed"
+	}
+
+	details := s.Details
+	if !strings.EqualFold(prefix, details) {
+		details = prefix + " : " + details
```
mastercactapus, line 128: I think, remove the leading space 
```suggestion
		details = prefix + ": " + details
```

graphql2/graphqlapp/alert.go
```diff
@@ -75,8 +76,29 @@ func (a *AlertLogEntry) notificationSentState(ctx context.Context, obj *alertlog
...
+	case notification.MessageStateSending:
+		prefix = "Sending"
+	case notification.MessageStateSent:
+		prefix = "Sent"
+	case notification.MessageStateDelivered:
+		prefix = "Delivered"
+	case notification.MessageStateFailedTemp:
+		prefix = "Failed"
+	case notification.MessageStateFailedPerm:
+		prefix = "Failed"
```
mastercactapus, line 123: These can be combined in Go:
```suggestion
	case notification.MessageStateFailedTemp, notification.MessageStateFailedPerm:
		prefix = "Failed"
```

notification/messagestatus.go
```diff
@@ -42,9 +42,12 @@ func (stat *MessageStatus) wrap(ctx context.Context, n *namedSender) *MessageSta
...
 
 const (
-	// MessageStateActive should be specified when a message is still active.
-	// This includes things like remotely queued, ringing, or in-progress calls.
-	MessageStateActive MessageState = iota
+	// MessageStateSending should be specified when a message is sending but has not been sent.
+	// This includes things like ringing, or in-progress calls.
+	MessageStateSending MessageState = iota
+
+	//MessageStatePending includes things like remotely queued.
```
mastercactapus, line 52: This should define pending from GoAlert's perspective, also add a leading space for line comments.

```suggestion
	// MessageStatePending indicates a message waiting to be sent.
```

notification/messagestatus.go
```diff
@@ -42,9 +42,12 @@ func (stat *MessageStatus) wrap(ctx context.Context, n *namedSender) *MessageSta
 type MessageState int
 
 const (
-	// MessageStateActive should be specified when a message is still active.
-	// This includes things like remotely queued, ringing, or in-progress calls.
-	MessageStateActive MessageState = iota
+	// MessageStateSending should be specified when a message is sending but has not been sent.
+	// This includes things like ringing, or in-progress calls.
```
mastercactapus, line 49: As stated in another comment, this still includes remotely queued. Think of the remote queue as an implementation detail of the other service. From GoAlert's perspective, the message is on-its way (we've handed off control).
```suggestion
	// This includes things like remotely queued, ringing, or in-progress calls.
```

notification/twilio/message.go
```diff
@@ -94,8 +95,10 @@ func (msg *Message) messageStatus(id string) *notification.MessageStatus {
 		status.State = notification.MessageStateDelivered
 	case MessageStatusSent, MessageStatusUndelivered:
 		status.State = notification.MessageStateSent
+	case MessageStatusAccepted, MessageStatusQueued:
+		status.State = notification.MessageStatePending
```
mastercactapus, line 99: We can drop this case -- any state from Twilio that isn't final should drop to `Sending` below
```suggestion
```

PR #788, opened 2020-07-13
TelTextField: add new phone number input field component with live validation

KaiHuaLin:

**Description:**
This PR introduces new component, <TelTextField />, which is a TextField specific for phone number field. All phone number fields (add contact method, Admin Toolbox, and Admin config) have converted to <TelTextField />

### DIFF
--- a/web/src/app/admin/AdminFieldComponents.tsx
+++ b/web/src/app/admin/AdminFieldComponents.tsx
@@ -67,13 +67,6 @@ export function StringInput(props: InputProps): JSX.Element {
   if (type === 'tel') {
     return (
       <TelTextField
-        InputProps={{
-          startAdornment: (
-            <InputAdornment position='start' style={{ marginBottom: '0.1em' }}>
-              +
-            </InputAdornment>
-          ),
-        }}
         onChange={(e: React.FormEvent<HTMLInputElement>) =>
           onChange((e.target as HTMLInputElement).value)
         }
--- a/web/src/app/admin/AdminFieldComponents.tsx
+++ b/web/src/app/admin/AdminFieldComponents.tsx
@@ -9,7 +9,7 @@ import VisibilityOff from '@material-ui/icons/VisibilityOff'
 import TelTextField from '../util/TelTextField'
 
 interface InputProps {
-  configtype: string
+  type: string
   name: string
   value: string
   password?: boolean
@@ -24,7 +24,7 @@ export const StringListInput = (props: InputProps): JSX.Element => {
       {value.map((val, idx) => (
         <Grid key={idx} item xs={12}>
           <StringInput
-            configtype={props.configtype}
+            type={props.type}
             value={val}
             name={val ? props.name + '-' + idx : props.name + '-new-item'}
             onChange={(newVal) =>
@@ -47,7 +47,7 @@ export const StringListInput = (props: InputProps): JSX.Element => {
 
 export function StringInput(props: InputProps): JSX.Element {
   const [showPassword, setShowPassword] = useState(false)
-  const { onChange, password, ...rest } = props
+  const { onChange, password, type = 'text', ...rest } = props
 
   const renderPasswordAdornment = (): JSX.Element | null => {
     if (!props.password) return null
@@ -64,23 +64,20 @@ export function StringInput(props: InputProps): JSX.Element {
     )
   }
 
-  if (props.configtype === 'tel') {
+  function handleChange(number: string): void {
+    onChange('+' + number)
+  }
+
+  if (type === 'tel') {
     return (
-      <TelTextField
-        fullWidth
-        InputProps
-        onChange={(e: React.FormEvent<HTMLInputElement>) =>
-          onChange((e.target as HTMLInputElement).value)
-        }
-        {...rest}
-      />
+      <TelTextField fullWidth InputProps onChange={handleChange} {...rest} />
     )
   }
   return (
     <Input
       fullWidth
       autoComplete='new-password' // chrome keeps autofilling them, this stops it
-      type={password && !showPassword ? 'password' : 'text'}
+      type={password && !showPassword ? 'password' : type}
       onChange={(e) => onChange(e.target.value)}
       endAdornment={renderPasswordAdornment()}
       {...rest}
--- a/web/src/app/admin/AdminFieldComponents.tsx
+++ b/web/src/app/admin/AdminFieldComponents.tsx
@@ -65,14 +65,7 @@ export function StringInput(props: InputProps): JSX.Element {
   }
 
   if (type === 'tel') {
-    return (
-      <TelTextField
-        onChange={(e: React.FormEvent<HTMLInputElement>) =>
-          onChange((e.target as HTMLInputElement).value)
-        }
-        {...rest}
-      />
-    )
+    return <TelTextField onChange={(e) => onChange(e.target.value)} {...rest} />
   }
   return (
     <Input
--- a/web/src/app/admin/AdminFieldComponents.tsx
+++ b/web/src/app/admin/AdminFieldComponents.tsx
@@ -6,8 +6,10 @@ import InputAdornment from '@material-ui/core/InputAdornment'
 import Switch from '@material-ui/core/Switch'
 import Visibility from '@material-ui/icons/Visibility'
 import VisibilityOff from '@material-ui/icons/VisibilityOff'
+import TelTextField from '../util/TelTextField'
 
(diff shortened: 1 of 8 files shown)


web/src/app/util/TelTextField.js
```diff
@@ -0,0 +1,67 @@
...
+})
+
+export default function TelTextField({ InputProps, ...props }) {
+  const classes = useStyles()
+  const { value, error } = props
+  const phoneNumber = '+' + value
+
+  // check validation of the input phoneNumber through graphql
+  const { data } = useQuery(isValidNumber, {
+    pollInterval: 1000,
```
mastercactapus, line 36: We can leave this out, it should use the default rather than be hard-coded.

web/src/app/util/TelTextField.js
```diff
@@ -0,0 +1,67 @@
...
+    variables: { number: phoneNumber },
+  })
+
+  // fetch validation
+  const valid = _.get(data, 'phoneNumberInfo.valid', null)
+
+  // add live validation icon to the right of the textfield as an endAdornment
+  const iprops = Object.assign(
```
mastercactapus, line 44: Use the spread operator instead:

```
const iprops = {
  endAdornment: //..
  ...InputProps
}
```

web/src/app/util/TelTextField.js
```diff
@@ -0,0 +1,67 @@
...
+      endAdornment: (
+        <InputAdornment position='end'>
+          {valid ? (
+            // change to makeStyle
+            <Check className={classes.valid} />
+          ) : error ? (
+            <Close className={classes.invalid} />
+          ) : (
+            <div />
+          )}
```
mastercactapus, line 55: Use a let variable & `if`/`else if` block instead of double ternary, it makes for a flatter and simpler read/verify for a reviewer or the next person to work on it.

```js
let adorn
if (valid) {
  adorn = <Check className={classes.valid} />
} else if (error) {
  adorn = <Close className={classes.invalid} />
}

let iprops = InputProps
if (adorn) {
  iprops = {
    endAdornment: <InputAdornment position='end'>{adorn}</InputAdornment>,
    ...InputProps
  }
}
```

web/src/app/admin/AdminSMSSend.tsx
```diff
@@ -86,9 +92,14 @@ export default function AdminSMSSend(): JSX.Element {
...
-                <TextField
-                  onChange={(e) =>
-                    setToNumber(e.target.value.replace(/[^0-9]/g, ''))
+                <TelTextField
+                  onChange={(e: React.FormEvent<HTMLInputElement>) =>
+                    setToNumber(
+                      (e.target as HTMLInputElement).value.replace(
+                        /[^0-9]/g,
+                        '',
+                      ),
```
mastercactapus, line 101: Move this logic to the component itself so that someone using the component doesn't need to remember they need this for consistent behavior.

web/src/app/admin/AdminFieldComponents.tsx
```diff
@@ -6,8 +6,10 @@ import InputAdornment from '@material-ui/core/InputAdornment'
 import Switch from '@material-ui/core/Switch'
 import Visibility from '@material-ui/icons/Visibility'
 import VisibilityOff from '@material-ui/icons/VisibilityOff'
+import TelTextField from '../util/TelTextField'
 
 interface InputProps {
+  configtype: string
```
mastercactapus, line 12: Just name this `type` to match the HTML attribute, make it optional with a fixed value of `tel` (since that's the only supported one ATM).

web/src/app/util/TelTextField.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  return (
+    <div>
+      <TextField
+        {...props}
+        InputProps={iprops}
+        type={props.type || 'tel'}
+        onChange={handleChange}
+      />
```
mastercactapus, line 81: Value should start with a prefix, but we will need to remove it (the `+`) before passing it to the text field

web/src/app/admin/AdminFieldComponents.tsx
```diff
@@ -60,11 +63,29 @@ export function StringInput(props: InputProps): JSX.Element {
...
+  if (type === 'tel') {
+    return (
+      <TelTextField
+        InputProps={{
+          startAdornment: (
+            <InputAdornment position='start' style={{ marginBottom: '0.1em' }}>
+              +
+            </InputAdornment>
+          ),
+        }}
```
Forfold, line 76: I think it would be worth having this be built into `TelTextField`, rather than passed in with InputProps, since we will always be handling the `+` within it now.

Like how you added in the end adornment

web/src/app/util/TelTextField.js
```diff
@@ -0,0 +1,99 @@
...
+
+  // fetch validation
+  const valid = _.get(data, 'phoneNumberInfo.valid', null)
+
+  let adorn
+  if (valid) {
+    adorn = <Check className={classes.valid} />
+  } else if (props.error) {
+    adorn = <Close className={classes.invalid} />
+  }
```
mastercactapus, line 58: This query will only return an error if there is a network problem, but not on validation error.
```suggestion
  if (props.error) {
    // no adornment if we can't check validity
  } else if (valid) {
    adorn = <Check className={classes.valid} />
  } else if (valid === false) {
    adorn = <Close className={classes.invalid} />
  }
```

If we don't have an error, then we can use the server's response, which will indicate if the number is valid or not.

web/src/app/util/TelTextField.js
```diff
@@ -0,0 +1,99 @@
...
+        +
+      </InputAdornment>
+    ),
+  }
+
+  // if has inputProps from parent commponent, spread it in the iprops
+  if (props.InputProps !== undefined) {
+    iprops = {
+      ...props.InputProps,
+      ...iprops,
```
mastercactapus, line 72: Switch these, always give the caller the final say
```suggestion
      ...iprops,
      ...props.InputProps,
```

web/src/app/util/TelTextField.js
```diff
@@ -0,0 +1,99 @@
...
+    fill: 'green',
+  },
+  invalid: {
+    fill: 'red',
+  },
+})
+
+export default function TelTextField({ ...props }) {
+  const classes = useStyles()
+  const [phoneNumber, setPhoneNumber] = useState(null)
```
mastercactapus, line 32: Start this as an empty string
```suggestion
  const [phoneNumber, setPhoneNumber] = useState('')
```

web/src/app/util/TelTextField.tsx
```diff
@@ -0,0 +1,102 @@
...
+const useStyles = makeStyles({
+  valid: {
+    fill: 'green',
+  },
+  invalid: {
+    fill: 'red',
+  },
+})
+
+export default function TelTextField({ ...props }): JSX.Element {
```
mastercactapus, line 30: Missing type info:

```suggestion
export default function TelTextField(props: TextFieldProps & {value: string}): JSX.Element {
```

`TextFieldProps` comes from material UI, the `value:string` bit is because we require a string type here whereas the `TextField` defines it as unknown.

web/src/app/util/TelTextField.tsx
```diff
@@ -0,0 +1,102 @@
...
+  if (adorn) {
+    iprops = {
+      endAdornment: <InputAdornment position='end'>{adorn}</InputAdornment>,
+      ...iprops,
+    }
+  }
+
+  // remove unwanted character
+  function handleChange(e: React.ChangeEvent<HTMLInputElement>): void {
+    e.target.value = '+' + e.target.value.replace(/[^0-9]/g, '')
```
mastercactapus, line 89: Since `TextField` defines `onChange` as optional, let's just make a simple short-circuit check.
```suggestion
  function handleChange(e: React.ChangeEvent<HTMLInputElement>): void {
    if (!props.onChange) return
    e.target.value = '+' + e.target.value.replace(/[^0-9]/g, '')
```

web/src/app/util/TelTextField.tsx
```diff
@@ -0,0 +1,102 @@
...
+  let adorn
+  if (props.value === '+') {
+    adorn = ''
+  } else if (valid) {
+    adorn = <Check className={classes.valid} />
+  } else if (valid === false) {
+    adorn = <Close className={classes.invalid} />
+  }
+
+  let iprops: object
```
mastercactapus, line 62: Define this as input props instead of object
```suggestion
  let iprops: Partial<InputProps>
```

PR #798, opened 2020-07-16
graphql: remove remaining references to old v1 graphql api

mastercactapus:

**Description:**
This PR removes remaining references to the "legacy" graphql client that uses the old `/v1` API.


web/src/app/util/Options.js
```diff
@@ -16,7 +16,6 @@ import Dialog from '@material-ui/core/Dialog'
 import DialogTitleWrapper from '../dialogs/components/DialogTitleWrapper'
```
Forfold, line 16: Can we delete this file, in favor of the newer `OtherActions`?

mastercactapus: I started to, but it's used in the alerts stuff where the code depends on the mutation behaviors so I decided that'd be better in it's own PR.

PR #802, opened 2020-07-17
ui/utils: convert RequireConfig to typescript

mastercactapus:

**Description:**
This PR converts `RequireConfig.js` to `RequireConfig.tsx`


devtools/configtsidgen/main.go
```diff
@@ -0,0 +1,45 @@
+package main
+
+import (
+	"flag"
```
Forfold, line 4: Is this Go file + the makefile changes + schema changes relevant for converting RequireConfig to Typescript?

mastercactapus: They are so that the typescript definition of possible config IDs are available for the RequireConfig tools. Alternatively, we could change to an enum in `schema.graphql` but since that file is not auto-generated I put the type generation output in the typescript definition file instead.

We could do the former in a subsequent PR and/or look into having a multi-file schema -- I made the assumption this would be a good compromise -- no API/schema changes, but have the config IDs available for typescript checking.

dctalbot (approved):

dctalbot (approved):

web/src/app/util/RequireConfig.tsx
```diff
@@ -102,44 +116,48 @@ export function useConfig() {
...
+  const config = useConfig()
+  const { isAdmin } = useSessionInfo()
 
   if (wantIsAdmin && !isAdmin) {
     return elseValue
   }
 
-  const cfgOpt = config.find((c) => c.id === configID)
-  if (configID && !test(cfgOpt?.value)) {
+  if (configID && !test(config[configID])) {
```
Forfold, line 158: We want the changes from the previous PR:

```suggestion
  const cfgOpt = config.find((c) => c.id === configID)
  if (configID && !test(cfgOpt?.value)) {
```

Forfold (changes requested):

web/src/app/util/RequireConfig.tsx
dctalbot: That change is accounted for. See https://github.com/target/goalert/pull/802/commits/47b3eb9e3efcfb81de2ddf1010c7c634f9948f1d

Forfold (approved):
lgtm

PR #813, opened 2020-07-22
smoketest: add tests for alerts query with regard to the new `includeNotified` flag

jakegleason:

**Description:**
This PR adds tests for the following scenarios in notified alerts:

- Show alerts by all services - Notified = true, FavoritesOnly = false
- Show alerts where notified is true, and favorites is true 
- Show alerts where notified is false, and favorites is true


smoketest/graphqlnotifiedalerts_test.go
```diff
@@ -0,0 +1,153 @@
...
+			}) {
+				nodes {
+					id
+				}
+			}
+		}`, &alerts2)
+
+	if len(alerts2.Alerts.Nodes) != 1 {
+		t.Errorf("got %d alerts; want 1", len(alerts2.Alerts.Nodes))
+	}
```
Forfold, line 134: I think we may need to think a little more on this test case, as the UI _does_ state that it displays both sets of alerts together:

image

and looking at the code for what is set, it appears that both `includeNotified` and `favoritesOnly` are set to `true`:

```
// default to favorites only, unless viewing alerts from a service's page
favoritesOnly: !props.serviceID && !allServices,
includeNotified: !props.serviceID, // keep service list alerts specific to that service
```

I think we will still have to use the user that is created in the harness (`bcefacc0-4764-012d-7bfb-002500d5decb` from `harness/graphql.go`), but I'm not sure why we were seeing foreign constraint errors when trying to add a contact method in the init SQL logic. Perhaps the user isn't created yet when the harness runs the initSql query? @mastercactapus I would appreciate your insight as well!

mastercactapus: That is correct, the user get's created after `initSQL` -- initSQL runs at a particular migration level (specified in the `NewHarness` func) but creating a new user uses the User store method, which has to run after all the other migrations have been applied.

It would be possible to make the user ID available earlier but a little complicated (we'd have to add the user at an early migration, then go to init level, then finish, then edit/update the user to match what we need).

What might be better, is to make a gql "client" to replace all the `doQL` calls and allow specifying the user id

That way we can setup initSQL and then use THAT user to make the calls

jakegleason: Thanks for the feedback. I am not sure how to setup a gql "client." Is there an existing similar file where this has already been done I could look at?

mastercactapus: I think the easiest thing, for now, is a graphQL method on the harness where the user ID can be passed in. Maybe in its own PR though. Then the existing one can just call using a fixed user ID and this can specify it's own

smoketest/graphqlnotifiedalerts_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+/*
+ * This file tests that the alerts GraphQL query shows the proper amount
+ * of results when flipping between "includeNotified" and "favoritesOnly"
+ * query options.
+ *
+ * Service 1: Notified alert
+ * Service 2: Favorited alert
+ */
+func TestNotifiedAlerts(t *testing.T) {
```
mastercactapus, line 19: Use `// line-style` and start with `TestNotifiedAlerts` to document functions.

smoketest/graphqlnotifiedalerts_test.go
```diff
@@ -0,0 +1,171 @@
...
+		query {
+			service(id: "%s") { 
+				isFavorite 
+			}	
+		}
+	`, h.UUID("sid2")), &s)
+
+	if s.Service.IsFavorite != true {
+		t.Fatalf("ERROR: ServiceID %s IsUserFavorite=%t; want true", h.UUID("sid2"), s.Service.IsFavorite)
+	}
```
mastercactapus, line 101: We can do this in the `user_favorites` table in init sql -- no need to use/assert that the `setFavorite` mutation works when we're only testing the `alerts` query.

smoketest/graphqlnotifiedalerts_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// create alerts against both services (notifed version & favorited version)
+	h.CreateAlert(h.UUID("sid"), "notified alert")
+	h.CreateAlert(h.UUID("sid2"), "favorited alert")
```
mastercactapus, line 105: These can be done in init SQL, we're only asserting the query works and will simplify the test.

smoketest/graphqlnotifiedalerts_test.go
```diff
@@ -0,0 +1,171 @@
...
+			nodes {
+				id
+				summary
+			}
+		}
+	}`, &alerts1)
+
+	if len(alerts1.Alerts.Nodes) != 1 {
+		t.Errorf("got %d alerts; want 1", len(alerts1.Alerts.Nodes))
+	}
```
mastercactapus, line 132: Use the assertion library

```go
assert.Len(t, alerts1.Alerts.Nodes, 1, "alerts query")
```

smoketest/graphqlnotifiedalerts_test.go
```diff
@@ -0,0 +1,171 @@
...
+				nodes {
+					id
+					summary
+				}
+			}
+		}`, &alerts2)
+
+	if len(alerts2.Alerts.Nodes) != 2 {
+		t.Errorf("got %d alerts; want 2", len(alerts2.Alerts.Nodes))
+	}
```
mastercactapus, line 152: Same as above

smoketest/graphqlnotifiedalerts_test.go
```diff
@@ -0,0 +1,171 @@
...
+			nodes {
+				id
+				summary
+			}
+		}
+	}`, &alerts3)
+
+	if len(alerts3.Alerts.Nodes) != 2 {
+		t.Errorf("got %d alerts; want 2", len(alerts3.Alerts.Nodes))
+	}
```
mastercactapus, line 170: Same

smoketest/graphqlnotifiedalerts_test.go
jakegleason: Changed comment to the style you requested. Also, started it with `TestNotifiedAlerts`.

smoketest/graphqlnotifiedalerts_test.go
jakegleason: There is now an `insert into user_favorites ` in the init sql.

smoketest/graphqlnotifiedalerts_test.go
jakegleason: Replaced these with two inserts into the `alerts` table. Retested the code, still passing all the tests.

smoketest/graphqlnotifiedalerts_test.go
jakegleason: Replaced with the assertion library version.

smoketest/graphqlnotifiedalerts_test.go
jakegleason: Replaced with the assertion library version.

smoketest/graphqlnotifiedalerts_test.go
jakegleason: Replaced with the assertion library version, and recommitted.

smoketest/graphqlnotifiedalerts_test.go
```diff
@@ -0,0 +1,135 @@
...
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/target/goalert/smoketest/harness"
+)
+
+// TestNotifiedAlerts tests that the alerts GraphQL query shows the proper amount of results when flipping between "includeNotified" and "favoritesOnly" query options.
+// Service 1: Notified alert, Service 2: Favorited alert
+
```
mastercactapus, line 13: Remove the blank line for the comment to be "attached" to the func

smoketest/graphqlnotifiedalerts_test.go
```diff
@@ -0,0 +1,135 @@
...
+	insert into services (id, escalation_policy_id, name) 
+	values
+		({{uuid "sid"}}, {{uuid "eid"}}, 'service'),
+		({{uuid "sid2"}}, {{uuid "eid2"}}, 'service 2');
+
+		insert into alerts (service_id, summary, dedup_key) 
+		values
+			({{uuid "sid"}}, 'Notified alert', 'auto:1:foo'),
+			({{uuid "sid2"}}, 'Favorited alert', 'auto:1:bar');
```
mastercactapus, line 46: nit: fix indentation

smoketest/graphqlnotifiedalerts_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+	insert into user_favorites (user_id, tgt_service_id)
+	values
+		({{uuid "user"}}, {{uuid "sid2"}});
+
+
```
mastercactapus, line 52: nit: cleanup extra whitespace

smoketest/graphqlnotifiedalerts_test.go
jakegleason: Removed blank line.

smoketest/graphqlnotifiedalerts_test.go
jakegleason: Cleaned up extra whitespace.

PR #827, opened 2020-07-29
ui: test notification dialog

### LINKED ISSUE #438: ui: add error dialog support when testing a contact method
Currently, when a user clicks to test their contact method and the mutation fails, no error message is displayed (`web/src/app/users/UserContactMethodList.js`).

A dialog should be displayed similar to other error dialogs within the application if the user attempts to test their contact method and the mutation fails.

Quynhd07:

**Description:**
This PR creates a dialog that displays realtime delivery details and status of the test notification. A graphql query was added to fetch the details/status of the latest text message. The details also changes color depending on the status. 

Closes #438


user/contactmethod/store.go
```diff
@@ -466,3 +477,39 @@ func (db *DB) FindAll(ctx context.Context, userID string) ([]ContactMethod, erro
 
 	return scanAll(rows)
 }
+
+func (db *DB) FindLastStatus(ctx context.Context, cmID string) (*n.MessageStatus, error) {
+	err := validate.UUID("Contact Method ID", cmID)
+	if err != nil {
+		return nil, err
+	}
```
mastercactapus, line 485: Permissions check should go before validation

graphql2/graphqlapp/alert.go
```diff
@@ -42,33 +42,8 @@ func (a *AlertLogEntry) Message(ctx context.Context, obj *alertlog.Entry) (strin
...
-		return nil, nil
-	}
-
-	s, err := (*App)(a).FindOneNotificationMessageStatus(ctx, meta.MessageID)
-	if err != nil {
-		return nil, errors.Wrap(err, "find alert log state")
-	}
-
-	var status graphql2.AlertLogStatus
+func GetNotificationState(s notification.MessageStatus) (*graphql2.NotificationState, error) {
```
mastercactapus, line 45: Since this is a helper method, let's not make it public. We can do a couple other things to make it clearer
- name it something declarative like `notificationStateFromStatus` instead of using `Get`
- remove the `error` from the return signature since it's not used/needed.

```suggestion
func notificationStateFromStatus(s notification.MessageStatus) *graphql2.NotificationState {
```

graphql2/schema.graphql
```diff
@@ -97,6 +97,9 @@ type Query {
 
   # Returns a Slack channel with the given ID.
   slackChannel(id: ID!): SlackChannel
+
+  # Returns status of sendtest with given contact method id
+  sendTestStatus(cmID: ID!): NotificationState
```
mastercactapus, line 102: This implies it's sending a message. I'm thinking since contact methods are rate-limited by both test & verify messages -- should we mirror that here?

```suggestion
  testVerifyStatus(id: ID!): NotificationState
```

I'm unsure if we should use `cmID` or `id` (to match other APIs) we do indicate in the documentation/comment it's a contact method ID.

graphql2/schema.graphql
```diff
@@ -97,6 +97,9 @@ type Query {
 
   # Returns a Slack channel with the given ID.
   slackChannel(id: ID!): SlackChannel
+
+  # Returns status of sendtest with given contact method id
```
mastercactapus, line 101:
```suggestion
  # Returns the status of the latest test or verify message for the given contact method ID.
```

notification/store.go
```diff
@@ -297,6 +297,28 @@ func (db *DB) VerifyContactMethod(ctx context.Context, cmID string, code int) er
 	return nil
 }
 
+func GetMessageStatusState(lastStatus string, hasNextRetry bool) MessageState {
```
mastercactapus, line 300: Same as above, since it's a helper declaratively describe what it does and then make it private (lower case first letter).

```suggestion
func messageStateFromStatus(lastStatus string, hasNextRetry bool) MessageState {
```

notification/store.go
```diff
@@ -328,26 +350,12 @@ func (db *DB) FindManyMessageStatuses(ctx context.Context, ids ...string) ([]Mes
...
-			s.State = MessageStateDelivered
-		case "failed", "bundled": // bundled message was not sent (replaced) and should never be re-sent
-			// temporary if retry
-			if hasNextRetry {
-				s.State = MessageStateFailedTemp
-			} else {
-				s.State = MessageStateFailedPerm
-			}
-		default:
+		state := GetMessageStatusState(lastStatus, hasNextRetry)
```
mastercactapus, line 353: Just use `s.State` instead of declaring a new variable
```suggestion
		s.State = GetMessageStatusState(lastStatus, hasNextRetry)
```

user/contactmethod/store.go
```diff
@@ -29,6 +31,7 @@ type Store interface {
 	EnableByValue(context.Context, Type, string) error
 	DisableByValue(context.Context, Type, string) error
 
+	FindLastStatus(ctx context.Context, cmID string) (*n.MessageStatus, error)
```
mastercactapus, line 34: This belongs in the notification store rather than in contact methods

user/contactmethod/store.go
```diff
@@ -4,8 +4,10 @@ import (
 	"context"
 	"database/sql"
 	"encoding/json"
+	"fmt"
 	"time"
 
+	n "github.com/target/goalert/notification"
```
mastercactapus, line 10: Don't alias packages unless it's needed -- `n` won't be obvious 200 lines down.

As a general rule, a variable/identifier's length/descriptiveness should coincide with how far away from their declaration they are used to give context. A single letter is good for something that is used on a few lines (i.e. one screen).

Also, package imports should be consistent wherever possible.

user/contactmethod/store.go
```diff
@@ -466,3 +477,39 @@ func (db *DB) FindAll(ctx context.Context, userID string) ([]ContactMethod, erro
...
+	if err != nil {
+		return nil, err
+	}
+
+	err = permission.LimitCheckAny(ctx, permission.User)
+	if err != nil {
+		return nil, err
+	}
+
+	tx, err := db.db.BeginTx(ctx, nil)
```
mastercactapus, line 492: Since we're running a single query -- I don't think we need a transaction. We also need to be careful of locks around the `outgoing_messages` table since it gets a lot of important activity.

One thing this API change introduces is the ability to run queries unthrottled directly against it. They are selects & short-lived -- but using a transaction turns it into a minimum 3 round trips instead of one.

user/contactmethod/store.go
```diff
@@ -466,3 +477,39 @@ func (db *DB) FindAll(ctx context.Context, userID string) ([]ContactMethod, erro
...
+	if err != nil {
+		return nil, err
+	}
+	defer tx.Rollback()
+
+	var s n.MessageStatus
+	var lastStatus string
+	var hasNextRetry bool
+	row := wrapTx(ctx, tx, db.findLastStatus).QueryRowContext(ctx, cmID)
```
mastercactapus, line 501: `wrapTx` is a helper for when a func can take an optional transaction (e.g. `tx` could be nil). As written this code will always have a `tx` so `tx.StmtContext(db.findLastStatus)` would be more concise.

user/contactmethod/store.go
```diff
@@ -466,3 +477,39 @@ func (db *DB) FindAll(ctx context.Context, userID string) ([]ContactMethod, erro
...
+	}
+
+	state := n.GetMessageStatusState(lastStatus, hasNextRetry)
+	if state == -1 {
+		return nil, fmt.Errorf("unknown last_status %s", lastStatus)
+	}
+
+	s.State = state
+
+	return &s, nil
```
mastercactapus, line 514: While we can do away with the transaction -- you should always commit within the same scope you call `BeginTx` (other than special circumstances).

If you do use a tx for a read operation, the results are only guaranteed valid (for whatever isolation level is chosen) if the commit succeeds.

user/contactmethod/store.go
```diff
@@ -124,6 +128,13 @@ func NewDB(ctx context.Context, db *sql.DB) (*DB, error) {
 				DELETE FROM user_contact_methods
 				WHERE id = any($1)
 			`),
+		findLastStatus: p.P(`
+			SELECT last_status, status_details, next_retry_at notnull
+			FROM outgoing_messages
+			WHERE contact_method_id = $1 AND message_type = 'test_notification'
+			ORDER BY created_at DESC
+			LIMIT 1
```
mastercactapus, line 136: I wonder if we should store the message ID in the contact method table as part of the send operation rather than needing to search

Has there been any perf. testing/analyze done with stage data? I don't think there is any type of index on `created_at` and with this being a mostly unbounded API -- that could cause big problems.

Let's have a chat on our options

user/contactmethod/store.go
```diff
@@ -466,3 +477,39 @@ func (db *DB) FindAll(ctx context.Context, userID string) ([]ContactMethod, erro
...
 	return scanAll(rows)
 }
+
+func (db *DB) FindLastStatus(ctx context.Context, cmID string) (*n.MessageStatus, error) {
+	err := validate.UUID("Contact Method ID", cmID)
+	if err != nil {
+		return nil, err
+	}
+
+	err = permission.LimitCheckAny(ctx, permission.User)
```
mastercactapus, line 487: Somehow we're going to want to ensure the requester is the same user as the contact method we're looking up

web/src/app/users/SendTestDialog.js
```diff
@@ -0,0 +1,124 @@
...
+const useStyles = makeStyles({
+  statusOk: {
+    color: '#218626',
+  },
+  statusWarn: {
+    color: '#867321',
+  },
+  statusError: {
+    color: '#862421',
+  },
```
mastercactapus, line 45: Where are these from? is there a shared source we can use instead?

web/src/app/users/SendTestDialog.js
```diff
@@ -0,0 +1,124 @@
...
+}
+
+SendTestDialog.propTypes = {
+  messageID: p.string.isRequired,
+  onClose: p.func,
+  disclaimer: p.string,
+  title: p.string,
+  subtitle: p.string,
+  sendTestMutationStatus: p.object,
+}
```
mastercactapus, line 124: Can we use typescript types instead of propTypes?

web/src/app/users/UserContactMethodList.js
```diff
@@ -118,20 +107,21 @@ export default function UserContactMethodList(props) {
...
         label: 'Send Test',
-        onClick: () =>
+        onClick: () => {
           sendTest({
             variables: {
               id: cm.id,
             },
-          }),
+          })
+          setShowSendTestByID(cm.id)
```
mastercactapus, line 116: If we always show the dialog, maybe we can just send the message from there?

We could see when the last message was sent and even indicate to the user the 1 minute cooldown (e.g. with a retry button or something)

web/src/app/users/SendTestDialog.js
```diff
@@ -0,0 +1,124 @@
...
+  )
+}
+
+SendTestDialog.propTypes = {
+  messageID: p.string.isRequired,
+  onClose: p.func,
+  disclaimer: p.string,
+  title: p.string,
+  subtitle: p.string,
+  sendTestMutationStatus: p.object,
```
mastercactapus, line 123: This is a good indicator that the mutation should live in here -- we should define the entire structure explicitly if it's going to be a prop -- but this component needs to react to what's happening with the mutation, thus should probably own it instead of having it passed in

graphql2/schema.graphql
mastercactapus: On second thought maybe we can expand this a bit (pseudocode):

```graphql
lastTestVerify(id: ID!) {
  type: test | verify
  startedAt: ISOTimestamp
  state: NotificationState!
}
```

This would give use better information on what options are available to the user and the information we can provide.

When the dialog opens we could send a message if the last one was sent > 5m ago, otherwise we can display a "test message sent 3m ago" with a "Retry" button that's enabled after 1m -- getting rid of the "rate limit exceeded" error that provides no feedback.

web/src/app/users/SendTestDialog.js
Quynhd07: these are the same colors from the alert log entries

graphql2/graphqlapp/alert.go
Quynhd07: when I removed the error from the return signature, I got an error from the return statement of "too many arguments to return"

notification/store.go
Quynhd07: this helper function is also used in the contact method store. I changed the name but shouldn't it stay public?

graphql2/graphqlapp/alert.go
mastercactapus: You'll need to update all the return statements to match

notification/store.go
mastercactapus: We should probably move the method that's accessing outgoing_messages into the notification store instead anyhow. We're only using the contact method to lookup the message ID.

web/src/app/users/SendTestDialog.js
mastercactapus: Maybe we can move them to be exported from `statusStyles.ts`?

Maybe something like


```ts
export const textColors = {
  statusOk: '#218626',
  // ...
}
```

and then just import for both places

graphql2/graphqlapp/contactmethod.go
```diff
@@ -32,6 +32,39 @@ func (a *ContactMethod) FormattedValue(ctx context.Context, obj *contactmethod.C
...
+func (a *ContactMethod) LastTestMessageState(ctx context.Context, obj *contactmethod.ContactMethod) (*graphql2.NotificationState, error) {
+	t := obj.LastTestVerifyAt()
+	if t.IsZero() {
+		return nil, nil
+	}
+
+	status, _, err := a.NotificationStore.LastMessageStatus(ctx, "test_notification", obj.ID, t)
+	if status == nil {
+		return nil, nil
+	}
```
mastercactapus, line 44: This will always ignore the error, put this after the `err != nil` check

graphql2/graphqlapp/contactmethod.go
```diff
@@ -32,6 +32,39 @@ func (a *ContactMethod) FormattedValue(ctx context.Context, obj *contactmethod.C
 	return formatted, nil
 }
 
+func (a *ContactMethod) LastTestMessageState(ctx context.Context, obj *contactmethod.ContactMethod) (*graphql2.NotificationState, error) {
+	t := obj.LastTestVerifyAt()
+	if t.IsZero() {
+		return nil, nil
+	}
+
+	status, _, err := a.NotificationStore.LastMessageStatus(ctx, "test_notification", obj.ID, t)
```
mastercactapus, line 41: Use the type defined in the `notification` package instead of hard-coding a string.

graphql2/graphqlapp/contactmethod.go
```diff
@@ -32,6 +32,39 @@ func (a *ContactMethod) FormattedValue(ctx context.Context, obj *contactmethod.C
...
+
+	return notificationStateFromStatus(*status), nil
+}
+func (a *ContactMethod) LastVerifyMessageState(ctx context.Context, obj *contactmethod.ContactMethod) (*graphql2.NotificationState, error) {
+	t := obj.LastTestVerifyAt()
+	if t.IsZero() {
+		return nil, nil
+	}
+
+	status, _, err := a.NotificationStore.LastMessageStatus(ctx, "verification_message", obj.ID, t)
```
mastercactapus, line 57: Same as above, use the type from the package instead of the string

notification/store.go
```diff
@@ -16,7 +16,7 @@ import (
 	"github.com/target/goalert/util/sqlutil"
 	"github.com/target/goalert/validation"
 	"github.com/target/goalert/validation/validate"
-
+
```
mastercactapus, line 19: Run `goimports` and remove the extra line

notification/store.go
```diff
@@ -29,6 +29,9 @@ type Store interface {
 	VerifyContactMethod(ctx context.Context, cmID string, code int) error
 	Code(ctx context.Context, id string) (int, error)
 	FindManyMessageStatuses(ctx context.Context, ids ...string) ([]MessageStatus, error)
+
+	// LastMessageStatus will return the MessageStatus and creation timestamp of the message matching the filter critera.
```
mastercactapus, line 33:
```suggestion
	// LastMessageStatus will return the MessageStatus and creation time of the most recent message of the requested type for the provided contact method ID, if one was created from the provided from time.
```

notification/store.go
```diff
@@ -29,6 +29,9 @@ type Store interface {
...
+	LastMessageStatus(ctx context.Context, typ string, cmID string, from time.Time) (*MessageStatus, time.Time, error)
```
mastercactapus, line 34: Use a discrete type instead of arbitrary string for the `typ` parameter.

notification/store.go
```diff
@@ -297,6 +313,28 @@ func (db *DB) VerifyContactMethod(ctx context.Context, cmID string, code int) er
...
+		return MessageStateDelivered
+	case "failed", "bundled": // bundled message was not sent (replaced) and should never be re-sent
+		// temporary if retry
+		if hasNextRetry {
+			return MessageStateFailedTemp
+		} else {
+			return MessageStateFailedPerm
+		}
+	default:
+		return -1
```
mastercactapus, line 334: It looks like both places we use this we are checking for `-1` and converting it to an error. Instead, let's just have this func return an error as a second argument so we can be explicit and return the same error for both.

graphql2/graphqlapp/contactmethod.go
Quynhd07: The notification messageType doesn't match any of the message_type in outgoing_messages. I originally had notification.MessageTypeTest but got a query error so I changed it to the exact message_type string.

notification/store.go
Quynhd07: I couldn't use the discrete type, MessageType from notification, because of the same reason above

graphql2/graphqlapp/contactmethod.go
mastercactapus: Right, they are incompatible since the Go `int` does not work as a `string` that can be cast to the enum in postgres.

To make `MessageType` work with the DB driver, we need to implement Scan and Value

https://golang.org/pkg/database/sql/#Scanner
https://golang.org/pkg/database/sql/driver/#Valuer

There's a few places we've needed to do this, here's an example:
https://github.com/target/goalert/blob/master/alert/status.go

For `Value`, use a switch to return the correct string, and for `Scan` a switch from the DB string to the `MessageType`. Once those methods exist you should no longer get a query error while scanning.

notification/message.go
```diff
@@ -22,3 +27,31 @@ const (
...
+	case 1:
+		return "alert_status_update", nil
+	case 2:
+		return "test_notification", nil
+	case 3:
+		return "verification_message", nil
+	case 4:
+		return "alert_notification_bundle", nil
+	case 5:
+		return "alert_status_update_bundle", nil
```
mastercactapus, line 44: Use the message type values (e.g. `MessageTypeAlertBundle` instead of `5`), not the `int` values.

notification/message.go
```diff
@@ -22,3 +27,31 @@ const (
...
+	}
+	return nil, fmt.Errorf("could not process unknown type for %s", s)
+}
+
+func (s *MessageType) Scan(value interface{}) error {
+	switch t := value.(type) {
+	case int:
```
mastercactapus, line 51: The type from the DB will be a `string` not an `int`, this will need to have a switch statement like `Value` but inverted.

Here's a suggestion to start:
```go
str, ok := value.(string)
if !ok {
  return fmt.Errorf("could not process unknown type %t", value)
}

switch str {
  case "alert_notification_bundle":
    *s = MessageTypeAlertBundle
  // ...
```

web/src/app/users/SendTestDialog.tsx
```diff
@@ -0,0 +1,153 @@
...
+  if (loading || sendTestStatus.loading) return <Spinner text='Loading...' />
+
+  return (
+    <Dialog open onClose={onClose}>
+      <DialogTitle>{title}</DialogTitle>
+
+      <DialogContent>
+        <DialogContentText>
+          GoAlert is sending{' '}
+          {contactMethodType === 'SMS' ? 'an SMS' : 'a VOICE'} to{' '}
```
mastercactapus, line 121: Maybe `GoAlert is sending a test SMS message to {toNumber}` and `GoAlert is making a test voice call to {toNumber}` as the two messages?

"sending a VOICE" doesn't seem super clear

web/src/app/users/SendTestDialog.tsx
```diff
@@ -0,0 +1,153 @@
...
+          {contactMethodToNumber} from {contactMethodFromNumber}
+        </DialogContentText>
+        {details && (
+          <DialogContentText className={getTestStatusClass(status)}>
+            {toTitleCase(details)}
+          </DialogContentText>
+        )}
+        {!details && (
+          <DialogContentText className={classes.statusError}>
+            Try again in one minute.
```
mastercactapus, line 131: Let's make this a little friendlier/suggestive:
```suggestion
            Couldn't send a message yet, please try again after about a minute.
```

PR #828, opened 2020-07-30
notices: display a warning when an EP is not assigned to a service

Forfold:

**Description:**
This PR creates a new API within the graphql schema for a Notice. A Notice should be rendered from a target's (ep, service, etc) details page above the header card if any are present. This PR implements the Notice API, and an example of notices being used for escalation policies.

**Screenshots:**
_Using mock data to display what multiple notices looks like_
Notice Design


graphql2/graphqlapp/escalationpolicy.go
```diff
@@ -323,6 +324,27 @@ func (ep *EscalationPolicy) Steps(ctx context.Context, raw *escalation.Policy) (
 	return ep.PolicyStore.FindAllSteps(ctx, raw.ID)
 }
 
+func (ep *EscalationPolicy) Notices(ctx context.Context, raw *escalation.Policy) ([]graphql2.Notice, error) {
+	_notices, _ := ep.PolicyStore.FindAllNotices(ctx, raw.ID)
```
mastercactapus, line 328: Missing error check

graphql2/schema.graphql
```diff
@@ -1006,3 +1008,15 @@ type AuthSubject {
   subjectID: ID!
   userID: ID!
 }
+
+type Notice {
+  type: NoticeType!
+  message: String!
+  details: String
```
mastercactapus, line 1015: I think we should make this required (since it's a non-structural output) -- it can still be empty and thus falsy.
```suggestion
  details: String!
```

notice/notice.go
```diff
@@ -0,0 +1,16 @@
...
+	Type    NoticeType `json:"type"`
+	Message string     `json:"message"`
+	Details string     `json:"details"`
+}
+
+// NoticeType represents the level of severity of a Notice.
+type NoticeType int
+
+const (
+	Warning NoticeType = iota
```
mastercactapus, line 13: Leave a comment indicating `Warning` is intentionally the default/zero-value

notice/notice.go
```diff
@@ -0,0 +1,16 @@
+package notice
+
+type Notice struct {
+	Type    NoticeType `json:"type"`
+	Message string     `json:"message"`
+	Details string     `json:"details"`
```
mastercactapus, line 6: We can leave out the `json` field tags as they are unused.

web/src/app/details/DetailsPage.js
```diff
@@ -185,10 +191,13 @@ export default function DetailsPage(props) {
     </Grid>
   )
 }
+
 DetailsPage.propTypes = {
   title: p.string,
   details: p.string,
 
+  notices: p.array, // todo: use type Notice when converted to ts
```
mastercactapus, line 199: Define the full type here
```suggestion
  notices: p.arrayOf(p.shape({/* ... */}), // todo: use type Notice when converted to ts
```

web/src/app/details/Notices.tsx
```diff
@@ -0,0 +1,107 @@
...
+      case 1:
+        return classes.secondGridItem
+      case props.notices.length - 1:
+        return classes.lastGridItem
+      default:
+        return classes.gridItem
+    }
+  }
+
+  function renderAlert(notice: Notice, index: number): JSX.Element {
```
mastercactapus, line 74: Can we call this `renderNotice`? Although the component we're using happens to be called `Alert` for our code/API/purposes let's be consistent with "notice" to differentiate.

web/src/app/details/Notices.tsx
```diff
@@ -0,0 +1,107 @@
...
+  },
+})
+
+interface NoticesProps {
+  notices: Notice[] // checks against .length are safe as notices is required
+}
+
+export default function Notices(props: NoticesProps): JSX.Element {
+  const classes = useStyles()
+  const [alertsExpanded, setAlertsExpanded] = useState(false)
```
mastercactapus, line 39: Let's be as consistent as we can be with using "notice"
```suggestion
  const [noticesExpanded, setNoticesExpanded] = useState(false)
```

graphql2/schema.graphql
```diff
@@ -845,6 +845,8 @@ type EscalationPolicy {
 
   assignedTo: [Target!]!
   steps: [EscalationPolicyStep!]!
+
+  notices: [Notice!]
```
mastercactapus, line 849: Same as elsewhere, since it's an output, we can make it "required".

It's only included if asked for, but if asked for you should get an empty array if there are none.

```suggestion
  notices: [Notice!]!
```

escalation/store.go
```diff
@@ -45,6 +46,7 @@ type PolicyStore interface {
 	FindAllPoliciesBySchedule(ctx context.Context, scheduleID string) ([]Policy, error)
 	FindManyPolicies(ctx context.Context, ids []string) ([]Policy, error)
 	DeleteManyPoliciesTx(ctx context.Context, tx *sql.Tx, ids []string) error
+	FindAllNotices(ctx context.Context, policyID string) ([]notice.Notice, error)
```
mastercactapus, line 49: I feel like maybe all the "notice" type things should live in their own store -- EP store can stick to find/update on the policy itself, but then we can have one place for lookup logic like this.

One way that may be important is that notice lookup logic may get somewhat expensive in the future, and we might be caching them for short time periods -- it would get hard to maintain to spread that across all the different stores rather than keeping those concerns inside a single package for finding "notices".

notice/store.go
```diff
@@ -0,0 +1,57 @@
...
+	var notices = make([]Notice, 1)
+	if numServices == 0 {
+		notices[0].Type = Warning
+		notices[0].Message = "Not assigned to a service"
+		notices[0].Details = "To receive alerts for this configuration, assign this escalation policy to its relavent service."
+		return notices, nil
+	}
+
+	// no results
+	return nil, nil
```
mastercactapus, line 56: Convention is to build on-to and then return a slice, that way adding new values can just be independent if-blocks. It also means there's no allocation unless there are notices (hopefully the more rare case).

Also as a side note, always use `:=` if you are assigning and declaring (e.g. only use either `var notices []Notice` or `notices := make([]notice, 1)`).

Since warning is the default, we can omit it as well.

```suggestion
	var notices []Notice
	if numServices == 0 {
		notices = append(notices, Notice{
			Message: "Not assigned to a service",
			Details: "To receive alerts for this configuration, assign this escalation policy to its relevant service.",
		}
	}

	return notices, nil
```

smoketest/graphqlnotices_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+	epID := h.UUID("eid")
+	var notices struct {
+		EscalationPolicy struct {
+			Notices []struct {
+				Type    string `json:"type"`
+				Message string `json:"message"`
+				Details string `json:"details"`
+			} `json:"notices"`
+		} `json:"escalationPolicy"`
```
mastercactapus, line 49: json tags shouldn't be needed since the fields match

smoketest/graphqlnotices_test.go
```diff
@@ -0,0 +1,68 @@
...
+					message
+					details
+    			}
+  			}
+		}
+	`, epID), &notices)
+
+	if len(notices.EscalationPolicy.Notices) < 1 {
+		t.Fatalf("ERROR: retrieved length of notices = %d; want %d", len(notices.EscalationPolicy.Notices), 1)
+	}
```
mastercactapus, line 67: Use the assertion library

```suggestion

	assert.Len(t, notices, 1, "retrieved notices")
```

notice/notice.go
```diff
@@ -0,0 +1,17 @@
+package notice
+
+type Notice struct {
+	Type    NoticeType
+	Message string
+	Details string
+}
+
+// NoticeType represents the level of severity of a Notice.
+type NoticeType int
```
mastercactapus, line 10: Since we're already in qualified in the package `notice` just call this `Type` -- it will be used as `notice.Type` everywhere and `notice.NoticeType` stutters.

notice/notice.go
```diff
@@ -0,0 +1,17 @@
...
+
+// Defaults to Warning when unset
+const (
+	Warning NoticeType = iota
```
mastercactapus, line 14: Add `Type` as a prefix to these `TypeWarning`, `TypeError` etc...

graphql2/graphqlapp/escalationpolicy.go
```diff
@@ -323,6 +324,30 @@ func (ep *EscalationPolicy) Steps(ctx context.Context, raw *escalation.Policy) (
 	return ep.PolicyStore.FindAllSteps(ctx, raw.ID)
 }
 
+func (ep *EscalationPolicy) Notices(ctx context.Context, raw *escalation.Policy) ([]graphql2.Notice, error) {
```
mastercactapus, line 327: Can we map this to `notice.Notice` instead (inside `gqlgen.yml`)? We could probably just do `return ep.NoticeStore.FindAllPolicyNotices(ctx, raw.ID)` that way without needing to loop over them to convert.

The only one we may have trouble with is the `Type` but we can map that with a marshaler method (can add UnmarshalGQL method like in assignment/targetttype.go) if it complains.

smoketest/graphqlnotices_test.go
Forfold: Gotcha, I wasn't sure because of the lowercase/title-case discrepancy

notice/notice.go
```diff
@@ -0,0 +1,59 @@
+package notice
+
+//go:generate go run golang.org/x/tools/cmd/stringer -type Type
```
mastercactapus, line 3: Add an entry in the `Makefile` so this gets run as needed.

```
notification/type_string.go: notice/notice.go
	go generate ./notice
```

notice/store.go
```diff
@@ -0,0 +1,56 @@
...
+func (s *Store) FindAllPolicyNotices(ctx context.Context, policyID string) ([]Notice, error) {
+	err := validate.UUID("EscalationPolicyStepID", policyID)
+	if err != nil {
+		return nil, err
+	}
+
+	err = permission.LimitCheckAny(ctx, permission.Admin, permission.User)
+	if err != nil {
+		return nil, err
+	}
```
mastercactapus, line 39: Permission first, then validate parameters. Also `permission.User` implicitly includes `Admin`

app/initstores.go
```diff
@@ -2,6 +2,7 @@ package app
 
 import (
 	"context"
+	"github.com/target/goalert/notice"
 	"net/url"
 
 	"github.com/target/goalert/alert"
```
mastercactapus, line 8: goimports

smoketest/graphqlnotices_test.go
```diff
@@ -0,0 +1,67 @@
+package smoketest
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/target/goalert/smoketest/harness"
+	"testing"
```
mastercactapus, line 7: Run `goimports`

Should result in something like:
```
	"encoding/json"
	"fmt"
	"testing"

	"github.com/target/goalert/smoketest/harness"
```

app/initstores.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"net/url"
 
+	"github.com/target/goalert/notice"
+
```
mastercactapus, line 8: nit: remove extra whitespace -- don't forget to run goimports to sort after

notice/store.go
```diff
@@ -0,0 +1,56 @@
...
+	return &Store{
+		findServicesByPolicyID: p.P(`
+			SELECT COUNT(*)
+			FROM services
+			WHERE escalation_policy_id = $1
+		`),
+	}, p.Err
+}
+
+// Sets a notice for a Policy if it is not assigned to any services
```
mastercactapus, line 29: Should start with `FindAllPolicyNotices`

You can use golint to check: `go run golang.org/x/lint/golint ./notice`

app/initstores.go
Forfold: Interesting, this was the result of `goimports -w app/initstores.go`

PR #831, opened 2020-08-03
engine/message: append alert log failure when Twilio is disabled

KaiHuaLin:

**Description:**
When twilio is disabled and trying to send message to the user, it will notify the user in the alert log that twilio is disabled.
**Screenshots:**
Screen Shot 2020-08-03 at 2 32 22 PM

### DIFF
--- a/alert/log/store.go
+++ b/alert/log/store.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/target/goalert/config"
 	"github.com/target/goalert/integrationkey"
 	"github.com/target/goalert/notificationchannel"
 	"github.com/target/goalert/permission"
@@ -266,6 +267,8 @@ func (db *DB) logAny(ctx context.Context, tx *sql.Tx, insertStmt *sql.Stmt, id i
 		return err
 	}
 
+	fmt.Println(_type)
+
 	var classExtras []string
 	switch _type {
 	case _TypeAcknowledgeAll:
@@ -287,6 +290,7 @@ func (db *DB) logAny(ctx context.Context, tx *sql.Tx, insertStmt *sql.Stmt, id i
 	}
 
 	src := permission.Source(ctx)
+	fmt.Println(src)
 	if src != nil {
 		switch src.Type {
 		case permission.SourceTypeNotificationChannel:
@@ -320,7 +324,12 @@ func (db *DB) logAny(ctx context.Context, tx *sql.Tx, insertStmt *sql.Stmt, id i
 			}
 			if _type == TypeNoNotificationSent {
 				// no CMID for no notification sent
-				r.subject.classifier = "no immediate rule"
+				cfg := config.FromContext(ctx)
+				if !cfg.Twilio.Enable {
+					r.subject.classifier = "Twilio Disabled"
+				} else {
+					r.subject.classifier = "no immediate rule"
+				}
 				break
 			}
 			var cmType contactmethod.Type
--- a/alert/log/store.go
+++ b/alert/log/store.go
@@ -267,8 +267,6 @@ func (db *DB) logAny(ctx context.Context, tx *sql.Tx, insertStmt *sql.Stmt, id i
 		return err
 	}
 
-	fmt.Println(_type)
-
 	var classExtras []string
 	switch _type {
 	case _TypeAcknowledgeAll:
@@ -290,7 +288,6 @@ func (db *DB) logAny(ctx context.Context, tx *sql.Tx, insertStmt *sql.Stmt, id i
 	}
 
 	src := permission.Source(ctx)
-	fmt.Println(src)
 	if src != nil {
 		switch src.Type {
 		case permission.SourceTypeNotificationChannel:
--- a/alert/log/store.go
+++ b/alert/log/store.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/target/goalert/config"
 	"github.com/target/goalert/integrationkey"
 	"github.com/target/goalert/notificationchannel"
 	"github.com/target/goalert/permission"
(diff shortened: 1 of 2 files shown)


alert/log/store.go
```diff
@@ -320,7 +321,12 @@ func (db *DB) logAny(ctx context.Context, tx *sql.Tx, insertStmt *sql.Stmt, id i
...
 			if _type == TypeNoNotificationSent {
 				// no CMID for no notification sent
-				r.subject.classifier = "no immediate rule"
+				cfg := config.FromContext(ctx)
+				if !cfg.Twilio.Enable {
+					r.subject.classifier = "Twilio Disabled"
+				} else {
+					r.subject.classifier = "no immediate rule"
+				}
 				break
```
mastercactapus, line 330: We don't need this anymore do we?

engine/message/db.go
```diff
@@ -45,6 +45,8 @@ type DB struct {
 	failDisabledCM *sql.Stmt
 	alertlogstore  alertlog.Store
 
+	twilioDisabled *sql.Stmt
```
mastercactapus, line 48: name this something more descriptive like `failSMSVoice`

engine/message/db.go
```diff
@@ -277,6 +279,23 @@ func NewDB(ctx context.Context, db *sql.DB, c *Config, a alertlog.Store) (*DB, e
...
+				set
+					last_status = 'failed',
+					last_status_at = now(),
+					status_details = 'twilio disabled',
+					cycle_id = null,
+					next_retry_at = null
+				from user_contact_methods cm
+				where
+					msg.last_status = 'pending' and
+					msg.message_type != 'verification_message'
```
mastercactapus, line 294: We need to join against the contact method table and assert the type, anything going to voice or SMS will fail if Twilio is not configured

```suggestion
				join user_contact_methods cm on cm.id = msg.contact_method_id
				where
					msg.last_status = 'pending' and
					cm.type in ('SMS', 'VOICE')
```

engine/message/db.go
```diff
@@ -277,6 +279,23 @@ func NewDB(ctx context.Context, db *sql.DB, c *Config, a alertlog.Store) (*DB, e
...
+				returning msg.id as msg_id, alert_id, msg.user_id, cm.id as cm_id
+			) select distinct msg_id, alert_id, user_id, cm_id from disabled
```
mastercactapus, line 296: I think the only reason this was needed is we weren't joining properly, it should be enough to just use the `update...returning`

engine/message/db.go
```diff
@@ -575,6 +594,49 @@ func (db *DB) _SendMessages(ctx context.Context, send SendFunc, status StatusFun
...
+	// if twilio is disable, create an entry to notify the user
+	cfg := config.FromContext(ctx)
+	if !cfg.Twilio.Enable {
+		rows, err := tx.Stmt(db.twilioDisabled).QueryContext(execCtx)
+		if err != nil {
+			return errors.Wrap(err, "check for failed message")
+		}
+		defer rows.Close()
+
+		type record struct {
```
mastercactapus, line 606: use `msgMeta` below by raising it above, along with the `var msgs msgMeta` declaration

Both of these can just append to the same slice that way and we can use the single `for _, m := range msgs {` below

engine/message/db.go
```diff
@@ -575,6 +594,49 @@ func (db *DB) _SendMessages(ctx context.Context, send SendFunc, status StatusFun
 		return errors.Wrap(err, "clear disabled status updates")
 	}
 
+	// if twilio is disable, create an entry to notify the user
+	cfg := config.FromContext(ctx)
+	if !cfg.Twilio.Enable {
+		rows, err := tx.Stmt(db.twilioDisabled).QueryContext(execCtx)
```
mastercactapus, line 600: Use `StmtContext` for new code.

```suggestion
		rows, err := tx.StmtContext(ctx, db.twilioDisabled).QueryContext(execCtx)
```

engine/message/db.go
```diff
@@ -590,13 +600,40 @@ func (db *DB) _SendMessages(ctx context.Context, send SendFunc, status StatusFun
...
+
+		for rows.Next() {
+			var msg msgMeta
+			err = rows.Scan(&msg.MessageID, &msg.AlertID, &msg.UserID, &msg.CMID)
+			if err != nil {
+				return errors.Wrap(err, "scan all failed messages")
+			}
+			msgs = append(msgs, msg)
+		}
+	} else {
```
mastercactapus, line 621: The `failDisabledCM` code must always be run, regardless of the state of Twilio -- should not be inside an `else`

engine/message/db.go
```diff
@@ -277,6 +279,21 @@ func NewDB(ctx context.Context, db *sql.DB, c *Config, a alertlog.Store) (*DB, e
...
+				last_status = 'failed',
+				last_status_at = now(),
+				status_details = 'twilio disabled',
+				cycle_id = null,
+				next_retry_at = null
+			from user_contact_methods cm
+			where
+				msg.last_status = 'pending' and
+				cm.type in ('SMS', 'VOICE')
+			returning msg.id as msg_id, alert_id, msg.user_id, cm.id as cm_id
```
mastercactapus, line 294: We need to ensure we're comparing `.type` on the contact method associated with the message.
```suggestion
			where
				msg.last_status = 'pending' and
				cm.type in ('SMS', 'VOICE') and
				cm.id = msg.id
			returning msg.id as msg_id, alert_id, msg.user_id, cm.id as cm_id
```

alert/log/store.go
mastercactapus: Since `TypeNoNotificationSent` is no longer used for the Twilio disabled code, we can revert this to the previous code

engine/message/db.go
```diff
@@ -277,6 +279,22 @@ func NewDB(ctx context.Context, db *sql.DB, c *Config, a alertlog.Store) (*DB, e
 			) select distinct msg_id, alert_id, user_id, cm_id from disabled
 		`),
 
+		failSMSVoice: p.P(`
+			update outgoing_messages msg
+			set
+				last_status = 'failed',
+				last_status_at = now(),
+				status_details = 'twilio disabled',
```
mastercactapus, line 287: Maybe we should generalize this a bit?

```suggestion
				status_details = 'SMS/Voice support not enabled by administrator',
```

PR #834, opened 2020-08-04
profile: conflicting contact method, user's name with link

alejandra-martinez:

**Description:**
added name of the user who owns the conflicting contact method and the error links to the conflicting users profile

allows users. to see who owns the conflicting contact method


user/contactmethod/store.go
```diff
@@ -261,6 +271,12 @@ func (db *DB) CreateTx(ctx context.Context, tx *sql.Tx, c *ContactMethod) (*Cont
 
 	_, err = wrapTx(ctx, tx, db.insert).ExecContext(ctx, n.ID, n.Name, n.Type, n.Value, n.Disabled, n.UserID)
 	if err != nil {
+		if strings.Contains(err.Error(), "user_contact_methods_type_value_key") {
+			var u string
+			row := db.findUserID.QueryRowContext(ctx, n.Type, n.Value)
+			row.Scan(&u)
+			return nil, validation.NewFieldError("Value", fmt.Sprintf("Contact method already exists for that type and value for user %v", u))
```
mastercactapus, line 278: This should be the format of `contact method already exists for that type and value; USER_ID=user-id-here`

web/src/app/forms/FormField.js
```diff
@@ -186,13 +186,24 @@ export class FormField extends React.PureComponent {
...
+                props.hint}
+            </FormHelperText>
+          )}
+        {!noError &&
+          (props.error || props.hint) &&
+          typeof props.error.message !== 'string' && (
+            <FormHelperText>
+              {(props.error && props.error.message) || props.hint}
+            </FormHelperText>
+          )}
```
mastercactapus, line 206: Rather than adding multiple ways of rendering errors, we should make this new one follow existing conventions

Quynhd07:
> As I'm looking at this, I wonder if it would make more sense to make a query to lookup a contact method by type, value and inform the user BEFORE submitting that there is a conflict, or at least using that for the error info rather than doing a query for the user ID and then having the UI do a second lookup anyhow.
Alternatively we could extend the users search to allow searching by contact method value (And optionally type).

Would this mean creating a new GraphQL query to look up the new contact method given the inputs? OR does it mean just mean just creating a SQL statement and using that logic before getting back the conflict? 

We think that the first approach is more simple and concise compared to extending the search functionality. Since this logic is only triggered when creating a new contact method, which only occurs when the user is setting up their new profile, which is not too often, it makes sense to have this close to the create contact method logic (all in the same place in code). Having said that, it could make sense to search all users if that is something that is used more frequently in the future.

mastercactapus:
It would be more simple and concise for this particular problem to make it's own query, but in the bigger picture I think extending user search is the better fit. The ability to lookup a user by phone number has been asked for in the past, though I don't think we would include any UI features for that as part of this.

---

Here's my thinking:

We already have `userContactMethod` query which takes an ID and returns a single contact method (which can then be used to query for the associated user) -- but we can't change that API in any breaking way. We could add type & value as parameters and require `id` or `type & value` though it feels a little messy -- the only thing we really want to solve is "does a user own this type/value combo?"

Adding a new query with search could be named `userContactMethods` but I'm not sure we really have a case for a paginated search of contact methods. And adding a `userContactMethodByValue` or similar specific query seems like a hack for a 1-off problem that I'd rather avoid in favor of a more general solution.

The user search isn't as direct a solution for this issue, but I think it might fit better into the overall API and has additional use.  Imagine an error found in Twilio logs with a phone number, and an admin may need to lookup who owns that number. In that way extending user search would be no different than searching for a user by email.

We could just add discrete fields and if we ever wanted to integrate it into search/list page, we could handle that on the client side.

`users(input: {cmType: SMS, cmValue: "+17635550123"})` 

---

If we go with the user search approach we only lose the ability to find the contact method's id & name by type/value (since you get a user back instead of the contact method) but it feels more like extending an existing query in a meaningful way that has future applications and all contact method management is done at the per-user level, rather than globally.

PR #842, opened 2020-08-11
auth: link mobile device API

mastercactapus:

**Description:**
This PR adds a new identity provider "authlink" that can be used to link a device to an existing session.


auth/authlink/store.go
```diff
@@ -0,0 +1,240 @@
...
+	for i := range val {
+		num, err := rand.Int(rand.Reader, big.NewInt(int64(len(codeAlphabet))))
+		if err != nil {
+			return "", err
+		}
+		val[i] = codeAlphabet[num.Int64()]
+	}
+
+	return string(val), nil
+}
```
Forfold, line 71: Like we talked about before briefly, could we separate this out to offer a digit only code? Also, I noticed that the `codeAlphabet` is missing `0` and `1`.

Perhaps something simple like this could work:
```
const codeAlphabet = "ABCDEFGHJKLMNPQRSTWXYZ"
const codeNums = "0123456789"

func genCode(n int, numOnly bool) (string, error) {
	var dict = codeAlphabet + codeNums
	if numOnly {
		dict = codeNums
	}
	val := make([]byte, n)
	for i := range val {
		num, err := rand.Int(rand.Reader, big.NewInt(int64(len(dict))))
		if err != nil {
			return "", err
		}
		val[i] = dict[num.Int64()]
	}

	return string(val), nil
}
```

auth/authlink/store.go
```diff
@@ -0,0 +1,240 @@
...
+	err := permission.LimitCheckAny(ctx, permission.UserSession)
+	if err != nil {
+		return nil, err
+	}
+	stat := Status{ID: uuid.NewV4().String()}
+	stat.ClaimCode, err = genCode(8)
+	if err != nil {
+		return nil, err
+	}
+	stat.VerifyCode, err = genCode(8)
```
Forfold, line 166: Same as above, maybe 4 or 6 so it's easier for the user to verify?

auth/authlink/store.go
```diff
@@ -0,0 +1,240 @@
...
+			where
+				id = $1 and
+				now() < expires_at and
+				claimed_at notnull and
+				verified_at notnull and
+				authed_at isnull
+			returning user_id
+		`),
+
+		reset: p.P(`delete from auth_link_codes where user_id = $1`),
```
Forfold, line 132: Is this intended to reset all sessions for a user, or should this be `where id = $1` to reset a single session? If the former, could we perhaps rename to `resetAll` for clarity?

auth/authlink/store.go
```diff
@@ -0,0 +1,240 @@
...
+		&stat.ExpiresAt,
+		&claim, &verify, &auth,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	stat.ClaimedAt = claim.Time
+	stat.VerifiedAt = verify.Time
+	stat.AuthedAt = auth.Time
```
Forfold, line 207: Can these values be scanned directly into `stat`?

auth/authlink/whitelist.go
```diff
@@ -0,0 +1,100 @@
...
+		if e.invalid {
+			continue
+		}
+		e.invalid = true
+		close(e.wait)
+	}
+}
+
+var (
+	ErrBadID = errors.New("no value with that ID")
```
Forfold, line 49: nit:
```suggestion
	ErrBadID = errors.New("no value with given ID")
```

auth/handler.go
```diff
@@ -305,8 +318,9 @@ func (h *Handler) handleProvider(id string, p IdentityProvider, refU *url.URL, w
 		return
 	}
 
-	var userID string
-	err = h.userLookup.QueryRowContext(ctx, id, sub.SubjectID).Scan(&userID)
+	if sub.UserID == "" {
+		err = h.userLookup.QueryRowContext(ctx, id, sub.SubjectID).Scan(&sub.UserID)
+	}
```
Forfold, line 323: I'm curious why we would want to look up a user if the user is set to empty. Could we skip the lookup entirely since we know the userID isn't present and return an appropriate error?

config/config.go
```diff
@@ -36,8 +36,9 @@ type Config struct {
 	}
 
 	Auth struct {
-		RefererURLs  []string `info:"Allowed referer URLs for auth and redirects."`
-		DisableBasic bool     `public:"true" info:"Disallow username/password login."`
+		RefererURLs     []string `info:"Allowed referer URLs for auth and redirects."`
+		DisableBasic    bool     `public:"true" info:"Disallow username/password login."`
+		DisableAuthLink bool     `public:"true" info:"Disallow linking mobile devices."`
```
Forfold, line 41: nit:
```suggestion
		DisableAuthLink bool     `public:"true" info:"Disallow linking external devices."`
```

because I _know_ we can get this working in a Tesla 😉

graphql2/graphqlapp/mutation.go
```diff
@@ -16,6 +18,44 @@ type Mutation App
 
 func (a *App) Mutation() graphql2.MutationResolver { return (*Mutation)(a) }
 
+func (a *Mutation) CreateAuthLink(ctx context.Context) (*graphql2.AuthLink, error) {
+
```
Forfold, line 22: nit: remove empty line

graphql2/graphqlapp/mutation.go
```diff
@@ -16,6 +18,44 @@ type Mutation App
...
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return &graphql2.AuthLink{
+		ID: stat.ID,
+
```
Forfold, line 43: nit: remove empty line

graphql2/graphqlapp/query.go
```diff
@@ -14,6 +14,20 @@ type Query App
...
+	if err != nil {
+		return nil, err
+	}
+	return &graphql2.AuthLinkStatus{
+		ID:        stat.ID,
+		ExpiresAt: stat.ExpiresAt,
+		Claimed:   !stat.ClaimedAt.IsZero(),
+		Verified:  !stat.VerifiedAt.IsZero(),
+		Authed:    !stat.AuthedAt.IsZero(),
+	}, nil
```
Forfold, line 28: Would it be worth mapping this in `gqlgen.yml` like we did for the notices work so we can return flat without recreating to match the schema?

auth/authlink/store.go
mastercactapus: It's missing anything that could be confused -- e.g. 0 & O, l I & 1, u & v etc..

Instead of switching to numbers entirely, what do you think if we shorten it and maybe just offer confirmation? (i.e. no typing)

auth/authlink/store.go
mastercactapus: A user can only link a single device at a time, and linking is locked to a specific session so there will only be 0 or 1, this reset's the user's link attempt, but is using the user_id instead of session in case it was started in another tab/browser

auth/authlink/store.go
mastercactapus: We can't because they can be null, we take Time and if it's null then it will just be the zero value. So we scan into `sql.NullTime` when pulling from the DB and then with stat switch to the Go-style zero value.

graphql2/graphqlapp/query.go
mastercactapus: Not unless we have a reason to want the timestamps and checking the zero-values in the UI -- this is mostly to get the `boolean` values mapped from `time.Time`

auth/handler.go
mastercactapus: `sub` is returned from the IdentityProvider. Most providers only know of the subject ID, (e.g. the GitHub ID for GitHub auth) and so we need to find which GoAlert UserID that maps to -- unless we already know the user, as with the `authlink` provider.

config/config.go
mastercactapus: Haha I'll make it a bit more generic, but I want it to be clear that it is used/required for the mobile app

PR #846, opened 2020-08-17
ui: fix bug shift time limit

### LINKED ISSUE #749: Adding shift time limit doesn't allow backspace for first character
**Describe the Bug:**
Time Limit defaults to "14". Backspacing only allows it to backspace to "1". Manually using the arrow button to increase the number works, but then the new number still cannot be backspaced. 

**Steps to Reproduce:**
1. Go to Schedules->Shifts->Three bars icon
2. Click on "Time Limit"
3. Click on "Specify..."
4. Backspace characters in the input box as far as possible
5. Increase manually from "1" to "3" with the arrow button
6. Backspace characters in the input box as far as possible

**Expected Behavior:**
All characters can be backspaced to produce an empty input box.

**Observed Behavior:**
The very first character is present in the input box.

**Screenshots/Stack Traces:**
Screen Shot 2020-06-28 at 10 35 27 PM

**Application Version:**
`
$ goalert version
Version: v0.24.0 
...
`

**Desktop:**
 - OS: MacOS Mojave 10.14.6
 - Browser: Firefox
 - Version: 76.0.1

Perhaps this is intentional to never allow an empty inbox. It seems like more of a nice to have. Any comments on whether or not this should be considered a bug are welcome and I can remove this.

pnengchu:

**Description:**
This PR fixes the bug found in Schedules > Shifts > Filter > Time Limit > Specify. The capability to backspace numerical values when "Specify" is selected was added as well as a conditional check defaulting user input values between 1-30 appropriately. 

Fixes #749


web/src/app/schedules/ScheduleShiftList.js
```diff
@@ -266,17 +267,33 @@ export default class ScheduleShiftList extends React.PureComponent {
...
-              days: clamp(1, 30, parseInt(e.target.value, 10)),
-            }).toISO(),
-          )
+          this.setState({ isClear: e.target.value === '' })
+          let val = e.target.value
+          if (parseInt(val, 10) > 30) {
+            val = '30'
+          } else if (parseInt(val, 10) < 1) {
+            val = '1'
+          }
```
mastercactapus, line 289: Doesn't look like this is used anymore

go.mod
```diff
@@ -67,7 +67,6 @@ require (
 	github.com/vektah/gqlparser/v2 v2.0.1
 	go.opencensus.io v0.22.2
 	golang.org/x/crypto v0.0.0-20200622213623-75b288015ac9
-	golang.org/x/mod v0.3.0 // indirect
```
mastercactapus, line 70: unintentional I assume?

web/src/app/schedules/ScheduleShiftList.js
```diff
@@ -266,17 +267,33 @@ export default class ScheduleShiftList extends React.PureComponent {
...
+          if (parseInt(e.target.value) > 0) {
+            this.props.handleSetDuration(
+              Duration.fromObject({
+                days: clamp(1, 30, parseInt(e.target.value, 10)),
+              }).toISO(),
+            )
+          }
```
mastercactapus, line 296: Invert and flatten this so that the `if` statement returns and the `handleSetDuration` runs outside

```js
if (failure condition) {
  return
}

// normal/success flow
this.props.handleSetDuration( ... )
```

web/src/app/schedules/ScheduleShiftList.js
Forfold: It looks to be working for me — clamping the value at `30` if typing say `144` or at `1` if typing say `-3`

pnengchu: Yes! beforehand, numerical values did not exceed 30 and that was handled by min & max props. This addition is not part of the acceptance criteria but it solves the scenario of what if a user specifies 60 or 90 days out, which is more than the 30 days. I wanted to mirror the former behavior by adding in the above.

go.mod
pnengchu: yes, unintentional. I will address this and make sure it matches the master branch.

web/src/app/schedules/ScheduleShiftList.js
mastercactapus: Clamping is already handled by the `clamp` function below this; `val` is unused, so this section of code can be dropped. 

`make check` would have flagged it as an unused variable but it is being read by `parseInt` in these conditionals and then re-assigned, after which it is never referenced again.

pnengchu: Ah okay, I see what you mean now. This has been addressed and the changes have been made as suggested. Thanks for everyone's feedback!

web/src/app/schedules/ScheduleShiftList.js
pnengchu: addressed and implemented.

web/src/app/schedules/ScheduleShiftList.js
```diff
@@ -266,12 +267,23 @@ export default class ScheduleShiftList extends React.PureComponent {
...
         min={1}
         type='number'
+        onBlur={() => {
+          this.setState({ isClear: false })
+        }}
         onChange={(e) => {
+          this.setState({ isClear: e.target.value === '' })
+          if (parseInt(e.target.value) < 0) {
+            return
+          }
```
mastercactapus, line 286: I think this is also handled by the `clamp` function below since -1, 0, etc.. is going to be less than 1 (as written).

The only thing it won't handle though is `NaN`. The logic you had before `if (parseInt(...) > 0) { handleSetDuration() }` would implicitly check that it was a valid number as well as `> 0`. 

Basically in javascript the logical inverse of `parseInt(x) > 0` would be either `!parseInt(x) <= 0` or `Number.isNaN(parseInt(x)) || parseInt(x) <= 0`. But since we're "clamping" the value already, it's probably enough to do the `isNaN` check here to avoid the redundant check.

```suggestion
          if (Number.isNaN(parseInt(e.target.value, 10))) {
            return
          }
```

This should ensure invalid values/states are ignored.

pnengchu: thanks for the feedback. The solution proposed is addressed and implemented.

PR #853, opened 2020-08-24
auth: web ui to link with mobile

Forfold:

**Description:**
This PR adds a "Link mobile device" to the user's profile such that they can authenticate and login to the (wip) GoAlert mobile app. First, the user scans/types in the QR code value, then verifies the code seen on both the phone (output from curl) and on the UI. After this, the phone will hit the endpoint below to finalize authentication, once it sees that the code has been verified by the user.

Until implemented on mobile as well, to verify test against `localhost:3030/api/v2/identity/providers/link` with `key -> value` equal to `code -> console.log() output when clicking button`

**Screenshots:**


web/src/app/users/linkToMobile/LinkToMobile.tsx
```diff
@@ -0,0 +1,202 @@
...
+    if (claimed) {
+      setIndex(1)
+    }
+  }, [claimed])
+
+  useEffect(() => {
+    if (authed) {
+      setIndex(2)
+    }
+  }, [authed])
```
mastercactapus, line 107: Data changing already causes a re-render, pushing it back into state just causes a new one. Instead let's just calculate it directly:

```js
let index = 0
if (authed) index = 2
else if (claimed) index = 1
```

web/src/app/users/linkToMobile/LinkToMobile.tsx
```diff
@@ -0,0 +1,202 @@
...
+          </DialogContent>
+        ) : (
+          <VirtualizeAnimatedViews
+            index={index}
+            onChangeIndex={(i: number) => setIndex(i)}
+            slideRenderer={slideRenderer}
+          />
+        )}
+        <DialogActions>
+          {index === 2 ? (
```
mastercactapus, line 162: Maybe we should store this in a variable, or a constant to compare with so we don't need context to understand the ternary.

```js
index === lastStepIndex
```

or

```js
isDone ? (
```

web/src/app/users/linkToMobile/VerifyCodeFields.tsx
```diff
@@ -0,0 +1,140 @@
...
+}
+
+export default function VerifyCodeFields(
+  props: VerifyCodeFieldsProps,
+): JSX.Element {
+  const classes = useStyles()
+  const [numOne, setNumOne] = useState('')
+  const [numTwo, setNumTwo] = useState('')
+  const [numThree, setNumThree] = useState('')
+  const [numFour, setNumFour] = useState('')
```
mastercactapus, line 42: Can we make this a little more generic? Take a prop like `length`, store the code as a single string in state.

web/src/app/users/linkToMobile/VerifyCodeFields.tsx
```diff
@@ -0,0 +1,140 @@
...
+                  const inputThree = document.getElementById('numThree')
+                  if (inputThree) inputThree.focus()
+                }
+              }}
+              inputProps={{
+                maxLength: 1,
+                className: classes.textField,
+              }}
+            />
+          </Grid>
```
mastercactapus, line 107: Since these are all identical, we can just render them in a loop for whatever length we want/need entry to be. For the focus bit just use the index in the id (e.g. `verify-${idx}`)

The conditional on the "next" element will already handle the final field since `document.getElementById(verify-${idx+1})` will return null.

web/src/app/users/linkToMobile/VerifyCodeFields.tsx
Forfold: Will be resolved transitioning to verifying without re-typing

mastercactapus:
Instead of fixed colors per-column can we use color as a pattern hint?

Meaning 2 codes, even if similar will have different colors to make it obvious they don't match. Should be a fun problem to solve

Maybe use the code as a seed for a random sequence (so you get the same colors & order on both devices)

auth/authlink/store.go
```diff
@@ -53,7 +53,7 @@ type Status struct {
 	AuthedAt   time.Time
 }
 
-const codeAlphabet = "ABCDEFGHJKLMNPQRSTWXYZ23456789"
+const codeAlphabet = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
```
mastercactapus, line 56: It was intentional to completely leave out any symbols that could be confused depending on fonts: Il1, O0. U and V are probably different enough but were left out as well intentionally.

Since we can't easily control the fonts available (esp. in browsers) the thought was to just leave it out of scope and resolve the ambiguity by only using characters that are always visually distinguishable. As long as both devices use the same font we should be okay -- but if we add in the other characters we'll need to add finding the right font -- something that we can instead do at a later time without consequence.

web/src/app/users/linkToMobile/LinkToMobile.tsx
```diff
@@ -0,0 +1,266 @@
...
+
+  // getter for error messages
+  const urlParams = useSelector(urlParamSelector)
+  const errorMsg = urlParams('error', '')
+
+  // setter for error messages
+  const dispatch = useDispatch()
+  const setErrorMessage = (value: string): void => {
+    dispatch(setURLParam('error', value))
+  }
```
mastercactapus, line 91: Why are we sending the error to the address bar? are we reloading the page?

web/src/app/util/colors.ts
```diff
@@ -0,0 +1,166 @@
...
+    for (let j = 0; j < 3; j++) {
+      const rng = seedrandom((colorSeed() + j).toString())()
+      rgb[j] = Math.floor(((rng + 1) * 255) / 2)
+    }
+
+    colors.push(makeColorA11y(rgb))
+  }
+
+  return colors
+}
```
mastercactapus, line 27: I'm thinking we should keep the color generation as a script/tool but for this purpose we should define a fixed-pallet of pre-defined colors that are all different enough from the background, as well as from each other.

The reason is that given the same input, we need the exact same sequence of colors as an output for this and all future versions of GoAlert. Phone apps will update at a different schedule than individual GoAlert instances, and a single phone could talk to multiple GoAlert versions.

Changing this function, or any of its dependencies could cause a change in the output, and there would then be a discrepancy between devices. To avoid that we can generate a set of colors one time (maybe 8 or something) and then select them for a given code.

mastercactapus (changes requested):

auth/authlink/store.go
Forfold: I see, I added them back in because:
1. The verify code is now rendered to the user instead of needing to be typed, and
2. The claim code is used programmatically

My thought was that we originally had them left out because of human error in keying in the code, but if there is concern with font as well we can definitely keep those characters out.

web/src/app/users/linkToMobile/LinkToMobile.tsx
Forfold: I did it mainly because prop drilling through a bunch of steps kinda bothered me to set error based on the different query/mutation results, but I can certainly think of a better tree structure for the components.

web/src/app/util/colors.ts
Forfold: This is fair, I'd definitely like to keep the tool as I have plans to expand upon it as a palette generator for further use in GoAlert for different themes, while staying a11y compatible on our terms (no 3rd party library dependency). Perhaps I can pull this tool out since it's in a different file with @jakegleason's help on the unit tests and put it in a separate draft PR that we can work on for a **_Colors_** milestone. That would be fun 🙂 

---

Are you thinking then, with a little bit toned down on the randomness, that we cycle through a set of 8 colors or something? If we did a pick on an array of pre-determined colors, there's a good chance that the colors could be the same if the codes are different.

Perhaps this is all overkill, and a simple typography display of the code is enough, since colors as a concept aren't a11y in the first place because of colorblindness. 🤷‍♀️

Lemme know your thoughts @mastercactapus, would love to hear them

mastercactapus: The colors aren't just for readability -- they are meant as an additional hint/indicator that the two codes are different.

If all colors end up being the same, but if a single character of the code changes the sequence of colors would be very different, even if two codes are a single letter apart.

Basically, the user is matching codes and/or a sequence of colors. `ABCDEF` is very similar to `ABDCEF` at first glance, but the two would have wildly different color sequences which would help prevent a user from glancing and assuming they are the same.

Our pallet for this would be fixed and shared between web and mobile (if originally generated randomly) and:
- all colors should have an acceptable contrast from all other colors in the list
- all colors should have an acceptable contrast from a white background
- all colors should have an acceptable contrast from a dark background (i.e. respect dark mode in a browser and/or mobile)
- should _probably_ have at least the number of colors match the number of characters we're thinking for the code

mastercactapus:
Closing until we cycle back to mobile.

Issue #439, opened 2020-02-10
ui: clamp max values in numeric input values


Forfold:
**Is your feature request related to a problem? Please describe:**
Currently, most numeric input fields allow any value. When very large values are entered we are unable to provide a proper error message because it actually manifests as a syntax error while parsing the request (due to integer size limits vs. javascript using floats).

![image-2019-07-30-09-37-37-317](https://user-images.githubusercontent.com/11381794/74183505-b0558500-4c0a-11ea-9ba4-c150c108b132.png)

**Describe the solution you'd like:**
All numeric input fields have a max value enforced in the UI that prevents the parse error

PR #863, opened 2020-09-08
ui: clamp max values for numeric inputs

pnengchu:

**Description:**
In this PR I created the file NormalizeNumbers.js to handle the clamping of values. The function normalizeNumbers() handles the check to verify that a numeric input is within a specified range which prevents the parse int error in the ui. It is instantiated where type='number' and where mapOnChangeValue is present in the FormField component. 

Resolves : https://github.com/target/goalert/issues/439


mastercactapus:
To make this app-wide I think we should take a moment to plan out how this could work in a more general way.

Maybe we can make this the default behavior in `FormField` instead of importing it everywhere and setting values -- we could also use the attributes `min` and `max` to define that rather than needing to know/understand the `mapOnChangeValue`. min/max are already part of the DOM API so it would come with general web knowledge.

We can default them to `1` and `9000` for `type=number` for consistency in our app and then override only in the few places like escalation policy repeat (which should have a min and max value of 0 and 5, respectively).

Then most of the files would be unchanged and "just work" as well as any new future code.

mastercactapus (changes requested):
In testing this, it seems some of the forms break with defaults set -- as I'm looking through a lot of the fields in the app, it seems we have quite a few different requirements, and one where we probably don't want to enforce min/max at all.

So may need to re-think the defaults, and just define min/max in the places missing it instead. Searching for `type='number` I found the following places:

- AdminFieldComponents.tsx, the numeric input should have a min of zero (max 9000)
- PolicyStepForm.js delayMInutes needs a min of 1 (max 9000)
- RotationForm.js needs a min of 1 (max 9000)
- ScheduleShiftList -- already set, no changes
- HeartbeatMonitorForm -- already set, no changes
- UserContactMethodVerification -- should probably not have enforcement since it's a code input, we want type=number for accessibility, but otherwise the code is treated as a string rather than a min/max number.
- UserNotificationRuleForm needs a min of 0 (max 9000)
- WizardForm delayMinutes needs a min of 1 (max 9000)

Only required change for this would be removing `min/max` from `defaultProps` and then setting `min/max` in the relevant fields.

mastercactapus (changes requested):
RotationForm and WizardForm are each missing min/max on a field that needs it (see previous comment)

mastercactapus (dismissed):

web/src/app/wizard/WizardForm.js
```diff
@@ -137,7 +137,8 @@ export default class WizardForm extends React.PureComponent {
               required
               type='number'
               placeholder='15'
-              mapOnChangeValue={(value) => value.toString()}
+              min={1}
```
Forfold, line 140: I believe this can be 0, which means that it will cause the step to immediately escalate

web/src/app/escalation-policies/PolicyStepForm.js
```diff
@@ -265,6 +265,8 @@ export default class PolicyStepForm extends React.Component {
               required
               type='number'
               mapOnChangeValue={(value) => value.toString()}
+              min={1}
```
Forfold, line 268: Same thing: I believe this can be 0, which means that it will cause the step to immediately escalate

Forfold (changes requested):

web/src/app/wizard/WizardForm.js
pnengchu: addressed, changes made as requested.

web/src/app/escalation-policies/PolicyStepForm.js
pnengchu: addressed, changes made as requested.

Forfold (dismissed):

mastercactapus:
@Forfold @pnengchu Even though the UI code has a provision for `delayMinutes` on a step as zero, it's not actually allowed:
https://github.com/target/goalert/blob/master/escalation/step.go#L34

The behavior would be undefined if it were allowed, as some actions on the step could get notified and some may not -- notifications as well as escalations aren't guaranteed to happen on the next cycle, depending on load, so a step that immediately times out wouldn't have a guarantee of completing before being invalidated.

It also wouldn't be clear if the step should ideally be skipped or appended to the second one. So for all that it was decided to make the minimum 1, I think the UI just never had that code removed.

Forfold (dismissed):

mastercactapus (approved):

Forfold (approved):

PR #865, opened 2020-09-09
profile: conflicting contact method

Quynhd07:

**Description:**
When adding a new contact method, users will be able to check if there is a conflicting contact method. The error links to the existing user's profile.


web/src/app/forms/FormContainer.js
```diff
@@ -10,12 +10,15 @@ import { get, set } from 'lodash-es'
...
+
+    errors: p.oneOfType([
       p.shape({
-        field: p.string.isRequired,
         message: p.string.isRequired,
+        field: p.string.isRequired,
       }),
-    ),
+      p.node,
+    ]),
```
mastercactapus, line 20: Instead of overriding all of the error processing with a React node, maybe we can enhance it to support a link.

We should continue to display the message provided by the backend and attach it to the proper field, but we can add the link to the end.

> contact method already exists for that type and value [Admin McAdminFace](https://example.org)

```
p.shape({
  message: p.string.isRequired,
  field: p.string.isRequired,
  helpLink: p.shape({text: p.string.isRequired, to: p.string.isRequired})
})
```

web/src/app/users/UserContactMethodCreateDialog.js
```diff
@@ -24,6 +40,45 @@ export default function UserContactMethodCreateDialog(props) {
...
+      input: {
+        cmValue: cmValue.value,
+        cmType: cmValue.type,
+      },
+    },
+  })
+
+  const [userConflict, setUserConflict] = useURLParam('userConflict', '')
+
+  const resetUserConflict = useResetURLParams('userConflict')
```
mastercactapus, line 56: Don't use query params for this, use the error as the source of the query, then we don't need to track redundant data through state & params.

mastercactapus (changes requested):

web/src/app/forms/FormContainer.js
Forfold: On top of this, it would be helpful to append the link icon at the end of the field error so indicate it's different (and action can be taken on the text) to the user

mastercactapus (changes requested):
Just a couple small code things, I'll try to get a functional review/test this afternoon too

web/src/app/forms/FormField.js
```diff
@@ -186,14 +186,36 @@ export class FormField extends React.PureComponent {
...
+          this.renderFormHelperText(props.error, props.hint)}
       </FormControl>
     )
   }
+
+  renderFormHelperText(error, hint) {
+    if (error.helpLink) {
+      return (
+        <FormHelperText>
+          <AppLink to={error.helpLink}>
```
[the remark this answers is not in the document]
mastercactapus: ^ mostly because we're in a dialog, so if they click on the link, the back button won't take them back to their place

user/search.go
```diff
@@ -79,6 +101,8 @@ func (opts renderData) QueryArgs() []sql.NamedArg {
 		sql.Named("search", opts.SearchStr()),
 		sql.Named("afterName", opts.After.Name),
 		sql.Named("omit", sqlutil.UUIDArray(opts.Omit)),
+		sql.Named("CMValue", opts.CMValue),
```
arurao, line 104: In order to be consistent with naming conventions, should we change these to `cmvalue` and `cmtype`?

user/search.go
```diff
@@ -70,6 +83,15 @@ func (opts renderData) Normalize() (*renderData, error) {
 	if opts.After.Name != "" {
 		err = validate.Many(err, validate.Name("After.Name", opts.After.Name))
 	}
+	if opts.CMValue != "" {
+		err = validate.Many(err, validate.Phone("CMValue", opts.CMValue))
```
arurao, line 87: Should we change the variables to be `cmvalue` and `cmtype` below?

user/search.go
```diff
@@ -31,18 +35,27 @@ type SearchCursor struct {
...
+		AND not usr.id = any(:omit)
 	{{end}}
 	{{if .SearchStr}}
 		AND usr.name ILIKE :search
 	{{end}}
 	{{if .After.Name}}
 		AND lower(usr.name) > lower(:afterName)
 	{{end}}
+	{{ if .CMValue }}
+		AND ucm.value = :CMValue
```
arurao, line 54: as per the two comments below

arurao (changes requested):

user/search.go
Quynhd07: Caza suggested the change in one of the comments above. Since C and M are separate words, they should be upper case. I changed all the references to keep it consistent.

arurao: Ahh I see! That makes sense. We can keep them as it is.

web/src/cypress/integration/profile.ts
```diff
@@ -99,6 +99,26 @@ function testProfile(): void {
       cy.get('body').should('contain', `${name} (${type})`)
     })
 
+    it('should return error with link to conflicting user', () => {
```
Forfold, line 102: One thing we can do that I didn't think of while we were pairing is to use the 'users' fixture instead of the 'profile' fixture to create the first contact method. This would allow us to verify that clicking the link actually works (if we were to test clicking the link as is nothing would happen but the dialog closing since it's the same profile).

it would look something along the lines of:

```
cy.fixture('users').then(users => {
  cy.addContactMethod({ userID: users[0].id }).then(
    ...
    ...
    cy.dialog().contains('Contact method already exists for that type and value: ' + users[0].name).click()
    cy.get('[data-cy="details-heading"]').should('contain', users[0].name)
  })
})
```

arurao (approved):

user/search.go
```diff
@@ -3,14 +3,14 @@ package user
...
+	"github.com/pkg/errors"
 	"github.com/target/goalert/permission"
 	"github.com/target/goalert/search"
+	"github.com/target/goalert/user/contactmethod"
+	"github.com/target/goalert/util/sqlutil"
+	"github.com/target/goalert/validation"
 	"github.com/target/goalert/validation/validate"
-
-	"github.com/pkg/errors"
+	"text/template"
```
mastercactapus, line 13: Run goimports

user/search.go
```diff
@@ -22,6 +22,10 @@ type SearchOptions struct {
 	Omit []string `json:"o,omitempty"`
 
 	Limit int `json:"-"`
+	//CMValue is matched against the user's contact method phone number.
```
mastercactapus, line 25: Blank line before the comment line, and space before the text
```suggestion
	Limit int `json:"-"`

	// CMValue is matched against the user's contact method phone number.
```

web/src/app/forms/FormContainer.js
```diff
@@ -10,12 +10,15 @@ import { get, set } from 'lodash-es'
 export class FormContainer extends React.PureComponent {
   static propTypes = {
     value: p.object,
+
     errors: p.arrayOf(
       p.shape({
-        field: p.string.isRequired,
-        message: p.string.isRequired,
+        message: p.string,
+        field: p.string,
```
mastercactapus, line 17: These are still required

web/src/app/users/UserContactMethodCreateDialog.js
```diff
@@ -60,8 +100,7 @@ export default function UserContactMethodCreateDialog(props) {
       loading={loading}
       errors={nonFieldErrors(error)}
       onClose={props.onClose}
-      // wrapped to prevent event from passing into createCM
-      onSubmit={() => createCM()}
+      onSubmit={createCM}
```
mastercactapus, line 103: This needs to be wrapped so we don't pass the event as an argument to `createCM`, keep the comment

mastercactapus (changes requested):

web/src/app/users/UserContactMethodCreateDialog.js
```diff
@@ -40,18 +64,34 @@ export default function UserContactMethodCreateDialog(props) {
   })
 
   const { loading, error } = createCMStatus
-  const fieldErrs = fieldErrors(error)
   const { title = 'Create New Contact Method', subtitle } = props
 
+  let fieldErrs = fieldErrors(error)
+  if (!queryLoading && data?.users?.nodes.length > 0) {
```
mastercactapus, line 70: Missing `?` after nodes

```suggestion
  if (!queryLoading && data?.users?.nodes?.length > 0) {
```

web/src/cypress/integration/profile.ts
```diff
@@ -99,6 +99,27 @@ function testProfile(): void {
...
+              name: c.word({ length: 8 }),
+              type: contactMethod.type,
+              value: contactMethod.value,
+            })
+            cy.dialogClick('Submit')
+            cy.dialogContains(
+              'Contact method already exists for that type and value: ' +
+                prof.name,
+            ).click()
+          },
```
mastercactapus, line 118: We should assert that it takes the user to the correct page, or at least assert on the link address (instead of clicking)

mastercactapus (changes requested):
Assuming nothing has changed with dependencies, can we revert the changes to `go.mod` and `go.sum`?

arurao (approved):
LGTM

arurao (approved):

web/src/app/forms/FormField.js
```diff
@@ -186,14 +186,36 @@ export class FormField extends React.PureComponent {
...
+      return (
+        <FormHelperText>
+          <AppLink to={error.helpLink} newTab data-cy='error-help-link'>
+            {error.message.replace(/^./, (str) => str.toUpperCase())}
+          </AppLink>
+        </FormHelperText>
+      )
+    }
+
+    if (error.message) {
```
mastercactapus, line 207: Same

mastercactapus (changes requested):

web/src/app/forms/FormField.js
```diff
@@ -186,14 +186,36 @@ export class FormField extends React.PureComponent {
...
-        {!noError && (props.error || props.hint) && (
-          <FormHelperText>
-            {(props.error &&
-              props.error.message.replace(/^./, (str) => str.toUpperCase())) ||
-              props.hint}
-          </FormHelperText>
-        )}
+        {!noError &&
+          (props.error || props.hint) &&
+          this.renderFormHelperText(props.error, props.hint)}
```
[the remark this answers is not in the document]
mastercactapus: Would this not work?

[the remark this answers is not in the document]
mastercactapus: Ah, I see, unfortunate naming. Remember to leave a note for the reviewer if a change request has questions or couldn't be made for some reason -- otherwise it seems like it may have been missed.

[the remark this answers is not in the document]
Quynhd07: gotcha!

mastercactapus (approved):

arurao (approved):

PR #872, opened 2020-09-17
Cypress: update to 5.2.0

alejandra-martinez:

**Description:**
This PR updates Cypress to 5.2.0, which adds built-in test retries.

Brings Cypress up to date with current version.


web/src/cypress/support/index.ts
```diff
@@ -36,9 +34,9 @@ import './sql'
 import './form'
 import './dialog'
 
-Cypress.env('RETRIES', 2)
+Cypress.env('retries', 1)
```
mastercactapus, line 37: Was it intentional to change retries to 1, down from 2?

web/src/cypress/tsconfig.json
```diff
@@ -1,5 +1,8 @@
 {
   "extends": "../app/tsconfig",
   "include": ["**/*"],
-  "types": ["./types/*", "../schema.d.ts"]
+  "types": ["./types/*", "../schema.d.ts"],
+  "compilerOptions": {
+    "esModuleInterop": true
+  }
```
mastercactapus, line 7: This is already set in `../app/tsconfig` so we can remove it here

mastercactapus (changes requested):

web/src/cypress/support/index.ts
alejandra-martinez: yep! thats what the migration doc says. https://docs.cypress.io/guides/references/migration-guide.html#Configure-test-retries-in-the-configuration-file

mastercactapus: It looks like their example is just misleading, we want 3 attempts and we'd still need to set retries to 2 in order to get that behavior -- however it looks like we can delete this line and instead add `"retries": 2` to `cypress.json` based on that doc you linked.

alejandra-martinez: gotcha, ok, all fixed!

mastercactapus (approved):

m17ch (approved):
LGTM 👍

PR #877, opened 2020-09-22
feature/temp-schedules: fix bugs with invalid user IDs

mastercactapus:

**Description:**
This PR adds a cached list of valid user IDs for validating fixed shift data. The IDs are currently refreshed once per minute.

It uses the groupcache package, meaning future work can allow a single call for user IDs within an entire cluster (rather than per instance).

It also fixes an issue where invalid user IDs within fixed shift data would cause the engine to abort due to the on-call insert failing.

### DIFF
--- a/engine/cleanupmanager/update.go
+++ b/engine/cleanupmanager/update.go
@@ -98,7 +98,7 @@ func (db *DB) update(ctx context.Context) error {
 	lookup := lookupMap(currentUsers)
 	for _, dat := range m {
 		cleanupScheduleData(&dat.Data, lookup, now)
-		rawData, err := json.Marshal(dat)
+		rawData, err := json.Marshal(dat.Data)
 		if err != nil {
 			return err
 		}
--- a/schedule/storefixedshifts.go
+++ b/schedule/storefixedshifts.go
@@ -13,15 +13,24 @@ import (
 	"github.com/target/goalert/validation/validate"
 )
 
-func validateShifts(fname string, max int, shifts []FixedShift) error {
+func (store *Store) validateShifts(ctx context.Context, fname string, max int, shifts []FixedShift) error {
 	if len(shifts) > max {
 		return validation.NewFieldError(fname, "too many shifts defined")
 	}
+
+	check, err := store.usr.UserExists(ctx)
+	if err != nil {
+		return err
+	}
+	defer check.Done()
 
 	for i, s := range shifts {
 		err := validate.UUID(fmt.Sprintf("%s[%d].UserID", fname, i), s.UserID)
 		if err != nil {
 			return err
+		}
+		if !check.UserExistsString(s.UserID) {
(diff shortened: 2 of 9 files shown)


m17ch (dismissed):
👍

mastercactapus:
note: the cypress tests are failing on the base feature branch -- will be addressed in a separate PR

schedule/storefixedshifts.go
```diff
@@ -64,6 +73,24 @@ func (store *Store) FixedShiftGroups(ctx context.Context, tx *sql.Tx, scheduleID
 		}
 	}
 
+	check, err := store.usr.UserExists(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer check.Done()
+
+	for i, tmp := range data.V1.TemporarySchedules {
```
Forfold, line 82: nit: comment that we are removing users that no longer exist here for future maintenance

user/store.go
```diff
@@ -129,7 +148,16 @@ func NewDB(ctx context.Context, db *sql.DB) (*DB, error) {
 				provider_id = $2 AND 
 				subject_id = $3
 		`),
-	}, p.Err
+	}
+	if p.Err != nil {
+		return nil, p.Err
+	}
+
+	store.userExist <- make(map[uuid.UUID]struct{}, 5000)
```
Forfold, line 156: would it be excessive to allocate enough capacity with some variation of `select count(*) users`?

engine/cleanupmanager/update.go
```diff
@@ -98,7 +98,7 @@ func (db *DB) update(ctx context.Context) error {
 	lookup := lookupMap(currentUsers)
 	for _, dat := range m {
 		cleanupScheduleData(&dat.Data, lookup, now)
-		rawData, err := json.Marshal(dat)
+		rawData, err := json.Marshal(dat.Data)
```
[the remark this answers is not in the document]
mastercactapus: No, because `m` is of type `[]schedData` which is a slice of structs (non-pointers), if it was `[]*schedData` then it would be possible that `dat` was `nil` and cause a panic, but there's no possibility of that here.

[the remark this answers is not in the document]
Forfold: Gotcha, thanks for the clarification :)

user/store.go
mastercactapus: It wouldn't be worth the lookup/roundtrip to the DB at startup just for that. The map will auto-expand if it goes beyond this value though. 5k will only use up approx. 80KiB.

Honestly though, it might not be worth pre-allocating in the first place -- it's arbitrary, doesn't have much performance impact (one-time deal), and thus might just cause confusion/clutter. Will remove.

Forfold (approved):
lgtm!

m17ch (approved):

Issue #867, opened 2020-09-15
User session cleanup


mastercactapus:
**Is your feature request related to a problem? Please describe:**
User session tokens expire after 30 days, however the sessions remain valid indefinitely.

**Describe the solution you'd like:**
Now that we track last access time, we should introduce the following changes to address this:
- A migration to set `NOT NULL` on the last access column, setting the default value to `now()`
- An addition to the cleanup manager of the engine to delete session rows with last access > 30 days

**Describe alternatives you've considered:**
There's not a great way without direct/manual DB access/cleanup

PR #878, opened 2020-09-22
engine/cleanupmanager: user session cleanup

pnengchu:

**Description:**
User sessions greater than 30 days is removed from the database and a migration sets the last access time column to not null and default value to now(). 

Fixes https://github.com/target/goalert/issues/867


migrate/migrations/20200922140909-session-last-access-time.sql
```diff
@@ -0,0 +1,9 @@
+-- +migrate Up
+ALTER TABLE auth_user_sessions
+    ALTER COLUMN last_access_at SET NOT NULL,
+    ALTER COLUMN last_access_at SET DEFAULT now();
```
mastercactapus, line 4: `DEFAULT` needs to be set before `NOT NULL`. Production tables will have null values causing this migration to fail otherwise.

For migrate down, keep the order you have (drop not null, then drop default).

mastercactapus (changes requested):

migrate/migrations/20200922140909-session-last-access-time.sql
pnengchu: addressed, changes made as requested.

mastercactapus (dismissed):

mastercactapus (approved):

m17ch (approved):
LGTM 👍

PR #886, opened 2020-09-24
setFixedShifts: validate time ranges strictly

mastercactapus:

**Description:**
This PR updates the validation on the setFixedShifts mutation to require time ranges and values to be correct. Previously all values would be merged together and the resulting output would be cleaned/normalized and then used.

This change ensures that UI bugs or invalid API calls return with an error rather than accepted as best-effort for this mutation.


m17ch (approved):
LGTM 👍

arurao (approved):

PR #890, opened 2020-09-25
engine/messages: track first-alert by destination type

mastercactapus:

**Description:**
This PR updates the priority message sorting to track first-alert status per-destination-type. This is intended to prevent issues where contact methods with high throughput, but potentially low visibility (e.g. email) from de-prioritizing someone getting say a phone call or text before other services are notified using those methods.

**Additional Context**
GoAlert will prioritize the first alert notification to a service above all others, if the service has had no outgoing notifications within the last 2 minutes. This is intended to "guarantee" that at least one on-call user of a service will receive at least one notification for at least one alert when/if the system is at total capacity for sending messages.

As an example, SMS and VOICE calls are allowed by Twilio at a rate of 1 per second. When there are more notifications pending than this limit, prioritization is used to decide which notifications are the most critical to keep things flowing.

With this change the first-alert priority will be segmented per type, meaning an on-call user for a service should receive at least 1 SMS message, before another service gets a second SMS -- even if the first service had a user notified by phone call, email, Slack, etc..

### DIFF
--- a/engine/message/queue.go
+++ b/engine/message/queue.go
@@ -28,11 +28,17 @@ type queue struct {
 	pending map[notification.DestType][]Message
 	now     time.Time
 
+	firstAlert  map[destID]struct{}
 	serviceSent map[string]time.Time
 	userSent    map[string]time.Time
 	destSent    map[notification.Dest]time.Time
 
 	mx sync.Mutex
+}
+
+type destID struct {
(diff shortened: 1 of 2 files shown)


Forfold (approved):
lgtm!

arurao (approved):
LGTM

PR #891, opened 2020-09-25
alerts: add toggle to alerts list page for full timestamps

Forfold:

**Description:**
Adds a toggle to the alerts list filter to switch between displaying the duration since created or the timestamp when created.

**Additional changes:**
- Fixes text overflow into timestamp on alerts list
- Fixes text overflow into timestamp on alerts event log list

**Screenshots:**


web/src/app/actions/alerts.js
```diff
@@ -22,6 +22,11 @@ export function setAlertsAllServicesFilter(bool) {
   return setURLParam('allServices', bool ? '1' : null)
 }
 
+// setAlertsAllServicesFilter will set the alert list to include all services.
+export function setAlertsShowAsTimestampsFilter(bool) {
+  return setURLParam('timestamps', bool ? '1' : null)
```
mastercactapus, line 27: maybe `fullTime` or `absTime`? timestamps is a bit ambiguous

web/src/app/alerts/components/AlertsListFilter.js
```diff
@@ -113,6 +134,17 @@ export default class AlertsListFilter extends Component {
         <Grid item xs={12} className={classes.gridItem}>
           <FormControl>
             {favoritesFilter}
+            <FormControlLabel
+              control={
+                <Switch
+                  aria-label='Show full timestamps toggle'
+                  data-cy='toggle-timestamps'
```
mastercactapus, line 141: Give this a name and drop the `data-cy`

web/src/app/util/timeFormat.js
```diff
@@ -1,5 +1,15 @@
...
+// fortmateTimeLocale returns a formatted time stamp from a given ISO string
+// @params: iso: string, type: 'short' | 'full'
+// in a full, medium, or short locale. Defaults to medium.
+export function formatTimeLocale(iso, type) {
+  let locale = DateTime.DATETIME_MED
+  if (type === 'short') locale = DateTime.DATETIME_SHORT
+  if (type === 'full') locale = DateTime.DATETIME_FULL
+  return DateTime.fromISO(iso).toLocaleString(locale)
+}
+
```
mastercactapus, line 12: I think breaking this out is unnecessary indirection, it's just mapping arbitrary strings to the already defined constants in `DateTime` but doing no work itself. So instead of knowing to use `DATETIME_FULL` the caller has to read this file and know to pass `full` rather than just reading the luxon documentation.

Just call `DateTime.fromISO(iso).toLocaleString(DateTime.DATETIME_FULL)` from the site -- a little copying is better than a little dependency. If we were doing alterations or something it would warrant it's own function, but since this is a pass-through it doesn't give us much.

mastercactapus (changes requested):
Did you consider other placement options? We put "Show in <timezone>" in the filter box on schedules, but on the alert event log we put it above the timestamps.

This is probably fine, but I'd be curious on others' thoughts.

web/src/app/util/timeFormat.js
Forfold: Gotcha, sounds good — My main reasoning was for maintenance. If down the road we change the timestamps in one spot we would likely want to do the same in other places, based on our discussions of consistent timestamp display

web/src/app/actions/alerts.js
Forfold: Sounds good, I'll do fullTime (or whatever `DATETIME` format we land on) as the URL param

mastercactapus:
@Forfold sounds reasonable to me

web/src/app/util/timeFormat.js
mastercactapus: The issue is we're just swapping the constants `DateTime.DATETIME_SHORT` with `'short'` the format is already dictated by the locale, not by us, which is why I don't think the indirection is useful.

If we were writing our own format strings, I'd agree, but since we're using locale I think it's better to rely on luxon documentation than both ours and luxon.

As for maintenance, readability is going to help even if writing it takes slightly longer the first time. The luxon call tells you the expected format and you can look directly at their documentation without opening another file to find out we're using FULL and SHORT
`DateTime.fromISO(iso).toLocaleString(full ? DateTime.DATETIME_FULL : DateTime.DATETIME_SHORT)`

Forfold: You make a good point: it's a lot easier to use the luxon API directly than searching our utils for something that may or may not be there. We'll just have to stay on top of ourselves if/when we update timestamp formats in the future.

dctalbot (approved):

PR #910, opened 2020-10-05
ui: fix RequireConfig regression

Forfold:

**Description:**
This PR adds additional type checks within `RequireConfig` to ensure the value is being assessed properly. Fixes an issue where integration keys using Mailgun will always display as disabled, even if Mailgun is enabled.


web/src/app/util/RequireConfig.js
```diff
@@ -126,14 +128,18 @@ export default function RequireConfig(props) {
   if (wantIsAdmin && !isAdmin) {
     return elseValue
   }
-  if (configID && !test(config[configID])) {
+
+  const cfgOpt = config.find((c) => c.id === configID)
+  if (configID && !test(cfgOpt.value)) {
```
mastercactapus, line 133: null check:
```suggestion
  if (configID && !test(cfgOpt?.value)) {
```

web/src/app/util/RequireConfig.js
```diff
@@ -126,14 +128,18 @@ export default function RequireConfig(props) {
...
     return elseValue
   }
 
+  if (!children) return null
   return React.Children.map(children, (child) =>
     React.cloneElement(child, _.omit(rest, Object.keys(child.props))),
   )
 }
+
```
mastercactapus, line 142: The falsy check is redundant: `React.Children.map` already handles this internally.

https://reactjs.org/docs/react-api.html#reactchildrenmap
```suggestion
  return React.Children.map(children, (child) =>
    React.cloneElement(child, _.omit(rest, Object.keys(child.props))),
  )
}
```

web/src/app/util/RequireConfig.js
```diff
@@ -60,7 +60,9 @@ function parseValue(type, value) {
 
 function isTrue(value) {
   if (Array.isArray(value)) return value.length > 0
-
+  if (value === 'true' || value === 'false') {
+    return value === 'true'
+  }
```
mastercactapus, line 65: This is just a nit, but the only exception we really need to code for is `'false'`

`if (value === 'false') return false`

mastercactapus (changes requested):

web/src/app/util/RequireConfig.js
Forfold: That makes sense, but I was getting an error when children wasn't given to the component because children was undefined or something to that extent, iirc (the proptypes for `RequireConfig` do not have children as required)

mastercactapus: The prop type may need to be adjusted -- otherwise there's a PR (though stale) to convert RequireConfig to typescript which should resolve any remaining type complaints in the console when it gets merged.

arurao (approved):
LGTM

web/src/app/util/RequireConfig.js
Forfold: Addressed by using `<React.Fragment>{children}</React.Fragment>` as in the `convert to typescript` PR for this file. Seems like a sound approach, since the parent will have access to all props, there's no point to pass the props to `RequireConfig` and have it spread when it clones if the props can be passed to that child much more transparently from the origin component.

ie:

```
const { foo, bar } = props

return (
  <RequireConfig>
    <SomeComponent foo={foo} bar={bar} />
  </RequireConfig>
)
```

Forfold:
@dctalbot 
> I'd say let's get #802 merged first

I think it's more important to get this PR merged first as #802 likely won't get in for the next milestone, and this is a bug we see in current production

dctalbot (approved):

arurao (approved):

mastercactapus (approved):

PR #925, opened 2020-10-07
material select: fix onBlur handler

### LINKED ISSUE #920: material select: input not cleared on blur (regression)
**Describe the Bug:**
When a material select field loses focus the filter input should be blank (or the selected value for single-selects).

**Steps to Reproduce:**
1. Go to Services page
2. Click add `+` button
3. Type a search string like "foo" in the Escalation Policy field
4. Click on a different field
5. Observe the search string persists

**Expected Behavior:**
The search string should clear when the field loses focus

**Observed Behavior:**
The search string persists when the field loses focus

**Application Version:**
Output of `goalert version` and/or version information from view-source on the UI.

```
$ goalert version
Version:   v0.25.0-424-ga808c9e7
GitCommit: a808c9e7682516350be1ea4d4b7badc66d98dfda (clean)
BuildDate: 2020-10-07T07:59:00-05:00
GoVersion: go1.15.2 (gc)
Platform:  linux/amd64
Migration: session-last-access-time (#200)
```

mastercactapus:

**Description:**
This PR addresses a recent regression where the onBlur handler for material selects no longer worked properly.

Fixes #920

### DIFF
--- a/web/src/app/selection/MaterialSelect.tsx
+++ b/web/src/app/selection/MaterialSelect.tsx
@@ -151,7 +151,7 @@ export default function MaterialSelect(
           setInputValue('')
         }
       }}
-      onBlur={getLabel}
+      onBlur={() => setInputValue(multiple || !value ? '' : getLabel())}
       loading={isLoading}
       getOptionLabel={(option) => option?.label ?? ''}
       options={options}
--- a/web/src/app/selection/MaterialSelect.tsx
+++ b/web/src/app/selection/MaterialSelect.tsx
@@ -85,21 +85,30 @@ export default function MaterialSelect(
     name,
     noOptionsText,
     onChange,
-    onInputChange,
+    onInputChange = () => {},
     options: _options,
     placeholder,
     required,
     value,
   } = props
 
-  const getLabel = (): string =>
-    Array.isArray(value) ? value[0]?.label ?? '' : value?.label ?? ''
+  // getInputLabel will return the label of the current value.
+  //
+  // If in multi-select mode an empty string is always returned as selected values
+  // are never preserved in the input field (they are chips instead).
+  const getInputLabel = (): string =>
+    multiple || Array.isArray(value) ? '' : value?.label || ''
 
-  const [inputValue, setInputValue] = useState(multiple ? '' : getLabel())
+  const [inputValue, _setInputValue] = useState(getInputLabel())
+
+  const setInputValue = (input: string) => {
+    _setInputValue(input)
+    onInputChange(input)
+  }
   useEffect(() => {
     if (multiple) return
     if (!value) setInputValue('')
-    if (!inputValue && value) setInputValue(getLabel())
+    if (!inputValue && value) setInputValue(getInputLabel())
   }, [value, multiple])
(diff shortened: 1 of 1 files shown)


web/src/app/selection/MaterialSelect.tsx
```diff
@@ -151,7 +151,7 @@ export default function MaterialSelect(
           setInputValue('')
         }
       }}
-      onBlur={getLabel}
+      onBlur={() => setInputValue(multiple || !value ? '' : getLabel())}
```
Forfold, line 154: `getLabel` should handle returning either an empty string or the current value's label based on if it's set to multiple or not:

```
const getLabel = (): string => Array.isArray(value) ? value[0]?.label ?? '' : value?.label ?? ''
```

```suggestion
      onBlur={() => setInputValue(getLabel())}
```

Forfold (changes requested):

web/src/app/selection/MaterialSelect.tsx
mastercactapus: Ah, then the original `onBlur` handler may have been correct and `getLabel` is broken instead.

I'll adjust and add a comment to `getLabel`

Forfold: Gotcha, sounds good — That probably sounds like the case, I had originally made the `getLabel` function as a simple helper since I was using that logic in more than 1 place with the previous changes that likely caused the regression

Forfold: I think you would only need to call `setInputValue` as the net new change, as the bug looks like it was because we weren't actually doing anything onBlur except saying "hey this is the new string lol"

mastercactapus: It might be a little deeper, because there's also when to call `onInputChange` which updates the query to the server which isn't working properly either. But calling it all the time also poses an issue because we want the search string to be empty if the input == the selected value.

With what I have pushed up now, the only issue is you get "No options" on a single-select with a selected value. I *think* that's the only outstanding bit.

Going to take a break from it for a bit and then look at it again.

web/src/app/selection/MaterialSelect.tsx
```diff
@@ -85,21 +85,30 @@ export default function MaterialSelect(
...
+  //
+  // If in multi-select mode an empty string is always returned as selected values
+  // are never preserved in the input field (they are chips instead).
+  const getInputLabel = (): string =>
+    multiple || Array.isArray(value) ? '' : value?.label || ''
 
-  const [inputValue, setInputValue] = useState(multiple ? '' : getLabel())
+  const [inputValue, _setInputValue] = useState(getInputLabel())
+
+  const setInputValue = (input: string) => {
```
dctalbot, line 104:
```suggestion
  const setInputValue = (input: string): void => {
```
lint

dctalbot (changes requested):

dctalbot (approved):

Forfold (approved):

Forfold:
Created #958 to continue this work

PR #927, opened 2020-10-08
feature request: allow searching for users by phone number

### LINKED ISSUE #895: Allow searching for users by phone number
Currently we are limited to searching the list of users by username. I want to be able to filter users by phone number similarly to how search by label on services is implemented.

Add support to search by phone number via filter drop down and search text field.

Quynhd07:

**Description:**

This PR allows for user search by phone number via filter drop down and search text field. 
Close #895


web/src/app/lists/QueryList.tsx
```diff
@@ -76,6 +76,7 @@ export interface QueryListProps extends ControlledPaginatedListProps {
   // note: The `input.search` and `input.first` parameters are included by default, but can be overridden
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   variables?: any
+  mapVariables?: (vars: any) => any
```
mastercactapus, line 79: Need to add the correct type (`OperationVariables`) and add a comment for what this prop does.

web/src/app/users/UserPhoneNumberFilterContainer.js
```diff
@@ -0,0 +1,48 @@
...
+          label='Search by Phone Number'
+          helperText='Please provide your country code e.g. +1 (USA)'
+          type='tel'
+        />
+      </Grid>
+    </FilterContainer>
+  )
+}
+
+UserPhoneNumberFilterContainer.propTypes = {
```
mastercactapus, line 43: Since this is net-new let's make it tsx and use typescript types instead of propTypes

mastercactapus (changes requested):

web/src/app/users/UserPhoneNumberFilterContainer.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { useDispatch, useSelector } from 'react-redux'
+
+interface UserPhoneNumberFilterContainerProps {
+  anchorRef?: Ref<HTMLElement>
+}
+
+export default function UserPhoneNumberFilterContainer(
+  props: UserPhoneNumberFilterContainerProps,
+): JSX.Element {
+  const searchParam = String(useSelector(searchSelector))
```
mastercactapus, line 17: `searchSelector` always returns a string (default is `''`) so there's no need to cast, however, the types in `url.ts` are a bit ambiguous currently (should probably fix that) so we can just indicate the type for now
```
  const searchParam = useSelector(searchSelector) as string
```

However, a better option would be to use the hook for params directly:
```suggestion
  const [searchParam, setSearchParam] = useURLParam('search', '')
```

mastercactapus (changes requested):
Can we get a quick cypress test for the happy path on this?

arurao (dismissed):
LGTM

mastercactapus:
I think this is okay without a verification server or fuzzy searching for initial implementation -- that can be subsequent work if there's interest.

If we can get the happy path tested with existing testing tools, conflicts resolved, and comments addressed it should be good to merge.

mastercactapus (approved):

mastercactapus (approved):

arurao (approved):

PR #936, opened 2020-10-13
mailgun: address validation fix

mastercactapus:

**Description:**
This PR fixes an issue where an invalid recipient address would cause a `500` response instead of the expected `406` when a message is to be rejected for failing validation.

**Screenshots:**
image

### DIFF
--- a/smoketest/mailgunalerts_test.go
+++ b/smoketest/mailgunalerts_test.go
@@ -90,6 +90,15 @@ func TestMailgunAlerts(t *testing.T) {
 
 	h.Twilio(t).Device(h.Phone("1")).ExpectSMS("second alert")
 
+	v.Set("recipient", "w"+h.UUID("intkey")+"@"+cfg.Mailgun.EmailDomain)
+	resp, err = http.PostForm(h.URL()+"/api/v2/mailgun/incoming", v)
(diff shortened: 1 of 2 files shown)


smoketest/mailgunalerts_test.go
```diff
@@ -90,6 +90,15 @@ func TestMailgunAlerts(t *testing.T) {
 
 	h.Twilio(t).Device(h.Phone("1")).ExpectSMS("second alert")
 
+	v.Set("recipient", "w"+h.UUID("intkey")+"@"+cfg.Mailgun.EmailDomain)
```
dctalbot, line 93:
```suggestion
	v.Set("recipient", "junk"+h.UUID("intkey")+"@"+cfg.Mailgun.EmailDomain)
```
is this the idea here?

dctalbot (changes requested):

smoketest/mailgunalerts_test.go
mastercactapus: yeah, I added the example we had in production which was a "w" prefix

dctalbot (approved):

Forfold (approved):

PR #937, opened 2020-10-13
temporary schedules: out of bounds timestamp validation

### LINKED ISSUE #930: temporary schedules: display indication when a shift is out of bounds
When setting up a temporary schedule, if a shift is out-of bounds (e.g. start is before temp. schedule start and/or end is after temp. schedule end) there is no indication that there is a problem.

When submitting, the backend will reject the mutation but the error message isn't particularly helpful for a user.

The list of shifts on the second step should have some indication there is a problem.
"This shift extends beyond the start and/or end of this temporary schedule"

A help icon could have a tooltip with additional information:
"You can click the delete icon and update the times, or go back to the previous step and extend the span of this temporary schedule to correct this."

The flow is still functional but if a user gets into this situation there's no clear indication of what steps to take to correct it.

**Additional context:**
- part of #862

KatieMSB:

**Description:**
Disables add button on temp schedules when timestamp fields have out of bounds errors.

Fixes #930

### DIFF
--- a/web/src/app/schedules/temp-sched/TempSchedAddShiftForm.tsx
+++ b/web/src/app/schedules/temp-sched/TempSchedAddShiftForm.tsx
@@ -15,7 +15,15 @@ const useStyles = makeStyles({
   },
 })
 
-export default function TempSchedAddShiftForm(): JSX.Element {
+type TempSchedAddShiftFormProps = {
+  start: string
+  end: string
+}
+
+export default function TempSchedAddShiftForm({
+  start,
+  end,
+}: TempSchedAddShiftFormProps): JSX.Element {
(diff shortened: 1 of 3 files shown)


mastercactapus (changes requested):
As-is, this doesn't look to resolve issue #930 

The intent is not to prevent the user from adding out-of-bounds shifts, but to identify them and indicate it be fixed. A user may add a shift that is out of bounds, and then go to the previous step and update the start/end times appropriately, which should be an allowed flow.

Conversely, a user may add "valid" shifts, and then change the start and end times to where one or more shifts are now invalid.

In all cases the user should see an indication of what is wrong and some direction on how to correct it.

PR #938, opened 2020-10-13
fix: handle empty default values in fallback datetime picker

dctalbot:

**Description:**
Currently, the fallback non-native datetime picker component defaults to `now` when given an empty value such as an empty string.

**Screenshots:**
Screen Shot 2020-10-13 at 2 29 16 PM

### DIFF
--- a/web/src/app/util/ISOPickers.js
+++ b/web/src/app/util/ISOPickers.js
@@ -24,9 +24,7 @@ function useISOPicker(
   const params = useSelector(urlParamSelector)
   const zone = timeZone || params('tz', 'local')
   const dtValue = value ? DateTime.fromISO(value, { zone }) : null
-  const [inputValue, setInputValue] = useState(
-    value ? dtValue.toFormat(format) : '',
-  )
+  const [inputValue, setInputValue] = useState(dtValue?.toFormat(format) ?? '')
 
   // parseInput takes input from the form control and returns a DateTime
   // object representing the value, or null (if invalid or empty).
--- a/web/src/app/util/ISOPickers.js
+++ b/web/src/app/util/ISOPickers.js
@@ -91,7 +91,7 @@ function useISOPicker(
   const FallbackIcon = type === 'time' ? AccessTime : DateRange
   return (
     <Fallback
-      value={dtValue}
+      value={value ? dtValue : null}
       onChange={(v) => handleChange({ target: { value: v } })}
       showTodayButton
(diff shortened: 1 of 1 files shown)


web/src/app/util/ISOPickers.js
```diff
@@ -35,7 +35,7 @@ function useISOPicker(
     const dt = DateTime.fromFormat(input, format, { zone })
     if (dt.isValid) {
       if (type === 'time') {
-        return dtValue.set({
+        return DateTime.fromISO(value, { zone }).set({
```
mastercactapus, line 38: This won't work if `value` is null/empty.

This is the case when we have a time-only picker that starts with no value.

We can call that unsupported for now and make value required for `type="time"` instead of making assumptions here. We _could_ use the current timestamp, convert it to the zone, and then set hour & minute, but it might not work as expected -- and again we don't currently have a use-case for it.

Alternatively, add this towards the top, and then we can always do `dtValue.set` here.
```js
if (type === 'time' && !dtValue) throw new Error('initial value is required for type=time')
```

While we have a use-case for an empty date & time (or even an empty date) -- having an empty clock time is a little less meaningful and should probably be up to the parent to handle things like zone changes & DST by picking the appropriate reference date.

mastercactapus (changes requested):

mastercactapus (changes requested):
Tried setting the default override timestamps to empty and it causes a crash:
image


```
ISOPickers.js:64 Uncaught TypeError: Cannot read property 'toFormat' of null
    at ISOPickers.js:64
    at commitHookEffectListMount (react-dom.development.js:19765)
    at commitPassiveHookEffects (react-dom.development.js:19803)
    at HTMLUnknownElement.callCallback (react-dom.development.js:189)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:238)
    at invokeGuardedCallback (react-dom.development.js:293)
    at flushPassiveEffectsImpl (react-dom.development.js:22885)
    at unstable_runWithPriority (scheduler.development.js:653)
    at runWithPriority$1 (react-dom.development.js:11062)
    at flushPassiveEffects (react-dom.development.js:22852)
```

Also looks like we may need a workaround/fix for placeholder text for this to work:
image

Forfold (approved):
Functionally tested in Safari and Firefox, lgtm!

mastercactapus (approved):

PR #941, opened 2020-10-14
Refactor using generated types

dctalbot:

**Description:**

Usage: `Fixes #<issue number>`, or `Fixes (paste link of issue)`.

**Out of Scope:**

**Screenshots:**

**Additional Info:**


mastercactapus:
What problem are we attempting to solve?

Specifically, the schedule ID shouldn't be part of `value` -- it's never editable/changeable and only comes into play when the final mutation is sent, I feel like we're mixing concerns here. 

Things like `ScheduleTZFilter` or delete confirmation and anything that we may need to reference the parent schedule needs it to be a prop and not a dynamic value.

PR #944, opened 2020-10-14
temp schedules: timestamp bounds validation

### LINKED ISSUE #930: temporary schedules: display indication when a shift is out of bounds

KatieMSB:

**Description:**

- added validation for start/end date/time so end must be after start
- display all shifts in shift list, doesn't filter by start/end
- added warning icon next to out of bounds shifts
- added floating warning in bottom left for out of bounds shifts

Fixes #930

### DIFF
--- a/web/src/app/schedules/temp-sched/TempSchedDialog.tsx
+++ b/web/src/app/schedules/temp-sched/TempSchedDialog.tsx
@@ -65,6 +65,7 @@ export default function TempSchedDialog({
             key={key}
             stepText='STEP 1 OF 2'
             scheduleID={scheduleID}
+            value={value}
           />
         )
       case 1:
--- a/web/src/app/schedules/temp-sched/TempSchedDialog.tsx
+++ b/web/src/app/schedules/temp-sched/TempSchedDialog.tsx
@@ -10,7 +10,7 @@ import { bindKeyboard, virtualize } from 'react-swipeable-views-utils'
 import SwipeableViews from 'react-swipeable-views'
 import TempSchedAddShiftsStep from './TempSchedAddShiftsStep'
 import TempSchedTimesStep from './TempSchedTimesStep'
-import { isAfter, isBefore } from '../../util/luxon-helpers'
+import { parseInterval } from '../../util/shifts'
 // allows changing the index programatically
 const VirtualizeAnimatedViews = bindKeyboard(virtualize(SwipeableViews))
 
@@ -42,8 +42,9 @@ export default function TempSchedDialog({
     ),
   })
 
+  const schedInterval = parseInterval(value)
   const hasInvalidShift = value.shifts.some(
-    (s) => isAfter(value.start, s.start) || isBefore(value.end, s.end),
+    (s) => !schedInterval.engulfs(parseInterval(s)),
   )
 
   const shiftErrors = hasInvalidShift
--- a/web/src/app/schedules/temp-sched/TempSchedDialog.tsx
+++ b/web/src/app/schedules/temp-sched/TempSchedDialog.tsx
@@ -52,6 +52,7 @@ export default function TempSchedDialog({
         {
           message:
             'One or more shifts extend beyond the start and/or end of this temporary schedule',
+          nonSubmit: step !== 1,
         },
       ]
     : []
--- a/web/src/app/schedules/temp-sched/TempSchedDialog.tsx
+++ b/web/src/app/schedules/temp-sched/TempSchedDialog.tsx
@@ -10,6 +10,7 @@ import { bindKeyboard, virtualize } from 'react-swipeable-views-utils'
 import SwipeableViews from 'react-swipeable-views'
 import TempSchedAddShiftsStep from './TempSchedAddShiftsStep'
 import TempSchedTimesStep from './TempSchedTimesStep'
+import { isAfter, isBefore } from '../../util/luxon-helpers'
 // allows changing the index programatically
 const VirtualizeAnimatedViews = bindKeyboard(virtualize(SwipeableViews))
 
@@ -40,6 +41,19 @@ export default function TempSchedDialog({
       _.pick(s, 'start', 'end', 'userID'),
     ),
   })
+
+  const hasInvalidShift = value.shifts.some(
+    (s) => isAfter(value.start, s.start) || isBefore(value.end, s.end),
+  )
+
+  const shiftErrors = hasInvalidShift
+    ? [
+        {
+          message:
+            'One or more shifts extend beyond the start and/or end of this temporary schedule',
+        },
+      ]
+    : []
 
   const [submit, { loading, error, data }] = useMutation(mutation, {
     onCompleted: () => onClose(),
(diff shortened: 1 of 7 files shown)


m17ch:
Looks great, @KatieMSB & @dctalbot  - the only thing that came up in my functional review was the `Shift Duration` field highlighting red/error (see screenshot below). 

image

To reproduce:
1. Create temporary schedule for 24h window
2. Fill with 3x 8hr shifts
3. After third shift is populated, you should be able to reproduce screenshot

The temp sched CAN be created (it does not prevent submission), but just something strange I noticed.

Thoughts?

m17ch (changes requested):
Left a note in conversation - let me know your thoughts or if you need more details.

KatieMSB:
> Left a note in conversation - let me know your thoughts or if you need more details.

@m17ch Yea we talked about this briefly when we ran into it and decided to keep it for now, and the more I think about it, it might make sense to keep it. I was thinking that if a user is adding shifts and see the warning then they could make changes before having to add and delete a shift because it went out of bounds. Similarly they can add a shift that is out of bounds then go back to step 1 and change the start/end date if they're creating a new temp schedule.

m17ch (approved):
Functionally validated -- LGTM 👍

Forfold:
@KatieMSB @m17ch 
> ..see the warning then they could make changes before having to add and delete

This makes sense to me, show the warning that it is out of bounds on the FormFields, but still allow adding to the right

web/src/app/schedules/temp-sched/TempSchedDialog.tsx
```diff
@@ -41,6 +42,19 @@ export default function TempSchedDialog({
     ),
   })
 
+  const hasInvalidShift = value.shifts.some(
+    (s) => isAfter(value.start, s.start) || isBefore(value.end, s.end),
+  )
```
mastercactapus, line 47: use the Interval API here

```js
const schedInterval = parseInterval(value)
const hasInvalidShift = value.shifts.some(s => !schedInterval.engulfs(parseInterval(s)))
```

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
```diff
@@ -49,13 +65,20 @@ export default function TempSchedShiftsList({
         DateTime.fromISO(s.start, { zone }),
         DateTime.fromISO(s.end, { zone }),
       ),
+      isValid: !isAfter(start, s.start) && !isBefore(end, s.end),
```
mastercactapus, line 68: Same as above use the Interval API.

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
```diff
@@ -49,13 +65,20 @@ export default function TempSchedShiftsList({
...
     }))
 
     if (!shifts.length) return []
 
+    const scheduleDawn = DateTime.fromISO(start, { zone }).startOf('day')
+    const scheduleDusk = DateTime.fromISO(end, { zone }).endOf('day')
+
+    const firstShiftDawn = shifts[0].start.startOf('day')
+    const lastShiftDusk = shifts[shifts.length - 1].end.endOf('day')
```
mastercactapus, line 77: This won't give you the last shift, they are sorted by start time, not end time. We will need to reduce through the list to get the last end time.

mastercactapus, line 77: These can be `firstShiftStart` and `lastShiftEnd` -- skip the start/end of day bit and just do it once for the display span. after calling `min` and `max`

mastercactapus (changes requested):

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
```diff
@@ -65,20 +66,27 @@ export default function TempSchedShiftsList({
...
 
-    const scheduleDawn = DateTime.fromISO(start, { zone }).startOf('day')
-    const scheduleDusk = DateTime.fromISO(end, { zone }).endOf('day')
+    const lastShift = shifts.reduce(
+      (result, candidate) => (candidate.end > result.end ? candidate : result),
+      shifts[0],
+    )
+
+    const scheduleStart = DateTime.fromISO(start, { zone })
+    const scheduleEnd = DateTime.fromISO(end, { zone })
```
mastercactapus, line 82: We can just use `schedInterval.start` and `.end` below

mastercactapus (changes requested):
- when going back to step 1 the "Next" button remains as "Retry"

m17ch (dismissed):

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
```diff
@@ -49,13 +66,24 @@ export default function TempSchedShiftsList({
         DateTime.fromISO(s.start, { zone }),
         DateTime.fromISO(s.end, { zone }),
       ),
+      isValid: schedInterval.engulfs(
+        parseInterval({ start: s.start, end: s.end }),
```
mastercactapus, line 70:
```suggestion
        parseInterval(s),
```

web/src/app/schedules/temp-sched/TempSchedTimesStep.tsx
```diff
@@ -7,24 +7,33 @@ import {
...
 
 export default function TempSchedTimesStep({
   scheduleID,
   stepText,
+  value,
 }: TempSchedTimesStepProps): JSX.Element {
   const classes = useStyles()
 
+  function isValid(): Error | null {
+    if (DateTime.fromISO(value.start) < DateTime.fromISO(value.end)) return null
```
mastercactapus, line 33: Since this returns an error not a bool call it `validate` -- `is` implies a boolean value.

For the comparison we can use `isBefore` now that it's in a helper lib

web/src/app/util/luxon-helpers.js
```diff
@@ -31,3 +31,13 @@ export function getEndOfWeek(luxonDateTime = DateTime.local()) {
     })
     .endOf('day')
 }
+
+// compare two iso datetime strings
+export function isAfter(a, b) {
```
mastercactapus, line 36: Since these are now in a separate file, let's add a little more context/hinting to the names, like `isISOBefore` and `isISOAfter`.

Also, it might be better to have this in `util/shifts.ts`, since these don't actually consume or produce any luxon types. Though honestly, we don't have great organization around utility functions in the first place. :shrug: 

Lastly, let's add proper comments for these

```
// isISOAfter
// Compares two ISO timestamps, returning true if `a` occurs after `b`.
```

mastercactapus (changes requested):

mastercactapus (approved):

m17ch (approved):

PR #948, opened 2020-10-15
ui: miscellaneous usability improvements

dctalbot:

**Description:**
- fixed overflow scroll bug on shifts list
- made the manual entry toggle an anchor tag
- updated green temp shift color to a darker shade for a11y

**Additional Info:**
I found a nice tool to check for color contrast: https://snook.ca/technical/colour_contrast/colour.html#fg=FFFFFF,bg=094F13

### DIFF
--- a/web/src/app/schedules/temp-sched/TempSchedAddShiftForm.tsx
+++ b/web/src/app/schedules/temp-sched/TempSchedAddShiftForm.tsx
@@ -55,12 +55,12 @@ export default function TempSchedAddShiftForm(): JSX.Element {
             name='end'
             hint={
               <Typography
+                component='a'
                 className={classes.typography}
                 variant='caption'
-                color='textSecondary'
                 onClick={() => setManualEntry(false)}
               >
-                Configure as duration?
+                Configure as duration
               </Typography>
             }
           />
@@ -90,9 +90,9 @@ export default function TempSchedAddShiftForm(): JSX.Element {
             min={0.25}
             hint={
               <Typography
+                component='a'
                 className={classes.typography}
                 variant='caption'
-                color='textSecondary'
                 onClick={() => setManualEntry(true)}
               >
                 Configure as date/time?
--- a/web/src/app/schedules/temp-sched/TempSchedAddShiftForm.tsx
+++ b/web/src/app/schedules/temp-sched/TempSchedAddShiftForm.tsx
@@ -1,22 +1,13 @@
-import { Grid, TextField, Typography, makeStyles } from '@material-ui/core'
+import { Grid, TextField } from '@material-ui/core'
 import { DateTime } from 'luxon'
 import React, { useState } from 'react'
 import { FormField } from '../../forms'
 import { UserSelect } from '../../selection'
+import { AppLink } from '../../util/AppLink'
 import { ISODateTimePicker } from '../../util/ISOPickers'
 import { Value } from './sharedUtils'
 
-const useStyles = makeStyles({
-  typography: {
-    '&:hover': {
-      cursor: 'pointer',
-      textDecoration: 'underline',
-    },
-  },
-})
-
 export default function TempSchedAddShiftForm(): JSX.Element {
-  const classes = useStyles()
   const [manualEntry, setManualEntry] = useState(false)
 
   return (
@@ -54,14 +45,15 @@ export default function TempSchedAddShiftForm(): JSX.Element {
             label='Shift End'
             name='end'
             hint={
-              <Typography
-                className={classes.typography}
-                variant='caption'
-                color='textSecondary'
-                onClick={() => setManualEntry(false)}
+              <AppLink
+                to='#'
+                onClick={(e) => {
+                  e.preventDefault()
+                  setManualEntry(false)
+                }}
               >
-                Configure as duration?
-              </Typography>
+                Configure as duration
+              </AppLink>
             }
           />
         ) : (
@@ -89,14 +81,15 @@ export default function TempSchedAddShiftForm(): JSX.Element {
             }}
             min={0.25}
             hint={
-              <Typography
-                className={classes.typography}
-                variant='caption'
-                color='textSecondary'
-                onClick={() => setManualEntry(true)}
+              <AppLink
+                to='#'
+                onClick={(e) => {
+                  e.preventDefault()
+                  setManualEntry(true)
+                }}
               >
-                Configure as date/time?
-              </Typography>
+                Configure as date/time
+              </AppLink>
             }
           />
         )}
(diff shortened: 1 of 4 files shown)


KatieMSB (approved):

web/src/app/schedules/temp-sched/TempSchedAddShiftForm.tsx
```diff
@@ -54,14 +45,15 @@ export default function TempSchedAddShiftForm(): JSX.Element {
             label='Shift End'
             name='end'
             hint={
-              <Typography
-                className={classes.typography}
-                variant='caption'
-                color='textSecondary'
-                onClick={() => setManualEntry(false)}
+              <AppLink
```
mastercactapus, line 48: AppLink isn't for styling, it's meant for link/nav/tab things, could render it as an `<a>` tag if we want it as a "link" though (that's all AppLink is doing visually).

mastercactapus (changes requested):

web/src/app/schedules/temp-sched/TempSchedAddShiftForm.tsx
dctalbot: Yeah, I wasn't quite sure what to do here since the linter doesn't like `<a role="button">`. I saw we were using `AppLink` for some copy-to-clipboard "links" so I went ahead and grabbed it. 

alternatives:
- use `<button>` which requires a lot of styling
- use `<Typography>` which is less semantic and requires some styling

mastercactapus: Based on the link from the linter, we should use `<button>` and add the appropriate styles for accessibility. It looks like they have them a little lower on the page that can be copy/pasted for the most part.

https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md

web/src/app/schedules/temp-sched/TempSchedAddShiftForm.tsx
```diff
@@ -55,12 +55,12 @@ export default function TempSchedAddShiftForm(): JSX.Element {
...
               <Typography
+                component='a'
                 className={classes.typography}
                 variant='caption'
-                color='textSecondary'
                 onClick={() => setManualEntry(false)}
               >
-                Configure as duration?
+                Configure as duration
               </Typography>
```
mastercactapus, line 64: The linter complained for a reason, now that we know it's an accessibility issue, let's fix this properly.

mastercactapus (changes requested):

web/src/app/util/ClickableText.tsx
```diff
@@ -0,0 +1,50 @@
...
+    <button
+      onClick={(e) => {
+        e.preventDefault()
+        onClick()
+      }}
+      className={classes.button}
+    >
+      {text}
+    </button>
+  )
```
mastercactapus, line 47: We can just make the `onClick` a passthrough with other props -- that will make it usable for things like tooltips or things that require the event without needing to define them directly and any other props the caller might need (e.g. like `data-cy` or `aria-*`.

We could also use `children` instead of text, `<ClickableText>foobar</ClickableText>` but I think `<ClickableText text="foobar" />` is fine for it's intended use.

```suggestion
  const {text, ...rest} = props
  return (
    <button
      className={classes.button}
      {...rest}
    >
      {text}
    </button>
  )
```

mastercactapus (changes requested):

web/src/app/util/ClickableText.tsx
dctalbot:
```js
const {text, onClick, ...rest} = props
  return (
    <button
      className={classes.button}
     onClick={(e) => {
        e.preventDefault()
        onClick()
      }}
      {...rest}
    >
      {text}
    </button>
  )

```

How about something like this? That way we can easily require an `onClick` handler, which should be required as per the component's name, and we can always prevent the default "form submit" action from taking place

Forfold (changes requested):
It looks like the delete step no longer shows because the list item width overflows past the container width:

image

web/src/app/util/ClickableText.tsx
```diff
@@ -0,0 +1,50 @@
+import { makeStyles } from '@material-ui/core'
+import React from 'react'
+
+const karakurenai = '#cd1831'
```
Forfold, line 4: What's this name in reference to? It pulls up the SAO wiki for me 😂

dctalbot: That must be it then 😁. I just looked to see if the hex code had a name and that was the first thing I found

Forfold (approved):

m17ch (approved):
LGTM 👍

Forfold (approved):

web/src/app/util/ClickableText.tsx
```diff
@@ -0,0 +1,55 @@
...
+  text,
+  onClick,
+  ...props
+}: ClickableTextProps): JSX.Element {
+  const classes = useStyles()
+  return (
+    <button
+      onClick={(e) => {
+        e.preventDefault()
+        onClick()
```
mastercactapus, line 45: Do we need `preventDefault` with a button? If not we can remove `onClick` and leave it to the "...rest".
Otherwise, we should be passing the event to the handler.

mastercactapus (changes requested):

web/src/app/util/ClickableText.tsx
dctalbot: Strictly speaking, no, it is not required here. However, in my opinion since this thing does not resemble a button (it resembles a link), it shouldn't function as a default button would. When this thing is inside a form, the programmer has to do either:

```js
 <ClickableText
    type='button'
    text='Configure as date/time'
    onClick={() => setManualEntry(true)}
    />
```

or

```js
<ClickableText
  text='Configure as date/time'
  onClick={(e) => {
    e.preventDefault()
    setManualEntry(true)
  }}
/>
```


I feel like the component should just take care of it, but I'm open to changing it if there's a compelling argument where preventing form submit / default behavior isn't desired

web/src/app/util/ClickableText.tsx
mastercactapus: Missed this comment with my recent review, submit should only happen if `type=submit` -- and if that's set we shouldn't be overriding it. It's up to the renderer to decide if that's the behavior they want.

We have little need to require an `onClick` handler, as it's just a pass-through. It is a button, so we can probably assume there will be one, but if say the parent has the `onClick` handler instead of the button in some scenario and the button is just for visuals -- that shouldn't be a problem.

That said, if we want it explicitly in prop-types it should be defined as optional, passed a SyntheticEvent, etc.. so it follows the standard API.

---

The purpose of this component is just a link-styled button, so having it be a passthrough to a button that's pre-styled seems like it would be the minimal required API.

web/src/app/util/ClickableText.tsx
dctalbot: Actually thought of a third option: have `type="button"`  by default. This will allow full control of the element's behavior. Implementing now...

m17ch (approved):

Forfold (approved):

mastercactapus (approved):

Issue #929, opened 2020-10-12
temporary schedules: warning when creating or editing an override that overlaps a temp. schedule


mastercactapus:
**Is your feature request related to a problem? Please describe:**
As a temporary schedule takes precedence over all rules/overrides, it may be confusing for a user when creating/editing an override where it will not have any affect.

**Describe the solution you'd like:**
Display a warning notice at the top of the override dialog when the override time span overlaps one or more temporary schedules.

**Describe alternatives you've considered:**
Without this there may just be an influx of user questions/confusion.

**Additional context:**

- part of #862 
- use the list of temporary schedules from the current schedule to determine start & end times

Example of a notice:
![image](https://user-images.githubusercontent.com/595010/95765861-07b76380-0c78-11eb-871c-4339a0949b33.png)

PR #950, opened 2020-10-15
temporary schedules: warning when creating or editing an override that overlaps a temp. schedule

pnengchu:

**Description:**
When a user edits or adds a temporary schedule, a warning appears above the form when the start or end time overlaps with a temporary schedule. A new hook was added, 'userOverrideNotices' to handle the query of start and end times and returns a warning to the FormDialog component if there is an existing overlap with a temporary schedule. Other supporting files were also modified to support this new functionality. 

Fixes https://github.com/target/goalert/issues/929

**Out of Scope:**
ScheduleOverrideEditDialog.js is out of scope because it is a class component, the useOverrideNotices hook can only be implemented in a React component. 

**Screenshots:**
image

A warning banner appears in the dialog box when the start or end time of an override overlaps with a temporary schedule.

### DIFF
--- a/web/src/app/details/Notices.tsx
+++ b/web/src/app/details/Notices.tsx
@@ -31,20 +31,25 @@ const useStyles = makeStyles({
 })
 
 interface NoticesProps {
-  notices: Notice[] // checks against .length are safe as notices is required
+  notices?: Notice[]
 }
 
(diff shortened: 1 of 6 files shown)


mastercactapus (approved):

m17ch (approved):
lgtm

PR #952, opened 2020-10-19
ui: bump typescript and eslint packages

dctalbot:

**Description:**
- bump typescript to v4
- bump all eslint packages

closes #972


web/src/.eslintrc.js
```diff
@@ -47,12 +44,21 @@ module.exports = {
         ignoreRestSiblings: true,
       },
     ],
+    'no-use-before-define': 'off',
```
Forfold, line 47: I believe we left this rule on explicitly to enforce code consistency between any contributors

Forfold (changes requested):

web/src/.eslintrc.js
dctalbot: We override this rule on the following line, please see usage here: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-use-before-define.md#how-to-use

Forfold: Ah interesting, I didn't know that so thanks for the docs link 👍

Forfold (approved):

mastercactapus (approved):

PR #955, opened 2020-10-20
ui: upgrade apollo client

dctalbot:

**Description:**
Migration guide: https://www.apollographql.com/docs/react/migrating/apollo-client-3-migration/

I was able to automate a good amount of this
```sh
yarn add jscodeshift
npx jscodeshift -t ./imports.js --extensions js --ignore-pattern="./node_modules/*" --parser babylon .
npx jscodeshift -t ./imports.js --extensions ts --ignore-pattern="./node_modules/*" --parser ts .
npx jscodeshift -t ./imports.js --extensions tsx --ignore-pattern="./node_modules/*" --parser tsx .
yarn remove jscodeshift
```

- fix imports
- bundle node modules
- webpack: resolve `.mjs` files (ES modules)
- `web/src/app/apollo.js`: use read queries instead of cache redirects (deprecated) (achieves effectively the same functionality)
- `web/src/app/schedules/ScheduleShiftList.js`: convert to using apollo hooks. This is a significant rewrite and may be broken out into its own PR (#1027)
- `web/src/app/services/ServiceOnCallList.js`: due to an issue when deleting a service, we add a null check for data

Closes #819


mastercactapus (dismissed):

KatieMSB (changes requested):
Found just one catch, looks like UserCalendarSubscriptionList broke with the changes.
Screen Shot 2020-11-17 at 3 00 54 PM

KatieMSB (dismissed):

mastercactapus (changes requested):
There are still imports from the removed dependency `'graphql-tag'` that need to be updated before we can merge this:
```
$ grep "graphql-tag" -R web/src/app
web/src/app/services/ServiceLabelList.js:import gql from 'graphql-tag'
web/src/app/schedules/ScheduleRuleForm.js:import gql from 'graphql-tag'
web/src/app/schedules/ScheduleShiftList.js:import gql from 'graphql-tag'
web/src/app/schedules/ScheduleRuleList.js:import gql from 'graphql-tag'
web/src/app/schedules/ScheduleOverrideDeleteDialog.js:import gql from 'graphql-tag'
web/src/app/schedules/ScheduleCalendarQuery.js:import gql from 'graphql-tag'
web/src/app/schedules/ScheduleOverrideList.js:import gql from 'graphql-tag'
web/src/app/users/UserNotificationRuleList.js:import gql from 'graphql-tag'
web/src/app/escalation-policies/PolicyStepEditDialog.js:import gql from 'graphql-tag'
web/src/app/escalation-policies/PolicyStepsCard.js:import gql from 'graphql-tag'
web/src/app/escalation-policies/PolicyStepCreateDialog.js:import gql from 'graphql-tag'
web/src/app/main/components/ToolbarTitle.js:import gql from 'graphql-tag'
web/src/app/alerts/components/AlertDetails.js:import gql from 'graphql-tag'
web/src/app/wizard/WizardRouter.js:import gql from 'graphql-tag'
web/src/app/rotations/RotationUserList.js:import gql from 'graphql-tag'
web/src/app/util/Chips.js:import gql from 'graphql-tag'
```

web/src/app/admin/AdminLimits.tsx
```diff
@@ -1,16 +1,15 @@
+import { useQuery, gql } from '@apollo/client'
```
Forfold, line 1: One thing of note, could you have your script make it so this isn't being placed above our `React` imports on L1?

Forfold (changes requested):

web/src/app/admin/AdminLimits.tsx
dctalbot: I don't think so, at least not easily. If we want to enforce that, we should set up a linting rule in a separate PR (see https://dev.to/otamnitram/sorting-your-imports-correctly-in-react-213m)

mastercactapus: We do have an old codemod to sort imports you can use as an example of how it can be achieved inside the `codemods` directory.

It might be easier than manually moving them all

Forfold: I would say we should try and retain `React` on line 1 since that is one of the things we currently do as `eslint` says, and we can re-run our `jsimports` codemod when we restructure our folder coming up.

It looks like it happens when there's no `@apollo/client` present. Perhaps we can default to replacing one of the imports with the new name instead (i.e. `'@apollo/react-hooks'` -> `'@apollo/client'` in place).

Or maybe we can just insert on L2/after L1 instead? It doesn't necessarily have to be a pretty solution to get our codebase where we want it

dctalbot: Pretty it was not! Because ordering of imports matters, there aren't many tools that will "autofix" this sort of thing. Just ran it through:
```sh
#!/bin/bash
# for file in dir
#     if line 2 imports from react
#         swap lines 1 and 2

for filename in $(find ../app); do
    LINE=$(sed '2q;d' $filename)
    if [[ "$LINE" == *"from 'react'"* ]]; then
      printf '%s\n' 1m2 2-m1- w q | ed -s $filename
      echo $filename;
    fi
done
```

web/src/webpack.prod.config.js
```diff
@@ -20,7 +20,7 @@ module.exports = (env) => ({
     filename: 'static/[name].js',
   },
   resolve: {
-    extensions: ['.ts', '.tsx', '.js', '.jsx', '.css'],
+    extensions: ['.mjs', '.ts', '.tsx', '.js', '.jsx', '.css'],
```
Forfold, line 23: I'm not familiar with this extension `.mjs`, could you help explain why this is needed to upgrade Apollo?

Forfold (changes requested):
When deleting a notification rule, I got the following warning:

```
Cache data may be lost when replacing the notificationRules field of a User object.

To address this problem (which is not a bug in Apollo Client), define a custom merge function for the User.notificationRules field, so InMemoryCache can safely merge these objects:

  existing: [{"__ref":"UserNotificationRule:7f12b4a2-b4d0-436b-adc9-a79d4a5f7ad2"}]
  incoming: []
```

web/src/webpack.prod.config.js
dctalbot: `.mjs` is more or less equivalent to `.js` in most contexts, it's just a signal that the code is packaged in ES format (as opposed to commonjs or AMD etc.) The new Apollo client uses this file extension.

Forfold (dismissed):
Walked through the code and functionally tested several queries and mutations. Aside from the console warnings to be resolved in a future Apollo release, LGTM 👍

mastercactapus (dismissed):

Forfold (approved):

mastercactapus (approved):

PR #957, opened 2020-10-21
docs: recommended dev tools

dctalbot:

**Description:**
Add section of recommended tools for development

### DIFF
--- a/docs/development-setup.md
+++ b/docs/development-setup.md
@@ -52,3 +52,14 @@ More information about smoketests can be found [here](../smoketest/README.md).
 All unit tests can be run with `make test`.
 
 UI Unit tests are found under the directory of the file being tested, with the same file name, appended with `.test.js`. They can be run independently of the Go unit tests with `make jest`. Watch mode can be enabled with `make jest JEST_ARGS=--watch`.
+
+## Toolbox
+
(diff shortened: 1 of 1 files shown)


docs/development-setup.md
```diff
@@ -52,3 +52,14 @@ More information about smoketests can be found [here](../smoketest/README.md).
 All unit tests can be run with `make test`.
 
 UI Unit tests are found under the directory of the file being tested, with the same file name, appended with `.test.js`. They can be run independently of the Go unit tests with `make jest`. Watch mode can be enabled with `make jest JEST_ARGS=--watch`.
+
+## Toolbox
+
+There is an abundance of free/open-source tooling out there, but these are some that work well for our team on mac at the time of writing
```
Forfold, line 58:
```suggestion
There is an abundance of free/open-source tooling out there, but below are some tools that work well for our team:
```

I think we can leave out `on mac` and `at time of writing` since there will always be alternative ways to run these apps (VMs or similar), and the timestamp will show on the readme, so viewers should have a good sense if the doc gets outdated

dctalbot: Hmm good points. I also don't want the mention of mac to put off any contributors. The reason I noted that is because Postico is only available on mac, and schema command relies on Docker For Mac.

The goal here is to document some shared knowledge which I suppose we can keep internal, but seems pretty harmless to put it out there

PR #962, opened 2020-10-22
oncall: handle calculations for 0 and 1 participant rotations

### LINKED ISSUE #960: temporary schedules: can't view shifts when rotation with 0-1 participants is assigned
**Describe the Bug:**
If a rotation with 0 or 1 participant is assigned to a schedule, the page breaks. This does not seem to occur if more than 1 participant is on a rotation.

**Steps to Reproduce:**
1. Create an empty rotation (or a rotation with 1 participant)
2. Create a schedule
3. Add the rotation to the schedule
4. Attempt to visit the shifts page via the quick links section

**Additional Context:**
This appears to be a backend bug. When querying from the graphql playground, `users` is returned as `null` and `userID` is set to the rotation's ID. This causes the UI to break, as the API does not expect the user object to ever be null if shifts are provided.

**Screenshots/Stack Traces:**

_Note: The input start/end date used above was the following:_
```
shifts(start:"2020-11-20T00:00:00.000Z" end:"2020-11-30T00:00:00.000Z")
```

**Application Version:**
`feature/temp-schedules` branch

mastercactapus:

**Description:**
This PR fixes issues with the new oncall calculation code with rotations with < 2 participants

Fixes #960 

**Additional Info:**
0-length rotations caused issues because an assumption was made that `.Rotation` would be populated in all cases. However empty rotations have no entry in `rotation_state` which causes them to be omitted from the original query. The fix for this was to ensure target type is user before assuming target id is a user id.

single-user rotations caused an infinite loop because the `UserID` method would short-circuit without advancing the rotation times. The fix for this is to not iterate through the rotation when there would be no handoff.

### DIFF
--- a/oncall/singlerulecalculator.go
+++ b/oncall/singlerulecalculator.go
@@ -1,6 +1,10 @@
 package oncall
 
-import "time"
+import (
+	"time"
+
+	"github.com/target/goalert/assignment"
+)
 
 // SingleRuleCalculator will calculate the currently active user.
 type SingleRuleCalculator struct {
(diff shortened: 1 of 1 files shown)


dctalbot (dismissed):
Functionally validated 👍🏻

dctalbot (approved):

Forfold (approved):
lgtm, functionally validated and working well

PR #963, opened 2020-10-22
feature/temp schedules: set min start date to now()

Forfold:

**Description:**
This PR ensures that the start date of a temporary schedule can not begin in the past.

**Additional Info:**
- Converts `ISOPickers` to Typescript
- Adds support for `min` and `max` dates in `ISOPickers`

**Screenshots:**

### DIFF
--- a/web/src/app/forms/FormField.js
+++ b/web/src/app/forms/FormField.js
@@ -36,8 +36,9 @@ export class FormField extends React.PureComponent {
     fieldName: p.string,
 
     // min and max values specify the range to clamp a int value
-    min: p.number,
-    max: p.number,
+    // will be passed to base component for input controls
+    min: p.oneOfType([p.number, p.string]),
+    max: p.oneOfType([p.number, p.string]),
 
     // used if name is set,
     // but the error name is different from graphql responses
--- a/web/src/app/schedules/temp-sched/TempSchedAddShiftForm.tsx
+++ b/web/src/app/schedules/temp-sched/TempSchedAddShiftForm.tsx
@@ -1,6 +1,7 @@
 import { Grid, TextField, Typography, makeStyles } from '@material-ui/core'
 import { DateTime } from 'luxon'
 import React, { useState } from 'react'
+import { useURLParam } from '../../actions'
 import { FormField } from '../../forms'
 import { UserSelect } from '../../selection'
 import { ISODateTimePicker } from '../../util/ISOPickers'
@@ -18,6 +19,8 @@ const useStyles = makeStyles({
 export default function TempSchedAddShiftForm(): JSX.Element {
   const classes = useStyles()
   const [manualEntry, setManualEntry] = useState(false)
+  const [zone] = useURLParam('tz', 'local')
+  const now = DateTime.local().setZone(zone).startOf('day').toFormat("yyyy-MM-dd'T'HH:mm:ss")
 
   return (
     <React.Fragment>
@@ -35,6 +38,7 @@ export default function TempSchedAddShiftForm(): JSX.Element {
           component={ISODateTimePicker}
           label='Shift Start'
           name='start'
+          min={now}
           mapOnChangeValue={(value: string, formValue: Value) => {
             if (!manualEntry) {
               const diff = DateTime.fromISO(value).diff(
@@ -53,6 +57,7 @@ export default function TempSchedAddShiftForm(): JSX.Element {
             component={ISODateTimePicker}
             label='Shift End'
             name='end'
+            min={now}
             hint={
               <Typography
                 className={classes.typography}
--- a/web/src/app/schedules/temp-sched/TempSchedAddShiftForm.tsx
+++ b/web/src/app/schedules/temp-sched/TempSchedAddShiftForm.tsx
@@ -20,7 +20,10 @@ export default function TempSchedAddShiftForm(): JSX.Element {
   const classes = useStyles()
   const [manualEntry, setManualEntry] = useState(false)
   const [zone] = useURLParam('tz', 'local')
-  const now = DateTime.local().setZone(zone).startOf('day').toFormat("yyyy-MM-dd'T'HH:mm:ss")
+  const now = DateTime.local()
+    .setZone(zone)
+    .startOf('day')
+    .toFormat("yyyy-MM-dd'T'HH:mm:ss")
 
   return (
     <React.Fragment>
--- a/web/src/app/schedules/temp-sched/TempSchedTimesStep.tsx
+++ b/web/src/app/schedules/temp-sched/TempSchedTimesStep.tsx
@@ -1,9 +1,5 @@
 import React from 'react'
-import {
-  Grid,
-  DialogContentText,
-  Typography,
-} from '@material-ui/core'
+import { Grid, DialogContentText, Typography } from '@material-ui/core'
 import { makeStyles } from '@material-ui/core/styles'
 import { FormField } from '../../forms'
 import { ISODateTimePicker } from '../../util/ISOPickers'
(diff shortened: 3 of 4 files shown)


web/src/app/util/ISOPickers.tsx
```diff
@@ -62,51 +94,54 @@ function useISOPicker(
...
-  // starts with label above textfield so format placeholder can be seen
-  const shrinkInputLabel = (p) => ({
-    ...(p?.InputLabelProps ?? {}),
-    shrink: true,
-  })
+  const handleInputChange = (e: ChangeEvent<HTMLInputElement>): void => {
+    const val = e.target.value
+    setInputValue(val)
+    handleChange(val)
+  }
```
mastercactapus, line 110: Why the separation/indirection here? The only effective line in this function is `setInputValue(val)` we can call it from `handleChange` to make the flow easier to follow

web/src/app/util/ISOPickers.tsx
```diff
@@ -62,51 +94,54 @@ function useISOPicker(
...
       <TextField
         type={type}
         value={inputValue}
-        onChange={handleChange}
+        onChange={handleInputChange}
+        inputProps={{ min, max }}
         {...otherProps}
-        InputLabelProps={shrinkInputLabel(otherProps)}
+        InputLabelProps={{
+          ...otherProps,
```
mastercactapus, line 121: This won't work properly, we need to expand `otherProps.InputLabelProps` -- _if_ it exists.

web/src/app/util/ISOPickers.tsx
```diff
@@ -62,51 +94,54 @@ function useISOPicker(
...
 
-  const extraProps = {}
+  const cypressProps: { [key: string]: { 'data-cy'?: string } } = {}
+  cypressProps.DialogProps = { 'data-cy': 'picker-fallback' }
   if (type !== 'time') {
-    extraProps.leftArrowButtonProps = { 'data-cy': 'month-back' }
-    extraProps.rightArrowButtonProps = { 'data-cy': 'month-next' }
+    cypressProps.leftArrowButtonProps = { 'data-cy': 'month-back' }
+    cypressProps.rightArrowButtonProps = { 'data-cy': 'month-next' }
   }
```
mastercactapus, line 133: Maybe call these `testingProps`? without context it would make it more clear to a new dev that may not be familiar with cypress

web/src/app/util/ISOPickers.tsx
```diff
@@ -62,51 +94,54 @@ function useISOPicker(
...
-      onChange={(v) => handleChange({ target: { value: v } })}
+      onChange={handleChange}
       showTodayButton
-      DialogProps={{
-        'data-cy': 'picker-fallback',
-      }}
+      minDate={min}
+      maxDate={max}
       InputProps={{
+        // @ts-ignore
```
mastercactapus, line 144: Why is this ignored?

web/src/app/util/ISOPickers.tsx
```diff
@@ -116,35 +151,38 @@ function useISOPicker(
           </InputAdornment>
         ),
       }}
-      {...extraProps}
+      {...cypressProps}
       {...otherProps}
-      InputLabelProps={shrinkInputLabel(otherProps)}
+      InputLabelProps={{
+        ...otherProps,
```
mastercactapus, line 157: Same as above, `otherProps.InputLabelProps` needs to be expanded

web/src/app/schedules/temp-sched/TempSchedAddShiftForm.tsx
```diff
@@ -18,6 +19,11 @@ const useStyles = makeStyles({
 export default function TempSchedAddShiftForm(): JSX.Element {
   const classes = useStyles()
   const [manualEntry, setManualEntry] = useState(false)
+  const [zone] = useURLParam('tz', 'local')
+  const now = DateTime.local()
+    .setZone(zone)
+    .startOf('day')
+    .toFormat("yyyy-MM-dd'T'HH:mm:ss")
```
mastercactapus, line 26: use ISO format, this is missing the offset so it's going to have issues being interpreted as a minimum value.
```suggestion
    .toISO()
```

web/src/app/schedules/temp-sched/TempSchedAddShiftForm.tsx
```diff
@@ -18,6 +19,11 @@ const useStyles = makeStyles({
 export default function TempSchedAddShiftForm(): JSX.Element {
   const classes = useStyles()
   const [manualEntry, setManualEntry] = useState(false)
+  const [zone] = useURLParam('tz', 'local')
+  const now = DateTime.local()
```
mastercactapus, line 23: This is going to change every time it renders, making it possible to set a value that becomes invalid depending on when the user clicks a button.

Maybe this would be a good case for state? Then minimum would basically be set on mount so it would advance through the day (e.g. if user doesn't refresh) but not "break" while they are filling out the form

web/src/app/schedules/temp-sched/TempSchedTimesStep.tsx
```diff
@@ -27,6 +25,11 @@ export default function TempSchedTimesStep({
   value,
 }: TempSchedTimesStepProps): JSX.Element {
   const classes = useStyles()
+  const [zone] = useURLParam('tz', 'local')
+  const now = DateTime.local()
+    .setZone(zone)
+    .startOf('day')
+    .toFormat("yyyy-MM-dd'T'HH:mm:ss")
```
mastercactapus, line 32:
```suggestion
    .toISO()
```

web/src/app/schedules/temp-sched/TempSchedTimesStep.tsx
```diff
@@ -27,6 +25,11 @@ export default function TempSchedTimesStep({
   value,
 }: TempSchedTimesStepProps): JSX.Element {
   const classes = useStyles()
+  const [zone] = useURLParam('tz', 'local')
+  const now = DateTime.local()
```
mastercactapus, line 29: Same as above, might be good to put this into readonly state

mastercactapus (changes requested):

web/src/app/util/ISOPickers.tsx
```diff
@@ -1,36 +1,68 @@
...
 import Modernizr from '../../modernizr.config'
 import { DateRange, AccessTime } from '@material-ui/icons'
 
-function hasInputSupport(name) {
+type ISOPickersProps = {
+  value?: string
+  onChange: (newValue: string) => void
+  timeZone?: string
+  min?: string // yyyy-MM-dd'T'HH:mm:ss
+  max?: string // yyyy-MM-dd'T'HH:mm:ss
```
dctalbot, line 23: > max and min attributes works with the following input types: date, datetime-local, month, time and week.

Let's remove the comments here to avoid any confusion

web/src/app/forms/FormField.js
```diff
@@ -36,8 +36,9 @@ export class FormField extends React.PureComponent {
     fieldName: p.string,
 
     // min and max values specify the range to clamp a int value
-    min: p.number,
-    max: p.number,
+    // will be passed to base component for input controls
```
dctalbot, line 39:
```suggestion
    // one of: number, range, date, datetime-local, month, time and week
```
See https://www.w3schools.com/tags/att_input_min.asp

dctalbot (changes requested):

web/src/app/schedules/temp-sched/TempSchedAddShiftForm.tsx
Forfold: I was finding that the min/max does not get recognized by native inputs unless in this specific format after having tried ISOs originally

Forfold:
Closing, will re-open when at a stable point to change the base to master. I've opened PR #964 for this feature to shorten review on this guy.

PR #965, opened 2020-10-26
temp schedules: update with latest master

Forfold:
Updates `feature/temp-schedules` with latest from master as it is a protected branch.


Forfold:
Closing to address conflict in un-protected branch

PR #968, opened 2020-10-26
fix: disable arrow keys on swipable temp schedule dialog

dctalbot:

**Description:**
Disable arrow key navigation on swipable dialog views since it is unexpected behavior which can lead to confusing states for a user e.g. bypassing form validation


Forfold (approved):
makes sense :) lgtm

mastercactapus (approved):

PR #974, opened 2020-10-27
ui: convert to controlled tooltip in calendar

Forfold:

**Description:**
This PR converts the tooltips for each shift in the calendar to appear when clicked. This allows more accurate viewing of shifts in busy calendars, and addresses a bug where the popover would display overtop dialogs if the mouse position does not change.

### DIFF
--- a/web/src/app/schedules/CalendarEventWrapper.js
+++ b/web/src/app/schedules/CalendarEventWrapper.js
@@ -226,10 +226,7 @@ export default class CalendarEventWrapper extends Component {
         title={this.renderInteractiveTooltip()}
       >
         {React.cloneElement(children, {
-          onClick: () => {
-            children.props.onClick() // toggles selected
-            this.handleClickTooltip() // toggles tooltip
-          },
+          onMouseUp: this.handleClickTooltip,
           onFocus: this.handleFocusTooltip,
           onBlur: this.handleBlurTooltip,
           role: 'button',
--- a/web/src/app/schedules/CalendarEventWrapper.js
+++ b/web/src/app/schedules/CalendarEventWrapper.js
@@ -63,8 +63,19 @@ export default class CalendarEventWrapper extends Component {
     onDeleteTempSched: p.func,
   }
 
+  // Handle showing tooltips with tab focus
+  // but also toggling via clicks event
+  // state is tracked separately as an additional
+  // click on the shift span will close the tooltip
+  state = {
+    eventClicked: false,
+    eventHasFocus: false,
+  }
+
   handleShowOverrideForm = (type) => {
     const { event, onOverrideClick } = this.props
+
+    this.handleBlurTooltip()
 
     onOverrideClick({
       variant: type,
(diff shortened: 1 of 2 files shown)


web/src/app/schedules/CalendarEventWrapper.js
```diff
@@ -190,7 +225,14 @@ export default class CalendarEventWrapper extends Component {
         }}
         title={this.renderInteractiveTooltip()}
       >
-        {children}
+        {React.cloneElement(children, {
+          onClick: () => {
+            children.props.onClick() // toggles selected
```
mastercactapus, line 230: children needs to be treated as opaque (don't assume anything about it)

To map items and set props use `React.Children.map`

It might be easier to use a container/wrapper and "catch" the propagated events instead though

web/src/cypress/integration/scheduleCalendar.ts
```diff
@@ -87,7 +87,7 @@ function testCalendar(screen: ScreenFormat): void {
   })
 
   it(`should view a shift's tooltip`, () => {
-    cy.get('div').contains(rot.users[0].name).trigger('mouseover')
```
mastercactapus, line 90: We should keep this and add a second assertion unless we're removing the ability to hover.

mastercactapus (changes requested):

web/src/cypress/integration/scheduleCalendar.ts
Forfold: That's correct, this PR would remove hovering so that a tooltip appears when clicked or focused (screen readers, tab, etc)

mastercactapus: We should discuss this first before making UX changes -- let's get the release blockers out of the way first

Forfold: I feel if we are making UX changes to error handling this late, it's okay to get something like this working better for our users too

Forfold:
@dctalbot Yeah, I was going to have it go into `feature/temp-schedules` so the `onMouseover()` tests are replaced (they don't seem the most reliable), but I think that will have to wait — The flicker fixes should hopefully resolve those issues anyways

web/src/cypress/integration/scheduleCalendar.ts
Forfold: I definitely agree that we should get release blockers out of the way first — David pointed out this should probably go into master instead so I'll get the base branch updated to give us more space to chat about this 👍

web/src/app/schedules/CalendarEventWrapper.js
Forfold: What I need to happen is trigger both the `onClick` given to the div by the calendar to change colors, and the `onClick` that we define to open the tooltip. I originally had a second div wrapping the child, but then the child div is never clicked on and only the top level `onClick` (to open the tooltip) would trigger. So we won't want to overwrite the `onClick` with a map, since we want both to trigger.

I'll do some research into event propagation from a child component. Maybe I can have a listener rather than another `onClick` inside of here.

Forfold: I found a good alternative that just uses a different click handler that doesn't conflict with the calendar events 🙂

Forfold:
Closing so we can merge separate from this feature: #1019

PR #975, opened 2020-10-27
fix: add shift form validation

dctalbot:

**Description:**
Current state: 
When filling out the Add Shift form, users may add an "invalid shift". The user must ultimately either:
- delete and create new shift
- edit schedule bounds

New state:
When filling out the Add Shift form, users may only add shifts that are valid with respect to the stateful temp sched start/end times.

Net effect:
This only changes the UX. If a user creates an invalid state at any point in time, the ultimate form submit will prevent form submission and provide relevant error messaging.

### DIFF
--- a/web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
+++ b/web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
@@ -140,8 +140,6 @@ export default function TempSchedAddShiftsStep({
 
     if (!isISOAfter(shift.end, shift.start)) {
       add('end', 'must be after shift start time')
-    }
-    if (!isISOBefore(shift.start, shift.end)) {
       add('start', 'must be before shift end time')
     }
     if (isISOBefore(shift.start, start)) {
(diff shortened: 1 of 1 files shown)


Forfold (approved):

web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
```diff
@@ -126,52 +126,29 @@ export default function TempSchedAddShiftsStep({
...
-    if (!isISOAfter(shift.end, shift?.start)) {
-      result.push({
-        field: 'end',
-        message: 'must be after shift start time',
-      } as FieldError)
+    if (!isISOAfter(shift.end, shift.start)) {
+      add('end', 'must be after shift start time')
+    }
+    if (!isISOBefore(shift.start, shift.end)) {
+      add('start', 'must be before shift end time')
```
mastercactapus, line 145: this is redundant, do a single check

mastercactapus (changes requested):

web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
Forfold: I'm not sure I agree with this change, as I think we should only display the error on the field they are typing in, rather than both.

Forfold: I'm not sure I agree with this change, as I would expect only the field I am typing in to error, rather than display on both.

Forfold (approved):

mastercactapus (approved):

PR #982, opened 2020-10-29
temp-schedules: shift list transitions

KatieMSB:

**Description:**
Added in/out transitions to the shift list for temp schedules.


web/src/app/lists/FlatList.tsx
```diff
@@ -0,0 +1,285 @@
...
+import { CSSTransition, TransitionGroup } from 'react-transition-group'
+
+const useStyles = makeStyles({
+  background: { backgroundColor: 'white' },
+  highlightedItem: {
+    borderLeft: '6px solid #93ed94',
+    background: '#defadf',
+  },
+  participantDragging: {
+    backgroundColor: '#ebebeb',
```
dctalbot, line 29: I'd put these hex codes in variables so we can name them. Otherwise they sort of function like "magic numbers"

web/src/app/lists/FlatList.tsx
```diff
@@ -0,0 +1,285 @@
...
+  const handleDragStart = (): void => {
+    // adds a little vibration if the browser supports it
+    if (window.navigator.vibrate) {
+      window.navigator.vibrate(100)
+    }
+  }
+
+  const handleDragEnd = (result: DropResult): void => {
+    if (result.destination) {
+      if (onReorder) {
```
dctalbot, line 107:
```suggestion
    if (result.destination && onReorder) {
```

web/src/app/lists/FlatList.tsx
```diff
@@ -0,0 +1,285 @@
...
+        onReorder(
+          // result.draggableId, : removed this as per new reorderList function
```
dctalbot, line 109:
```suggestion
```

web/src/app/lists/FlatList.tsx
```diff
@@ -0,0 +1,285 @@
...
+        component: AppLink,
+        to: item.url,
+        button: true,
+      }
+    }
+    return (
+      <ListItem
+        key={idx}
+        {...itemProps}
+        style={{ width: '100%' }}
```
dctalbot, line 145: include this in the makestyles declaration

web/src/app/lists/FlatList.tsx
```diff
@@ -0,0 +1,285 @@
...
+    return items.map((item: FlatListListItem, idx: number) => {
+      if ('subHeader' in item) {
+        return renderSubheaderItem(item, idx)
+      }
+      if (!onReorder) {
+        return renderItem(item, idx)
+      }
+      if (item.id) {
+        return (
+          <Draggable key={idx + item.id} draggableId={item.id} index={idx}>
```
dctalbot, line 215:
```suggestion
          <Draggable key={idx} draggableId={item.id} index={idx}>
```
I think index alone should suffice for a unique key

web/src/app/lists/FlatList.tsx
```diff
@@ -0,0 +1,285 @@
...
+    return (
+      <List {...listProps}>
+        {headerNote && (
+          <ListItem>
+            <ListItemText
+              disableTypography
+              secondary={
+                <Typography color='textSecondary'>{headerNote}</Typography>
+              }
+              style={{ fontStyle: 'italic' }}
```
dctalbot, line 255: Let's use a className here

web/src/package.json
```diff
@@ -101,6 +101,7 @@
     "react-redux": "7.2.1",
     "react-router-dom": "5.2.0",
     "react-swipeable-views": "0.13.9",
+    "react-transition-group": "^4.4.1",
```
dctalbot, line 104:
```suggestion
    "react-transition-group": "4.4.1",
```

web/src/package.json
```diff
@@ -126,6 +127,7 @@
     "@types/lodash-es": "4.17.3",
     "@types/luxon": "1.25.0",
     "@types/react": "16.9.52",
+    "@types/react-beautiful-dnd": "^13.0.0",
```
dctalbot, line 130:
```suggestion
    "@types/react-beautiful-dnd": "13.0.0",
```

dctalbot (changes requested):

dctalbot (changes requested):
I'm seeing some odd behavior with the section headers in the shifts list when there are multiple days displayed. 
Steps to reproduce:
- create a new temp schedule for this week
- add 4 users; 2 today, 2 tomorrow
- delete one user from today and notice that the section header for tomorrow is duplicated for a brief moment

web/src/app/lists/FlatList.tsx
```diff
@@ -0,0 +1,289 @@
...
+  // If specified, enables drag and drop
+  //
+  // onReorder(id, oldIndex, newIndex)
+  onReorder?: (oldIndex: number, newIndex: number) => void
+
+  // will render transition in list
+  transition?: boolean
+}
+
+export default function FlatList(props: FlatListType): JSX.Element {
```
Forfold, line 96: nit: we've been naming interfaces/types with "Props" appended to the end
```suggestion
export default function FlatList(props: FlatListProps): JSX.Element {
```

web/src/app/lists/FlatList.tsx
```diff
@@ -0,0 +1,289 @@
...
+  const {
+    onReorder,
+    emptyMessage,
+    headerNote,
+    items,
+    inset,
+    transition,
+    listProps,
+  } = props
```
Forfold, line 105: you can destructure this from within L96 to you can skip over the `props` variable creation

web/src/app/lists/FlatList.tsx
```diff
@@ -0,0 +1,289 @@
...
+
+  const classes = useStyles()
+
+  const handleDragStart = (): void => {
```
Forfold, line 109: Let's change this (and others here) to a function now that this will be a functional component, since we don't need to worry about binding to `this.` with arrow notation anymore.

Forfold (changes requested):
Have you played around with reversing the transition effect (sliding in from the right)? If so how did that experience feel?

KatieMSB:
> Have you played around with reversing the transition effect (sliding in from the right)? If so how did that experience feel?

Yea originally they slid in from the right, but since the fields that populate it are on the left it feels a bit better sliding from the left.

dctalbot (changes requested):
I'm still seeing the duplicate subheader issue, only now it's just a flicker

Screen Shot 2020-11-03 at 10 43 55 AM

dctalbot (approved):

web/src/app/lists/FlatList.tsx
```diff
@@ -0,0 +1,288 @@
...
+  // indent text of each list item if no icon is present
+  inset?: boolean
+
+  // If specified, enables drag and drop
+  //
+  // onReorder(id, oldIndex, newIndex)
+  onReorder?: (oldIndex: number, newIndex: number) => void
+
+  // will render transition in list
+  transition?: boolean
```
dctalbot, line 93:
```suggestion
  transition?: boolean
  
  // cypress test ID
  'data-cy'?: string
```

dctalbot (changes requested):
I think we should be explicit about `data-cy` being an allowed prop. Otherwise the implicit type of `...rest` is `{}`

web/src/app/lists/FlatList.tsx
Forfold: I think it would be easier to extend ListProps if we are going to spread the remaining props to `List`. This should handle allowing `data-cy`.

dctalbot: Yeah good call, that will be more future-proof. No need for a partial though. So @KatieMSB, you could try something like
1. `export interface FlatListProps extends ListProps`
2. remove `listProps` from the interface
3. destructure `...listProps` instead of `...rest`
4. so you end up with just `<List {...listProps}>`

Hopefully that makes sense 😁

Forfold: One thing to note is to update where dense is being set as well since it can be added flat to the component now

web/src/app/lists/FlatList.tsx
```diff
@@ -0,0 +1,289 @@
...
+  icon?: JSX.Element
+  secondaryAction?: JSX.Element | null
+  url?: string
+  id?: string
+}
+
+export interface ListProps {
+  dense: boolean
+}
+
```
dctalbot, line 74:
```suggestion
```

`import  { ListProps } from '@material-ui/core/List'`

sorry for miscommunication

web/src/app/lists/FlatList.tsx
```diff
@@ -0,0 +1,289 @@
...
+export type FlatListListItem = FlatListSub | FlatListItem
+
+export interface FlatListProps extends Partial<ListProps> {
```
dctalbot, line 77:
```suggestion
export interface FlatListProps extends ListProps {
```
we don't need a partial here

dctalbot (changes requested):

dctalbot (approved):

Forfold (approved):
looks great!

PR #983, opened 2020-10-29
cypress: use reasonable limits for calendar tests

Forfold:

**Description:**
This PR adjusts the `shiftLength` limit of the hourly rotations created while testing the schedule calendar because of a performance issue mainly noticeable within our CI testing environment. **Schedules with a rotation assignment handing off every hour may lead to performance issues and should still be addressed.** The decision was made to continue forward with this milestone while addressing the performance degradation in CI first.

### DIFF
--- a/web/src/cypress/integration/rotations.ts
+++ b/web/src/cypress/integration/rotations.ts
@@ -69,7 +69,7 @@ function testRotations(): void {
   describe('Details Page', () => {
     let rot: Rotation
     beforeEach(() =>
-      cy.createRotation({ count: 3 }).then((r: Rotation) => {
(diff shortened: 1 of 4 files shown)


dctalbot (approved):

KatieMSB (dismissed):

mastercactapus (changes requested):
This needs to account for the end of the month, see the TODO in `should view shifts`

One approach might be to have the calendar view always 1 month ahead

mastercactapus:
For clarification, the reason for the 1-hour rotation was so that in almost all cases, all users would appear on the calendar, even on the last day of the month.

mastercactapus (approved):

KatieMSB (approved):

PR #984, opened 2020-10-30
temp schedules: indicate intervals with no coverage

Forfold:

This PR enhances step 2 within the temporary schedule form to visually indicate intervals where there is no on-call coverage. It also adds the start and end times of the temporary schedule from the original typography above the form fields to inside the list, replacing with the supplemental temporary schedule information from step 1.


web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
```diff
@@ -1,27 +1,40 @@
...
 import { relativeDate } from '../../util/timeFormat'
-import _ from 'lodash'
-import Tooltip from '@material-ui/core/Tooltip/Tooltip'
 import { styles } from '../../styles/materialStyles'
 import { parseInterval } from '../../util/shifts'
 
 const useStyles = makeStyles((theme) => {
   return {
+    alert: {
+      margin: '8px 0 8px 0',
```
dctalbot, line 26: a11y: use `rem` units which will be proportional to system's fonts size

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
```diff
@@ -57,33 +71,99 @@ export default function TempSchedShiftsList({
...
+              severity='warning'
+            >
+              No coverage
+            </Alert>
+          ),
+        })
+      }
+
+      // checkCoverage will determine if there is a gap of 1 minute or more between the given datetimes
+      const checkCoverage = (s: DateTime, e: DateTime): boolean => {
```
dctalbot, line 144: maybe rename this to something like `needsCoverage` or `hasGap`? Not sure, but `checkCoverage` wasn't obvious to me and is doing the opposite of what I initially assumed

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
```diff
@@ -57,33 +71,99 @@ export default function TempSchedShiftsList({
...
+        return Interval.fromDateTimes(s, e).length('minutes') > 1
+      }
+
+      // craft list items for each day
```
dctalbot, line 148:
```suggestion
```

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
```diff
@@ -132,15 +212,66 @@ export default function TempSchedShiftsList({
           ),
         })
 
+        // prevents actions from rendering on each item if it's for the same shift
```
dctalbot, line 215:
```suggestion
```

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
```diff
@@ -132,15 +212,66 @@ export default function TempSchedShiftsList({
...
+        ) {
+          result.push({
+            render: () => (
+              <Alert
+                key={s.end.toISO() + '-no-middle-coverage'}
+                className={classes.alert}
+                severity='warning'
+              >
+                No coverage from {s.end.setZone(zone).toFormat('h:mm a')} to{' '}
+                {dayShifts[shiftIdx + 1].start.setZone(zone).toFormat('h:mm a')}
```
dctalbot, line 231: use a string template

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
```diff
@@ -57,33 +71,99 @@ export default function TempSchedShiftsList({
...
+      // checkCoverage will determine if there is a gap of 1 minute or more between the given datetimes
+      const checkCoverage = (s: DateTime, e: DateTime): boolean => {
+        return Interval.fromDateTimes(s, e).length('minutes') > 1
+      }
+
+      // craft list items for each day
+      dayShifts.forEach((s, shiftIdx) => {
+        // check start of day coverage for the first shift
+        // if on the first day, temp sched start is used
+        const _s = dayIdx === 0 ? DateTime.fromISO(start) : day.start
```
dctalbot, line 152: Same logic problem mentioned above probably applies here. We need to think about how to handle going between steps. There might be another solution altogether we haven't considered. Maybe for simplicity, we shouldn't allow going backwards to begin with.

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
```diff
@@ -132,15 +212,66 @@ export default function TempSchedShiftsList({
...
+                No coverage from {s.end.setZone(zone).toFormat('h:mm a')} to{' '}
+                {dayShifts[shiftIdx + 1].start.setZone(zone).toFormat('h:mm a')}
+              </Alert>
+            ),
+          })
+        }
+
+        // check end of day/temp sched coverage
+        // if on the last day, temp sched end is used
+        const _e = dayIdx === days.length - 1 ? DateTime.fromISO(end) : day.end
```
dctalbot, line 239: Similar issue as above, we need to account up until `max(lastShiftEnd, schedEnd)`

dctalbot (changes requested):
Animations with the colored alert items are really slick! We need to consider what happens when users don't follow conventional workflows. As is, they can get into some very awkward states. Might need a larger discussion about this stepper experience altogether. Maybe we can have both steps on one page? Or disallow "going backwards"?

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
```diff
@@ -57,33 +71,99 @@ export default function TempSchedShiftsList({
...
-
-    const firstShiftStart = shifts[0].start
-    const lastShiftEnd = lastShift.end
-
     const displaySpan = Interval.fromDateTimes(
-      DateTime.min(schedInterval.start, firstShiftStart).startOf('day'),
-      DateTime.max(schedInterval.end, lastShiftEnd).endOf('day'),
+      DateTime.fromISO(start).startOf('day'),
+      DateTime.fromISO(end).endOf('day'),
     )
```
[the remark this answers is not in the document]
Forfold: As is I wasn't able to reproduce this. When I follow steps 1-3, an error message shows at the bottom of step 1 displays `One or more shifts extend beyond the start and/or end of this temporary schedule`. In which case they would go and delete the shift and then they could redo from step 1.

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
Forfold: `dayShifts` is evaluated on every render, so it will have the most recent day index and start of schedule here if they change the start and come back.

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
Forfold: Since the notices are for previewing the coverage for a temporary schedule, I don't think we can use `lastShiftEnd` here in any case. We specifically want to let the user know if they have all the shift coverage needed within the start/end they set on step 1.

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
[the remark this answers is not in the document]
Forfold: Right now we display error messages but still allow clicking the button, and the button also gets stuck on retry if a request fails instead of back/next if needed. Definitely more follow-up work will be needed here

KatieMSB (commented):
Might not have been introduced this pr but seeing lots of console errors from the swipeable views
Screen Shot 2021-02-16 at 11 20 32 AM

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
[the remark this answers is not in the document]
Forfold: What I can do while we work on that UX for the buttons is still render the shifts that fall outside of coverage without notice info. The logic for the notices will need to stay the same for reasons I mentioned above so we won't be able to use `max` there, but we can still render them past that, and the error message will be displayed below. It's been more than a few months since I've looked at this body of work, but I'll check out what it'll take for something like that.

Forfold:
@KatieMSB I'm seeing that a lot now too, I think there was an update that may have caused it so I'll be sure to look into it before we merge this.

@dctalbot I think I found a good solution around needing to use next/back buttons at all times, but showing shift coverage properly. Let me know what you think!:

image

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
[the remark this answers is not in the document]
dctalbot: Yeah, rendering shifts that fall outside of the bounds of the schedule start and end times is what I was getting at, sorry if that was unclear. It's been a while for me too, but I remember there being some wonky edge cases if users go through the workflow in unconventional ways.

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
dctalbot: Got it, I think you're right here. Just need to make sure we are rendering all shifts regardless of their validity with respect to time start/end

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
```diff
@@ -57,33 +84,94 @@ export default function TempSchedShiftsList({
...
 
-    const firstShiftStart = shifts[0].start
-    const lastShiftEnd = lastShift.end
-
     const displaySpan = Interval.fromDateTimes(
-      DateTime.min(schedInterval.start, firstShiftStart).startOf('day'),
-      DateTime.max(schedInterval.end, lastShiftEnd).endOf('day'),
+      DateTime.min(schedInterval.start, firstShift).startOf('day'),
+      DateTime.max(schedInterval.end, lastShift).endOf('day'),
```
dctalbot, line 94: I was able to break it, got this error: `Error: min requires all arguments be DateTimes`

A few notes here:

- I'd prefer to use `firstShiftStart` and `lastShiftEnd` to be explicit about to the point in time the var refers to
- The implementation of `lastShift` here isn't correct - this code is getting the end of the shift with the latest start time, but that isn't necessarily the shift with the latest end time. We need to either sort by end time and grab the last shift, or use the reduce we had in the existing code
- `sortedShifts[0]` will break if there are no shifts, I think we still need to handle that here

dctalbot (changes requested):
Nice, that image above looks perfect!

I was able able to trigger a broken state like so:
1. Add start and end times e.g. Feb 18th through 20th
2. Click next, add a shift, click back
3. Change start time to be Feb 25th

Haven't dug into the code, but we might need to bail out early on the shifts rendering if there is an invalid schedule state

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
Forfold: 1. :+1: 
2. Gotcha, the code from before had seemed redundant as the shifts were already sorted, but you make a good point with that case. I'll add a comment above the reduce call so it's clear for future maintainers.
3. This is handled at the top of the function, returning with JSX to instruct the user to add a shift if `_shifts` is empty.

Forfold:
@dctalbot @KatieMSB All bugs/inconsistencies have been resolved so I've re-requested your reviews. Ty!!

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
```diff
@@ -45,45 +58,149 @@ export default function TempSchedShiftsList({
...
+                .setZone(zone)
+                .toFormat('h:mm a')}`}
+            </Alert>
+          ),
+        })
+      }
+
+      // for temp scheds with at least 1 shift
+      // render no coverage and continue if no shifts for the given day
+      if (!dayShifts.length && sortedShifts.length) {
```
dctalbot, line 169:
```suggestion
      if (dayShifts.length === 0 && sortedShifts.length > 0) {
```
Nit, my brain doesn't like `!length` for some reason but maybe it's just me!

web/src/app/schedules/temp-sched/sharedUtils.tsx
```diff
@@ -11,6 +11,11 @@ export type Shift = {
   start: string
   end: string
   userID: string
+
+  user?: {
+    id: string
+    name: string
+  }
```
dctalbot, line 18: Is there a way we can normalize this so it doesn't have both a `userID` and `user.id`?

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
```diff
@@ -132,15 +249,80 @@ export default function TempSchedShiftsList({
...
+          checkCoverage(s.end, dayEnd)
+        ) {
+          result.push({
+            render: () => (
+              <Alert
+                key={s.end.millisecond + '-no-end-coverage'}
+                className={classes.alert}
+                severity='warning'
+              >
+                No coverage after {s.end.setZone(zone).toFormat('h:mm a')}
```
dctalbot, line 299: I wonder if it makes sense to add a util for time formatting since we do this in 6 places, something like
```js
function simpleFormat(dt: Datetime) {
   return dt.setZone(zone).toFormat('h:mm: a')
}
```

dctalbot (changes requested):
Functionally validated! Left a few nits/comments. We should try to get the `Shift` typedef normalized, representing the data in the most minimal form w/o redundancy e.g. `userID` and `user.id`

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
dctalbot: On second thought, I'm not sure this helps much. Lmk what you think

Forfold: I think it makes sense to just call the API directly in this case, since the only shortening would mainly come from not needing to type out `setZone` a bunch, as the start var in some cases is still pretty long, i.e. `dayShifts[shiftIdx + 1].start`

web/src/app/schedules/temp-sched/sharedUtils.tsx
Forfold: I noticed the user object was coming from the select API, but I wasn't sure exactly how (couldn't pin down the code). I'll do some more digging to make sure our types are in the state we want 👍

web/src/app/schedules/temp-sched/TempSchedDialog.tsx
```diff
@@ -72,33 +72,30 @@ export default function TempSchedDialog({
...
+
+    return (
+      <TempSchedTimesStep
+        key={key}
+        stepText='STEP 1 OF 2'
+        scheduleID={scheduleID}
+        value={value}
+        edit={edit}
+      />
+    )
```
KatieMSB, line 98: Looks like we should be comparing the indexes for all cases instead of having a default return here. The field validation for the time step no longer shows the error messages but does still disable the next button on error.

Forfold: I am seeing field validation coming through on my end- was there a specific set of actions you did to reproduce?

KatieMSB: Looks like this is a chrome issue. So to recreate on chome, setting the end time before the start time disables the next button but shows no field validation.

KatieMSB: Updating my yarn to 1.22.10 seems to resolve this issue.

dctalbot (approved):

PR #995, opened 2020-11-05
fix: material select shows selected option in dropdown for single-select

### LINKED ISSUE #958: ui: "No options" is rendered when a single select field is open with a value set
> In single select dropdowns, "No options" is rendered when a value is set.

_Originally posted by @mastercactapus in https://github.com/target/goalert/pull/925#discussion_r501285735_

Additional Cypress tests should also be created. Currently, only clearing the fields for single and multiple selects is covered.

### LINKED ISSUE #994: material select: omits selected option from options dropdown in single-select mode
**Describe the Bug:**
Provide a clear, concise description of the issue and why you consider it to be a bug.

**Steps to Reproduce:**
1. Go to rotations, create, timezone
2. America/Chicago (default for me) does not appear in dropdown list since it's the selected value

**Expected Behavior:**
Selected option should appear in dropdowns

**Observed Behavior:**
Selected option does not appear in dropdowns

dctalbot:

**Description:**
This PR fixes an issue in the autocomplete select component in which the stateful selected option would not render as an option in the dropdown.

Closes #994 and closes #958

**Screenshots:**
Screen Shot 2020-11-05 at 11 53 26 AM

### DIFF
--- a/web/src/app/selection/MaterialSelect.tsx
+++ b/web/src/app/selection/MaterialSelect.tsx
@@ -105,22 +105,26 @@ export default function MaterialSelect(
     _setInputValue(input)
     onInputChange(input)
   }
+
+  const multi = multiple ? { multiple: true, filterSelectedOptions: true } : {}
+
   useEffect(() => {
     if (multiple) return
     if (!value) setInputValue('')
     if (!inputValue && value) setInputValue(getInputLabel())
   }, [value, multiple])
 
-  const multi = multiple ? { multiple: true } : {}
-
-  // merge selected values with options to avoid annoying mui warnings while dropdown is closed
+  // merge selected values with options to avoid annoying mui warnings
   // https://github.com/mui-org/material-ui/issues/18514
-  // be sure to keep the props `filterSelectedOptions` set to hide selected value from dropdown
-  // and `getOptionSelected` to ensure what shows as selected for all incoming values
-  let options: SelectOption[] = _options
-  if (value) {
-    // merge options with value
-    options = options.concat(Array.isArray(value) ? value : [value])
+  let options = _options
+  if (value && Array.isArray(value)) {
+    options = [...options, ...value]
+  } else if (
+    value &&
+    !Array.isArray(value) &&
+    options.every((opt) => opt.value !== value.value)
+  ) {
+    options = [value, ...options]
   }
 
   return (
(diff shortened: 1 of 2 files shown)


KatieMSB (changes requested):
Should probably add new Cypress tests as mentioned in #958.

KatieMSB (dismissed):

web/src/cypress/integration/materialSelect.ts
```diff
@@ -96,6 +96,22 @@ function testMaterialSelect(): void {
       })
     })
   })
+
+  describe('render options', () => {
+    it.only('should show selected value as an option in single-select mode', () => {
```
mastercactapus, line 101:
```suggestion
    it('should show selected value as an option in single-select mode', () => {
```

web/src/app/selection/MaterialSelect.tsx
```diff
@@ -105,22 +105,26 @@ export default function MaterialSelect(
...
-  if (value) {
-    // merge options with value
-    options = options.concat(Array.isArray(value) ? value : [value])
+  let options = _options
+  if (value && Array.isArray(value)) {
+    options = [...options, ...value]
+  } else if (
+    value &&
+    !Array.isArray(value) &&
+    options.every((opt) => opt.value !== value.value)
```
mastercactapus, line 125: Should we be doing this check for the array form too?

mastercactapus (changes requested):

web/src/app/selection/MaterialSelect.tsx
dctalbot: No, this handles the case for when the input is re-selected. Without this check, you'd end up with this:
Screen Shot 2020-11-06 at 10 09 49 AM
So we do an explicit check to say "only add value to the list of options if it's not already an option".
This isn't an issue in multi-select since there is no concept of "re-selecting" (`filterSelectedOptions` is turned on). Thanks for checking!

web/src/cypress/integration/materialSelect.ts
dctalbot: Thank you :)

web/src/app/selection/MaterialSelect.tsx
mastercactapus: Ty for the clarification!

mastercactapus (approved):

KatieMSB (approved):

Issue #923, opened 2020-10-07
ui: convert remaining class components to hooks


mastercactapus:
This is an issue to track progress on converting the remaining class components to React hooks.

They should be converted in javascript for review purposes (continue to use `PropTypes`). A separate issue will be created for converting to typescript if needed.

Decorators should also be removed. There is usually a corresponding hook to achieve the same functionality. 
For example, a class component using `@withWidth` and `props.width` can be written as a functional component that uses `const width = useWidth()`

Please keep PR size small, and run tests locally, to help speed up reviews.

- [x] app/admin/AdminDialog.js: AdminDialog (#996)
- [x] app/alerts/components/AlertDetails.js: AlertDetails (#997)
- [x] app/alerts/components/AlertsListControls.js: AlertsListControls (#998)
- [x] app/alerts/components/AlertsListFilter.js: AlertsListFilter (#999)
- [x] app/alerts/components/UpdateAlertsSnackbar.js: UpdateAlertsSnackbar (#1000)
- [x] app/alerts/pages/AlertDetailPage.js: AlertDetailPage (#1001)
- [x] app/dialogs/components/DialogContentError.js: DialogContentError (@KatieMSB #1004)
- [x] app/dialogs/components/DialogTitleWrapper.js: DialogTitleWrapper (@KatieMSB #1005)
- [x] ~~app/dialogs/components/DropDownMenu.js: DropdownMenu~~ (@KatieMSB #1156)
- [x] app/dialogs/FormDialog.js: FormDialog (@dctalbot  #1236)
- [x] app/documentation/components/IntegrationKeyAPI.js: IntegrationKeyAPI (@pnengchu https://github.com/target/goalert/pull/1020)
- [x] app/error-pages/Errors.js: GenericError (@KatieMSB #1009)
- [x] app/error-pages/Errors.js: ObjectNotFound (@KatieMSB #1009)
- [x] app/error-pages/Errors.js: PageNotFound (@KatieMSB #1009)
- [x] app/escalation-policies/PolicyCreateDialog.js: PolicyCreateDialog (@KatieMSB #1013)
- [x] app/escalation-policies/PolicyEditDialog.js: PolicyEditDialog (@KatieMSB #1016)
- [x] app/escalation-policies/PolicyForm.js: PolicyForm (@KatieMSB #1017)
- [x] app/escalation-policies/PolicyServicesCard.js: PolicyServicesCard (@KatieMSB #1022)
[...]

PR #1006, opened 2020-11-09
ui/hooks: drop down menu

KatieMSB:

**Description:**
Convert DropDownMenu.js to use hooks.

Part of #923


web/src/app/dialogs/components/DropDownMenu.js
```diff
@@ -1,65 +1,61 @@
...
+        onClose={() => {
+          setAnchorEl(null)
+        }}
+      >
+        {options.map((option) => (
+          <MenuItem
+            key={option.label}
+            onClick={() => {
+              setAnchorEl(null)
+              option.onClick()
```
mastercactapus, line 40: This was intended to be fired upon `onExited`, after animations have finished. IIRC we had a lot of test flakiness leading to it being implemented this way.

Might require a little creativity to get it to work with hooks. Or some careful thought/validation to see if it's still required.

mastercactapus (changes requested):

web/src/app/dialogs/components/DropDownMenu.js
Forfold: I'd be curious to see how this runs against our tests as is, without the use of classes/old setState

web/src/app/dialogs/components/DropDownMenu.js
```diff
@@ -1,65 +1,61 @@
...
-    )
-  }
+DropDownMenu.propTypes = {
+  options: p.arrayOf(
+    p.shape({
+      label: p.string,
+      onClick: p.func,
+    }),
+  ),
+  onExited: p.func,
```
Forfold, line 58: I don't think this is needed, but we can likely extend `Menu` when we convert this to TypeScript and have a spread anywho

Forfold (dismissed):

web/src/app/dialogs/components/DropDownMenu.js
mastercactapus: IIRC it was a focus-stealing issue -- dialog opening before the menu closed or something like that

mastercactapus: vague memories of some nightmarish test debugging come to mind

KatieMSB: This component isn't being utilized anywhere but I ran through some cypress scenarios locally and everything seemed to work as expected with the latest changes.

Forfold (changes requested):
Since you mentioned that this isn't really being utilized anywhere, what do you think of just deleting this file and updating the once reference we have to it (snippet below) in `DialogTitleWrapper` to use `OtherActions` from utils instead? It's the more recent version we've been using that should handle switching between mobile and desktop.

**`web/src/app/dialogs/components/DialogTitleWrapper.js`**
```
  let menu
  if (options && options.length > 0 && fullScreen) {
    menu = <DropDownMenu options={options} color='white' />
  } else if (options && options.length > 0) {
    menu = (
      <div className={classes.topRightActions}>
        <DropDownMenu options={options} />
      </div>
    )
  }
```

KatieMSB:
> Since you mentioned that this isn't really being utilized anywhere, what do you think of just deleting this file and updating the once reference we have to it (snippet below) in `DialogTitleWrapper` to use `OtherActions` from utils instead? It's the more recent version we've been using that should handle switching between mobile and desktop.
> 
> **`web/src/app/dialogs/components/DialogTitleWrapper.js`**
> 
> ```
>   let menu
>   if (options && options.length > 0 && fullScreen) {
>     menu = <DropDownMenu options={options} color='white' />
>   } else if (options && options.length > 0) {
>     menu = (
>       <div className={classes.topRightActions}>
>         <DropDownMenu options={options} />
>       </div>
>     )
>   }
> ```

Yea that makes sense, I can make a separate pr for that and go ahead and close this one if that sounds good.

Forfold:
Sure, whatever is easiest for you! I wouldn't mind staying on this PR + updating the title either

KatieMSB:
Closing in favor of #1156

PR #1016, opened 2020-11-11
ui/hooks: policy edit dialog

KatieMSB:

**Description:**
Converts PolicyEditDialog to use hooks.

Part of #923


web/src/app/escalation-policies/PolicyEditDialog.js
```diff
@@ -24,78 +23,58 @@ const mutation = gql`
...
-          />
-        }
-      />
-    )
-  }
+  return (
+    <FormDialog
+      title='Edit Escalation Policy'
+      loading={
+        (!data && editDialogQueryStatus) || editDialogMutationStatus.loading
```
mastercactapus, line 58:
```suggestion
        (!data && editDialogQueryStatus.loading) || editDialogMutationStatus.loading
```

mastercactapus (changes requested):

web/src/app/escalation-policies/PolicyEditDialog.js
```diff
@@ -24,78 +23,58 @@ const mutation = gql`
...
-    const { loading, error } = status
-    const { value } = this.state
-    const fieldErrs = fieldErrors(error)
+  const [editDialogMutation, editDialogMutationStatus] = useMutation(mutation, {
+    variables: {
+      input: {
+        id: props.escalationPolicyID,
+        name: (value && value.name) || defaultValue.name,
+        description: (value && value.description) || defaultValue.description,
+        repeat: (value && value.repeat.value) || defaultValue.repeat.value,
```
dctalbot, line 47: Not related to the conversion to use hooks, but I think there might be a bug here if `value.repeat.value === 0` and `defaultValue.repeat.value > 0`

alternatives:
```js
repeat: (value?.repeat?.value) ?? defaultValue.repeat.value,
```
```js
repeat: _.get(value, 'repeat.value', defaultValue.repeat.value)
```

if we go forward with this change, we should adjust the other field assignments as well for consistency

dctalbot (changes requested):
Curious what you think about this. Even if it's not a bug today, it feels fragile to me.

web/src/app/escalation-policies/PolicyEditDialog.js
KatieMSB: Yea not sure if there's a bug since the value is null until the drop down is set but I agree it could probably be a little cleaner.

web/src/app/escalation-policies/PolicyEditDialog.js
```diff
@@ -24,78 +23,58 @@ const mutation = gql`
   }
 `
 
-export default class PolicyEditDialog extends PureComponent {
-  static propTypes = {
-    escalationPolicyID: p.string.isRequired,
-    onClose: p.func,
+function PolicyEditDialog(props) {
+  const [value, setValue] = useState(null)
+  const { data, editDialogQueryStatus } = useQuery(query, {
```
dctalbot, line 28:
```suggestion
  const { data, editDialogQueryStatus } = useQuery(query, {
  pollInterval: 0,
```
Sorry I missed this the first time, before we had the `noPoll` prop turned on. I think this is the "hooks" analog for that

dctalbot (changes requested):

dctalbot (approved):

dctalbot (approved):

web/src/app/escalation-policies/PolicyEditDialog.js
```diff
@@ -24,78 +23,60 @@ const mutation = gql`
...
+    pollInterval: 0,
+    variables: { id: props.escalationPolicyID },
+  })
+  const defaultValue = {
+    id: props.escalationPolicyID,
+    name: data.escalationPolicy.name,
```
mastercactapus, line 34: This will break if the cache doesn't return a response with the first render call.

The previous code would have rendered a spinner in a random spot while waiting, but we can just render nothing while loading instead:
```js
if (editDialogQueryStatus.loading && !data.escalationPolicy) return null
```

mastercactapus (changes requested):

web/src/app/escalation-policies/PolicyEditDialog.js
mastercactapus: Since we can't put that `if` statement above this line (since hooks can't be called conditionally), lets add `?` to these `data` props

e.g.
`data?.escalationPolicy?.name`

We still want the `if (...) return null` below so that we're not filling fields with incorrect info, but if the cache lookup fails, or the data is otherwise invalid, this line and the rest of `defaultValue` will result in a thrown exception otherwise

mastercactapus (approved):

dctalbot (approved):

PR #1019, opened 2020-11-11
ui: convert to controlled tooltip in calendar

### LINKED ISSUE #1065: ui: calendar popover doesn't apply typography
Expected behavior: MUI typography styles are applied within calendar popover components.

Screen Shot 2020-11-24 at 12 30 40 PM

Forfold:

**Description:**
This PR converts the tooltips for each shift in the calendar to appear when selected (as opposed to the current behavior of appearing on hover), which will allow for more accurate viewing of shifts busy calendars. This approach also gives us the benefit of fixing a bug where the tooltip would display on top of dialogs if the mouse doesn't move (i.e. `z-index` layering issue).

**Changes**
- `CalendarEventWrapper` converted to a functional component, with unnecessary bloat removed
- Replaces the `popup` prop in the calendar for `showAllEvents` (alternative solution to resolve #1065)
- [Adds screen reader support](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#Required_JavaScript_Features) when tabbing through shifts
- Adds package `"@types/react-big-calendar": "0.24.8",`


dctalbot (dismissed):

mastercactapus (dismissed):
- There's an odd delay & shake when clicking on an event
- The color of the event changes (separate from focus bubble) and doesn't change back when clicking away
- clicking inside the tooltip makes it go away (e.g. no longer possible to copy text)
- events get lost leaving multiple tooltips open

image

Forfold:
I'm seeing some problems when trying to select an event via keyboard (tab to focus, w/ enter or spacebar to select). I'm going to label this PR as blocked while I discuss my findings over at https://github.com/jquense/react-big-calendar/issues/1838.

dctalbot (changes requested):
I think we should add a closing trigger for #1019 in the PR description, right?

Found an issue trying to close popups on the weekly view.
Screen Shot 2021-01-04 at 1 41 01 PM

When a popup is opened (via `return`, `spacebar`, or `click`), I expect to be able to `tab` through the options within the popup. The popup would then be closable with `esc`. I noticed this is an issue in prod so I'm OK merging this if we open up a separate issue for these accessibility bits.

Otherwise, I'm really excited to get this in. A huge improvement in experience on busy calendars IMO.

Forfold:
@dctalbot which issue are you referring to? Since #1019 is this PR.

Also, there is still some work around keeping the tooltip open so you can highlight/copy text. We may get the tab functionality for free once that's resolved 👍  I'll keep you posted!

dctalbot (changes requested):
I'm seeing an issue with viewing overflowed calendar events. I think they're supposed to be scrollable now with the `showAllEventsProp`, but I'm still seeing "+X more" however it doesn't do anything when clicked e.g.:
Screen Shot 2021-01-11 at 1 40 49 PM

Forfold:
@dctalbot Are you on the right branch (`cal-tooltip`)? I was originally working in `controlled-tooltip` though that was before when we were working within the feature branch.

web/src/app/schedules/CalendarEventWrapper.js
```diff
@@ -21,57 +20,36 @@ const styles = (theme) => ({
...
-}
 
-@withStyles(styles)
-@connect(mapStateToProps, null)
-export default class CalendarEventWrapper extends Component {
-  static propTypes = {
-    event: p.object.isRequired,
-    onOverrideClick: p.func.isRequired,
+  function handleKeyDown(event) {
+    const code = event.keyCode || event.which
```
dctalbot, line 45: Looks like these are deprecated in favor of `event.key` see [here](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent)

web/src/app/schedules/CalendarEventWrapper.js
```diff
@@ -21,57 +20,36 @@ const styles = (theme) => ({
...
+    if (code === 13 || code === 32) {
```
dctalbot, line 46: Nit: might be nice to use constants here to avoid having to do a google search
e.g. `if (code === SPACEBAR || code === ENTER) {`

web/src/app/schedules/CalendarEventWrapper.js
```diff
@@ -91,10 +69,9 @@ export default class CalendarEventWrapper extends Component {
    * shift as an override, if possible (not in the
    * past).
    */
-  renderInteractiveTooltip = () => {
-    const { classes, event } = this.props
-
+  function renderShiftInfo() {
```
dctalbot, line 72: This function's comment is no longer valid

web/src/app/schedules/CalendarEventWrapper.js
```diff
@@ -136,31 +114,50 @@ export default class CalendarEventWrapper extends Component {
...
-  }
+        {renderShiftInfo()}
+      </Popover>
+      {React.cloneElement(children, {
+        tabIndex: 0,
+        onClick: handleClick,
+        onKeyDown: handleKeyDown,
+        role: 'button',
+        'aria-pressed': open,
+        'aria-describedby': open ? id : undefined,
```
dctalbot, line 154:
```suggestion
        'aria-describedby': id,
```
I think this logic is already taken care of when you initialize `id`

dctalbot (changes requested):

dctalbot (dismissed):

dctalbot (approved):

mastercactapus (approved):

PR #1023, opened 2020-11-12
ui/hooks: policy services query

KatieMSB:

**Description:**
Converts PolicyServicesQuery.js to use hooks.

Part of #923

### DIFF
--- a/web/src/app/escalation-policies/PolicyServicesQuery.js
+++ b/web/src/app/escalation-policies/PolicyServicesQuery.js
@@ -4,7 +4,7 @@ import gql from 'graphql-tag'
 import PolicyServicesCard from './PolicyServicesCard'
 import { useQuery } from '@apollo/react-hooks'
 import Spinner from '../loading/components/Spinner'
-import { GenericError } from '../error-pages'
+import { GenericError, ObjectNotFound } from '../error-pages'
 
 const query = gql`
   query($id: ID!) {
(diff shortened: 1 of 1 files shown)


dctalbot (dismissed):
I think this component is simple enough that we can introduce `useQuery` here. We could also break that into a separate PR, up to you.

mastercactapus (changes requested):
Agree with @dctalbot let's use the hooks version for the apollo query

Forfold (approved):

dctalbot (dismissed):

web/src/app/escalation-policies/PolicyServicesQuery.js
```diff
@@ -16,22 +18,26 @@ const query = gql`
...
-        query={query}
-        render={({ data }) => (
-          <PolicyServicesCard
-            services={data.escalationPolicy.assignedTo || []}
-          />
-        )}
-        variables={{ id: this.props.escalationPolicyID }}
-      />
-    )
+  if (error) {
```
mastercactapus, line 30: We'll also need a check like: `if (!data.escalationPolicy) return <ObjectNotFound />` otherwise it will throw an error for a missing policy.

mastercactapus (changes requested):

dctalbot (approved):

mastercactapus (approved):

PR #1024, opened 2020-11-12
ui/hooks: user contact method delete dialog

alejandra-martinez:

**Description:**
Converts UserContactMethodDeleteDialog.js to use hooks.

Part of #923


web/src/app/users/UserContactMethodDeleteDialog.js
```diff
@@ -2,41 +2,41 @@ import React from 'react'
...
+  const { contactMethodID, ...rest } = props
 
-  render() {
-    return (
-      <Mutation mutation={mutation} onCompleted={this.props.onClose}>
-        {(commit, status) => this.renderDialog(commit, status)}
-      </Mutation>
-    )
-  }
+  const [deleteUserMethod, deleteMethodStatus] = useMutation(mutation, {
```
Forfold, line 16: `[deleteCM, deleteCMStatus]` might be better names. When I read "deleteUserMethod" I could see it interpreted it as a method that will delete a user when called

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
Forfold, line 1: Try running `make clean` and then `cd web/src && yarn` to remove your current dependencies and then regenerate the front end ones to get back in sync

Forfold (changes requested):

web/src/app/users/UserContactMethodDeleteDialog.js
alejandra-martinez: makes sense, all fixed!

dctalbot (approved):
Nice work! In the future, we should use a more descriptive branch name so it's clear what component is being worked on

Forfold (approved):

PR #1027, opened 2020-11-13
ui/hooks: schedule shift list

dctalbot:

**Description:**
This PR converts ScheduleShiftList to hooks

Part of #923 
Prereq for #955

### DIFF
--- a/web/src/app/schedules/ScheduleShiftList.js
+++ b/web/src/app/schedules/ScheduleShiftList.js
@@ -1,11 +1,8 @@
-import React from 'react'
+import React, { useMemo, useState } from 'react'
 import { DateTime, Duration, Interval } from 'luxon'
 import p from 'prop-types'
 import FlatList from '../lists/FlatList'
 import gql from 'graphql-tag'
-import { withQuery } from '../util/Query'
-import { connect } from 'react-redux'
-import { urlParamSelector } from '../selectors'
 import { relativeDate } from '../util/timeFormat'
 import {
   Card,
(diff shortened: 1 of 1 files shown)


web/src/app/schedules/ScheduleShiftList.js
```diff
@@ -55,115 +53,59 @@ const durString = (dur) => {
...
+  const [specifyDuration, setSpecifyDuration] = useState(false)
+  const [isClear, setIsClear] = useState(false)
+
+  const [duration, setDuration] = useURLParam('duration', 'P14D')
+  const [zone] = useURLParam('tz', 'local')
+  const [userFilter, setUserFilter] = useURLParam('userFilter', [])
+  const [activeOnly, setActiveOnly] = useURLParam('activeOnly', false)
+  const [_start, setStart] = useURLParam(
     'start',
     DateTime.fromObject({ zone }).startOf('day').toISO(),
```
mastercactapus, line 75: This should be using the memoized `start` value so the default doesn't keep changing (`mapStateToProps` was only evaluated once)

mastercactapus (changes requested):

web/src/app/schedules/ScheduleShiftList.js
dctalbot: Oh interesting, so to clarify: this change would prevent the default value from changing across the midnight boundary when "start of today" changes value. And it would also prevent having redundant calculations for "start of today"

```js
const defaultStart = useMemo(
    () => DateTime.fromObject({ zone }).startOf('day').toISO(),
    [],
  )

  const [_start, setStart] = useURLParam('start', defaultStart)
  const start = useMemo(
    () => (activeOnly ? DateTime.fromObject({ zone }).toISO() : _start),
    [activeOnly, _start],
  )
```

mastercactapus: Right, as long as the component is mounted, keep the default start as-is. We just don't want to change the interpretation of the URL parameter based on when a render happens, it makes for hard to detect/reproduce bugs and/or tests that fail around midnight.

Since start of day is based on the time zone though, that will need to be added as a dependency. However the `activeOnly` would be better to be just `.utc()` as there is no time of day, it just needs to be clear it's the "current timestamp".

The old code was misleading so we should fix it here -- otherwise it would seem like the second `useMemo` should track `zone` when it doesn't need it.

```js
  const defaultStart = useMemo(()=> DateTime.fromObject({ zone }).startOf('day').toISO(), [zone])
  const [_start, setStart] = useURLParam('start', defaultStart)
  const start = useMemo(
    () => (activeOnly ? DateTime.utc().toISO() : _start),
    [activeOnly, _start],
  )
```

mastercactapus (approved):

KatieMSB (approved):

PR #1030, opened 2020-11-16
ui/hooks: Icons

pnengchu:

**Description:**
converted class components to hooks in Icons.js

Part of https://github.com/target/goalert/issues/923

### DIFF
--- a/web/src/app/icons/components/Icons.js
+++ b/web/src/app/icons/components/Icons.js
@@ -1,90 +1,61 @@
-import React, { Component } from 'react'
+import React from 'react'
 import { PropTypes as p } from 'prop-types'
 import Tooltip from '@material-ui/core/Tooltip'
-import withStyles from '@material-ui/core/styles/withStyles'
 import AddIcon from '@material-ui/icons/Add'
 import TrashIcon from '@material-ui/icons/Delete'
 import WarningIcon from '@material-ui/icons/Warning'
-import except from 'except'
-import { styles } from '../../styles/materialStyles'
 import slackIcon from '../../public/slack.svg'
 import slackIconBW from '../../public/slack_monochrome_black.svg'
+import { makeStyles } from '@material-ui/core/styles'
 
-@withStyles(styles)
-export class Trash extends Component {
-  render() {
-    const { classes } = this.props
+const useStyles = makeStyles({
+  trashIcon: {
+    color: '#666',
+    cursor: 'pointer',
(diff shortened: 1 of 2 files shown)


web/src/app/icons/components/Icons.js
```diff
@@ -1,90 +1,61 @@
...
+
+const useStyles = makeStyles({
+  trashIcon: {
+    color: '#666',
+    cursor: 'pointer',
+    float: 'right',
+  },
+  warningColor: {
+    color: '#FFD602',
+  },
```
dctalbot, line 19: We should import these styles to keep the code DRY (don't repeat yourself). One example of where we do this is [here](https://github.com/target/goalert/blob/master/web/src/app/alerts/components/AlertsListFilter.js#L23)

web/src/app/icons/components/Icons.js
```diff
@@ -1,90 +1,61 @@
...
+})
+
+export function Trash() {
+  const classes = useStyles()
+
+  return <TrashIcon className={classes.trashIcon} />
```
dctalbot, line 25:
```suggestion
  return <TrashIcon className={classes.trashIcon} {...props} />
```
Let's pass through any `props` given to the component declaration. This makes the component more generic and flexible so we can do things like pass HTML attributes to the DOM e.g. `<Trash data-cy='my-trash-con' />`

dctalbot: On second thought, I just checked and we don't have any instances where we actually pass through any additional props, so I'm OK keeping this as is. We can add this back in later if needed.

dctalbot (changes requested):

web/src/app/icons/components/Icons.js
Forfold: I think in the essence of us moving away from that global style sheet, we could remove the references of `trashIcon` and `warningColor` from `globalStyles` and keep them flat with the component that uses it here. I checked to be sure and we don't reference those two styles in any other components

dctalbot: I can get on board with that too. Either delete them here, or delete them in the stylesheet, but don't have both

pnengchu: addressed, removed trashIcon and warningColor from styles/materialStyles.js

Forfold (approved):
lgtm

dctalbot (approved):

PR #1034, opened 2020-11-17
rotations: specify next handoff as a datetime

### LINKED ISSUE #993: Specify Rotation Start Date
While setting up a new 4 day rotation for our team, after the change the person "on call" is listed as having 2days, x hours, and y minutes left instead of 4days - (time since midnight today). This was surprising to me as i expected the rotation to start relative to today (when i made the change) instead of relative to some other unknown date time.

It would be useful to be able to set the "zero day" of the rotation. i.e. when calculating how much time until the next rotation, allow a user to set the "zero day" in that calculation. Currently it is unclear how that date is determined.

Example:
For an existing support rotation, i modify the rotation period to be 4 days, and i have 3 people in the rotation (A, B, C). 
Currently the rotation that is generated will look like so:
- A : 2 days x hours y minutes left 
- B: Starts at 12:00 am in 3 days
- C: Starts at 12:00 am in 5 days

It is unclear what date the remaining time the calculation is based on (what i am referring to as the "zero day" of the rotation.

My suggestion is that when changing the rotation period, you give the option to select a date for the rotation calculations to start relative to, or simply a checkbox to reset it to be relative to today at the "handoff time" (or another time such as midnight for cases where "handoff time" today will occur after the change is made). 
Example:
For an existing support rotation, i modify the rotation period to be 4 days, and i have 3 people in the rotation (A, B, C). I select the "zero day" of the rotation to be today.
Currently the rotation that is generated will look like so:
- A : 3 days x hours y minutes left 
- B: Starts at 12:00 am in 5 days
- C: Starts at 12:00 am in 7 days

All of the generated count downs are calculated relative to `(today at the handoff time) + (elapsed time since then) - 4 days`

As mentioned above the "zero day" reset could be made:
- as a simple checkbox and set to the most recent handoff to when the rotation schedule edit is made
- a date picker to set the relative date (may be useful for fixing rotations that were set up improperly by accident)

**Additional context:**
Not that i can think of.

dctalbot:

**Description:**
There is an issue described in #993 in which a rotation is created with an ambiguous "next rotation handoff time". This PR aims to address this by allowing users to select a specific datetime for an upcoming handoff.

Fixes #993 

**Screenshots:**
Screen Shot 2020-12-03 at 3 38 10 PM

**User-facing changes:**
- Removes weekday field for weekly rotations
- Default handoff time for a new rotation is the top of the upcoming hour
- Default handoff time for an existing rotation is the next upcoming handoff time
- Handoff time field is now an Initial Handoff Time datetime field
- Display next three upcoming handoffs (dynamically calculated based on form inputs) as a hint for what to expect

**Timezone rules:**
- If local zone is selected, option to configure in said zone is hidden
- If non-local zone is selected, option to configure in said zone is visible
- all times default to local zone, and are configured in local zone
- the timezone switch defaults to off and maintains its state (on / off) until a user interacts with it.


web/src/app/rotations/util.js
```diff
@@ -112,3 +112,35 @@ export function reorderList(_items, oldIndex, newIndex) {
   items.splice(newIndex, 0, _items[oldIndex]) // add dest to newIndex position
   return items
 }
+
+export function getNextHandoffs(count, handoffTime, rotationType, shiftLength) {
```
mastercactapus, line 116: This will need the correct configured zone to return the correct values.

mastercactapus (changes requested):
This will need a bit more work so that we handle timezones correctly, also the displayed times should probably show configured & local as we do on the details page:

image

web/src/app/rotations/util.js
dctalbot: I have the displayed values in this format since we were originally having users configure Handoff Time in their local timezone. Not sure what the alternative would be since I think ideally we would do something like 
```
Nov. 26, 2020 6:30PM EST (5:30 PM local)
```
but this is invalid across day boundaries i.e. local time may not necessarily be on the same date. So we would have to do
```
Nov. 26, 2020 6:30PM EST (Nov. 26, 2020 5:30PM local)
```
which is quite lengthy and verbose. Thoughts?

mastercactapus: I want to say we have a function that handles that discrepancy already (only show the day if there's a diff)

mastercactapus: That way it's only verbose on those edge cases, and at least it would always be clear & consistent

web/src/app/rotations/RotationForm.js
```diff
@@ -111,16 +108,49 @@ export default function RotationForm(props) {
...
+                  disableGutters
+                >
+                  <ListItemText
+                    primary={text}
+                    className={classes.noVerticalSpace}
+                  />
+                </ListItem>
+              )
+            })}
+          </List>
```
Forfold: Since you are creating a MUI list here but then removing the padding and such with:
- `dense`
- `disablePadding`
- `disableGutters`
- `className={classes.noVerticalSpace}`

We can just render as `Typography` as this will evaluate to multiple `p` tags under a `div`, which will stack as intended without padding, and removes the dependency on `makeStyles`.

```suggestion
          <Typography>
            <strong>Upcoming Handoff Times:</strong>
          </Typography>
          {nextHandoffs.map((text, i) => (
            <Typography key={i}>{text}</Typography>
          ))}
```

web/src/app/rotations/util.js
```diff
@@ -112,3 +112,39 @@ export function reorderList(_items, oldIndex, newIndex) {
...
+    hourly: 'hours',
+    daily: 'days',
+    weekly: 'weeks',
+  }
+  const luxonUnit = rotationTypeToLuxonUnit[rotationType]
+  const now = DateTime.utc()
+
+  let nextHandoff = DateTime.fromISO(handoffTime).setZone(inputTimeZone)
+  if (nextHandoff < now) {
+    while (nextHandoff < now) {
```
Forfold, line 134: Have you tested the performance impact of these? I'd be curious to see how it behaves if the start time is well in the past, as we allow any valid ISO timestamp be added as the initial handoff time within the DB.

Forfold (changes requested):
Typing in the handoff field causes the page to freeze momentarily

web/src/app/rotations/util.js
dctalbot: Can't think of a better way to calculate this while maintaining DST rules, but we can make sure the input is bound to some reasonable interval e.g. +/- 1 year

web/src/app/rotations/RotationForm.js
dctalbot: Trying to keep the elements semantic - Decided to go with `<Typography component='li' />`, with the wrapping grid item being a `ol`. Was definitely easier to manage the spacing this way.

web/src/app/rotations/util.js
mastercactapus: Given that the calculation is so complex, maybe it would be worthwhile to have a GraphQL query. As long as it's calculated on a separate bit of code there's going to be a real chance of discrepancies.

Something like:
```gql
nextRotationHandoffTimes(input:{start: "2020-12-01T14:07:48.172Z", zone: "America/Chicago", hours: 24, count: 3})
```

This will be more important in the future when rotations are part of schedule assignments/rules and things like weekday filters need to be taken into account. Best to write & test that code once I think.

dctalbot: The only concern I have with that is the latency, especially on a slow network. Even a small debounce interval might feel clunky. Thoughts?

mastercactapus: I don't think it would be an issue and I think we could get away without debouncing for this.

The actual calculation is negligible (most likely < millisecond since it's just a couple of math operations) so we're only talking network round-trip-time; in our environments, roughly 50ms. It's also not required/blocking for the user to continue and is similar to the phone number validation checkmark on those input fields.

Somewhat separately, I feel like it could always be visible, or visible if the start-time has been changed since I feel like that's almost always going to be helpful when updating that field. In any case, if it is showing we could put a spinner on it, in case the user is on a slow connection.

---

Alternatively, if we *do* opt for implementing it twice, take a look at the `timeutils` branch and `Rotation.EndTime` with the dependent code in `util/timeutil`. We could port that to js/luxon and implement the same set of unit tests, but it doesn't seem worth the hassle/maintenance to keep those in sync IMO

graphql2/graphqlapp/rotation.go
```diff
@@ -375,3 +375,27 @@ func (m *Mutation) UpdateRotation(ctx context.Context, input graphql2.UpdateRota
...
 }
+
+func (a *Query) UpcomingHandoffTimes(ctx context.Context, input *graphql2.UpcomingHandoffTimesInput) ([]time.Time, error) {
+	var result []time.Time
+
+	if input.Hours <= 0 || input.Count <= 0 {
+		return result, errors.New("invalid upcoming handoff times input")
+	}
+
+	loc, err := time.LoadLocation(input.TimeZone)
```
mastercactapus, line 386: This error needs to be checked and handled, also use the util.LoadLocation helper, shouldn't have made it past `make check`

graphql2/graphqlapp/rotation.go
```diff
@@ -375,3 +375,27 @@ func (m *Mutation) UpdateRotation(ctx context.Context, input graphql2.UpdateRota
 	}
 	return true, nil
 }
+
+func (a *Query) UpcomingHandoffTimes(ctx context.Context, input *graphql2.UpcomingHandoffTimesInput) ([]time.Time, error) {
+	var result []time.Time
+
+	if input.Hours <= 0 || input.Count <= 0 {
+		return result, errors.New("invalid upcoming handoff times input")
```
mastercactapus, line 383: This needs to be a field validation error for each. Use the validation helpers, otherwise, this is just going to come out as 500 unexpected error

graphql2/graphqlapp/rotation.go
```diff
@@ -375,3 +375,27 @@ func (m *Mutation) UpdateRotation(ctx context.Context, input graphql2.UpdateRota
 	}
 	return true, nil
 }
+
+func (a *Query) UpcomingHandoffTimes(ctx context.Context, input *graphql2.UpcomingHandoffTimesInput) ([]time.Time, error) {
+	var result []time.Time
+
+	if input.Hours <= 0 || input.Count <= 0 {
```
mastercactapus, line 382: We need to put a limit on `Count`

graphql2/schema.graphql
```diff
@@ -37,6 +37,8 @@ type Query {
   # Returns a paginated list of rotations.
   rotations(input: RotationSearchOptions): RotationConnection!
 
+  upcomingHandoffTimes(input: UpcomingHandoffTimesInput): [ISOTimestamp!]!
```
mastercactapus, line 40: This is rotation specific, maybe `nextRotationHandoffTimes` or something to that regard to match the existing `Rotation.nextHandoffTimes`?

web/src/app/rotations/RotationForm.js
```diff
@@ -1,54 +1,81 @@
...
+})
+
 export default function RotationForm(props) {
-  function dayOfWeek() {
-    const { start, timeZone } = props.value
-    return DateTime.fromISO(start, { zone: timeZone }).weekday
-  }
+  const { value } = props
+  const classes = useStyles()
+  const localZone = useMemo(() => DateTime.local().zone.name, [])
```
mastercactapus, line 44: Does this need memoization? The zone name should be stable

web/src/app/rotations/RotationForm.js
```diff
@@ -1,54 +1,81 @@
...
+  const [configInZone, setConfigInZone] = useState(false)
+  const configZone = configInZone ? value.timeZone : 'local'
+  const [upcomingHandoffs, setUpcomingHandoffs] = useState([])
```
mastercactapus, line 47: Don't use state for data, just render the data from apollo

web/src/app/rotations/RotationForm.js
```diff
@@ -1,54 +1,81 @@
...
+  useEffect(() => {
+    if (data?.upcomingHandoffTimes) {
+      const upcomingHandoffTimes = data.upcomingHandoffTimes.map((iso) =>
+        DateTime.fromISO(iso)
+          .setZone(configZone)
+          .toLocaleString(DateTime.DATETIME_FULL),
+      )
+      setUpcomingHandoffs(upcomingHandoffTimes)
+    }
+  }, [configInZone, data])
```
mastercactapus, line 78: As above, don't use the state as a cache, apollo already does that.

web/src/app/rotations/RotationForm.js
```diff
@@ -111,16 +127,62 @@ export default function RotationForm(props) {
...
           <FormField
             fullWidth
-            component={ISOTimePicker}
-            label='Handoff Time'
+            component={ISODateTimePicker}
+            timeZone={configZone}
+            label='Initial Handoff Time'
             name='start'
+            min={minStart.toISO()}
+            max={maxStart.toISO()}
```
mastercactapus, line 166: Maybe just `+/-` a year from the current time?

web/src/app/rotations/RotationForm.js
```diff
@@ -111,16 +127,62 @@ export default function RotationForm(props) {
...
+            Upcoming Handoff times:
+          </Typography>
+          {upcomingHandoffs.map((text, i) => (
+            <Typography
+              key={i}
+              component='li'
+              className={classes.listNone}
+              variant='body2'
+            >
+              {text}
```
mastercactapus, line 182: Just pull from data
```suggestion
              {DateTime.fromISO(iso).setZone(configZone).toLocaleString(DateTime.DATETIME_FULL)}
```

web/src/app/rotations/util.js
```diff
@@ -112,3 +112,12 @@ export function reorderList(_items, oldIndex, newIndex) {
...
 }
+
+export function getHours(count, unit) {
+  const lookup = {
+    hourly: 1,
+    daily: 24,
+    weekly: 24 * 7,
+  }
+  return lookup[unit] * count
+}
```
mastercactapus, line 123: If it lives here it should have documentation on what it does, supported units, etc... but feels like noise -- it's only used in one place and spreads the code/logic out unnecessarily.

web/src/app/util/ISOPickers.js
```diff
@@ -17,14 +17,16 @@ function hasInputSupport(name) {
...
-  { value, onChange, timeZone, ...otherProps },
+  { value, onChange, timeZone, min, max, ...otherProps },
   { format, truncateTo, type, Fallback },
 ) {
   const native = hasInputSupport(type)
   const params = useSelector(urlParamSelector)
   const zone = timeZone || params('tz', 'local')
-  const dtValue = DateTime.fromISO(value, { zone })
-  const [inputValue, setInputValue] = useState(dtValue.toFormat(format))
+  const dtValue = value ? DateTime.fromISO(value, { zone }) : null
```
mastercactapus, line 26: It looks like a few changes were mixed in for empty value support, but there's no use case here to actually test/validate in this PR, it would reduce noise to leave it out. I could be missing a use case though

mastercactapus (changes requested):

web/src/app/rotations/RotationForm.js
dctalbot: Thanks, this is an artifact from when I was doing the calculation client side

graphql2/schema.graphql
dctalbot: personally, I find those names a little too similar, but I would like to see "rotation" in the name. How about `upcomingRotationHandoffTimes` ?

web/src/app/rotations/RotationForm.js
dctalbot: Without this, the dialog flickers since data is `undefined` momentarily while the calculation is being performed. There's also the case in which the input field is in an invalid state e.g. empty and this cache allows us the render the "last valid input".

web/src/app/util/ISOPickers.js
dctalbot: In this PR, we actually want the field to _not_ be clearable. But there are other places in our app (at least in temporary schedules) where we do want clearable datetime inputs.

I will strip out the clearable bits from this PR but keep the `min` and `max` props which are used by the pickers.

graphql2/schema.graphql
mastercactapus: A reason for keeping them similar is that they are the exact same data, one a property of the rotation with params pre-filled and the other top-level.

I'd rather be consistent with the adjective across the UI than use "next" in some places and "upcoming" elsewhere.

Another approach could be to call this one `calcRotationHandoffTimes` and take an optional `from` parameter (timestamp) that defaults to the backend clock. It might be a good idea anyhow so that the times are always consistent with the user's machine.

web/src/app/rotations/RotationForm.js
mastercactapus: `data || []` will produce the same result without using the state if it's undefined.

Which input field becomes invalid? Generally, the form value should always remain valid and an invalid input should be reset to "last correct value" on blur -- sounds like a potential bug. Otherwise, it would be better to store the "last valid input" in the state and use that to fetch the data as a source of truth than to try to cache something down the chain.

dctalbot: I'm not sure that will help. The issue is that `data` would be populated, cleared, and repopulated causing us to render 3 lines of text, 0 lines of text, and then 3 lines of text again. So once `data` is populated, it should always be populated.

mastercactapus: Hmm. Maybe we can use placeholder text in the case the data is missing/invalid to solve for that. Showing the wrong times doesn't seem like the right answer either, even if it's temporary. We need to account for slow network and errors

There's also the issue of the flicker when the dialog first opens and we don't have anything. We can solve both with some placeholder text I think. Most of the time it's going to load quickly, and it would be good to give the user some indication that the data is fresh/valid anyhow.

Maybe a `Refreshing...`, `Loading...`, or `Calculating...` first line and the next two with some non-breaking space. We could also render in disabled style to make it feel really polished `disabledStyle={loading} errorStyle={error}` or something.

Maybe something like this?
`error ? [error, '\xa0', '\xa0'] :  (data || ['Calculating...', '\xa0', '\xa0'])`

dctalbot: Sounds good, I'll try a few things and report back

dctalbot: https://user-images.githubusercontent.com/17692467/102817471-76f6b580-4395-11eb-8b51-4bec984a89eb.mov

This is about the best I can do without using a stateful cache or debounce. Not perfect, but maybe good enough?

graphql2/schema.graphql
dctalbot: I made the change to `nextRotationHandoffTimes`. Not sure I fully understand your latter suggestion, but lmk if you'd like to go forward with that change too

graphql2/schema.graphql
```diff
@@ -655,6 +657,13 @@ input RotationSearchOptions {
   favoritesFirst: Boolean = false
 }
 
+input NextRotationHandoffTimesInput {
+  start: ISOTimestamp!
+  timeZone: String!
```
Forfold, line 662: What do you think of making this field optional, defaulting to the timeZone of the rotation as output?

graphql2/schema.graphql
```diff
@@ -655,6 +657,13 @@ input RotationSearchOptions {
...
+  hours: Int!
```
Forfold, line 663: I feel like `end: ISOTimestamp!` makes more sense instead of an `hours` duration, as this would match how we query for shifts and overrides within gql

web/src/app/rotations/RotationForm.js
```diff
@@ -1,54 +1,94 @@
...
+  flex: {
+    display: 'flex',
+  },
+  height7rem: {
+    height: '7rem',
+  },
+  noSpacing: {
+    margin: 0,
+    padding: 0,
+  },
```
Forfold, line 46: nit: when using classNames, I generally set the name relative to what I'm changing within the JSX, and let the CSS do the talking

For example (though you don't have necessarily to use these):

```suggestion
  handoffTimestamp: {
    listStyle: 'none',
  },
  handoffsTitle: {
    fontWeight: 'bolder',
  },
  tzContainer: {
    display: 'flex',
  },
  handoffsTitleContainer: {
    height: '7rem',
  },
  handoffsList: {
    margin: 0,
    padding: 0,
  },
```

web/src/app/rotations/RotationForm.js
```diff
@@ -1,54 +1,94 @@
...
+        timeZone: value.timeZone,
+        hours: getHours(value.shiftLength, value.type),
+        count: 3,
+      },
+    },
+  })
+
+  const isCalculating = !data || loading
+
+  const nextHandoffs = isCalculating
```
Forfold, line 85: If you put this in a `useEffect` where we only update the array when new data comes in from `nextRotationHandoffTimes`, we should see the flicker go away

web/src/app/rotations/RotationForm.js
```diff
@@ -1,54 +1,94 @@
...
-                {day}
-              </MenuItem>
-            )
-          })}
-        </TextField>
-      </Grid>
-    )
-  }
+export default function RotationForm(props) {
+  const { value } = props
```
Forfold, line 61: nit:
```suggestion
export default function RotationForm({ value }) {
```

Forfold (changes requested):

graphql2/graphqlapp/rotation.go
```diff
@@ -375,3 +375,36 @@ func (m *Mutation) UpdateRotation(ctx context.Context, input graphql2.UpdateRota
...
+		validate.Range("input.count", input.Count, 0, 20),
+		validate.Range("input.hours", input.Hours, 0, 99999),
+	)
+	if err != nil {
+		return result, err
+	}
+
+	loc, err := util.LoadLocation(input.TimeZone)
+	if err != nil {
+		return result, errors.Wrap(err, "failed to load location for "+input.TimeZone)
```
mastercactapus, line 394: Errors should be more direct like `load tz location: not found`. However in this case, the wrapper is redundant as `LoadLocation` will return the name if relevant already (e.g., `unknown time zone asdf`).

Additionally, this should be returned as a field validation error.

graphql2/graphqlapp/rotation.go
```diff
@@ -375,3 +375,36 @@ func (m *Mutation) UpdateRotation(ctx context.Context, input graphql2.UpdateRota
...
+		Type:        rotation.TypeHourly,
+	}
+
+	t := time.Now()
+	for len(result) < input.Count {
+		t = rot.EndTime(t)
+		result = append(result, t)
+	}
+
+	return result, err
```
mastercactapus, line 409: Use `return result, nil` here instead since the err has already been checked, for clairity.

graphql2/schema.graphql
mastercactapus: We don't have a rotation (no ID) as this is a top-level query

graphql2/schema.graphql
mastercactapus: I think this is a misunderstanding, this is the shift length not a duration for the query. Maybe we should call it `shiftLengthHours`?

web/src/app/rotations/RotationForm.js
mastercactapus: We moved away from that to keep from duplicating/double-caching data. We also don't want to show stale data while fetching with no indication of it being invalid (e.g., delayed network request).

This would be a good use-case for that early spinner design we had -- show the old/no spinner if the new comes within < 100ms otherwise spin for at least 1 second etc.. Normal use will be to update the handoff zero or one time and a flicker on that update is probably an okay compromise.

mastercactapus (changes requested):

graphql2/schema.graphql
dctalbot: Yeah, `end` doesn't really work here. We want X number of upcoming handoff times regardless of how far into the future they might be. I'm good with `shiftLengthHours`.

web/src/app/rotations/RotationForm.js
Forfold: Gotcha, that makes sense. We could alternatively find and replace only the characters that change, rather than the whole container. I'm not sure how it would play out, but could potentially work in our favor.

web/src/app/rotations/RotationForm.js
dctalbot: We pass `props` into `FormContainer` further down so this won't work

graphql2/schema.graphql
Forfold: From an API standpoint (black box), I'm not sure why I would need to provide the rotation's shift length when asking for the next handoff times. Wouldn't `count` be sufficient?

graphql2/schema.graphql
Forfold: Ah I see, I think that's what has been confusing me. When I see a query of `NextRotationHandoffTimesInput`, I expect to give it (or have available) a rotation ID. `Next` also sounds to me like there is also a rotation that already exists, and we are getting the next vals. I think different verbiage like `Calculate` or `Anticipate` would be more appropriate

graphql2/schema.graphql
dctalbot: I think you've got more context now as per the conversation above, but the thing to realize is this is basically a stand-alone utility function. It's more of a function than a query, but we wanted to push the computation to the server to keep our timezone rules consistent. 

It doesn't have a relationship with the rotation data model. It answers the question: "If I configure a rotation in this way, what would be the next X handoff times after `now`?"

graphql2/schema.graphql
mastercactapus: `calcRotationHandoffTimes` and `CalcRotationHandoffTimesInput` maybe?

web/src/app/rotations/RotationForm.js
mastercactapus: Not sure I know what you mean, React already reconciles the virtual DOM after render, and only updates the changed bits. The flicker is because we're rendering times, then the spinner, then the new times again in a short timeframe.

Maybe we can look into tools/methods of debouncing render for these situations? Maybe we should keep that out of scope of this PR though, I don't think the flicker on change detracts enough from the use to block it.

Forfold (approved):
lgtm!

graphql2/schema.graphql
dctalbot: This enhancement is being tracked here https://github.com/target/goalert/issues/1141

Forfold (approved):

mastercactapus (approved):

PR #1037, opened 2020-11-17
mobile: initialize with expo

Forfold:

**Description:**
This PR introduces the folder `/mobile` at root which holds the start of GoAlert's mobile application, built in React Native using Expo. I've included a home and a settings screen utilizing `react-navigation` (see screenshot below).

Run `yarn start` from within this directory to spin up the mobile app.

**Screenshots:**


Forfold:
I haven't heard of yarn workspaces before, so I'll check it out!

dctalbot (changes requested):
We should remove any changes under `web/`, resolve conflicts with master

Forfold:
Good catch, the lockfile must have slipped through the cracks from the reverts. I'll be sure to remove it and make sure `web` is in a clean state 👍

dctalbot (dismissed):
Functionally validated on:
- physical iPhone 7, iOS 13
- virtual iPhone 11, iOS 13
- virtual Pixel 2, Android 11

mastercactapus (changes requested):
Can we get this integrated with the Makefile?

Something like `make start MOBILE=1`? That should handle running `yarn` in the mobile directory (frozen lockfile) just like we do with the web UI, starting the backend, etc...

Also, it looks like `expo-cli` is missing as a required dependency for that to work properly -- we'll want to track it in package.json so we're not dealing with version mismatch issues, have to document/explain how to install it, dangers of using npm with sudo, etc...

Ideally, as long as docker/node/Go is installed, `make start MOBILE=1` should result in all necessary dependencies being installed and expo starting, the backend, and the web UI. We could have a separate start command, but I think in most cases (at least to start) you'll still need/want the web interface in addition to mobile for things like linking/login.

Lastly, I forget where we landed, but should this be going into a feature branch until we have the MVP working (login + alerts)?

mastercactapus:
Also, I tried playing with workspaces locally, seems to really speed things up, and would probably go a long way to simplify dep handling /w the Makefile

Forfold:
I'm also a fan of all of the above (`Makefile` additions, `expo-cli` tracking, and using a feature branch) 👍

Since we want to hold off on merging into master, how does adding those enhancements as follow-up issues/PRs sound so we can start to divvy up the work? I'll get a feature branch started and update the base of this PR

mastercactapus:
Closing due to age; will do a new spike when we cycle back to mobile.

PR #1050, opened 2020-11-19
ui/hooks: policy step card

KatieMSB:

**Description:**
Converts PolicyStepCard.js to use hooks.

Part of #923

### DIFF
--- a/web/src/app/escalation-policies/PolicyStepsCard.js
+++ b/web/src/app/escalation-policies/PolicyStepsCard.js
@@ -100,7 +100,7 @@ function PolicyStepsCard(props) {
       newIdx.current = null
     },
     onError: (err) => setError(err),
-    onUpdate: (cache, data) => onStepUpdate(cache, data),
+    update: (cache, { data }) => onStepUpdate(cache, data),
     optimisticResponse: { updateEscalationPolicy: true },
   })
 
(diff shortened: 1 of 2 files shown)


web/src/app/escalation-policies/PolicyStepsCard.js
```diff
@@ -1,80 +1,64 @@
...
+  const { escalationPolicyID, repeat, steps } = props
 
-  state = {
-    error: null,
-  }
+  const width = useWidth()
+
+  let oldID = null
+  let oldIdx = null
+  let newIdx = null
```
dctalbot, line 43: These won't be persisted across renders. Instead, we can use the `useRef` hook to create containers. See https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables

web/src/app/escalation-policies/PolicyStepsCard.js
```diff
@@ -105,42 +89,51 @@ export default class PolicyStepsCard extends Component {
     })
   }
 
-  handleDragStart = () => {
+  const [stepChange] = useMutation(mutation, {
```
dctalbot, line 92:
```suggestion
  const [updateEscalationPolicy] = useMutation(mutation, {
```
Nit maybe, but not immediately clear to me what `stepChange` means

dctalbot (changes requested):

dctalbot (approved):

mastercactapus (approved):

PR #1059, opened 2020-11-24
db: performance tweaks

mastercactapus:

**Description:**
This PR addresses a couple of recently uncovered performance/debug issues:
- Added an index specific for alert cleanup
- Tell the planner to avoid sequential scans where possible
- Wrap cleanup manager errors to aid in debugging if they slow down/timeout

**Additional Context:**
The alert cleanup query can start to bog down with a large number of alerts. To keep it operating efficiently a partial index was added. Debugging the issue would have been much easier if the timeout error included the operation being performed, so error wrapping was added.

There were cases where the `alert_logs` table was being scanned every engine cycle while processing status updates. Avoiding sequential scans brought the time from 6 seconds to 60ms. In just about all cases where we have a purpose-specific index we would want to have the planner use the index first so `enable_seqscan` has been set to `off` in the connection string. It was also found to speed up various other queries in the engine (/w default postgres config).


dctalbot (approved):

m17ch (approved):
LGTM 👍

PR #1063, opened 2020-11-24
rotations: warn of unexpected rotation advancement

dctalbot:

**Description:**
This PR logs a warning message if a rotation advances unexpectedly outside the bounds of the specified threshold (15 minutes)

### DIFF
--- a/engine/rotationmanager/advance.go
+++ b/engine/rotationmanager/advance.go
@@ -38,8 +38,7 @@ func calcAdvance(ctx context.Context, t time.Time, rot *rotation.Rotation, state
 		return nil
 	}
 
-	warningThreshold := time.Minute * -15
-	if !newStart.After(t.Add(warningThreshold)) {
+	if !newStart.After(t.Add(-15 * time.Minute)) {
 		log.Debugf(ctx, "unexpected rotation advancement")
 	}
 
--- a/engine/rotationmanager/advance.go
+++ b/engine/rotationmanager/advance.go
@@ -1,8 +1,11 @@
 package rotationmanager
 
 import (
-	"github.com/target/goalert/schedule/rotation"
+	"context"
 	"time"
+
+	"github.com/target/goalert/schedule/rotation"
+	"github.com/target/goalert/util/log"
 )
 
 type advance struct {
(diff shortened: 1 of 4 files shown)


engine/rotationmanager/advance.go
```diff
@@ -35,6 +38,11 @@ func calcAdvance(t time.Time, rot *rotation.Rotation, state rotation.State, part
 		return nil
 	}
 
+	warningThreshold := time.Minute * -15
+	if !newStart.After(t.Add(warningThreshold)) {
```
mastercactapus, line 42: Use `value * unit` for readability, also no need for a variable, it is only used once, and doesn't change
```suggestion
	if !newStart.After(t.Add(-15 * time.Minute)) {
```

engine/rotationmanager/advance.go
```diff
@@ -35,6 +38,11 @@ func calcAdvance(t time.Time, rot *rotation.Rotation, state rotation.State, part
...
+		log.Debugf(ctx, "unexpected rotation advancement")
```
mastercactapus, line 43: Include the amount of time so the log entry is more useful, and be more specific in the message:
```suggestion
		log.Debugf(ctx, "rotation advanced late (%s)", t.Sub(newStart).String())
```

mastercactapus (changes requested):

engine/rotationmanager/advance.go
```diff
@@ -35,6 +38,10 @@ func calcAdvance(t time.Time, rot *rotation.Rotation, state rotation.State, part
 		return nil
 	}
 
+	if !newStart.After(t.Add(-15 * time.Minute)) {
+		log.Debugf(ctx, "rotation advanced late (%s)", t.Sub(newStart).String())
```
mastercactapus, line 42: So I've been thinking on this, and maybe this should be logged as an error -- as it does indicate a failure mode -- the wrong person was active for awhile since it's behind, and we're catching up.

The original idea with Debug was to prevent unnecessary noise, but the only case I can think of in normal operation that would trigger this would be a config change, but that behavior would be unexpected. In all others, at least in a production environment, it would indicate some type of problem with rotation processing.

Fixing the config change behavior by updating the shift start time would prevent that scenario from occurring as well.

@m17ch What do you think?

mastercactapus (changes requested):

engine/rotationmanager/advance.go
m17ch: Yeh - I think this makes sense to log as an error if it's not something we expect to see a lot and want visibility into it.

mastercactapus (approved):

mastercactapus (changes requested):
Looks like we will need an exception in `smoketest/harness/backend.go` for the new "rotation advanced late" error. We should still log it but not as a test error.

mastercactapus (approved):

Forfold (approved):

PR #1069, opened 2020-11-25
ui/schedules: hide schedule tz toggle when local zone is the same

### LINKED ISSUE #1067: ui: remove timezone toggle displays in favor of helper text
We should hide the tz toggle switch if the target's timezone is the same as the user's local timezone, as toggling the switch results in no operation taking place.

dctalbot:

**Description:**
This PR introduces a `useTimeZone()` utility hook that provides information on the local system time zone and the configured `tz` URL query paramter

Fixes #1067 

Users will not see an option to toggle schedule options in their system's local time zone


web/src/app/schedules/ScheduleOverrideForm.js
```diff
@@ -92,13 +96,15 @@ export default function ScheduleOverrideForm(props) {
...
-        </Grid>
+        {!isLocalTZ && (
+          <Grid item xs={12} sm={12} md={6}>
+            {/* Purposefully leaving out of form, as it's only used for converting display times. */}
+            <ScheduleTZFilter
+              label={(tz) => `Configure in ${tz}`}
+              scheduleID={scheduleID}
+            />
+          </Grid>
+        )}
```
mastercactapus, line 107: This logic needs to be inside `ScheduleTZFilter`, I think what you intended was just a check if `schedule.timeZone === DateTime.local().zone.name`

mastercactapus (changes requested):
As implemented, the switch is removed from only the override form, which is different from the rest of the schedules dialogs.

It also seems broken; the switch is always missing, one exception is if you have the URL param set, you can turn it off, and then the switch disappears.

web/src/app/schedules/ScheduleOverrideForm.js
dctalbot: In this case, if we want to hide the grid item which is a requirement here IMO, the logic has to be elevated to this component. I'm also a bit averse to the idea of having a component conditionally render `null` which has been confusing and not-obvious in my experience

mastercactapus: Will continue in issue

PR #1073, opened 2020-11-29
ui/users: fix race condition; check for session readiness

### LINKED ISSUE #1072: ui/users: issue displaying user profiles
**Describe the Bug:**
UI will inconsistently fail to render User Details views in the application
Once failure is manifested, navigation to any page is broken until refreshing to a valid state.
Since I don't have a reliable way to reproduce, I suspect there may be a race condition that happens to be mitigated by Chrome. I believe this is related to the query for active sessions of a given user (as opposed to the session of the current user). Otherwise, we should review if any native browser API's are leveraged with respect to session management.

This bug seems to be mitigated in Chrome and more prevalent in Firefox.

**Steps to Reproduce:**
1. Navigate to a user details page
2. If no error, refresh.
3. If no error, repeat step 2.

**Expected Behavior:**
Consistent rendering of user details page.

**Observed Behavior:**
Screenshot below after navigating to user details page.

**Screenshots/Stack Traces:**
Screen Shot 2020-11-29 at 3 05 55 PM

**Application Version:**
```html
<meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
<meta http-equiv="x-goalert-version" content="v0.25.0-547-g12a97514">
<meta http-equiv="x-goalert-build-date" content="2020-11-03T16:42:00Z">
<meta http-equiv="x-goalert-git-commit" content="12a9751416a5ac53aa576954440861883b0111ab">
<meta http-equiv="x-goalert-git-tree-state" content="clean">
```

**Desktop:**
 - OS: 10.15.7
 - Browser Firefox
 - Version 84.0b5

**Additional Context:**
Confirmed by @arurao

dctalbot:

**Description:**
This fixes the issue linked below. The business logic for this code is the following: "you're only authorized to view your own sessions unless you're an admin"
There are two solutions for this:
1. Query everything but the sessions until it is confirmed that the current user is an admin. Then if confirmed, query for sessions.
2. Wait to confirm whether current user is an admin. Then perform appropriate query.

This PR implements the second solution which has a slower TTI, but is a more reliable experience. I'm open to more discussion on the tradeoffs here.

Fixes #1072

### DIFF
--- a/web/src/app/users/UserDetails.js
+++ b/web/src/app/users/UserDetails.js
@@ -96,18 +96,25 @@ function serviceCount(onCallSteps = []) {
 export default function UserDetails(props) {
   const classes = useStyles()
 
-  const { userID: currentUserID, isAdmin } = useSessionInfo()
+  const {
+    userID: currentUserID,
+    isAdmin,
+    ready: isSessionReady,
+  } = useSessionInfo()
   const [disclaimer] = useConfigValue('General.NotificationDisclaimer')
   const [createCM, setCreateCM] = useState(false)
   const [createNR, setCreateNR] = useState(false)
   const [showVerifyDialogByID, setShowVerifyDialogByID] = useState(null)
 
-  const { data, loading, error } = useQuery(
+  const { data, loading: isQueryLoading, error } = useQuery(
     isAdmin || props.userID === currentUserID ? profileQuery : userQuery,
     {
       variables: { id: props.userID },
+      skip: !isSessionReady,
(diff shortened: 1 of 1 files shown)


mastercactapus (approved):

Forfold (approved):
This solution works for me! It follows the same flow we use in much of the app, so I'm comfortable with this solution. Especially since like you said- it's reliable as we can simply utilize Apollo's `skip` option

PR #1087, opened 2020-12-01
repo: add stale bot config

Forfold:
Adds [Stale](https://github.com/apps/stale) to our repo which will manage stale issues and pull requests. I aimed to use some reasonable defaults for now as we are still growing within the open-source community:

- Issues and PRs will be labeled with (stale) after 30 days and a comment will be added.
- Issues will be closed after 1 year.
- PRs will be closed after 3 months.

With our current workflow and pace of incoming external support, this seems to be a good default for how we currently operate in real time when while managing our backlog.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,24 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+daysUntilStale: 30
+staleLabel: stale # Label to use when marking as stale
+limitPerRun: 30 # Limit the number of actions per hour, from 1-30. Default is 30
+exemptLabels:
+  - security
+
+pulls:
+  daysUntilClose: 90
+  markComment: >
+    This pull request has been automatically marked as stale because it has not had
+    recent activity. It will be closed if no further activity occurs. Thank you
+    for your contributions.
+  closeComment: false
+
+issues:
+  daysUntilClose: 365
+  markComment: >
+    This issue has been automatically marked as stale because it has not had
+    recent activity.
+  closeComment: >
+    This issue has been automatically closed because it has not had any activity within
+    the past year. Thank you for your contributions.
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,24 +1,18 @@
 # Configuration for probot-stale - https://github.com/probot/stale
 
-daysUntilStale: 30
 staleLabel: stale # Label to use when marking as stale
 limitPerRun: 30 # Limit the number of actions per hour, from 1-30. Default is 30
 exemptLabels:
   - security
 
 pulls:
-  daysUntilClose: 90
+  daysUntilStale: 30
   markComment: >
     This pull request has been automatically marked as stale because it has not had
-    recent activity. It will be closed if no further activity occurs. Thank you
-    for your contributions.
-  closeComment: false
+    recent activity. Thank you for your contributions.
 
 issues:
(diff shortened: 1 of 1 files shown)


Forfold:
I initially was thinking the same thing– only mark things as stale for tracking purposes. Though after some thought I think it would be best for us long-term to have these types of things managed for us, as we are moving towards using this repo (and GH projects) as the sole source for our work/productivity.

With how the GoAlert team has worked in the past, we did decide at one point that issues after a year aren't worth holding onto, as priorities often change. Looking at our internal tracking tool in its current state, it would take a good deal of effort IMO to ensure the stories listed are actually relevant, groomed or ungroomed.

While 90 days may seem harsh up front for closing a PR, any activity on it such as a comment or a commit will reset that timer.

I did think about the enhancements bit as well, and the reason I excluded it from the `exemptLabels` section is because enhancements are also prone to change– for instance at this point our alerts redesign issue is pretty stale now and no longer a priority for the team as we think about current customer wants. I'd be more in favor of reopening enhancements when we have the capacity to properly invest the time in it, since they will still exist within the repo. We will always have our closed issues and pull requests to filter through, but I think it's important to keep a clean backlog.

mastercactapus:
For exempt labels we should add in progress, and maybe like a pinned one, or "accepted" for enhancements in their example. For instance while SMTP isn't actively being worked on at the moment, there is in flight work to make it happen.

Alternatively we could start with just marking things as stale, since in our current state I feel like that would be the most beneficial option, and then just use that as things to revisit and decide manually if we should close or not.

One thing that's different for goalert, at least right now, compared to bigger projects is we don't have an unruly amount of issues/prs. If we had hundreds or thousands, automatically closing things would be absolutely necessary. But we can easily walk through every issue and PR we have without problem right now, so I don't see cleaning up / closing as a huge necessity atm. 

But we do struggle at times with things getting stale, and adding visibility to that could be really helpful.

$0.02

m17ch:
Yeh, I think adding the stale label (which would also cause us to receive a notification / have visibility) would be helpful, but I also don't think we need any automatic close rules (this may change in the future). 

After issues are marked stale, us getting into a regular routine of reviewing/pruning manually is probably a good habit for us to get in anyway to ensure things are up-to-date or still relevant.

Forfold:
All fair points, though I would love to hear from everyone else on the core team as well 😄 I definitely see what you guys mean in terms of closing vs. leaving stale since right now we still only have a couple of pages of issues and not even a second page of pull requests. And who's to say we can't change those settings in the future once we're all famous anyways 🥇

pnengchu:
I support this bot idea and also agree that issues should not be closed if labeled stale. Overall, adding a stale label brings visibility to which issues we can discuss as a team to decide which issues are relevant or should be closed. As a form of backlog grooming, we can go through which issues are no longer relevant or deserve priority. As of now, we have 40-something issues and this bot would help us stay organized and keep us engaged with bookkeeping.

Forfold:
@dctalbot @mastercactapus @m17ch @pnengchu 
Based on everyone's feedback, I'll go ahead and remove the config options for closing issues and PRs after a period of time.

Since issues are generally relevant to us for longer than 30 days, how does the following sound?

- Issues are marked stale after 6 months
- Pull requests are marked stale after 30 days

mastercactapus (approved):

arurao (approved):

PR #1103, opened 2020-12-09
engine/cleanupmanager: uncouple alert_logs from alert deletion

mastercactapus:

**Description:**
This PR addresses performance issues in the cleanup manager when deleting old alerts that have a large number of log entries.

- the `alert_logs` table no longer has foreign key constraints (previously SET NULL and CASCADE)
- cleanup manager will cleanup up to 1000 entries per cycle

Previously an action like deleting an alert (or by constraint, a service) or an integration key could time out due to the sheer number of log entries required to be updated or deleted. With this change `alert_logs` data will not change once added, other than being deleted when the associated alert no longer exists.

**Additional Context:**

Foreign key constraints were removed; all columns other than `alert_id` already could be null, so JOINs should work as before when something like an integration key is deleted and no longer has a matching entry.

There is a possibility for the log search query to return entries for alerts that have been deleted, but this should work normally with a `null` alert structure if requested.

The implementation determines the next 1000 `alert_log` ids and records the 1000th one for the next cycle. When it reaches the end it simply resets the counter to zero.

One weakness of this implementation is if the engine is restarted frequently, or more than 17 million rows are added per-day, some records may never get cleaned up. However, as intended to follow the alert cleanup, the oldest records are cleaned up first and are the most likely to be orphaned.


Forfold (approved):
Confirmed a lockout in master with an alert with 1,000,000 alert_logs:
```
Backend  (err): time="2020-12-17T10:51:12-08:00" level=error AuthSystemComponent=Engine Trigger=INTERVAL error="send outgoing messages: acquire exclusive locks: ERROR: canceling statement due to lock timeout (SQLSTATE 55P03)"
```

After switching branches, I functionally confirmed that the error was resolved and that 1,000 logs were cleaned up every engine cycle. Implementation LGTM!

arurao (approved):
LGTM

PR #1120, opened 2020-12-17
throttle: added bucketed throttling

### LINKED ISSUE #1084: SMS/Voice: Support bucketed throttling
SMS and Voice calls are set to 1 per-minute per-number maximum. This is fine for bursts, but even having as much as 16 calls in an hour is enough in some cases to get the From number temporarily blocked as spam for an entire carrier. Similar issues exist for SMS.

For example, a throttling model could be 1 call per minute, 3 per 15 minutes, 5 per hour.

Currently, there is no alternative other than hoping for the best :) 

**Additional context:**
This is a starting point, but additional research should be done to see real-world usage and any carrier recommendations that can be found.

Voice (per dest. number):
- 1 call per minute max 
- up to 3 per 15 minutes
- up to 5 per hour
- up to 10 per 3 hours

SMS (per dest. number):
- 1 per minute max
- up to 5 per 15 minutes
- up to 12 per hour
- up to 20 per 3 hours

Additionally, it could be helpful to add additional per-carrier global limits. Currently, the global limit is 1 per second for SMS and voice (each). If carrier identification is enabled in the GoAlert configuration, this could allow setting max limits (e.g., 100 calls per hour per-carrier or something) to further protect from unwanted filtering.

Note: It would be technically possible to delete & recreate a contact method to intentionally work-around this issue in the current way it's implemented. For this issue, however, that should be considered out-of-scope. This issue's primary goal is to account for normal usage while installing guards to prevent unwanted carrier filtering.

KatieMSB:

**Description:**
Adds bucketed throttling for sending messages following voice and SMS rules to help prevent carrier blocking.

Fixes #1084

**Additional Context**

Baseline (applies to all):
- 1 per minute

Status updates:
- 1 per 3 minutes
- 3 per 20 minutes
- 8 per 2 hours (rate limited)

Alerts (voice):
- 3 per 15 minutes
- 7 per hour (rate limited)
- 15 per 3 hours (rate limited)

Alerts (SMS):
- 5 per 15 minutes
- 11 per hour (rate limited)
- 21 per 3 hours (rate limited)


dctalbot (changes requested):
This looks really solid, nice work!

engine/message/db.go
```diff
@@ -763,65 +779,39 @@ func (db *DB) updateStuckMessages(ctx context.Context, statusFn StatusFunc) erro
...
-		case sent := <-resCh:
-			if !sent {
-				failures++
-			}
-			processing--
 		}
 	}
 
+	if failed {
+		return errors.New("one or more failures when sending")
```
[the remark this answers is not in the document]
mastercactapus: This would just add noise and wouldn't aid in debugging, each full error is already logged independently, this is just to indicate why the top-level method failed.

engine/message/db.go
```diff
@@ -763,65 +779,39 @@ func (db *DB) updateStuckMessages(ctx context.Context, statusFn StatusFunc) erro
...
 	}
-	defer close(sendCh)
 
-	var done bool
-	for {
-		if !done && sentCount() < toSend {
-			msg := q.NextByType(typ)
-			if msg == nil {
-				done = true
+	var failed bool
```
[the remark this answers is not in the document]
mastercactapus: It's preferred to use `var` with the type instead of `:=` when using the zero value (e.g., `false`, `0`).

```go
var failed bool
var count int
var buf bytes.Buffer

// vs.
failed := false
count := 0
buf := bytes.Buffer{}
// etc...
```

engine/message/db.go
```diff
@@ -360,21 +353,40 @@ func NewDB(ctx context.Context, db *sql.DB, c *Config, a alertlog.Store) (*DB, e
...
+
 func (db *DB) currentQueue(ctx context.Context, tx *sql.Tx, now time.Time) (*queue, error) {
-	rows, err := tx.Stmt(db.messages).QueryContext(ctx)
+	cutoff := now.Add(-maxMessageAge)
+	sentSince := db.lastSent
+	if sentSince.IsZero() {
+		sentSince = cutoff
+	}
+
+	// TODO: possibly dedup (e.g. multiple engines)
```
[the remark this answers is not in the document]
mastercactapus: This would be a net-new issue, so should be addressed here. IIRC it needs some attention to verify that we're not going to get the same message multiple times in `sentMessages` -- otherwise it would affect the limits.

The main issue is we are scanning for `sent_at >= <cutoff>`. If that returns a message sent after the transaction's `now` time we do want to include it, but the next tick will result in reading the same message again.

It may not be possible in practice right now, as the message module should operate exclusively. However, doing some deduplication via a map would ensure the counts are always accurate despite future changes.

engine/message/queue.go
```diff
@@ -23,6 +19,36 @@ var typePriority = map[Type]int{
 	TypeAlertStatusUpdateBundle: 4,
 }
 
+var perCMThrottle = ThrottleConfig{
```
mastercactapus, line 22: Maybe we should move these into `throttle.go` or `ratelimits.go` for visibility.

engine/message/throttle_test.go
```diff
@@ -0,0 +1,98 @@
...
+		SentAt:    n,
+	}
+
+	pending := throttle.InCooldown(msg)
+	require.Equal(t, pending, false)
+
+	throttle.Record(msg)
+
+	pending = throttle.InCooldown(msg)
+	require.Equal(t, pending, true)
```
mastercactapus, line 97: The assertion library order is `expected, actual` so these should be flipped `require.Equal(t, true, pending)`

engine/message/throttle_test.go
```diff
@@ -0,0 +1,98 @@
...
+		ServiceID: "Service B",
+		Dest:      notification.Dest{Type: notification.DestTypeSMS, ID: "SMS B"},
+		SentAt:    n.Add(15 * time.Minute),
+	}
+
+	throttle.Record(secondMsg)
+	require.NotNil(t, throttle)
+	require.Equal(t, throttle, &Throttle{
+		cfg:      cfg,
+		ignoreID: false,
```
mastercactapus, line 58: It's not super helpful to assert on the internal state/implementation of Throttle -- we only care that it works properly.

We can stick with the `InCooldown` test since that uses all relevant public methods (`NewThrottle` `Record` and `InCooldown`) and just rename it to `TestThrottle`.

We could actually change the package to `message_test` at the top of this file to ensure we are only testing the exported methods.

mastercactapus (changes requested):

arurao (approved):
Awesome work! :tada: LGTM

arurao (dismissed):

arurao (approved):
LGTM

dctalbot (approved):
Code looks good, but it feels like we should cover more scenarios in our tests. For example, I don't think the existing tests exercise the global throttle in a meaningful way.

engine/message/ratelimit.go
```diff
@@ -0,0 +1,52 @@
...
+// GlobalCMThrottle represents the rate limits for each notification type.
+var GlobalCMThrottle ThrottleConfig = ThrottleRules{{Count: 5, Per: 5 * time.Second}}
+
+// PerCMThrottle configures rate limits for individual contact methods.
+var PerCMThrottle ThrottleConfig
+
+func init() {
+	var perCM ThrottleConfigBuilder
+
+	// 1 per minute for all message types
```
dctalbot, line 18:
```suggestion
	// all message types
```

engine/message/ratelimit.go
```diff
@@ -0,0 +1,52 @@
...
+	perCM.AddRules([]ThrottleRule{{Count: 1, Per: time.Minute}})
+
+	// 1 per 15-minute for all status notifications
```
dctalbot, line 21:
```suggestion
	// status notifications
```

engine/message/ratelimit.go
```diff
@@ -0,0 +1,52 @@
...
+	perCM.
+		WithMsgTypes(notification.MessageTypeAlertStatus, notification.MessageTypeAlertStatusBundle).
+		AddRules([]ThrottleRule{
+			{Count: 1, Per: 3 * time.Minute},
+			{Count: 3, Per: 20 * time.Minute},
+			{Count: 8, Per: 120 * time.Minute, Smooth: true},
+		})
+
+	// alert-specific rules
```
dctalbot, line 30:
```suggestion
	// alert notifications
```

dctalbot (changes requested):

mastercactapus (approved):

engine/message/ratelimit.go
```diff
@@ -0,0 +1,52 @@
...
+
+	// alert notifications
+	alertMessages := perCM.WithMsgTypes(notification.MessageTypeAlert, notification.MessageTypeAlertBundle)
+
+	alertMessages.
+		WithDestTypes(notification.DestTypeVoice).
+		AddRules([]ThrottleRule{
+			{Count: 1, Per: time.Minute},
+			{Count: 3, Per: 15 * time.Minute},
+			{Count: 7, Per: time.Hour, Smooth: true},
```
arurao, line 38: Should we add a short comment here or upate the values in the PR description? If the values change because of `Smooth: true` might be helpful to have a one-liner.

arurao (changes requested):

mastercactapus (dismissed):

engine/message/ratelimit.go
mastercactapus: I'll update the PR description, the struct has a comment for what Smooth does already

engine/message/db.go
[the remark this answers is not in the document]
mastercactapus: this should now be resolved

mastercactapus (approved):

mastercactapus (dismissed):

dctalbot (dismissed):

arurao (approved):

dctalbot (approved):

PR #1147, opened 2020-12-28
cypress: upgrade to 6.2.0

dctalbot:

**Description:**
- Followed the [migration guide](https://docs.cypress.io/guides/references/migration-guide.html#Migrating-to-Cypress-6-0) from v5 to v6
- Reviewed full [changelog](https://docs.cypress.io/guides/references/changelog.html#6-2-0)
- Bug fix: multi select fields were not truly clearable. Now the input text will be cleared and the inline chips will be deleted when using `cy.form({inputName: ''})`
- Stability fix: added a `cy.reload()` before making assertions on the alerts page. While this may not be ideal, this is the precedent set in that test suite. I'm open to investigating this more, but IMO we are still exercising our code well enough. The main downside is the small perf impact for doing 3 page reloads

### DIFF
--- a/web/src/cypress/support/form.ts
+++ b/web/src/cypress/support/form.ts
@@ -146,7 +146,7 @@ function fillFormField(
               cy.get(selector).type(`{backspace}`)
             })
 
-          return cy.get(selector).clear()
+          return cy.get(selector)
         }
 
         if (DateTime.isDateTime(value)) {
--- a/web/src/cypress/support/form.ts
+++ b/web/src/cypress/support/form.ts
@@ -136,7 +136,18 @@ function fillFormField(
         .data('cyFallbackType')
 
       if (isSelect) {
-        if (value === '') return cy.get(selector).clear()
+        if (value === '') {
+          cy.get(selector).clear()
+
+          // clear chips on multi-select
+          el.parent()
+            .find('[data-cy="multi-value"] > svg')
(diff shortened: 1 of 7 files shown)


web/src/cypress/support/form.ts
```diff
@@ -136,7 +136,18 @@ function fillFormField(
...
+          cy.get(selector).clear()
+
+          // clear chips on multi-select
+          el.parent()
+            .find('[data-cy="multi-value"] > svg')
+            .each(() => {
+              cy.get(selector).type(`{backspace}`)
+            })
+
+          return cy.get(selector).clear()
```
Forfold, line 149: I'm curious why `cy.get(selector).clear()` is being called both before and after clearing the chips– were you seeing test failures without?

dctalbot: No failures. We need the first one in case there is outstanding text in the field (which is only possible if the field is active, but still an edge case to handle). The last clear is redundant as far as removing text, but I left it there in case cypress's `clear` method does anything interesting after the fact

dctalbot: I suppose the alternative would be to just return `cy.get(selector)`? That way we would keep it chainable

KatieMSB (approved):

Forfold (approved):

PR #1178, opened 2021-01-08
dev: add yarn check to make check

dctalbot:

**Description:**
This PR prevents unmet peer javascript dependencies from being merged


mastercactapus (changes requested):
Looks like using this is deprecated:
https://classic.yarnpkg.com/en/docs/cli/check/

It says to use `--check-files` with the `install` command instead

PR #1189, opened 2021-01-14
alerts smtp: add config validation 

arurao:

**Description:**
This PR adds validation for the config changes for smtp, when enabled. 
Validating that the `from` email address is a valid email and `SMTP` when enabled, makes sure that `from` and `address` fields are also specified.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -340,6 +340,7 @@ func (cfg Config) Validate() error {
 		),
 		validateEnable("SMTP", cfg.SMTP.Enable,
 			"From", cfg.SMTP.From,
+			"Address", cfg.SMTP.Address,
 		),
 	)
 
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -335,6 +338,9 @@ func (cfg Config) Validate() error {
 			"ClientID", cfg.OIDC.ClientID,
 			"ClientSecret", cfg.OIDC.ClientSecret,
 		),
+		validateEnable("SMTP", cfg.SMTP.Enable,
+			"From", cfg.SMTP.From,
+		),
```
mastercactapus, line 343: Also need address to connect to set:
```suggestion
			"From", cfg.SMTP.From,
			"Address", cfg.SMTP.Address,
		),
```

mastercactapus (changes requested):
We also need to validate that `Address` is set on

mastercactapus (approved):

KatieMSB (approved):

PR #1206, opened 2021-01-22
ui/schedules: hide time zone switch for schedules configured in local time zone

### LINKED ISSUE #1067: ui: remove timezone toggle displays in favor of helper text

dctalbot:

**Description:**
- `ScheduleTZFilter` converted to typescript
- custom labels using `label` prop
- optional `render` prop allows for wrapping switch element output
- hidden when system time zone === schedule's time zone
- modified column widths on the schedule override form to prevent default text from having an orphaned word

fixes #1067

Users will not be able to toggle time zones for schedules configured in the local zone (current behavior is effectively a no-op).


web/src/app/schedules/ScheduleRuleList.js
```diff
@@ -111,11 +111,16 @@ export default class ScheduleRuleList extends React.PureComponent {
...
         <PageActions>
-          <FilterContainer onReset={() => this.props.resetFilter()}>
-            <Grid item xs={12}>
-              <ScheduleTZFilter scheduleID={this.props.scheduleID} />
-            </Grid>
-          </FilterContainer>
+          <ScheduleTZFilter
+            scheduleID={this.props.scheduleID}
+            render={(children) => (
+              <FilterContainer onReset={() => this.props.resetFilter()}>
```
Forfold, line 117: What's going on here makes sense, but having the Wrapper/Container element nested in the Filter as a prop reads a little off. What is the main benefit in reversing the children logic here into a render prop?

Forfold (changes requested):

web/src/app/schedules/ScheduleRuleList.js
dctalbot: This is to prevent having an empty filter container since the only item in the FilterContainer is the time zone toggle. Because it's the only item, we can show/hide the filter container altogether.

Forfold:
Looks like we'll need to be watching if the timezone changes, if you change the timezone to your local zone (with the toggle currently set to true) and then re-open a form with the filter, it appears to get out of sync. I'm noticing this behavior in prod, too, though it can be resolved there by clicking the toggle twice

Screen Shot 2021-01-25 at 10 54 21 AM

Forfold:
I don't think we can reset the TZ filter on dialog close since we purposefully save that info in the URL so it stays consistent while viewing shifts/the calendar/etc. The TZ can also change while other people are working on the schedule, too.

We could likely poll for the timezone on the schedule/track TZ changes from the URL, and make the appropriate changes when either gets out of sync

Forfold:
We do intentionally leave the timezone in the URL for the whole schedule. I can see how not having it available as a filter from the main detail does make it awkward. Ultimately I feel a proper solution that I've mentioned a few times in the past would be a global setting for a user to choose which timezone to view anything in GoAlert in

mastercactapus:
I think this work may need to go through additional discussion first.

The current implementation (using the URL param) was intentional to ensure timestamp-in-timezone rendering is always consistent for the entire page (no possibility of mix/match) and persists in a refresh. It was also meant to be on the details page but the calendar lacks support so it's stuck local-only.

---

We should focus on: what problem are users facing today that we are trying to solve, and what would the ideal experience/solution be.

We can discuss other ideas as well that way, maybe a switch from one to the other isn't even the best approach in the first place. Then we can talk about what changes need to be made to achieve that without spinning during the review.

web/src/app/schedules/ScheduleTZFilter.tsx
```diff
@@ -13,25 +13,46 @@ const tzQuery = gql`
...
+  render?: (children: JSX.Element) => JSX.Element
+}
 
-export function ScheduleTZFilter(props) {
+export function ScheduleTZFilter(
+  props: ScheduleTZFilterProps,
+): JSX.Element | null {
   const [zone, setZone] = useURLParam('tz', 'local')
   const { data, loading, error } = useQuery(tzQuery, {
     pollInterval: 0,
```
Forfold, line 27: I think if we were to remove this line we would get the logic desired to fix the current bug + move forward with this PR

Forfold: Needs a slight bit more ^, but it's working so far with using an Effect as well to watch for changes:

```suggestion
  const [filterTZ, setFilterTZ] = useURLParam('tz', 'local')
  const resetTZ = useResetURLParams('tz')
  const { data, loading, error } = useQuery(tzQuery, {
    variables: { id: props.scheduleID },
  })
  const scheduleTZ = data?.schedule?.timeZone
  const localTZ = DateTime.local().zone.name

  useEffect(() => {
    if (filterTZ === localTZ || localTZ === scheduleTZ) {
      resetTZ()
    }
  }, [scheduleTZ])

  if (localTZ === scheduleTZ) return null
```

Issue #1239, opened 2021-02-03
sendit: update or use multiplexing alternative


Forfold:
**Is your feature request related to a problem? Please describe:**
The latest version of `github.com/hashicorp/yamux` is incompatible with `sendit` in its current state.

**Describe the solution you'd like:**
We'll need to either find an alternative multiplexing library, fix the compatibility issue with yamux and upgrade again, or create our own.

PR #1240, opened 2021-02-03
sendit: revert yamux version

Forfold:

**Description:**
Reverts `github.com/hashicorp/yamux` to a working version with sendit, as described in #1239


mastercactapus (approved):

m17ch:
Since we'll want yamux to stay on this version, should we add a comment in `go.mod` with something like:
``` // pinned version - see https://github.com/target/goalert/issues/1239 ``` 
or something similar (I foresee someone likely forgetting about this dependency when updating packages in the future.  

Thoughts @Forfold @mastercactapus ?

Forfold:
I was thinking about adding that but wasn't sure, so since you mentioned it I'll go ahead and add that 👍

m17ch (approved):

mastercactapus (approved):

Issue #1244, opened 2021-02-08
ui: in-app documentation has bad markdown formatting


dctalbot:
**Describe the Bug:**
In-app docs are formatted incorrectly

**Steps to Reproduce:**
1. Go to https://localhost:3030/docs
2. See error

**Expected Behavior:**
Proper formatting as markdown table

**Observed Behavior:**
See below

**Screenshots/Stack Traces:**
<img width="684" alt="Screen Shot 2021-02-08 at 3 04 21 PM" src="https://user-images.githubusercontent.com/17692467/107284629-b5b6fc80-6a23-11eb-8b3a-74f5813d1ddf.png">

**Application Version:**
`
fa7660842877a494d59edf0cfda612e913da7a5a
`

**Desktop:**
 - OS: Mac
 - Browser: Firefox 86

PR #1249, opened 2021-02-09
ui: fix markdown table rendering

dctalbot:

**Description:**
This PR fixes a regression with rendering tables with `react-markdown`. It now requires a separate dependency, `remark-gfm`

Fixes: #1244

### DIFF
--- a/web/src/app/documentation/components/IntegrationKeyAPI.js
+++ b/web/src/app/documentation/components/IntegrationKeyAPI.js
@@ -5,24 +5,16 @@ import Typography from '@material-ui/core/Typography'
 import markdownText from '../IntegrationKeys.md'
 import Markdown from '../../util/Markdown'
 
-function replaceAll(target, search, replacement) {
-  return target.split(search).join(replacement)
-}
-
-const replaceString = 'https://<example.goalert.me>'
-
 export default function IntegrationKeyAPI() {
-  const protocol = window.location.protocol || 'https:'
-  const host = window.location.host
+  const finalText = markdownText.replaceAll(
+    'https://<example.goalert.me>',
+    window.location.origin,
+  )
 
-  let finalText = markdownText
-  if (host) {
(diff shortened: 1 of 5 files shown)


web/src/app/documentation/components/IntegrationKeyAPI.js
```diff
@@ -5,24 +5,16 @@ import Typography from '@material-ui/core/Typography'
...
-
-const replaceString = 'https://<example.goalert.me>'
-
 export default function IntegrationKeyAPI() {
-  const protocol = window.location.protocol || 'https:'
-  const host = window.location.host
+  const finalText = markdownText.replaceAll(
+    'https://<example.goalert.me>',
+    window.location.origin,
+  )
```
mastercactapus, line 12: Not necessarily introduced by this PR, but this should be using the PublicURL from config, must have been missed when the HTTP prefix work was added

```suggestion
  const [pubURL] = useConfigValue('General.PublicURL')
  const finalText = pubURL
    ? markdownText.replaceAll('https://<example.goalert.me>', pubURL)
    : markdownText
```

mastercactapus (changes requested):

web/src/app/documentation/components/IntegrationKeyAPI.js
dctalbot: Just confirming the suggestion  - when public URL is blank, we want to display `https://<example.goalert.me>` in the docs?

mastercactapus: That was my intention, although as an alternative we could fall back to `origin + pathPrefix`, I'll let you decide which.

IIRC we don't have a way to get the generated/fallback PublicURL from the backend so that would be the next best thing since the prefix is embedded on the page.

mastercactapus (approved):

KatieMSB (approved):

PR #1252, opened 2021-02-10
schedule/rotation: timeutil package

mastercactapus:

**Description:**
This PR moves most time calculation logic into a new timeutil package. Careful attention has been paid to daylight savings changes.

Rules regarding DST are now well defined:
- Shift start times are processed as First of Day in wall-clock-time (e.g., BEFORE clocks are set back)
- Shift end times are processed as Last of Day in wall-clock-time (e.g., AFTER clocks are set back)
- Shift handoff during clocks set back WILL have overlap if the time is ambiguous
- Rotation handoff happens at First of Day when the time is ambiguous
- Hourly rotations are now aligned to wall-clock-time, meaning a 2-hour rotation will handoff on even hours, always (even during/after DST changes)


m17ch (approved):
LGTM 👍

KatieMSB (approved):

PR #1255, opened 2021-02-12
ui/hooks: page actions

kanish671:

**Description:**
Converted PageActions to use hooks

Part of #923

### DIFF
--- a/web/src/app/util/PageActions.js
+++ b/web/src/app/util/PageActions.js
@@ -55,7 +55,7 @@ export const PageActionProvider = (props) => {
       _pending.current = false
       debouncedSetActions.cancel()
     }
-  })
+  }, [])
 
   return (
     <PageActionsContext.Provider
@@ -85,7 +85,11 @@ const PageActionUpdater = (props) => {
       props.trackMount(false)
       props.setActions(null)
     }
-  })
+  }, [])
+
+  if (_mounted.current) {
+    props.setActions(props.children)
+  }
 
   return null
 }
--- a/web/src/app/util/PageActions.js
+++ b/web/src/app/util/PageActions.js
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect, useState } from 'react'
+import React, { useContext, useEffect, useRef, useState } from 'react'
 import p from 'prop-types'
 import { debounce } from 'lodash'
 
@@ -15,14 +15,13 @@ export const PageActionContainer = () => {
 
 export const PageActionProvider = (props) => {
   const [actions, setActions] = useState(null)
-
-  let _mountCount = 0
-  let _mounted = false
-  let _pending = null
+  const _mountCount = useRef(0)
+  const _mounted = useRef(false)
+  const _pending = useRef(null)
 
   const _setActions = (actions) => {
-    if (!_mounted) {
-      _pending = actions
+    if (!_mounted.current) {
+      _pending.current = actions
       return
     }
 
@@ -33,12 +32,12 @@ export const PageActionProvider = (props) => {
 
   const updateMounted = (mount) => {
     if (mount) {
-      _mountCount++
+      _mountCount.current++
     } else {
-      _mountCount--
+      _mountCount.current--
     }
 
-    if (_mountCount > 1 && global.console && console.error) {
+    if (_mountCount.current > 1 && global.console && console.error) {
       console.error(
         'PageActions: Found more than one <PageActions> component mounted within the same provider.',
       )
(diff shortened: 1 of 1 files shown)


web/src/app/util/PageActions.js
```diff
@@ -8,104 +8,90 @@ const PageActionsContext = React.createContext({
...
+  const [actions, setActions] = useState(null)
 
-  componentWillUnmount() {
-    this._mounted = false
-    this._pending = false
-    this.setActions.cancel()
-  }
+  let _mountCount = 0
+  let _mounted = false
+  let _pending = null
```
Forfold, line 21: I think the way these are evaluated in classes vs. hooks is a bit different than it seems- `this.foo = 'bar'` is a synchronous update, and that value will live in the "background" of the component allowing it to retain its value on every re-render. With hooks though, these will get reset back to `0`, `false`, and `null` on every re-render.

I think we can get around this in hooks if we initialize the variables outside of the component and use the `useEffect` hook, as it will only update the values on mount and unmount:

```js
let _mountCount = 0
let _mounted = false

const MyComponentExample = () => {
  useEffect(() => {
    // componentDidMount code here

    return () => {
      // componentWillUnmount code here
    }
  }, [])

  // ...
}
```

mastercactapus: Having them outside of the component could introduce really odd side-effects bugs/leaks, we can use `useRef` though -- https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables

And then `useEffect(..., [])` to handle the mount/unmount triggers.

kanish671: Hey thanks! I'll take a look at it and make the changes

kanish671: Sure, I'll take a look at it.

kanish671: I've made some changes. please take a look and let me know if it looks good.

web/src/app/util/PageActions.js
```diff
@@ -8,104 +8,91 @@ const PageActionsContext = React.createContext({
...
+    if (_pending.current) {
+      debouncedSetActions(_pending.current)
+      _pending.current = false
+    }
+    return () => {
+      _mounted.current = false
+      _pending.current = false
+      debouncedSetActions.cancel()
+    }
+  })
```
mastercactapus, line 58: Adding an empty dependency array at the end will ensure this `useEffect` only runs once on mount and once on unmount -- otherwise, it's going to call `debouncedSetActions` and `.cancel` on every render.
```suggestion
  }, [])
```

web/src/app/util/PageActions.js
```diff
@@ -8,104 +8,91 @@ const PageActionsContext = React.createContext({
...
 
-  render() {
-    if (this._mounted) {
-      this.props.setActions(this.props.children)
+    return () => {
+      _mounted.current = false
+      props.trackMount(false)
+      props.setActions(null)
     }
+  })
```
mastercactapus, line 88: Same as above, just add an empty array to have it only run on mount/unmount.

```suggestion
  }, [])
```

mastercactapus (changes requested):

kanish671:
@mastercactapus Addressed the review comments. Also, I had missed out a piece of code that was causing bugs when I was testing locally. Added that too.

kanish671:
@mastercactapus @Forfold is there anything else that needs to be done here?

web/src/app/util/PageActions.js
```diff
@@ -8,104 +8,95 @@ const PageActionsContext = React.createContext({
...
+    if (_pending.current) {
+      debouncedSetActions(_pending.current)
+      _pending.current = false
+    }
+    return () => {
+      _mounted.current = false
+      _pending.current = false
+      debouncedSetActions.cancel()
+    }
+  }, [])
```
Forfold, line 58: nit: move this useEffect block to the top of the component, under the `const _pending = useRef(null)` line

Forfold (approved):
LGTM! Only have 1 comment regarding formatting, so approving for CI + smoketests

web/src/app/util/PageActions.js
kanish671: I can move it under `const debouncedSetActions = debounce(_setActions)`. Otherwise,  eslint complains about using `debouncedSetActions` before defining it.

kanish671:
@mastercactapus could you also review this again?

mastercactapus (approved):

kanish671:
Thank you! 😄

Issue #871, opened 2020-09-17
Call alert script with arguments


hugoreadonlyuser:
**Is your feature request related to a problem? Please describe:**
We would like to  to send alert messages to whatsapp and or telegram . For instance use smooch api with a python script that receives the message and its values via arguments from goalert.
Currently we are locked on twilio which cannot configured to use whatsapp instead of SMS.

**Describe the solution you'd like:**
Having a media type alert or custom script that can be configured with arguments would open to use any message providers.

m17ch:
Hi @hugoreadonlyuser -- thanks for the enhancement suggestion.

Just gathering some additional data / feedback on what you have in mind for this:

You're not necessarily looking for direct WhatsApp or Telegram support, but a generic method of sending a webhook (with the alert details payload) to external services?

Let me know if you have any additional details for what you're thinking or if I am misinterpreting the request.  Thanks again!

forrayz:
Hi sorry I was using our hugo service acc for post :) A generic webhook method would be better. Twilio vendor lock is a show stopper for us.

ekeih:
Hi, we are also very interested in this. We would like to use goalert to send generic emails (smtp) and SMS with other providers. Maybe my thoughts about this provide some additional background that could help during decisions. I see the following options to support a broader range of outgoing alert target:

1. Implement SMTP handler for outgoing alerts. #45 
2. Implement handlers for other SMS providers one by one.
3. Implement an exec handler, that runs an external script for outgoing alerts.
4. Implement a webhook handler for outgoing alerts.

I think 1. would be cool, because it is already very generic and therefore could be part of the goalert core. 2. sounds like a never ending maintenance nightmare. 3. and 4. are universal solutions that would allow anyone to implement their own alert handlers without the need to change the goalert core. 3. reminds me of the [NotificationCommand of Icinga2](https://icinga.com/docs/icinga2/latest/doc/09-object-types/#objecttype-notificationcommand) and 4. of the [webooks in Prometheus Alertmanager](https://prometheus.io/docs/alerting/latest/configuration/#webhook_config). I think 3. is a bit easier for the enduser, but 4. is better suited for distributed systems like Kubernetes.

Personally I would love it if goalert implements 1., 3. and 4., but I would already be happy if just 3. or 4. works. However you decide, thank you very much for all the work you put into goalert and have a nice day 🙂

mastercactapus:
So maybe for MVP:
[...]

PR #1269, opened 2021-02-17
webhook as a contact method

dctalbot:

**Description:**
This PR provides a new Webhook contact method that will send alerts as POST requests to a URL specified by a user.

Part of #871 

**Screenshots:**
Screen Shot 2021-02-17 at 3 51 11 PM
Screen Shot 2021-02-17 at 3 57 42 PM
Screen Shot 2021-02-17 at 4 29 10 PM

Users will have new CM option, if enabled in admin config. See above images.

add `WEBHOOK` to `enum ContactMethodType`

**Additional Info:**
Reviewers may find it useful to read the in-app documentation (web/src/app/documentation/sections/Webhooks.md)

If they wish to validate functionality, I recommend using webhook.site


config/config.go
```diff
@@ -113,6 +113,11 @@ type Config struct {
 		Password string `password:"true" info:"Password for authentication."`
 	}
 
+	Webhook struct {
+		Enable    bool     `public:"true" info:"Enables webhook as a contact method."`
+		AllowList []string `public:"true" info:"If set, allows webhooks for these domains only."`
```
mastercactapus, line 118: Let's call this `AllowedURLs` following the naming of some of the other config options.

config/config_test.go
```diff
@@ -6,6 +6,21 @@ import (
...
+
+		assert.True(t, cfg.ValidWebhookURL("http://api.example.com"))
+	})
+	t.Run("checks if webhook urls are allowed", func(t *testing.T) {
+		var cfg Config
+		cfg.Webhook.AllowList = append(cfg.Webhook.AllowList, "http://api.example.com")
+
+		assert.True(t, cfg.ValidWebhookURL("http://api.example.com:5555/path"))
+		assert.False(t, cfg.ValidWebhookURL("http://example.com"))
+	})
```
mastercactapus, line 21: These can just be run flat
```suggestion
	var cfg Config

	assert.True(t, cfg.ValidWebhookURL("http://api.example.com"))

	cfg.Webhook.AllowList = append(cfg.Webhook.AllowList, "http://api.example.com")

	assert.True(t, cfg.ValidWebhookURL("http://api.example.com:5555/path"))
	assert.False(t, cfg.ValidWebhookURL("http://example.com"))
```

user/contactmethod/contactmethod.go
```diff
@@ -32,14 +32,16 @@ func (c ContactMethod) Normalize() (*ContactMethod, error) {
...
+		validate.OneOf("Type", c.Type, TypeSMS, TypeVoice, TypeEmail, TypePush, TypeWebhook),
 	)
 
 	switch c.Type {
 	case TypeSMS, TypeVoice:
 		err = validate.Many(err, validate.Phone("Value", c.Value))
 	case TypeEmail:
 		err = validate.Many(err, validate.Email("Value", c.Value))
+	case TypeWebhook:
+		err = validate.Many(err, validate.URL("Value", c.Value))
```
mastercactapus, line 44:
```suggestion
		err = validate.Many(err, validate.AbsoluteURL("Value", c.Value))
```

graphql2/graphqlapp/contactmethod.go
```diff
@@ -72,6 +73,14 @@ func (q *Query) UserContactMethod(ctx context.Context, id string) (*contactmetho
 
 func (m *Mutation) CreateUserContactMethod(ctx context.Context, input graphql2.CreateUserContactMethodInput) (*contactmethod.ContactMethod, error) {
 	var cm *contactmethod.ContactMethod
+
+	cfg := config.FromContext(ctx)
+
+	// checks webhook url against allow list set in config
+	if input.Type == "WEBHOOK" && !cfg.ValidWebhookURL(input.Value) {
```
mastercactapus, line 80:
```suggestion
	if input.Type == contactmethod.TypeWebhook && !cfg.ValidWebhookURL(input.Value) {
```

graphql2/graphqlapp/contactmethod.go
```diff
@@ -72,6 +73,14 @@ func (q *Query) UserContactMethod(ctx context.Context, id string) (*contactmetho
...
+		return cm, validation.NewFieldError("value", "Webhook URL not included in allow list.")
```
mastercactapus, line 81: Maybe something like "URL disallowed by administrator", an end-user will most likely not know about the "allow list".

notification/webhook/sender.go
```diff
@@ -0,0 +1,101 @@
...
+	"strconv"
+
+	"github.com/target/goalert/notification"
+)
+
+type Sender struct{}
+
+type WebhookAlert struct {
+	AlertID     int    `json:",omitempty"`
+	AlertType   string `json:",omitempty"`
```
mastercactapus, line 18:
```suggestion
	Type   string `json:",omitempty"`
```

notification/webhook/sender.go
```diff
@@ -0,0 +1,101 @@
...
+
+	case notification.AlertStatusBundle:
+		payload.AlertType = "AlertStatusBundle"
+		payload.AlertID = m.AlertID
+		payload.Count = m.Count
+		payload.LogEntry = m.LogEntry
+
+	default:
+		return nil, errors.New("message type not supported")
+	}
```
mastercactapus, line 84: I almost wonder if this should just go in a "renderMessage" method so we don't need to break out the `postWithBody` logic.

We can also leave out the static summary/details text since this is an API call (and not an end-user message) for test/verify. Also, returning an inline struct will also help to make it more scannable visually.

```go
func renderMessage(msg notification.Message) (*WebhookAlert, error) {
	switch m := msg.(type) {
	case notification.Test:
		return &WebhookAlert{
			Type: TypeTest
		}, nil
	case notification.Verification:
		return &WebhookAlert{
			Type: TypeVerification,
			Code: strconv.Itoa(m.Code),
		}, nil
	case notification.Alert:
		return &WebhookAlert{
			Type: TypeAlert,
			AlertID: m.AlertID,
			Summary: m.Summary,
			Details: m.Details,
		}, nil
	default:
		return nil, fmt.Errorf("unknown message type %T", m)
	}

}
```

notification/webhook/sender.go
```diff
@@ -0,0 +1,101 @@
...
+	"errors"
+	"net/http"
+	"strconv"
+
+	"github.com/target/goalert/notification"
+)
+
+type Sender struct{}
+
+type WebhookAlert struct {
```
mastercactapus, line 16: Generally, it's best to avoid "stutters" in names, this will be used as `webhook.WebhookAlert` maybe `POSTData` or `NotificationData` -- something descriptive to its use. Also, add comments for exported names.

notification/webhook/sender.go
```diff
@@ -0,0 +1,101 @@
...
+		payload.AlertType = "AlertStatusBundle"
+		payload.AlertID = m.AlertID
+		payload.Count = m.Count
+		payload.LogEntry = m.LogEntry
+
+	default:
+		return nil, errors.New("message type not supported")
+	}
+
+	jsonbytes, err := json.Marshal(payload)
```
mastercactapus, line 86: nit/convention: use `data` instead to represent a relevant set of bytes.
```suggestion
	data, err := json.Marshal(payload)
```

notification/webhook/sender.go
```diff
@@ -0,0 +1,101 @@
...
+
+func NewSender(ctx context.Context) *Sender {
+	return &Sender{}
+}
+
+// Send will send an alert for the provided message type
+func (s *Sender) Send(ctx context.Context, msg notification.Message) (*notification.MessageStatus, error) {
+
+	postWithBody := func(body string) (*http.Response, error) {
+		req, err := http.NewRequest("POST", msg.Destination().Value, bytes.NewBufferString(body))
```
mastercactapus, line 36: If we move the switch statement out we can move this inline, additionally we can use the `[]byte` directly instead of casting to a string and also skip the `WithContext` by using `NewRequestWithContext`.

```suggestion
		req, err := http.NewRequestWithContext(ctx, "POST", msg.Destination().Value, bytes.NewBuffer(data))
```

smoketest/webhook_test.go
```diff
@@ -0,0 +1,73 @@
...
+	AlertType string
+	Summary   string
+	Details   string
+}
+
+func TestWebhookAlert(t *testing.T) {
+	t.Parallel()
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var alert WebhookTestingAlert
```
mastercactapus, line 25: Send this into a channel so we can wait for it below, and then assert on the fields in the test instead of this handler

web/src/app/documentation/sections/Webhooks.md
```diff
@@ -0,0 +1,67 @@
...
+```
+{
+    "AlertType": "Verification",
+    "Summary": "Verification Message",
+    "Details": "This is a verification message from GoAlert",
+    "Code": "283917"
+}
+```
+
+### Test Message (from profile page)
```
mastercactapus, line 16: We should add short descriptions of the message types along with why/when you get them.
```suggestion
### Test Message

Triggered on the profile page by clicking "Send Test"
```

mastercactapus (changes requested):

notification/webhook/sender.go
dctalbot: I inlined the `postWithBody` code after the switch, added inline structs, and added string constants for the `POSTData` `Type` field. I haven't broken out the switch into a `renderMessage` function since I feel that indirection outweighs the value it would bring. The code seems simple enough as is IMO. Lmk what you think!

config/config.go
```diff
@@ -191,6 +196,19 @@ func (cfg Config) CallbackURL(path string, mergeParams ...url.Values) string {
...
+	if len(cfg.Webhook.AllowedURLs) == 0 {
+		return true
+	}
+	for _, u := range cfg.Webhook.AllowedURLs {
+		if strings.HasPrefix(whURL, u) {
+			return true
+		}
+	}
+	return false
+}
```
mastercactapus, line 210: Webhooks commonly need auth info, this will break if just using prefix matching. Let's use a helper method (so we can help the referrer check too)

- account for implicit ports (e.g., `http` and `:80` as well as `https` and `:443`)
- ignore auth info, and reject it from config validation
- host is case-insensitive
- path is prefix checked itself
- query parameters are matched via the source URL

```go
func MatchURL(baseURL, testURL string) (bool, error) {

}
```

```go
func TestMatchURL(t *testing.T) {
	// true
	MatchURL("http://example.com/", "HTTP://example.COM")
	MatchURL("http://example.com/", "HTTP://foo:bar@example.COM:80")
	MatchURL("http://example.com/?notAllowedQueryParam=&requiredQueryParam=1", "http://example.com?requiredQueryParam=1")

	// false
	MatchURL("http://example.com/?notAllowedQueryParam=&requiredQueryParam=1", "http://example.com")
	MatchURL("http://example.com/?notAllowedQueryParam=", "http://example.com?notAllowedQueryParam=1")
	MatchURL("https://example.com", "http://example.com")
```

We can't really restrict paths more than we have since we already have the prefix precedent in referrer URLs (e.g., trailing `/`) so those can be just a direct match.

Handling the implicit ports & capitalization should also help with some auth referrer issues we've seen in the past.

notification/webhook/sender.go
```diff
@@ -0,0 +1,117 @@
...
+	case notification.AlertStatusBundle:
+		payload = &POSTData{
+			Type:     TypeAlertStatusBundle,
+			Count:    m.Count,
+			AlertID:  m.AlertID,
+			LogEntry: m.LogEntry,
+		}
+
+	default:
+		return nil, errors.New("message type not supported")
```
mastercactapus, line 89: Add the type in the message to aid in debugging if this case is ever hit

notification/webhook/sender.go
```diff
@@ -0,0 +1,117 @@
...
+const (
+	TypeTest              POSTDataType = "Test"
+	TypeVerification      POSTDataType = "Verification"
+	TypeAlert             POSTDataType = "Alert"
+	TypeAlertBundle       POSTDataType = "AlertBundle"
+	TypeAlertStatus       POSTDataType = "AlertStatus"
+	TypeAlertStatusBundle POSTDataType = "AlertStatusBundle"
+)
+
+// POSTData is a union of all possible message types, should be populated accordingly
```
mastercactapus, line 28: Describe what its purpose/use is, rather than describing the structure. I wonder if calling it `Message` would be the more descriptive name?

For example, `Message is sent as JSON to all webhook endpoints by a Sender. See the Sender.Send method for details on what fields are sent per MessageType.`

smoketest/webhook_test.go
```diff
@@ -0,0 +1,78 @@
...
+func TestWebhookAlert(t *testing.T) {
+	t.Parallel()
+
+	ch := make(chan WebhookTestingAlert)
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var alert WebhookTestingAlert
+
+		data, err := ioutil.ReadAll(r.Body)
+		assert.Nil(t, err)
```
mastercactapus, line 30: Do a `require.Nil` so it panics/exits immediately, otherwise, it will hang until the test times out.

mastercactapus (changes requested):

graphql2/graphqlapp/contactmethod.go
```diff
@@ -72,6 +73,13 @@ func (q *Query) UserContactMethod(ctx context.Context, id string) (*contactmetho
 
 func (m *Mutation) CreateUserContactMethod(ctx context.Context, input graphql2.CreateUserContactMethodInput) (*contactmethod.ContactMethod, error) {
 	var cm *contactmethod.ContactMethod
+
+	cfg := config.FromContext(ctx)
+
+	if input.Type == contactmethod.TypeWebhook && !cfg.ValidWebhookURL(input.Value) {
+		return cm, validation.NewFieldError("value", "URL disallowed by administrator")
```
mastercactapus, line 80:
```suggestion
		return nil, validation.NewFieldError("value", "URL disallowed by administrator")
```

graphql2/graphqlapp/contactmethod.go
```diff
@@ -72,6 +73,13 @@ func (q *Query) UserContactMethod(ctx context.Context, id string) (*contactmetho
 
 func (m *Mutation) CreateUserContactMethod(ctx context.Context, input graphql2.CreateUserContactMethodInput) (*contactmethod.ContactMethod, error) {
 	var cm *contactmethod.ContactMethod
+
+	cfg := config.FromContext(ctx)
+
```
mastercactapus, line 78: nit: remove extra blank line

smoketest/webhook_test.go
```diff
@@ -0,0 +1,81 @@
...
+	t.Parallel()
+
+	ch := make(chan WebhookTestingAlert)
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var alert WebhookTestingAlert
+
+		data, err := ioutil.ReadAll(r.Body)
+		require.Nil(t, err)
+		assert.Nil(t, err)
```
mastercactapus, line 32: The assertion is redundant since `require` will panic
```suggestion
```

smoketest/webhook_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+		err = json.Unmarshal(data, &alert)
+		require.Nil(t, err)
+		assert.Nil(t, err)
```
mastercactapus, line 36: Same

mastercactapus (changes requested):

config/config.go
```diff
@@ -191,6 +196,101 @@ func (cfg Config) CallbackURL(path string, mergeParams ...url.Values) string {
...
+
+	test, err := url.Parse(testURL)
+
+	if err != nil {
+		return false, err
+	}
+
+	// ignore auth info
+	test.User = nil
+	base.User = nil
```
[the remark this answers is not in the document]
KatieMSB: No, I was doing some extra testing but not necessary anymore

config/config.go
```diff
@@ -191,6 +196,87 @@ func (cfg Config) CallbackURL(path string, mergeParams ...url.Values) string {
...
+
+	addImplicitPort(base)
+	addImplicitPort(test)
+
+	// use full host that includes port by default, i.e example.com:8080
+	baseHost := base.Host
+	testHost := test.Host
+
+	// host/port check
+	if !strings.EqualFold(baseHost, testHost) {
```
mastercactapus, line 241: no need for temp vars here
```suggestion

	// host/port check
	if !strings.EqualFold(base.Host, test.Host) {
```

config/config.go
```diff
@@ -191,6 +196,87 @@ func (cfg Config) CallbackURL(path string, mergeParams ...url.Values) string {
...
+		}
+	}
+	base, err := url.Parse(baseURL)
+
+	if err != nil {
+		return false, err
+	}
+	test, err := url.Parse(testURL)
+
+	if err != nil {
```
mastercactapus, line 229: nit: whitespace before err checks
```suggestion
	base, err := url.Parse(baseURL)
	if err != nil {
		return false, err
	}

	test, err := url.Parse(testURL)
	if err != nil {
```

config/config.go
```diff
@@ -191,6 +196,87 @@ func (cfg Config) CallbackURL(path string, mergeParams ...url.Values) string {
...
+func (cfg Config) ValidWebhookURL(whURL string) bool {
+	if len(cfg.Webhook.AllowedURLs) == 0 {
+		return true
+	}
+	for _, u := range cfg.Webhook.AllowedURLs {
+		matched, err := MatchURL(whURL, u)
+		if err != nil {
+			return false
+		}
+		return matched
```
mastercactapus, line 275: only jump out if `matched` is true, otherwise this will reject the first non-match even if allowed via a subsequent entry in `AllowedURLs`

config/config.go
```diff
@@ -206,13 +292,19 @@ func (cfg Config) ValidReferer(reqURL, ref string) bool {
...
 	}
 
 	for _, u := range cfg.Auth.RefererURLs {
-		if strings.HasPrefix(ref, u) {
-			return true
+		matched, err := MatchURL(u, ref)
+		if err != nil {
+			return false
 		}
+		return matched
```
mastercactapus, line 307: Same as above, since we're ranging over a list, only return true if `matched` is true

notification/webhook/sender.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	req.Header.Add("Content-Type", "application/json")
+
+	_, err = http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	return &notification.MessageStatus{ID: msg.ID(), State: notification.MessageStateSent, ProviderMessageID: msg.ID()}, nil
```
mastercactapus, line 109: Leave `ProviderMessageID` blank since we don't have one -- this would be something the webhook server returns to us. It's used for Twilio as the ID of the SMS or Slack for the Slack message ID.

```suggestion
	return &notification.MessageStatus{ID: msg.ID(), State: notification.MessageStateSent}, nil
```

mastercactapus (changes requested):

config/config.go
```diff
@@ -191,6 +196,84 @@ func (cfg Config) CallbackURL(path string, mergeParams ...url.Values) string {
...
+		return false, nil
+	}
+
+	// path check
+	if len(base.Path) > 1 && !strings.HasPrefix(test.Path, base.Path) {
+		return false, nil
+	}
+
+	// query check
+	if len(test.Query()) > 0 || len(base.Query()) > 0 {
```
mastercactapus, line 251: I think this conditional might be redundant; the `range` as well as `Get` method will already handle empty sets inside `compareQueryValues`.

config/config.go
```diff
@@ -191,6 +196,84 @@ func (cfg Config) CallbackURL(path string, mergeParams ...url.Values) string {
...
+	if len(cfg.Webhook.AllowedURLs) == 0 {
+		return true
+	}
+	for _, u := range cfg.Webhook.AllowedURLs {
+		matched, err := MatchURL(whURL, u)
+		if err != nil {
+			return false
+		}
+		if matched {
+			return matched
```
mastercactapus, line 271: nit: return an explicit `true` here to make the code more scannable (i.e., quick glance to see all possible `return true` locations) -- for the same reason we'd `return value, nil` after an err check instead of `return value, err`
```suggestion
			return true
```

config/config.go
```diff
@@ -206,12 +289,20 @@ func (cfg Config) ValidReferer(reqURL, ref string) bool {
...
 
 	for _, u := range cfg.Auth.RefererURLs {
-		if strings.HasPrefix(ref, u) {
-			return true
+		matched, err := MatchURL(u, ref)
+		if err != nil {
+			return false
+		}
+		if matched {
+			return matched
```
mastercactapus, line 305: nit: same as above, since we checked that it was true, be explicit with `return true` for readability.
```suggestion
			return true
```

mastercactapus (changes requested):

mastercactapus:
I think we may need to decide on a documented hard-deadline for webhook calls. Since they can be set arbitrarily by application users, they could represent a potential DOS vector.

Example:
1. User sets their own webhook URL to a target with a firewall that drops packets (perhaps by mistake)
2. Calls to notify the user fail after the engine cycle hard 30s timeout
3. No other notifications can get sent since the cycle deadline was reached
4. Notification is scheduled for retry
5. Delay of notifications to other application users occurs

Granted, it wouldn't be *quite* that easy but the potential is there. We could set something like a 3-second deadline for the POST request, and add that to the markdown documentation.

Thoughts?

Forfold (changes requested):
So far I've functionally tested everything I can think of except for bundled alerts and things are looking really good. A couple of things to note though, mainly around how we are handling outgoing_messages for new webhook notifications:

1. I'm seeing "Failed: send deadline expired" on all sends and alerts in the alert log after the first successful webhook cm alert of GoAlert goes out. Notifications are still hitting the webhook, but we'll need to handle the `provider_msg_id` column in `outgoing_messages` differently so we can store the event properly in `alert_logs` and avoid logging false positives
    - The error: `send message: update message status: ERROR: duplicate key value violates unique constraint \"idx_outgoing_messages_provider_msg_id\`
    - You'll want to set that value to null for status updates, and to `webhook:something-unique` for new alerts and escalations- this is how we're handling the Slack notification method as well
2. A table view under docs with a full list of the json properties being sent out would be nice. I'm imagining something like the tables MUI provides when you need to reference the API for a specific component.
    - It would helpful to know the default values (e.g., if details is empty should the user expect it to come in as `""`, or expect it to be omitted from the request entirely?), character length limit per field, etc

Edit: To avoid `"webhook:"` being written to `outgoing_messages` for status updates rather than `null`, one thing we had to do in the Slack environment was update [this line](https://github.com/target/goalert/blob/c46e48733247afc0ff31f25d10bea102c75f431a/notification/messagestatus.go#L37) to:

```
if s.ProviderMessageID != "" {
  s.ProviderMessageID = n.name + ":" + s.ProviderMessageID
}
```

graphql2/graphqlapp/contactmethod.go
```diff
@@ -72,6 +73,12 @@ func (q *Query) UserContactMethod(ctx context.Context, id string) (*contactmetho
 
 func (m *Mutation) CreateUserContactMethod(ctx context.Context, input graphql2.CreateUserContactMethodInput) (*contactmethod.ContactMethod, error) {
 	var cm *contactmethod.ContactMethod
+	cfg := config.FromContext(ctx)
+
+	if input.Type == contactmethod.TypeWebhook && !cfg.ValidWebhookURL(input.Value) {
+		return nil, validation.NewFieldError("value", "URL disallowed by administrator")
```
Forfold, line 79: nit: something like "blacklisted" might be more clear

```suggestion
		return nil, validation.NewFieldError("value", "URL blacklisted by an administrator")
```

web/src/app/users/UserContactMethodList.js
```diff
@@ -134,6 +135,20 @@ export default function UserContactMethodList(props) {
     )
   }
 
+  function getSubText(cm) {
+    if (cm.type === 'WEBHOOK') {
+      const { origin } = new URL(cm.formattedValue)
+      return (
+        <React.Fragment>
+          <React.Fragment>{`${origin} (`}</React.Fragment>
```
Forfold, line 143: Duplicate fragment here?

config/config.go
```diff
@@ -191,6 +196,82 @@ func (cfg Config) CallbackURL(path string, mergeParams ...url.Values) string {
...
+
+	// path check
+	if len(base.Path) > 1 && !strings.HasPrefix(test.Path, base.Path) {
+		return false, nil
+	}
+
+	// query check
+	if !compareQueryValues(base.Query(), test.Query()) {
+		return false, nil
+	}
```
Forfold, line 253: Even though the majority of the functions using this `MatchURL` function return `bool` values, I think since this function returns `error` along with bool we should return some reason strings alongside `false`, i.e.:
- "host or ports do not match"
- "schemes do not match
- "paths do not match"
- "query values do not match"

KatieMSB: That's what it was originally but it was changed to return nil in the case that we need to find URL's that return true in a list and only want the error when there's an invalid URL. https://github.com/target/goalert/pull/1269#discussion_r593289489

graphql2/graphqlapp/contactmethod.go
mastercactapus: In this case, it would be "not whitelisted" as it doesn't currently support blacklisting -- I think that's why it was phrased as a neutral term

dontlaugh: Alternatives:
URL blocked by administrator?
URL excluded by administrator?

mastercactapus: Maybe just "not allowed"?

migrate/migrations/20210518145011-webhook-user-contact-method-type.sql
```diff
@@ -0,0 +1,6 @@
+-- +migrate Up notransaction
+
+ALTER TYPE enum_user_contact_method_type ADD VALUE IF NOT EXISTS 'WEBHOOK';
+
+-- +migrate Down
+
```
Forfold, line 6: Missing removal of type in migrate down

Forfold (changes requested):
Functionally tested- Everything works great! Added one comment regarding the migration.

migrate/migrations/20210518145011-webhook-user-contact-method-type.sql
dctalbot: IIRC this is a limitation in pg. We would have to drop and create the enum or something to that effect

Forfold: We should probably do that then, so if we have to rollback this isn't left in the DB even though the feature doesn't exist

mastercactapus: A while back we switched to just leaving these in going forward; re-creating and swapping the type has other potential side-effects that weren't worth the risk. The migration test has a special case to ignore this specifically:
```go
// enumOK handles checking for safe enum differences. This case is that migrate
// up can add, but migrate down will not remove new enum values.
//
// migrate down can't safely remove enum values, but it's safe for new ones
// to exist. So we simply check that all original items exist.
func enumOK(got, want string) bool {
```

Forfold (dismissed):

Forfold (dismissed):

Forfold (dismissed):

dontlaugh:
Are there any critical blockers to this? I would love to use this feature :blush:

mastercactapus:
We're waiting to get #1655 in so we don't have a new API go out with something that's about to be removed (status update bundles).

mastercactapus (approved):

arurao (approved):
LGTM

PR #1276, opened 2021-02-22
ui: fix markdown url rendering

### LINKED ISSUE #1268: ui/users: regression in rendering user emails
**Steps to Reproduce:**
1. Go to a user profile
2. See below

**Expected Behavior:**
Render the email address as either plain text or as a "mailto" link

**Observed Behavior:**
See blow

**Screenshots/Stack Traces:**
Screen Shot 2021-02-17 at 1 53 58 PM

**Application Version:**
`761a0b7b878a7a2d269b78016096b7c44266af6c`

dctalbot:

**Description:**
This PR fixes a regression in displaying markdown links likely introduced by `remark-gfm`

Previously, emails were rejected by `safeURL` and rendered as full plain text (see below image). Tel numbers behaved similarly.

Emails are now rendered as `mailto` links if the email address matches the label, else full plain text. Tel numbers behave similarly but use `tel`. User details are not rendered in markdown.
**Issue this fixes:**
fixes #1268 

Before:
Screen Shot 2021-02-17 at 1 53 58 PM

After:
Screen Shot 2021-02-17 at 1 58 40 PM

Screen Shot 2021-02-23 at 11 42 50 AM
Screen Shot 2021-02-23 at 11 42 56 AM

Screen Shot 2021-02-23 at 1 33 02 PM
Screen Shot 2021-02-23 at 1 32 46 PM


web/src/app/util/Markdown.js
```diff
@@ -16,6 +16,12 @@ export default function Markdown(props) {
         if (node.type !== 'link') return true
         if (node.children[0].type !== 'text') return true // only validate text labels
         if (safeURL(node.url, node.children[0].value)) return true
+        if (node.url.startsWith('mailto:')) {
+          // do not yield native mailto link, use plain text instead
+          node.type = 'text'
+          delete node.url
+          return true
+        }
```
mastercactapus, line 24: Not sure it's necessary to bar the links everywhere in-app -- this will also prevent `mailto` links from working in descriptions and alert details when added intentionally by users of the application.

I think the goal was only to fix it on the user details page by making it render as text.

mastercactapus (changes requested):

web/src/app/util/Markdown.js
dctalbot: This is an app-wide bug. In the current state, users are already prevented from intentionally adding `mailto` links anywhere in the app since they are rejected by `safeURL`. I think the `remark-gfm` plugin now gives us the capability to use `mailto`, but I didn't think we were turning it on. 

I couldn't think of a compelling scenario wherein allowing `mailto:` links would enhance UX. The `safeURL` logic is also app-wide, so we are already limiting users' abilities to a degree. Happy to entertain other opinions on this :)

dctalbot: If we _do_ allow custom schemas, we should consider `tel:`, `slack:`, other protocol handlers. And then we need to decide which parts of the app are worthy of these features, and which are not. If User Details is the only concern for now, maybe it shouldn't be in markdown to begin with.

mastercactapus: For profile maybe something like `noMarkdown` as a flag for `DetailsPage`

As for the others, I wasn't aware that other protocols were broken, that seems like a bug. As far as URLs that are not allowed, it was intentional to render them as raw markdown rather than hide any information.

The primary concern was labels that don't match the URL. In that case a fix could be to allow the label to omit the protocol in the URL so that things like both `[http://example.com](http://example.com)` and `[example.com](http://example.com)` work

web/src/app/util/safeURL.ts
```diff
@@ -2,6 +2,12 @@
 //
 // It tries to determine if the label is misleading.
 export function safeURL(url: string, label: string): boolean {
+  if (url.startsWith('mailto:')) {
+    const email = url.split(':')[1]
```
mastercactapus, line 6: `:` is allowed in the local part of an address if it's in a quoted string, which could break this
```suggestion
    const email = url.substr(7)
```

mastercactapus (changes requested):
There's no special support required, they just render as anchor tags. We should support `tel` being an on-call app, though that's probably as far as we need to go. Slack supports linking most things through normal http/https links.

A simple check that the number matches should be good enough for `tel`, and should allow decent integration when on a mobile device.
```go
  if (url.startsWith('tel:')) {
    const phone = url.substr(4)
    return phone === label && /^\+?[\d\- ]+$/.test(phone)
  }
```

web/src/app/util/safeURL.ts
dctalbot: Thanks for checking, I should've known this would be weird 🙃

web/src/app/users/UserDetails.js
```diff
@@ -198,6 +198,8 @@ export default function UserDetails(props) {
       <DetailsPage
         title={user.name + (svcCount ? ' (On-Call)' : '')}
         details={user.email}
+        noMarkdown
+        detailsStyles={{ marginBottom: '1rem' }}
```
mastercactapus, line 202: Instead of adding a new prop for this, can we bake it into the `DetailsPage` and just fix the margin when `noMarkdown` is set, automatically? The caller of `DetailsPage` shouldn't need to handle/override internal styles when using supported props.

mastercactapus (changes requested):

web/src/app/users/UserDetails.js
dctalbot: We could, but I think coupling this margin style to `noMarkdown` is a false relationship. But since there is only one component with the `noMakdown` prop, it would work fine and we could refactor later if needed.

dctalbot: https://github.com/target/goalert/pull/1290 aims to address this so I will remove the style changes in this PR

Forfold (approved):
functionally tested in alert details/email rendering properly on profile, lgtm!

mastercactapus (approved):

PR #1290, opened 2021-03-01
ui: remove dead css and refactor markdown css

dctalbot:

**Description:**
This PR removes and refactors the css for the markdown component. Converting those styles to be "CSS in JS", we scope that css to the component which will prevent it from being left behind if and when we remove the in-app documentation.

Markdown text elements will have less vertical space between each other

**Note**
This PR previously contained changes to the bundling configuration. In order to limit scope, those changes were reverted.


web/src/app/util/Markdown.js
```diff
@@ -2,14 +2,63 @@ import React from 'react'
...
+      lineHeight: '1.42857143',
+      color: '#333',
+      wordBreak: 'break-all',
+      wordWrap: 'break-word',
+      backgroundColor: '#f5f5f5',
+      border: '1px solid #ccc',
+      borderRadius: '4px',
+    },
+    '& code': {
+      padding: '2px 4px',
```
Forfold, line 37: are the CSS properties from the global file used here? From the classname on [L12](https://github.com/target/goalert/pull/1290/files#diff-3a39745961e924d7b6600b2aa05f7daaef4304372542dfa5aa9c8cfd6930154bL12) previously it looks like only the `.react-markdown td` and `.react-markdown th` styles would be used for this component

dctalbot: Yeah, they're used in the markdown code blocks. It wasn't scoped to that class initially, but we don't have any other `pre` or `code` elements in the app so IMO it belongs here.

Forfold: That makes sense, thanks for the clarification!

m17ch (approved):
lgtm

Forfold (changes requested):
Comparing with dev, it looks like we're losing some styling on the code snippets. I verified that headers are rendering properly. Tables look weird, but they look weird in Prod/Dev too so that can be a separate issue w/ an isolated fix. Can we avoid using this CSS altogether in favor of something more maintainable?

Dev as of today:
Screen Shot 2021-03-12 at 10 51 42 AM

PR Screenshot:
Screen Shot 2021-03-12 at 10 51 52 AM

web/src/app/util/Markdown.js
```diff
@@ -2,14 +2,63 @@ import React from 'react'
...
+      margin: '0 0 10px',
+      fontSize: '13px',
+      lineHeight: '1.42857143',
+      color: '#333',
+      wordBreak: 'break-all',
+      wordWrap: 'break-word',
+      backgroundColor: '#f5f5f5',
+      border: '1px solid #ccc',
+      borderRadius: '4px',
+    },
```
Forfold, line 35: Everything is rendering as expected now, though imho I think we should be using the defaults from this library instead of this here and just focus on getting the code block elements rendered with the proper colors. To be blunt, I'm not really in favor of maintaining things like `lineHeight: '1.42857143'` if we can avoid it

Forfold: Doing some functional testing to see the differences, outside of the code elements I'm only seeing the CSS here compact the lines together.

dctalbot: I'm guessing that lineheight was ripped from some computed source a while back along with the fontsize and padding, maybe others. Probably best to open an issue for touching up all markdown styles so we don't bikeshed on this too much

Forfold: Since this PR is for removing the dead CSS but also refactoring, I'd be fine with making those changes here. We should be able to do away with everything except the color bits for `pre`/`code` and be in a good state. I went ahead and double-checked what we'll need to keep (also adjusting one px value to rem). Will comment with the commit suggestion

web/src/app/util/Markdown.js
```diff
@@ -2,14 +2,63 @@ import React from 'react'
...
+    '& pre code': {
+      padding: 0,
+      fontSize: 'inherit',
+      color: 'inherit',
+      whiteSpace: 'pre-wrap',
+      backgroundColor: 'inherit',
+      borderRadius: 0,
+    },
+  },
```
Forfold, line 51:
```suggestion
    markdown: {
    '& pre': {
      display: 'block',
      padding: '0.375rem',
      color: '#333',
      backgroundColor: '#f5f5f5',
      border: '1px solid #ccc',
      borderRadius: '4px',
    },
    '& code': {
      color: '#c7254e',
      backgroundColor: '#f9f2f4',
    },
    '& pre code': {
      fontSize: 'inherit',
      color: 'inherit',
      whiteSpace: 'pre-wrap',
      backgroundColor: 'inherit',
    },
  },
```

web/src/app/util/Markdown.js
Forfold: I did some further digging, and [this code](https://git.target.com/goalert/goalert-ui/commit/5b304fdc0da5191306a66538a785357725f45b48) is a remnant from when we were using a completely different markdown library altogether (way back at the beginning of 2018)- so I definitely think we're good here to remove

m17ch (dismissed):

Forfold (changes requested):
It appears that none of the css defined in `Markdown.js` gets applied within `AlertDetails`

Forfold (approved):
lgtm!

m17ch (approved):

PR #1318, opened 2021-03-11
engine/rotationmanager: safely migrate hourly rotations to new handoff calculation

mastercactapus:

**Description:**
This PR adds a `version` column to `rotation_state` to allow migrating to the new time calculation code safely.

With the change to hourly rotations strictly following clock-time it became possible for some rotations to rotate immediately with the new engine code. This PR fixes the issue by doing a final "should it advance?" check using the old handoff calculation code for hourly rotations, advancing as necessary, then setting the `shift_start` time to `now()`. This ensures that the current user will remain on-call until the next handoff time after the deployment without side effects.

Active shifts will end based on the new handoff calculation logic, but without retro-actively advancing due to prior-shift calculation differences.

**Additional Context**
The `rotation` processing module has had it's version bumped from `1 -> 2` to prevent calculation conflicts with the prior engine version.


migrate/migrations/20210309144821-rotation-module-update.sql
```diff
@@ -0,0 +1,20 @@
...
+
+UPDATE engine_processing_versions
+SET version = 2
+WHERE type_id = 'rotation';
+
+ALTER TABLE rotation_state
+ADD COLUMN version INT DEFAULT 1 NOT NULL;
+
+ALTER TABLE rotation_state
+ALTER COLUMN version SET DEFAULT 2;
```
Forfold, line 11: Are these redundant since they are in the same migration? As in, would it make more sense to only do `DEFAULT 2 NOT NULL`?

engine/rotationmanager/legacyadvance.go
```diff
@@ -0,0 +1,53 @@
...
+	"time"
+
+	"github.com/target/goalert/schedule/rotation"
+)
+
+func addHours(t time.Time, n int) time.Time {
+	return time.Date(t.Year(), t.Month(), t.Day(), t.Hour()+n, t.Minute(), t.Second(), t.Nanosecond(), t.Location())
+}
+
+func addHoursAlwaysInc(t time.Time, n int) time.Time {
```
Forfold, line 13: maybe `addHoursAbs` would be more appropriate? that way a dev would know it would do `|n|` on neg, rather than say `0` or `1`

```suggestion
func addHoursAbs(t time.Time, n int) time.Time {
```

Forfold (changes requested):

migrate/migrations/20210309144821-rotation-module-update.sql
mastercactapus: Adding the column with a default will set all existing/old rows to version 1, then we change the default to 2 so that any new rows start at version 2. That lets us migrate just the existing data

engine/rotationmanager/legacyadvance.go
mastercactapus: It was named because it guarantees time will always increment, but I'm going to prefix them with `legacy` and comment as such. These aren't meant to be used going forward -- they were copied 1-to-1 from old code for compatibility during the migration only.

migrate/migrations/20210309144821-rotation-module-update.sql
Forfold: Gotcha, I comprehended it as it setting the default to 1 for all existing rows, but then immediately after changing all of them to 2. This makes more sense 👍

Forfold (approved):

Forfold (approved):

m17ch (approved):

PR #1354, opened 2021-03-24
notification/slack: fix missing Slack scope error

### LINKED ISSUE #1246: Slack scope error is incorrect
**Describe the Bug:**
When a slack app has incorrect scopes, GoAlert returns an "only channels supported" error hiding the "missing_scope" error.

**Steps to Reproduce:**
1. Create a Slack app with missing scopes
2. Configure in the Admin page
2. Attempt to add a channel to an EP
3. Observe an "only channels supported" error in the UI
4. Observe no error from the backend

**Expected Behavior:**
The UI should display a more accurate message (like Permission Denied) and the backend should log a detailed message from the real error Slack provides.

**Observed Behavior:**
Misleading error with no useful information logged.

dctalbot:
- [x]  Added comments in the code, where necessary.

**Description:**
Example log message: 
- `error="Slack: missing_scope; need one of channels:read,groups:read,mpim:read,im:read but got app_mentions:read"`
- `error="unable to authenticate to Slack: token_revoked"`

Fixes #1246 

- User will see "Permission Denied" instead of "Only Channels Supported" when Slack app has misconfigured scopes
- User will see "Permission Denied" instead of "User account must be linked" when goalert is unable to authenticate to the Slack app

### DIFF
--- a/notification/slack/channel.go
+++ b/notification/slack/channel.go
@@ -250,7 +250,7 @@ func (s *ChannelSender) loadChannels(ctx context.Context) ([]Channel, error) {
 		if !resData.OK {
 			acceptedScopes := resp.Header.Get("X-Accepted-Oauth-Scopes")
 			providedScopes := resp.Header.Get("X-Oauth-Scopes")
-			log.Log(ctx, errors.New("Slack app scopes must include one of: "+acceptedScopes+"; got: "+providedScopes))
+			log.Log(ctx, errors.New("Slack app scopes must include one of: ["+acceptedScopes+"]; got: ["+providedScopes+"]"))
 			return nil, wrapError(resData.Error, "list Slack channels")
 		}
 
--- a/notification/slack/channel.go
+++ b/notification/slack/channel.go
@@ -232,7 +232,7 @@ func (s *ChannelSender) loadChannels(ctx context.Context) ([]Channel, error) {
 			return nil, errors.New("non-200 response from Slack: " + resp.Status)
 		}
 
-		var resData struct {
+		var respBody struct {
 			OK       bool
 			Error    string
 			Channels []Channel
@@ -241,23 +241,23 @@ func (s *ChannelSender) loadChannels(ctx context.Context) ([]Channel, error) {
 			} `json:"response_metadata"`
 		}
 
-		err = json.NewDecoder(resp.Body).Decode(&resData)
+		err = json.NewDecoder(resp.Body).Decode(&respBody)
 		resp.Body.Close()
 		if err != nil {
 			return nil, errors.Wrap(err, "parse JSON")
 		}
 
-		if !resData.OK {
-			return nil, wrapError(resData.Error, "list Slack channels")
+		if !respBody.OK {
+			return nil, wrapError(respBody.Error, "list Slack channels")
 		}
 
-		channels = append(channels, resData.Channels...)
+		channels = append(channels, respBody.Channels...)
 
-		if resData.Meta.NextCursor == "" {
+		if respBody.Meta.NextCursor == "" {
 			break
 		}
 
-		v.Set("cursor", resData.Meta.NextCursor)
+		v.Set("cursor", respBody.Meta.NextCursor)
 	}
 
 	for i := range channels {
--- a/notification/slack/channel.go
+++ b/notification/slack/channel.go
@@ -159,7 +159,7 @@ func (s *ChannelSender) loadChannel(ctx context.Context, channelID string) (*Cha
 	}
 
 	if !resData.OK {
-		return nil, fmt.Errorf("lookup Slack channels: %w", &apiError{msg: resData.Error, header: resp.Header})
+		return nil, fmt.Errorf("lookup Slack channel: %w", &apiError{msg: resData.Error, header: resp.Header})
 	}
 
 	return &Channel{
--- a/notification/slack/channel.go
+++ b/notification/slack/channel.go
@@ -61,7 +61,14 @@ type apiError struct {
 	header http.Header
 }
 
-func (err apiError) Error() string { return err.msg }
+func (err apiError) Error() string {
+	if err.msg == "missing_scope" {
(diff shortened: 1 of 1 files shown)


notification/slack/channel.go
```diff
@@ -248,6 +248,9 @@ func (s *ChannelSender) loadChannels(ctx context.Context) ([]Channel, error) {
 		}
 
 		if !resData.OK {
+			acceptedScopes := resp.Header.Get("X-Accepted-Oauth-Scopes")
+			providedScopes := resp.Header.Get("X-Oauth-Scopes")
+			log.Log(ctx, errors.New("Slack app scopes must include one of: ["+acceptedScopes+"]; got: ["+providedScopes+"]"))
```
mastercactapus, line 253: Let's make this a little more specific, and only log the scope info IF it's a scope error
```suggestion
		if !resData.OK {
			if resData.Error == "missing_scope" {
				acceptedScopes := resp.Header.Get("X-Accepted-Oauth-Scopes")
				providedScopes := resp.Header.Get("X-Oauth-Scopes")
				log.Log(ctx, fmt.Errorf("list Slack channels: missing_scope; need one of %v but got %v", acceptedScopes, providedScopes))
			}
```

mastercactapus: Though, it might be better to pass `resp.Header` to `wrapError` and put this code inside the `missing_scope` case instead of here -- that way we log missing scope error for ALL potential slack calls, and don't have two places we're handling `missing_scope`.

mastercactapus (changes requested):

notification/slack/channel.go
dctalbot: Agreed, I think when I tried that it felt messy passing through ctx and header for only one switch case. Might be able to refactor wraperror to handle generic logs

mastercactapus: Yeah, a more proper way would probably be to make a custom error type we can embed the data with, then handle logging/mapping of the error as part of sanitizing (e.g. see how `sqlutil.MapError` or `errutil.HTTPError` work). The one difference here is we want to have a specific message for the user and in this one special case log the config issue.

```go
type apiError struct {
	msg string
	headers http.Header
}
func (err apiError) Error() string{return err.msg}
func mapError(ctx context.Context, err error)error {
	var apiErr apiError
	if !errors.As(err, &apiErr) {
		log.Log(ctx,err)
		return slackError("unexpected error")
	}

	switch apiErr.msg {
	case "missing_scope":
		log.Log(ctx, fmt.Errorf(/*... apiErr.headers ...*/))
		return slackError("Permission Denied")
	}

	log.Log(ctx, err)
	return slackError("unexpected error")
}

func loadChannel() {
	return nil,  fmt.Errorf("list slack channels: %w",&apiError{msg: respData.Error, headers: resp.Headers})
}
func Channel() {
	foo,err:=loadChannel()
	if err != nil {
		return nil, mapError(ctx, err)
	}
}
```

That being said, I wouldn't worry much about it just being a switch statement, it's more about keeping ownership and concerns separate. I think a custom error type would be better than adding a bunch of arguments though.

dctalbot: I went ahead with the refactor, thanks for the tip! Updated PR description to reflect new state.

notification/slack/channel.go
```diff
@@ -55,21 +56,33 @@ type Channel struct {
...
-func (err slackError) ClientError() bool { return true }
-func wrapError(errMsg, details string) error {
-	switch errMsg {
+	switch apiError.msg {
 	case "missing_scope":
-		// happens if the ID is for a user
-		return validation.NewFieldError("ChannelID", "Only channels supported.")
+		acceptedScopes := apiError.header.Get("X-Accepted-Oauth-Scopes")
+		providedScopes := apiError.header.Get("X-Oauth-Scopes")
+		log.Log(ctx, fmt.Errorf("Slack: missing_scope; need one of %v but got %v", acceptedScopes, providedScopes))
```
mastercactapus, line 76: It may be useful to include the context of the original error and then just add the information to the end:

```suggestion
		log.Log(ctx, fmt.Errorf("%w; need one of %v but got %v", err, acceptedScopes, providedScopes))
```

This would read something like: `lookup Slack channel: missing_scope; need one of ...` so we know where it came from before hitting `mapError`

notification/slack/channel.go
```diff
@@ -55,21 +56,33 @@ type Channel struct {
...
+		return validation.NewFieldError("ChannelID", "Permission Denied.")
 	case "channel_not_found":
 		return validation.NewFieldError("ChannelID", "Invalid Slack channel ID.")
 	case "invalid_auth", "account_inactive", "token_revoked", "not_authed":
-		return slackError("User account must be linked.")
+		log.Log(ctx, fmt.Errorf("unable to authenticate to Slack: %v", apiError.msg))
```
mastercactapus, line 81: This one we can just log the original, as it will already contain the context of which API call we attempted, as well as the specific error like `token_revoked`.

```suggestion
		log.Log(ctx, err)
```

mastercactapus (changes requested):
Really liking how the error mapping turned out!

notification/slack/channel.go
mastercactapus: An alternative that might be cleaner could be to put the logic in apiError

```go
func (err apiError) Error() string{
  if err.msg == "missing_scope" {
    // accepted/provided from err.header
    return fmt.Sprintf("missing_scope; need one of %v but got %v", acceptedScopes, providedScopes)
}

// in `mapError`
switch apiError.msg {
// can group all permission denied errors this way
case "missing_scope", "token_revoked", "etc...":
  log.Log(ctx, err)
  return validation.NewError("Permission Denied")
}
```

dctalbot: Nice, I agree this is better

mastercactapus (approved):

pnengchu (approved):

PR #1358, opened 2021-03-25
slack: introduce interactivity on alert messages

### LINKED ISSUE #19: slack/alerts: add interactivity and post updates to message

An alert went off. The notification went to Slack first and then to SMS and, later, voice. I acknowledged it via SMS. While I received an acknowledgment from SMS, nothing went into Slack. Someone else looked at the already-addressed problem because they didn't see the acknowledgment SMS arrive.

I got an SMS for closure, but didn't in Slack.

I'd like to see minimally a new Slack message saying that the issue has been acknowledged and eventually closed, one message for each event.

Adding a reaction or a message in-thread or both would reduce clutter. Updating the message would strongly indicate the status but someone could miss the edit, thus the need for a message in-thread.

Ideally, Slack's button workflow would very much work to our advantage. A button for acknowledgment when clicked does the same as acknowledging via SMS and other buttons for escalation, all-hands, manual closure, etc. could accentuate and enable a Slack-based workflow similar to other (proprietary) tools on the market.

I considered highlights and manually adding the ack and closure in the thread. The text of the alert is so generic that it's risky to set something like `alert` as a highlighted word, so I've set the name of the service in the alert as a highlight. In my case, it's said rarely enough that I should not get false-positive notifications in another channel.

**Additional context:**

None pertinent right now.

Forfold:

**Description:**
This PR allows users to interact with an alert message from Slack. A user will have the option to: acknowledge, escalate or close an alert, as well as opening the alert on the GoAlert website.

If a user is not authenticated, they will be prompted within Slack to link with GoAlert, utilizing Slack's Oath 2.0 flow.

Closes #19

**How to test:**
Currently there's a bit of a checklist you need to run through before you can locally test this feature:
1. `$ make regendb`
2. `$ make start`
3. Open localhost:3030 and login as an admin
4. Configure sendit if not already
5. Create a Slackbot at api.slack.com (or use one of the GoAlert Sandbox ones from our testing workspace)
    - Under "Basic Information", copy the **Client ID**, **Client Secret**, and **Signing Secret** to your GoAlert admin config
    - Under "Interactivity & Shortcuts", enable and set your request URL to: `https://sendit.dev.target.goalert.me/YOUR_HTTP_PREFIX/api/v2/slack/actions`
    - Under "Oath & Permissions", copy the **Bot User OAuth Token** to your GoAlert admin config
        - Set a Redirect URL to `https://sendit.dev.target.goalert.me/YOUR_HTTP_PREFIX/api/v2/slack/auth`
6. Save GoAlert admin config
7. Install and invite your Slack bot to the channel you would like to test in
8. Back in GoAlert (still on localhost), go to Escalation Policies and create an EP where **repeat is set to 0**. It also helps to have the name start with Aa so it's at the top of the list.
9. Create a step on the EP with your slack channel added
10. Create a service, selecting our new EP
11. Stop GoAlert and restart with Sendit
12. Begin testing by creating alerts for the service you just created

**Remaining todos for this PR:**
- Alert Log:
  - Show user's name in logs (i.e. "Acknowledged by Nathaniel Cook (Slack)")
  - Show if a new message to Slack is pending
  - Show pending status updates in original Alert notification logs
    - i.e. "Pending status update" instead of delivered, as seen below:
    - Screen Shot 2021-03-25 at 10 24 05 AM

- Tests
- Ensure bundled alert flow remains the same
- Name in alert log from action

**Out of Scope:**
- Adding interactivity to bundled alerts
- Showing details as a popup (Show Details button)
- Showing the last event log within Alert message

**Screenshots:**
TBD

TBD

TBD

**Additional Info:**
- OAth tokens do not expire
- Need to think about an icon for the bot, and if/what colors and emojis we want to use for the buttons.


engine/statusupdatemanager/insertchannelmessages.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"context"
+
+	"github.com/target/goalert/util/log"
+
+	"github.com/pkg/errors"
+)
+
+func (db *DB) updateChannels(ctx context.Context, all bool, alertID *int) error {
```
dctalbot, line 11:
```suggestion
func (db *DB) updateChannels(ctx context.Context) error {
```

engine/statusupdatemanager/insertusermessages.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"context"
+
+	"github.com/target/goalert/util/log"
+
+	"github.com/pkg/errors"
+)
+
+func (db *DB) updateUsers(ctx context.Context, all bool, alertID *int) error {
```
dctalbot, line 11:
```suggestion
func (db *DB) updateUsers(ctx context.Context) error {
```

notification/slack/actions.go
```diff
@@ -0,0 +1,191 @@
...
+
+// Payload represents the relevant payload information sent from Slack
+type Payload struct {
+	ResponseURL string `json:"response_url"`
+	Actions     []Action
+	Channel     slack.Channel
+}
+
+// Action represents the information given from an action event within Slack
+// i.e. clicking to acknowledge an alert from slack
```
dctalbot, line 35:
```suggestion
// e.g. clicking to acknowledge an alert from slack
```

notification/slack/actions.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	// don't process url buttons
+	for _, action := range payload.ActionCallback.BlockActions {
+		if action.ActionID == "openLink" || action.ActionID == "auth" {
+			return
+		}
+	}
+
+	intrnlErr := func() { http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError) }
```
dctalbot, line 114:
```suggestion
	serverErr := func() { http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError) }
```

notification/slack/actions.go
```diff
@@ -0,0 +1,191 @@
...
+		// actions may come in batches, range over
+		for _, action := range payload.ActionCallback.BlockActions {
+			alertID, err := strconv.Atoi(action.Value)
+			if err != nil {
+				intrnlErr()
+				return
+			}
+
+			// add source info to ctx to write to alert log
+			ncID, _, err := h.c.AlertLogStore.FindByValue(ctx, nil, payload.Channel.ID)
```
dctalbot, line 142: Does this include the user or only the channel? We probably want both I'd imagine

notification/slack/blocks.go
```diff
@@ -0,0 +1,96 @@
...
+
+func alertIDAndStatusSection(id int, status string) *slack.HeaderBlock {
+	var s string
+	if status == "triggered" {
+		s = "Unacknowledged"
+	} else if status == "active" {
+		s = "Acknowledged"
+	} else {
+		s = "Closed"
+	}
```
dctalbot, line 19: We should be explicit about marking anything as closed. Is there a "closed" status we can check for?
Default case should be "Unacknowledged" or an error

dctalbot (changes requested):

notification/slack/channel.go
```diff
@@ -267,51 +269,66 @@ func (s *ChannelSender) loadChannels(ctx context.Context) ([]Channel, error) {
...
-	vals.Set("channel", msg.Destination().Value)
 	switch t := msg.(type) {
 	case notification.Alert:
-		vals.Set("text", fmt.Sprintf("Alert: %s\n\n<%s>", t.Summary, cfg.CallbackURL("/alerts/"+strconv.Itoa(t.AlertID))))
+		a.Summary = t.Summary
+		a.ID = t.AlertID
+		a.Status = alert.StatusTriggered
 	case notification.AlertBundle:
-		vals.Set("text", fmt.Sprintf("Service '%s' has %d unacknowledged alerts.\n\n<%s>", t.ServiceName, t.Count, cfg.CallbackURL("/services/"+t.ServiceID+"/alerts")))
+		a.Summary = fmt.Sprintf("Service '%s' has %d unacknowledged alerts.", t.ServiceName, t.Count)
```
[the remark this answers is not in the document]
Forfold: I think we should be good here since bundles should always come in as 1+n alerts, though this does remind me that we need to ensure we aren't filling in all the blocks for bundled messages since they won't work with bundles as is.

alert/log/store.go
```diff
@@ -287,20 +304,24 @@ func (db *DB) logAny(ctx context.Context, tx *sql.Tx, insertStmt *sql.Stmt, id i
...
+				return errors.Wrap(err, "lookup contact method type")
 			}
-
 			switch ncType {
 			case notificationchannel.TypeSlack:
 				r.subject.classifier = "Slack"
+				var c sql.NullString
+				c.String = "Puppies!"
+				c.Valid = true
+				r.subject.channelName = c
```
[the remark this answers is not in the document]
Forfold: Looks like this is an artifact/not needed anymore. Channel name on the Entry is getting set properly elsewhere now

notification/slack/blocks.go
Forfold: We do have a closed status to check for as well, and in this case we're hoping that everything is 1-to-1. What do you think of a switch, and on default panicking? We should also probably use the `AlertStatus` type instead of `string` as the func parameter

notification/slack/actions.go
Forfold: Right now this only includes channel, but yeah since this is sourced from a user's action within Slack I think we will need both for the alert log

notification/slack/blocks.go
Forfold: https://github.com/target/goalert/pull/1358/commits/1f80a4ef6cd2a51ac7ee8cd4e27e4c5053c162ab#diff-f5ee0dc56091cb36e70eb6e95a5bc4906ef20cd69cf14f7de917018578c1332cR13

notification/slack/actions.go
Forfold: I'm wondering if, since we will need both the User's name and the Channel's name, if we'll need to create a new source type, like "SourceTypeNotificationChannelUser" with two ID fields to use instead of the normal 1.

notification/slack/blocks.go
dctalbot: Good call, that looks great!

notification/slack/actions.go
Forfold: Per internal discussion, moving forward with the user's name and leaving out the channel for MVP:
"Acknowledged by Nathaniel Cook (Slack)"

This is acceptable because other log events do not show phone numbers or emails, i.e. "Acknowledged by Nathaniel Cook (SMS)"

PR #1364, opened 2021-03-29
ui: add time zone abbriviation in filter

### LINKED ISSUE #1066: ui: add tz abbreviation to TZ toggle
It would be useful if we appended the time zone short code to the filter toggle used throughout GoAlert, i.e. "Show Times in America/Chicago (CDT)"

Ensure CDT vs. CST and so on is accurately rendered.

**Screenshots:**

KatieMSB:

**Description:**
Adds time zone abbreviation in ScheduleTZFilter.js

Fixes #1066

**Screenshots:**
Screen Shot 2021-03-29 at 1 31 47 PM

### DIFF
--- a/web/src/app/schedules/ScheduleTZFilter.js
+++ b/web/src/app/schedules/ScheduleTZFilter.js
@@ -22,15 +22,16 @@ export function ScheduleTZFilter(props) {
     variables: { id: props.scheduleID },
   })
 
-  let label, tz, short
+  let label, tz
   if (error) {
     label = 'Error: ' + (error.message || error)
   } else if (!data && loading) {
     label = 'Fetching timezone information...'
   } else {
     tz = data.schedule.timeZone
-    short = DateTime.fromObject({ zone: tz }).toFormat('ZZZZ')
-    label = props.label ? props.label(tz) : `Show times in ${tz} (${short})`
+    const short = DateTime.fromObject({ zone: tz }).toFormat('ZZZZ')
+    const tzName = tz === short ? tz : tz + ` (${short})`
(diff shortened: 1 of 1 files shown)


web/src/app/schedules/ScheduleTZFilter.js
```diff
@@ -21,14 +22,15 @@ export function ScheduleTZFilter(props) {
...
-  let label, tz
+  let label, tz, short
   if (error) {
     label = 'Error: ' + (error.message || error)
   } else if (!data && loading) {
     label = 'Fetching timezone information...'
   } else {
     tz = data.schedule.timeZone
-    label = props.label ? props.label(tz) : `Show times in ${tz}`
+    short = DateTime.fromObject({ zone: tz }).toFormat('ZZZZ')
```
mastercactapus, line 32: We can move the `let` keyword here since it doesn't need to be used outside this block. Also we should add a condition to not repeat if the `short === tz` (e.g., UTC)

web/src/app/schedules/ScheduleTZFilter.js
```diff
@@ -21,14 +22,15 @@ export function ScheduleTZFilter(props) {
...
+    label = props.label ? props.label(tz) : `Show times in ${tz} (${short})`
```
mastercactapus, line 33: We should pass the short bit into `props.label` as well so we display consistently.

```suggestion
    let tzName = tz
    if (short !== tz) tz += ` (${short})`
    label = props.label ? props.label(tzName) : `Show times in ${tzName}`
```

mastercactapus (changes requested):

mastercactapus (approved):

m17ch (approved):

PR #1377, opened 2021-04-01
ui/forms: copy values into mutable object

dctalbot:

**Description:**
This PR will copy form values into a mutable object. Without this, we are unable to initialize forms with values from Apollo that may be frozen in the cache

Users can toggle existing checkboxes when editing a schedule assignment

**Additional Info:**
This may introduce a performance hit unfortunately, but for the time being correctness is more important for UX.


Forfold (dismissed):
Functionally verified and LGTM. I didn't notice any visible performance change

mastercactapus (approved):

Forfold (approved):

PR #1380, opened 2021-04-01
ui: lodash -> lodash-es

dctalbot:

**Description:**
This PR replaces `lodash` with `lodash-es`, the main benefit being that the library will run in strict mode


mastercactapus:
Instead of swapping the import to catch issues with `set` maybe we can just add `"use strict";` to the top of the `app.js` bundle -- running the entire application & all libraries in strict mode.

mastercactapus:
What about something like this: https://github.com/mistadikay/strict-loader

It was an intentional decision to switch to `lodash` from places that used `lodash-es` so that we only had a single copy (other libraries use `lodash` as well) and because of the build times/complexity of trying to use babel for `node_modules` with varying tools.

If the only reason for switching is because that version happens to use strict mode, I'd rather adopt a solution we can apply to other libraries as well.

PR #1392, opened 2021-04-06
oncall: fix shift calculation with gaps

mastercactapus:

**Description:**
This PR fixes two issues with the new `oncall` code:
- Out-of-order overrides would cause shift calculation to fail, they are now sorted like history entries
- Calculating future shifts would sometimes connect to current on-call shifts, time between `now()` and `Start` is now processed to determine actual start times


pnengchu (approved):

m17ch (commented):
lgtm

m17ch (approved):

PR #1394, opened 2021-04-07
migrate: no migrations for api only

mastercactapus:

**Description:**
This PR disables migrations when in `--api-only` mode. This ensures only instances running an engine loop can apply a new migration, ensuring messages always go out.

Additionally, retry messages are suppressed when applying migrations. Only the final error will be reported (if it fails).

### DIFF
--- a/app/cmd.go
+++ b/app/cmd.go
@@ -114,7 +114,7 @@ var RootCmd = &cobra.Command{
 		if cfg.APIOnly {
 			err = migrate.VerifyAll(log.EnableDebug(ctx), cfg.DBURL)
 			if err != nil {
-				return errors.Wrap(err, "apply migrations")
+				return errors.Wrap(err, "verify migrations")
 			}
 		} else {
 			s := time.Now()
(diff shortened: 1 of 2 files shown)


app/cmd.go
```diff
@@ -111,13 +111,17 @@ var RootCmd = &cobra.Command{
...
 
-		s := time.Now()
-		n, err := migrate.ApplyAll(log.EnableDebug(ctx), cfg.DBURL)
-		if err != nil {
-			return errors.Wrap(err, "apply migrations")
-		}
-		if n > 0 {
-			log.Logf(ctx, "Applied %d migrations in %s.", n, time.Since(s))
+		if cfg.APIOnly {
+			err = migrate.VerifyAll(log.EnableDebug(ctx), cfg.DBURL)
```
KatieMSB, line 115:
```suggestion
			err = migrate.VerifyAll(log.EnableDebug(ctx), cfg.DBURL)
			if err != nil {
				return errors.Wrap(err, "verify migrations")
			}
```
Missing error check here that's causing `make check` to fail

KatieMSB (changes requested):

app/cmd.go
mastercactapus: Well, that's embarrassing. fixed

KatieMSB (approved):

m17ch (approved):

PR #1408, opened 2021-04-12
temp-schedules: fix shift list transitions

KatieMSB:

**Description:**
- Fixes bug where no coverage notices were not rendering properly in the shift list
- Fixes bug where shifts weren't splitting when overlapping multiple days.

### DIFF
--- a/web/src/app/lists/FlatList.tsx
+++ b/web/src/app/lists/FlatList.tsx
@@ -63,12 +63,12 @@ const useStyles = makeStyles({
 export interface FlatListSub {
   id?: string
   subHeader: string
-  noTransition?: boolean
 }
 
 export interface FlatListNotice extends Notice {
   id?: string
   icon?: JSX.Element
+  transition?: boolean
 }
 export interface FlatListItem {
   title?: string
@@ -219,8 +219,8 @@ export default function FlatList({
           <CSSTransition
             key={'notice_' + item.id + idx}
             timeout={500}
-            exit={!item.noTransition}
-            enter={!item.noTransition}
+            exit={false || item.transition}
+            enter={false || item.transition}
             classNames={{
               enter: classes.slideEnter,
               enterActive: classes.slideEnterActive,
--- a/web/src/app/lists/FlatList.tsx
+++ b/web/src/app/lists/FlatList.tsx
@@ -61,6 +61,7 @@ const useStyles = makeStyles({
 })
 
 export interface FlatListSub {
+  id?: string
   subHeader: string
 }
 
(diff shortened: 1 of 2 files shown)


web/src/app/lists/FlatList.tsx
```diff
@@ -61,7 +61,9 @@ const useStyles = makeStyles({
 })
 
 export interface FlatListSub {
+  id?: string
   subHeader: string
+  noTransition?: boolean
```
dctalbot, line 66: I wonder if the default behavior for subheaders should be to _not_ transition, and we could provide a `transition` prop to turn it on instead. That way you would also avoid the double negatives later on when trying to do `!item.noTransition`

```suggestion
  transition?: boolean
```

web/src/app/lists/FlatList.tsx
```diff
@@ -202,16 +204,23 @@ export default function FlatList({
...
           </CSSTransition>
         )
       }
       if ('type' in item) {
         return (
           <CSSTransition
-            key={'notice_' + (item.id || idx)}
+            key={'notice_' + item.id + idx}
             timeout={500}
+            exit={false || item.transition}
```
dctalbot, line 222:
```suggestion
            exit={item.transition}
```
false or true => true
false or false => false

web/src/app/lists/FlatList.tsx
```diff
@@ -202,16 +204,23 @@ export default function FlatList({
...
+            enter={false || item.transition}
```
dctalbot, line 223:
```suggestion
            enter={item.transition}
```

dctalbot (changes requested):

web/src/app/lists/FlatList.tsx
dctalbot: Oh, never mind just realized this is in typescript. I probably would have used `Boolean(item.transition)`, but this works too!

dctalbot (approved):

web/src/app/lists/FlatList.tsx
mastercactapus: This will always return what's after the `||`, to convert it to bool either use `Boolean(item.transition)` or `!!item.transition` alternatively swapping the order to `item.transition || false` would also give the intended effect of returning `false` for all falsy values.


example: `console.log(false || 0)` prints `0` not `false`.

dctalbot: yeah nice catch, so `false || undefined` yielded `undefined` and was interpreted as falsey. In that case, if typescript isn't complaining, could probably just use `item.transition`

dctalbot (approved):

arurao (approved):

PR #1414, opened 2021-04-13
add teamID field to SlackChannel graphql entity

### LINKED ISSUE #1338: ui/util: links on slack chips are inconvenient if logged into alternate workspace
**Describe the Bug:**
Links on Slack chips are not necessarily routed to correct channel-in-workspace

**Steps to Reproduce:**
1. Authenticate for an alternate workspace in browser
2. Click on a slack chip
3. Redirect will display something like "unable to find channel in current workspace"
4. User must authenticate for the correct workspace to proceed

**Expected Behavior:**
An uninterrupted UX when clicking on slack chip

**Observed Behavior:**
See above

**Application Version:**
`e7bbe84b25ffde243dbaad0daa0037720a3064ea`

**Additional Context:**
We should provide the `team` query parameter in the redirect URL
https://api.slack.com/reference/deep-linking#deep-linking-into-your-slack-app__opening-a-channel-by-name-or-id

dctalbot:

**Description:**
This PR adds a teamID paramter to `notification_channels.meta` and the URL on Slack Chip components

Fixes #1338 

User's Slack browser sessions will not interfere with the "click to open slack channel" feature

Adds a `teamID` field to `SlackChannel`


notification/slack/channel.go
```diff
@@ -44,15 +45,17 @@ func NewChannelSender(ctx context.Context, cfg Config) (*ChannelSender, error) {
 		chanTht: newThrottle(time.Minute / 50),
 		listTht: newThrottle(time.Minute / 50),
 
-		listCache: newTTLCache(250, time.Minute),
-		chanCache: newTTLCache(1000, 15*time.Minute),
+		listCache:   newTTLCache(250, time.Minute),
+		chanCache:   newTTLCache(1000, 15*time.Minute),
+		teamIDCache: newTTLCache(1, 5*time.Minute),
```
mastercactapus, line 50: This should probably just be cached based on the access token value, rather than time

notification/slack/channel.go
```diff
@@ -149,10 +162,15 @@ func (s *ChannelSender) loadChannel(ctx context.Context, channelID string) (*Cha
 		return nil, wrapError(resData.Error, "lookup Slack channel")
 	}
 
-	return &Channel{
-		ID:   resData.Channel.ID,
-		Name: "#" + resData.Channel.Name,
-	}, nil
+	if teamIDStr, ok := teamID.(*string); !ok {
+		return nil, wrapError(resData.Error, "lookup Slack channel")
+	} else {
```
mastercactapus, line 167: Don't use `else` if we're returning, also this check should happen up where we're actually doing the lookup

notification/slack/channel.go
```diff
@@ -98,8 +101,18 @@ func (s *ChannelSender) Channel(ctx context.Context, channelID string) (*Channel
...
 	cfg := config.FromContext(ctx)
 
+	teamID, ok := s.teamIDCache.Get(cfg.Slack.AccessToken)
+	if !ok {
+		teamID, err = GetTeamID(ctx)
+		if err != nil {
+			return nil, err
+		}
+		s.teamIDCache.Add(cfg.Slack.AccessToken, teamID)
+	}
```
mastercactapus, line 114: We should handle this in an atomic way so we don't potentially spam requests instead of one

notification/slack/util.go
```diff
@@ -0,0 +1,36 @@
...
+	"net/http"
+
+	"github.com/target/goalert/config"
+)
+
+type Meta struct {
+	TeamID string `json:"team_id"`
+}
+
+func GetTeamID(ctx context.Context) (*string, error) {
```
mastercactapus, line 15: Make this private and just return a string instead of a pointer. As a helper method, pass the token directly instead of pulling values from context so dependencies/arguments are clear.

```suggestion
func lookupTeamIDForToken(ctx context.Context, token string) (string, error) {
```

notification/slack/util.go
```diff
@@ -0,0 +1,36 @@
...
+	"context"
+	"encoding/json"
+	"net/http"
+
+	"github.com/target/goalert/config"
+)
+
+type Meta struct {
+	TeamID string `json:"team_id"`
+}
```
mastercactapus, line 13: This should probably just go inside the `getTeamID` func for parsing the return value

web/src/app/util/Chips.tsx
```diff
@@ -105,7 +106,33 @@ export function ScheduleChip(props: WithID<ChipProps>): JSX.Element {
 }
 
 export function SlackChip(props: WithID<ChipProps>): JSX.Element {
-  const { id: channelID } = props
+  const { id: channelID, ...rest } = props
+
+  const query = gql`
+    query($id: ID!) {
+      slackChannel(id: $id) {
+        teamID
```
mastercactapus, line 114: Include `id` in the requested fields so that Apollo's cache can differentiate, additionally we can set the policy to `cache-first`

mastercactapus (changes requested):
We should pull out the DB changes since they are no longer required

notification/slack/channel.go
dctalbot: Can you please clarify this a bit? Not sure how we would potentially spam requests here

notification/slack/util.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+)
+
+func lookupTeamIDForToken(ctx context.Context, token string) (string, error) {
+	type RespBody struct {
+		TeamID string `json:"team_id"`
+	}
```
mastercactapus, line 12: nit: declare this inline by the var below

```go
var body struct { TeamID string `json:"team_id"` }
err = json.NewDecoder( // ...
```

notification/slack/channel.go
```diff
@@ -98,8 +102,18 @@ func (s *ChannelSender) Channel(ctx context.Context, channelID string) (*Channel
...
 	cfg := config.FromContext(ctx)
 
+	teamID, ok := s.teamIDCache.Get(cfg.Slack.AccessToken)
+	if !ok {
+		teamID, err = lookupTeamIDForToken(ctx, cfg.Slack.AccessToken)
+		if err != nil {
+			return nil, err
+		}
+		s.teamIDCache.Add(cfg.Slack.AccessToken, teamID)
+	}
```
mastercactapus, line 115: We should wrap this in a mutex so we only fetch once if we get concurrent calls. Also, we could probably drop the `lru` usage since it doesn't get us much.
```suggestion
	s.teamMx.Lock()
	if s.teamID == "" || s.token != cfg.Slack.AccessToken {
		// token changed or teamID missing
		id, err := lookupTeamIDForToken(ctx, cfg.Slack.AccessToken)
		if err != nil {
			s.teamMx.Unlock() // always unlock
			return nil, err
		}
		
		// update teamID and token after fetching succeeds
		s.teamID = id
		s.token = cfg.Slack.AccessToken
	}
	teamID := s.teamID
	// unlock after copying current teamID
	s.teamMx.Unlock()
```

mastercactapus (changes requested):

notification/slack/channel.go
dctalbot: Nice, I was also considering getting rid of the LRU but still wanted the store. This looks great!

notification/slack/channel.go
mastercactapus: This method can be called multiple times by concurrent HTTP requests, for multiple `channelID`s, all of which will have `ok` being false at first, then all calling `GetTeamID` at the same time.

notification/slack/channel.go
```diff
@@ -150,8 +170,9 @@ func (s *ChannelSender) loadChannel(ctx context.Context, channelID string) (*Cha
 	}
 
 	return &Channel{
-		ID:   resData.Channel.ID,
-		Name: "#" + resData.Channel.Name,
+		ID:     resData.Channel.ID,
+		Name:   "#" + resData.Channel.Name,
+		TeamID: s.teamID,
```
mastercactapus, line 175: We can't access `s.teamID` outside of the mutex, otherwise, there's the potential for concurrent read & write access. Assigning a copy to a variable before `Unlock` can work around this.

```go
teamID := s.teamID
s.teamMx.Unlock()

return &Channel{
	TeamID: teamID,
}
```

mastercactapus (changes requested):

mastercactapus (approved):

pnengchu (dismissed):

mastercactapus (dismissed):

devtools/mockslack/authtest.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+// ServeAuthTest serves a request to the `auth.test` API call.
+//
+// https://slack.com/api/auth.test
+func (s *Server) ServeAuthTest(w http.ResponseWriter, req *http.Request) {
+	var respData struct {
+		TeamID string `json:"team_id"`
+	}
+	respData.TeamID = genID(idChars, 9)
```
mastercactapus, line 14: This should be persistent, maybe add `TeamID` to `state` and add a teamID gen method to `gen` that begins with `T`

mastercactapus (changes requested):

pnengchu (approved):

mastercactapus (approved):

PR #1420, opened 2021-04-14
temp-schedules: always display start/end times in shift list

KatieMSB:

**Description:**
Shift list in the temporary schedule now always displays the start/end times and days with no coverage.

### DIFF
--- a/web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
+++ b/web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
@@ -104,17 +104,6 @@ export default function TempSchedShiftsList({
 
     const result: FlatListListItem[] = []
 
-    // render no coverage/get started below start time if no shifts
-    if (!sortedShifts.length) {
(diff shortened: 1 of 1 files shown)


dctalbot (changes requested):
I'm seeing the blue notice drop down into the list before transitioning out. Can we just remove that notice now that we are displaying warnings from the start?

KatieMSB:
> I'm seeing the blue notice drop down into the list before transitioning out. Can we just remove that notice now that we are displaying warnings from the start?

Yea since it's pretty clear there's no coverage at the start I think it's fair to remove it.

dctalbot (changes requested):
Seeing a similar issue when deleting a shift, it drops to the bottom of the list before transitioning out

Forfold (approved):
lgtm! i hope my comments around rendering all the shifts proved helpful haha

dctalbot (approved):
Let's get this in for demo, can polish up the remaining bits later :)

PR #1438, opened 2021-04-21
dev: lock build-env version

mastercactapus:

**Description:**
Updates Dockerfiles and task files to use a versioned tag of the build-env image.

### DIFF
--- a/devtools/ci/tasks/scripts/codecheck.sh
+++ b/devtools/ci/tasks/scripts/codecheck.sh
@@ -20,6 +20,33 @@ if [ "$PKG_JSON_VER" != "$DOCKERFILE_VER" ]; then
   exit 1
 fi
 
+# assert build-env versions are identical
+BUILD_ENV_VER=go1.16.3-postgres13
+for file in $(find devtools -name 'Dockerfile*')
+do
+  if ! grep -q "goalert/build-env" "$file"; then
+    continue
+  fi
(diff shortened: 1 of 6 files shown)


devtools/ci/tasks/scripts/codecheck.sh
```diff
@@ -20,6 +20,33 @@ if [ "$PKG_JSON_VER" != "$DOCKERFILE_VER" ]; then
...
+    echo "  $(grep goalert/build-env "$file")"
+    exit 1
+  fi
+done
+for file in $(find devtools -name '*.yml')
+do
+  if ! grep -q "goalert/build-env" "$file"; then
+    continue
+  fi
+  if ! grep -q "goalert/build-env, tag: $BUILD_ENV_VER" "$file"; then
```
dctalbot, line 42:
```suggestion
  if ! grep -q "tag: $BUILD_ENV_VER" "$file"; then
```
yml objects aren't necessarily inline

dctalbot (approved):

devtools/ci/tasks/scripts/codecheck.sh
mastercactapus: True, but that's the reason for the previous line's check -- putting it on multiple lines will cause the check to fail so it would need to be investigated/fixed.

pnengchu (approved):

PR #1452, opened 2021-04-23
Refactoring RotationAddUserDialog to use hooks

kanish671:

**Description:**
Refactored `web/src/app/rotations/RotationAddUserDialog.js` to use hooks.

Part of #923

### DIFF
--- a/web/src/app/rotations/RotationAddUserDialog.js
+++ b/web/src/app/rotations/RotationAddUserDialog.js
@@ -1,7 +1,6 @@
-import React from 'react'
-import { gql } from '@apollo/client'
-import { PropTypes as p } from 'prop-types'
-import { Mutation } from '@apollo/client/react/components'
+import React, { useState } from 'react'
+import { gql, useMutation } from '@apollo/client'
+import p from 'prop-types'
 import { fieldErrors, nonFieldErrors } from '../util/errutil'
 import UserForm from './UserForm'
 import FormDialog from '../dialogs/FormDialog'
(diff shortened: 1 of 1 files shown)


web/src/app/rotations/RotationAddUserDialog.js
```diff
@@ -12,66 +11,54 @@ const mutation = gql`
...
-      users: [],
-    }
-
-    return (
-      <Mutation mutation={mutation}>
-        {(commit, status) => this.renderDialog(defaultValue, commit, status)}
-      </Mutation>
-    )
-  }
+  const [updateRotationMutation, updateRotationMutationStatus] = useMutation(
```
Forfold, line 26:
```suggestion
  const [updateRotationMutation, { loadng, error, data }] = useMutation(
```

We can inline loading, error, and data here so we don't have to call a long var name later on

web/src/app/rotations/RotationAddUserDialog.js
```diff
@@ -12,66 +11,54 @@ const mutation = gql`
...
+const RotationAddUserDialog = (props) => {
+  const { rotationID, onClose } = props
+  const [value, setValue] = useState(null)
+  const defaultValue = {
+    users: [],
   }
+  // append to users array from selected users
+  const users = []
+  const userIDs = (value && value.users) || defaultValue.users
+  props.userIDs.forEach((u) => users.push(u))
```
Forfold, line 23: Let's call `userIDs` flat (destructered from props like the other prop variables on [L15](https://github.com/target/goalert/pull/1452/files?diff=split&w=1#diff-4fe778655b2db0c3985ad8d1b6c82ffd2a29d801955e9fc7c9f35720f2ea42faR15)), and rename the local variable to something like `uids`.

web/src/app/rotations/RotationAddUserDialog.js
```diff
@@ -12,66 +11,54 @@ const mutation = gql`
...
 `
 
-export default class RotationAddUserDialog extends React.Component {
-  static propTypes = {
-    rotationID: p.string.isRequired,
-    userIDs: p.array.isRequired,
-    onClose: p.func.isRequired,
+const RotationAddUserDialog = (props) => {
+  const { rotationID, onClose } = props
+  const [value, setValue] = useState(null)
```
Forfold, line 16:
```suggestion
  const [value, setValue] = useState({
    users: []
  })
```

It might be worth setting the defaultValue as the default state so we don't have to have the separate checks throughout

web/src/app/rotations/RotationAddUserDialog.js
```diff
@@ -12,66 +11,54 @@ const mutation = gql`
...
-            value={this.state.value || defaultValue}
-            onChange={(value) => this.setState({ value })}
-          />
-        }
-      />
-    )
-  }
+RotationAddUserDialog.propTypes = {
+  rotationID: p.string.isRequired,
+  userIDs: p.array.isRequired,
```
Forfold, line 60:
```suggestion
  userIDs: p.arrayOf(p.string).isRequired,
```

Forfold (changes requested):

Forfold (approved):
lgtm, thank you for making my recommended changes!

dctalbot (approved):

PR #1476, opened 2021-04-29
temp schedules: preserve unknown fields

mastercactapus:

**Description:**
This PR is future-proofing by ensuring that unknown fields (e.g. added by a future new version of GoAlert) are preserved in schedule data when processing.

This is done by ensuring changes to temp schedules (read-edit-writes to the `schedule_data` table) pass through a merging function. It merges objects from the root against the existing raw JSON. Arrays are intentionally left out, *any* changes to array structs require a new version node or field. This uses the same code that is used for the same purpose in the `config` package, which has been moved to a `jsonutil` package to be shared.


config/store.go
```diff
@@ -302,12 +303,7 @@ func (s *Store) updateConfigTx(ctx context.Context, tx *sql.Tx, fn func(Config)
 		return 0, err
 	}
 
-	data, err := json.Marshal(newCfg)
-	if err != nil {
-		return 0, errors.Wrap(err, "marshal config")
-	}
-
-	data, err = mergeJSON(cfg.data, data)
```
Forfold, line 310: We should be able to safely delete this func now, too (currently only being referenced here and in a test) now that `Merge()` has been created

m17ch (approved):
LGTM 👍

schedule/storetemporaryschedules.go
```diff
@@ -116,7 +117,12 @@ func (store *Store) updateFixedShifts(ctx context.Context, tx *sql.Tx, scheduleI
 		return err
 	}
 
-	rawData, err = json.Marshal(data)
+	if len(rawData) > 0 {
```
Forfold, line 120: Do we need to be unmarshalling on [L109](https://github.com/target/goalert/pull/1476/files#diff-171bca7f32612fd22236d2f6213d0509513e843a62673c8f1e66b2eeb0688740R109) above since we do so inside of `Merge()`?

schedule/storetemporaryschedules.go
```diff
@@ -116,7 +117,12 @@ func (store *Store) updateFixedShifts(ctx context.Context, tx *sql.Tx, scheduleI
...
+		// preserve unknown fields
+		rawData, err = jsonutil.Apply(rawData, data)
+	} else {
+		rawData, err = json.Marshal(data)
+	}
```
Forfold, line 125: Is this check needed since in `Merge()` if the dst is empty, we handle creating a new map with just the new data?

util/jsonutil/apply.go
```diff
@@ -0,0 +1,12 @@
+package jsonutil
+
+import "encoding/json"
+
+// Apply will recursively merge a src value into the dst JSON document.
+func Apply(dst []byte, src interface{}) ([]byte, error) {
```
Forfold, line 6: Since we aren't doing something like `data.Apply(rawData)`, would it be more appropriate to include this in logic `Merge()`? It appeared to me that `applyValues` is the only function here with the potential to be recursive

schedule/storetemporaryschedules.go
mastercactapus: Yes, because we need the struct populated (if `rawData` isn't empty) for the `apply` function that makes whatever necessary changes.

schedule/storetemporaryschedules.go
mastercactapus: Correct, we use the new data as-is if there was nothing before and only do the merge if we had JSON to begin with.

util/jsonutil/apply.go
mastercactapus: `Apply` is a convenience method that can be used in-place of `json.Marshal` but with a source document. Merge takes two already-marshalled byte slices.

config/store.go
mastercactapus: Good catch, I think I mistakenly re-introduced these two files when juggling `git stash`-es.

schedule/storetemporaryschedules.go
Forfold: I mean, would we be able to do something like

```suggestion
rawData, err = jsonutil.Apply(rawData, data)
```

since the result should be the same?

mastercactapus: Just added a unit test, it handles when `rawData` is nil properly, so I'll remove the `if` check to simplify.

util/jsonutil/apply.go
Forfold: I was mainly wondering if `jsonutil.Merge(src, dst)`  would make more sense over `jsonutil.Apply(src, dst)`, since `Apply` just passes itself through to `Merge`. Could we do the marshalling in Merge without needing them separated out?

m17ch (approved):

util/jsonutil/apply.go
mastercactapus: IMO `Merge` would imply combining two same-type things (e.g. two JSON documents, two structs, arrays, etc..) so that's why I chose to make a new function `Apply`. As in: `Apply this *to* the JSON document` vs `Merge these two documents`.

The only real reason for `Apply` is it seemed to make the code a little more readable by getting rid of the marshaling noise by being able to just call `jsonutil.Apply(srcDoc, newStruct)` since that's common enough case. `Merge` could potentially be made private, but didn't seem necessary as long as they both have comments/documentation on what does what.

Forfold (approved):

PR #1483, opened 2021-04-30
users: fix fkey and rotation issues when deleting a user

mastercactapus:

**Description:**
Fixes issues when deleting a user:
- User is removed from all rotations prior to deleting the row (works around rotation trigger conflicts for multi-row)
- Attempt is retried up to 5 times in case of temporary or fkey errors

### DIFF
--- a/user/store.go
+++ b/user/store.go
@@ -6,10 +6,12 @@ import (
 	"errors"
 	"fmt"
 	"sync/atomic"
+	"time"
 
 	"github.com/golang/groupcache"
 	uuid "github.com/satori/go.uuid"
 	"github.com/target/goalert/permission"
+	"github.com/target/goalert/retry"
 	"github.com/target/goalert/util"
 	"github.com/target/goalert/util/sqlutil"
 	"github.com/target/goalert/validation/validate"
@@ -45,14 +47,19 @@ type DB struct {
 
 	ids *sql.Stmt
 
-	insert  *sql.Stmt
-	update  *sql.Stmt
-	delete  *sql.Stmt
+	insert *sql.Stmt
+	update *sql.Stmt
+
 	findOne *sql.Stmt
 	findAll *sql.Stmt
 
-	findMany   *sql.Stmt
-	deleteMany *sql.Stmt
+	findMany *sql.Stmt
+
+	deleteOne          *sql.Stmt
+	userRotations      *sql.Stmt
+	rotationParts      *sql.Stmt
+	updateRotationPart *sql.Stmt
+	deleteRotationPart *sql.Stmt
 
 	findOneForUpdate *sql.Stmt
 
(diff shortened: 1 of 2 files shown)


user/store.go
```diff
@@ -95,18 +102,22 @@ func NewDB(ctx context.Context, db *sql.DB) (*DB, error) {
...
 			FROM users
 			WHERE id = any($1)
 		`),
-		deleteMany: p.P(`DELETE FROM users WHERE id = any($1)`),
+
+		deleteOne: p.P(`DELETE FROM users WHERE id = $1`),
+		userRotations: p.P(`
+			SELECT DISTINCT rotation_id
+			FROM rotation_state state
+			WHERE EXISTS (SELECT 1 FROM rotation_participants WHERE rotation_id = state.rotation_id AND user_id = $1)
```
dctalbot, line 116:
```suggestion
			SELECT DISTINCT a.rotation_id 
			FROM rotation_state a 
			JOIN rotation_participants b 
			ON a.rotation_id = b.rotation_id 
			WHERE parts.user_id = $1;
```
Haven't done any perf comparison, but this feels more conventional

dctalbot: Is there a reason we couldn't do `select distinct rotation_id from rotation_participants where user_id = $1`, or are the extra checks just for good measure?

user/store.go
```diff
@@ -227,18 +253,143 @@ func (db *DB) Insert(ctx context.Context, u *User) (*User, error) {
...
+	rows, err := tx.StmtContext(ctx, db.rotationParts).QueryContext(ctx, rotationID)
+	if err != nil {
+		return fmt.Errorf("query participants: %w", err)
+	}
+	defer rows.Close()
+	for rows.Next() {
+		var p part
+		err = rows.Scan(&p.ID, &p.UserID)
+		if err != nil {
+			return fmt.Errorf("scan participant %d: %w", len(participants), err)
```
dctalbot, line 335: I'm curious why we are logging `len(participants)` here as opposed to the `p.Id`?

mastercactapus: It's to log the index/position number since we're processing them in order, tells us if it failed on the 1st vs 3rd scan. Since the scan failed, we don't have `p.ID` and it would just be empty.

dctalbot: Makes sense, I should've caught that :)

dctalbot (changes requested):
Saw this when trying to delete six users at once. We can dig into the core issue some time next week
```
Backend  (err): time="2021-04-30T11:42:28-05:00" level=error AuthSystemComponent=Engine SQLErrDetails="Key (user_id)=(225a77da-01fb-4406-90c9-4610ab70465b) is not present in table \"users\"." Trigger=INTERVAL error="Engine.EscalationManager: update ep step on-call: ERROR: insert or update on table \"ep_step_on_call_users\" violates foreign key constraint \"ep_step_on_call_users_user_id_fkey\" (SQLSTATE 23503)"
```

dctalbot (approved):
After discussing this morning, we are OK ignoring the logged error. It is a red herring and not an indication of any real breakage. A separate body of work on the engine will address this in the future

arurao (approved):

Issue #1247, opened 2021-02-08
allow managing users from admin panel


mastercactapus:
**Is your feature request related to a problem? Please describe:**
It is not currently possible to do anything other than adding users via the CLI for basic auth.

**Describe the solution you'd like:**
Perform the following from within the GoAlert web UI:
- [x] Add users (@dctalbot)
- [x] Delete users (@dctalbot and @KatieMSB)
- [x] Change user roles (@arurao)
- [ ] Allow users to change their password when Basic auth is enabled from Profile
- [ ] Consolidate UI into centralized dashboard - may be out of scope/needs more discussion on design

**Describe alternatives you've considered:**
These steps are only currently possible with direct DB access and knowledge of GoAlert internals.

Forfold:
> - Allow users to change their password when Basic auth is enabled from Profile

This would be a good candidate for #1068

PR #1506, opened 2021-05-06
users: change user role

arurao:

**Description:**
This PR provides the ability for admins to change user roles. 

Part of #1247

Admins will be able to see 'Edit' icon on the user details page. 

Screen Shot 2021-05-11 at 4 54 34 PM

After that, there will be an option to check/uncheck the admin role option. 

Screen Shot 2021-05-10 at 3 26 37 PM

Checking it will set the user as admin, unchecking will set the role to user. 
If the logged-in user is already an admin, unchecking the checkbox will show a warning notice mentioning the change in role and it's effects.

Screen Shot 2021-05-10 at 3 27 38 PM

A split logic has been applied to the `updateUser` mutation. 
Whenever the `role` field is specified, the logic will go to a different store method that updates the user role, else it will go to the store update a user logic.
This decision was made since even though there is currently no UI for updateUser fields like name, email, etc that might change in the future (and is possible today via the graphql playground) editing user fields like name, email, etc could be done by the user themselves. But in the case of changing user's role, this operation is strictly limited to admins only. No non-admin user should be allowed to make themselves as an admin. So, in order to keep the two different permission checks separate for updating user fields and changing user role, the logic is split up accordingly.


graphql2/schema.graphql
```diff
@@ -972,6 +973,11 @@ input UpdateUserInput {
   statusUpdateContactMethodID: ID
 }
 
+input SetUserRoleInput {
+  id: ID!
+  role: UserRole!
+}
```
mastercactapus, line 979: From `GraphQL` we can probably stick with `updateUser` and we can switch off of the `role` field in the backend. Keep `SetUserRole` in the `user.Store` though

user/store.go
```diff
@@ -23,6 +23,7 @@ type Store interface {
 	InsertTx(context.Context, *sql.Tx, *User) (*User, error)
 	Update(context.Context, *User) error
 	UpdateTx(context.Context, *sql.Tx, *User) error
+	SetUserRoleTx(ctx context.Context, tx *sql.Tx, u *User) error
```
mastercactapus, line 26: This should probably just take an `id` and `role` instead of a full user struct

user/store.go
```diff
@@ -102,6 +103,13 @@ func NewDB(ctx context.Context, db *sql.DB) (*DB, error) {
 			WHERE id = $1
 		`),
 
+		setUserRole: p.P(`
+			UPDATE users
+			SET
+				role = $2
+			WHERE id = $1
+		`),
```
mastercactapus, line 111: nit: just one-line this since it's so short

```suggestion
		setUserRole: p.P(`UPDATE users SET role = $2 WHERE id = $1`),
```

user/store.go
```diff
@@ -410,6 +418,24 @@ func (db *DB) UpdateTx(ctx context.Context, tx *sql.Tx, u *User) error {
...
+func (db *DB) SetUserRoleTx(ctx context.Context, tx *sql.Tx, u *User) error {
+	n, err := u.Normalize()
+	if err != nil {
+		return err
+	}
+
+	err = permission.LimitCheckAny(ctx, permission.System, permission.Admin)
+	if err != nil {
+		return err
+	}
```
mastercactapus, line 430: Always check permissions first, then validation

web/src/app/users/UserEditDialog.tsx
```diff
@@ -0,0 +1,106 @@
...
+                    <TableCell padding='checkbox'>
+                      <Checkbox
+                        checked={state.isAdmin}
+                        onChange={handleChange}
+                        name='isAdmin'
+                      />
+                    </TableCell>
+                  </Hidden>
+                </TableBody>
+              </Table>
```
mastercactapus, line 97: I think a table is overkill here, I think just a single checkbox and label will do

mastercactapus (changes requested):
A few high-level things we could

web/src/app/users/UserEditDialog.tsx
```diff
@@ -0,0 +1,106 @@
...
+          variables: {
+            input: {
+              id: props.userID,
+              role: state.isAdmin ? 'admin' : 'user',
+            },
+          },
+        })
+      }
+      notices={
+        props.role === 'admin' && state.isAdmin === false
```
mastercactapus, line 65: We should only show this notice if it's the current user

Forfold (changes requested):
<details>
<summary>
Edit: Fixed in latest commit
</summary>

I got this error when trying to update a user other than myself to admin, in the UI console:

Screen Shot 2021-05-06 at 1 42 10 PM

The network tab showed this response on the request:
```
{
  "errors": [
    {
      "message": "Cannot query field \"setUserRole\" on type \"Mutation\".",
      "locations": [{ "line": 2, "column": 3 }],
      "extensions": { "code": "GRAPHQL_VALIDATION_FAILED" }
    },
    {
      "message": "Unknown type \"SetUserRoleInput\".",
      "locations": [{ "line": 1, "column": 1 }],
      "extensions": { "code": "GRAPHQL_VALIDATION_FAILED" }
    }
  ],
  "data": null
}
```
</details>

web/src/cypress/integration/users.ts
```diff
@@ -50,6 +50,20 @@ function testUsers(screen: ScreenFormat): void {
...
+  describe('Page Actions', () => {
+    it('should edit a user role', () => {
+      cy.adminLogin()
+
+      cy.fixture('users').then((users) => {
+        cy.visit(`/users/${users[0].id}`)
+
+        cy.pageAction('Edit')
+        cy.get('[type="checkbox"]').check()
+        cy.dialogFinish('Confirm')
```
Forfold, line 63: Since we don't have any indication on the user page if someone is an admin or not, can we open the dialog again and validate that the checkbox is still filled as a sanity check?

web/src/app/users/UserEditDialog.tsx
```diff
@@ -0,0 +1,84 @@
+import React from 'react'
+import { gql, useMutation } from '@apollo/client'
+import Spinner from '../loading/components/Spinner'
+import FormDialog from '../dialogs/FormDialog'
+import { useSessionInfo } from '../util/RequireConfig'
+import { Checkbox, FormControlLabel } from '@material-ui/core'
```
Forfold, line 6: nit, move material import above local files:
```suggestion
import { gql, useMutation } from '@apollo/client'
import { Checkbox, FormControlLabel } from '@material-ui/core'
import Spinner from '../loading/components/Spinner'
import FormDialog from '../dialogs/FormDialog'
import { useSessionInfo } from '../util/RequireConfig'
```

web/src/app/users/UserEditDialog.tsx
```diff
@@ -0,0 +1,84 @@
...
+  role: string
+  onClose: () => void
+}
+
+function UserEditDialog(props: UserEditDialogProps): JSX.Element {
+  const { ready: isSessionReady, userID: currentUserID } = useSessionInfo()
+
+  const [state, setState] = React.useState({
+    isAdmin: props.role === 'admin',
+  })
```
Forfold, line 26: this can be simplified from an object to
```suggestion
  const [adminChecked, setAdminChecked] = React.useState(isAdmin)
```

would also have to update the `state.isAdmin` references below though, and make sure that `isAdmin` from `useSessionInfo` is called beforehand for scope

web/src/app/users/UserEditDialog.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  const [editUser, editUserStatus] = useMutation(mutation, {
+    onCompleted: props.onClose,
+  })
+
+  function handleChange(e: React.ChangeEvent<HTMLInputElement>): void {
+    setState({ ...state, [e.target.name]: e.target.checked })
```
Forfold, line 33:
```suggestion
    setAdminChecked(e.target.checked)
```

web/src/app/users/UserDetails.js
```diff
@@ -254,6 +262,13 @@ export default function UserDetails(props) {
           </Grid>
         }
       />
+      {showEdit && (
+        <UserEditDialog
+          onClose={() => setShowEdit(false)}
+          userID={props.userID}
+          role={user.role}
```
Forfold, line 269: To simplify the interface of this component, we have the ability to get the role of the user from the `useSessionInfo` call on L22 in `UserEditDialog.js`:

```
const { ready: isSessionReady, userID: currentUserID } = useSessionInfo()
```

->

```
const {
  isAdmin,
  ready: isSessionReady,
  userID: currentUserID
} = useSessionInfo()
```

web/src/app/users/UserDetails.js
```diff
@@ -19,14 +21,14 @@ import Spinner from '../loading/components/Spinner'
...
-import PageActions from '../util/PageActions'
+import UserEditDialog from './UserEditDialog'
 import UserDeleteDialog from './UserDeleteDialog'
-import OtherActions from '../util/OtherActions'
 
 const userQuery = gql`
   query userInfo($id: ID!) {
     user(id: $id) {
       id
+      role
```
Forfold, line 31: I think we can safely remove `role` from the queries here if we use our `useSessionInfo` hook (more info below on this)

Forfold (changes requested):

web/src/app/users/UserEditDialog.tsx
```diff
@@ -0,0 +1,84 @@
...
+                message: 'Updating role to User',
+                details:
+                  'If you remove your admin privileges you will need to log in as a different admin to restore them.',
+              },
+            ]
+          : []
+      }
+      form={
+        <FormControlLabel
+          label='admin'
```
Forfold, line 70: One last thing,
```suggestion
          label='Admin'
```

web/src/app/users/UserEditDialog.tsx
arurao: We would not be able to use `isAdmin` here from `useSessionInfo`. The reason being `isAdmin` checks whether the logged in user is an admin or not, whereas the use case for the variable here is the check the role of the user whose details page is being viewed.

web/src/app/users/UserDetails.js
arurao: As mentioned in an earlier comment, there needs to be a way to propogate the `role` of the user whose details page is being viewed. The `isAdmin` from `useSessionInfo` is if the logged in user is an admin or not, vs the role of the user for details page. That's why I thought querying for `role` and passing it onto `UserEditDialog` would be the best option.

web/src/app/users/UserEditDialog.tsx
Forfold: Ooooh that makes sense- I was only thinking in the context of the profile page for the permissions of opening the panel. I didn't realize that that could differ when editing someone else's page. That makes sense, ty!

Forfold (dismissed):

Forfold (approved):

Forfold (approved):

mastercactapus (approved):

PR #1536, opened 2021-05-13
ui/users: create user dialog

dctalbot:

**Description:**
Currently, goalert admins can only create users for basic auth via the CLI. This PR brings that functionality into the UI.

Part of #1247 

**Screenshots:**
Screen Shot 2021-05-18 at 8 30 17 AM

Screen Shot 2021-05-17 at 5 20 31 PM

Admins will have a create user fab on the users list page

```
  createUser(input: CreateUserInput!): User
```

```
input CreateUserInput {
  username: String!
  password: String!
  name: String
  email: String
  role: UserRole
}
```

**Additional Info:**
- This functionality may move into a central user management page in the future

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -12,6 +12,7 @@ import (
 	alertlog "github.com/target/goalert/alert/log"
 	"github.com/target/goalert/app/lifecycle"
 	"github.com/target/goalert/auth"
+	"github.com/target/goalert/auth/basic"
 	"github.com/target/goalert/auth/nonce"
 	"github.com/target/goalert/calendarsubscription"
 	"github.com/target/goalert/config"
@@ -84,6 +85,7 @@ type App struct {
 	AlertStore    alert.Store
 	AlertLogStore alertlog.Store
 
+	BasicStore            *basic.Store
 	UserStore             user.Store
 	ContactMethodStore    contactmethod.Store
 	NotificationRuleStore notificationrule.Store
--- a/app/app.go
+++ b/app/app.go
@@ -85,7 +85,7 @@ type App struct {
 	AlertStore    alert.Store
 	AlertLogStore alertlog.Store
 
-	BasicStore            *basic.Store
+	AuthBasicStore        *basic.Store
 	UserStore             user.Store
 	ContactMethodStore    contactmethod.Store
 	NotificationRuleStore notificationrule.Store
--- a/app/initgraphql.go
+++ b/app/initgraphql.go
@@ -19,7 +19,7 @@ func (app *App) initGraphQL(ctx context.Context) error {
 
 	app.graphql2 = &graphqlapp.App{
 		DB:                app.db,
-		BasicStore:        *app.BasicStore,
+		AuthBasicStore:    *app.AuthBasicStore,
 		UserStore:         app.UserStore,
 		CMStore:           app.ContactMethodStore,
 		NRStore:           app.NotificationRuleStore,
--- a/app/initgraphql.go
+++ b/app/initgraphql.go
@@ -19,6 +19,7 @@ func (app *App) initGraphQL(ctx context.Context) error {
 
 	app.graphql2 = &graphqlapp.App{
 		DB:                app.db,
+		BasicStore:        *app.BasicStore,
 		UserStore:         app.UserStore,
 		CMStore:           app.ContactMethodStore,
 		NRStore:           app.NotificationRuleStore,
--- a/app/initgraphql.go
+++ b/app/initgraphql.go
@@ -19,7 +19,7 @@ func (app *App) initGraphQL(ctx context.Context) error {
 
 	app.graphql2 = &graphqlapp.App{
 		DB:                app.db,
-		AuthBasicStore:    *app.AuthBasicStore,
+		AuthBasicStore:    app.AuthBasicStore,
 		UserStore:         app.UserStore,
 		CMStore:           app.ContactMethodStore,
 		NRStore:           app.NotificationRuleStore,
--- a/app/initstores.go
+++ b/app/initstores.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/target/goalert/alert"
 	alertlog "github.com/target/goalert/alert/log"
+	"github.com/target/goalert/auth/basic"
 	"github.com/target/goalert/auth/nonce"
 	"github.com/target/goalert/calendarsubscription"
 	"github.com/target/goalert/config"
@@ -133,6 +134,13 @@ func (app *App) initStores(ctx context.Context) error {
 	}
 	if err != nil {
 		return errors.Wrap(err, "init service store")
+	}
+
+	if app.BasicStore == nil {
+		app.BasicStore, err = basic.NewStore(ctx, app.db)
+	}
+	if err != nil {
+		return errors.Wrap(err, "init basic auth store")
 	}
 
 	if app.UserStore == nil {
--- a/app/initstores.go
+++ b/app/initstores.go
@@ -136,8 +136,8 @@ func (app *App) initStores(ctx context.Context) error {
 		return errors.Wrap(err, "init service store")
 	}
 
-	if app.BasicStore == nil {
-		app.BasicStore, err = basic.NewStore(ctx, app.db)
+	if app.AuthBasicStore == nil {
+		app.AuthBasicStore, err = basic.NewStore(ctx, app.db)
 	}
 	if err != nil {
 		return errors.Wrap(err, "init basic auth store")
--- a/graphql2/graphqlapp/app.go
+++ b/graphql2/graphqlapp/app.go
@@ -16,6 +16,7 @@ import (
 	"github.com/target/goalert/alert"
 	alertlog "github.com/target/goalert/alert/log"
 	"github.com/target/goalert/auth"
+	"github.com/target/goalert/auth/basic"
 	"github.com/target/goalert/calendarsubscription"
 	"github.com/target/goalert/config"
 	"github.com/target/goalert/escalation"
@@ -50,6 +51,7 @@ import (
 
 type App struct {
 	DB             *sql.DB
+	BasicStore     basic.Store
 	UserStore      user.Store
 	CMStore        contactmethod.Store
 	NRStore        notificationrule.Store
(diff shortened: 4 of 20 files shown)


mastercactapus:
A couple of initial thoughts:
- For role, maybe just an `Admin` checkmark in the dialog to replicate the `--admin` flag in CLI rather than making it a multi-step process
- Still allow creating a user if Basic auth is disabled, but add a notice to the dialog that they will be unable to login until it's turned on

web/src/app/users/UserCreateDialog.tsx
```diff
@@ -0,0 +1,115 @@
...
+  email: '',
+  isAdmin: false,
+}
+
+interface UserCreateDialogProps {
+  onClose: () => void
+}
+
+function UserCreateDialog(props: UserCreateDialogProps): JSX.Element {
+  const [value, setValue] = useState(initialValue)
```
Forfold, line 29:
```suggestion
interface UserCreateDialogProps {
  onClose: () => void
}

function UserCreateDialog(props: UserCreateDialogProps): JSX.Element {
  const [value, setValue] = useState({
  username: '',
  password: '',
  email: '',
  isAdmin: false,
})
```

web/src/app/users/UserCreateDialog.tsx
```diff
@@ -0,0 +1,115 @@
...
+      input: {
+        username: value.username,
+        password: value.password,
+        email: value.email,
+        role: value.isAdmin ? 'admin' : 'user',
+      },
+    },
+  })
+
+  if (data?.createUser) {
```
Forfold, line 42:
```suggestion
  if (!loading && data?.createUser) {
```

web/src/app/users/UserCreateDialog.tsx
```diff
@@ -0,0 +1,115 @@
...
+          value={value}
+          errors={fieldErrors(error)}
+          onChange={(val: typeof value) => setValue(val)}
+        >
+          <Grid container spacing={2}>
+            <Grid item xs={12}>
+              <FormField
+                fullWidth
+                component={TextField}
+                label='Username'
```
Forfold, line 76: we can leave out label if it's the same as the name (titleCase is handled within FormField)
```suggestion
```

web/src/app/users/UserCreateDialog.tsx
```diff
@@ -0,0 +1,115 @@
...
+                name='username'
+                required
+              />
+            </Grid>
+            <Grid item xs={12}>
+              <FormField
+                fullWidth
+                component={TextField}
+                label='Password'
```
Forfold, line 85:
```suggestion
```

web/src/app/users/UserCreateDialog.tsx
```diff
@@ -0,0 +1,115 @@
...
+                name='password'
+                required
+              />
+            </Grid>
+            <Grid item xs={12}>
+              <FormField
+                fullWidth
+                component={TextField}
+                label='Email'
```
Forfold, line 94:
```suggestion
```

web/src/app/users/UserCreateDialog.tsx
```diff
@@ -0,0 +1,115 @@
...
+              />
+            </Grid>
+            <Grid item xs={12}>
+              <FormField
+                fullWidth
+                component={TextField}
+                label='Password'
+                name='password'
+                required
+              />
```
Forfold, line 88: Can we add a second field confirming the password input?

As well as setting the type of the field to password so that the text is hidden, with the option to toggle and see the plaintext

web/src/app/users/UserCreateDialog.tsx
```diff
@@ -0,0 +1,115 @@
...
+                  'This user will be unable to log in until basic auth is enabled.',
+              },
+            ]
+          : []
+      }
+      form={
+        <FormContainer
+          value={value}
+          errors={fieldErrors(error)}
+          onChange={(val: typeof value) => setValue(val)}
```
Forfold, line 69:
```suggestion
          onChange={(val: typeof value) => setValue(val)}
          optionalLabels
```

What are your thoughts on flipping the require logic to `optionalLabels`? This would render the username and password labels without the `*`s, and render email's label as `Email (optional)`

web/src/app/users/UserCreateDialog.tsx
```diff
@@ -0,0 +1,115 @@
...
+        >
+          <Grid container spacing={2}>
+            <Grid item xs={12}>
+              <FormField
+                fullWidth
+                component={TextField}
+                label='Username'
+                name='username'
+                required
+              />
```
Forfold, line 79: Username error: `Can only contain lower-case letters and digits`

We should probably allow spaces, as full names are valid users in GoAlert

Forfold (changes requested):

web/src/app/lists/QueryList.tsx
```diff
@@ -79,6 +84,8 @@ export interface QueryListProps extends ControlledPaginatedListProps {
   mapVariables?: (vars: OperationVariables) => OperationVariables
 }
 
+export type QueryListProps = Omit<_QueryListProps, 'items'>
```
Forfold, line 87: Nice 😎, this tripped me up a couple of times working in Typescript with some of the QueryLists in the past

web/src/app/users/UserCreateDialog.tsx
dctalbot: Sure, that's a nice idea!

web/src/app/users/UserCreateDialog.tsx
dctalbot: I see we are also doing that for the create escalation policy dialog, works for me!

dctalbot: Although now I see we are _not_ doing that for the create server dialog. I think the asterisks are more conventional?

dctalbot: I will look for a material design guide re this

Forfold: Yeah I'm not sure- sounds good 👍

web/src/app/users/UserCreateDialog.tsx
dctalbot: Added the password confirm field, but decided against the visibility toggle. Implementation got a bit hairy and I didn't feel the potential value justified the complexity.

web/src/app/users/UserCreateDialog.tsx
dctalbot: > Required text indicator
> 
> To indicate that a field is required, display an asterisk (*) next to the label text and mention near the form that asterisks indicate required fields.
> 
> - If some fields are required, indicate all required ones
> - If most fields are required, indicate optional fields by displaying the word “optional” in parentheses next to the label text
> - If required text is colored, that color should also be used for the asterisk
> 
> 

Based on that, I think we've got it right

web/src/app/users/UserCreateDialog.tsx
dctalbot: Yeah, not sure why this is our existing behavior
@mastercactapus shall we loosen the requirements on this?

mastercactapus: `username` (auth) isn't the same as the user's name/display name. In this case, it's referring to basic auth (i.e. user/pass) specifically and not the display name when that error comes back.

It's set that way to remove ambiguity and things like 2 usernames that appear the same due to whitespace differences, etc.. Name or Display Name should probably be its own field. The command line just set's the display name to be the login username as a convenience, but they are stored and validated separately. `username`, which is typed in the login form, is just stricter

dctalbot: Thanks, that makes total sense. I'll look into adding an optional Display Name field

graphql2/graphqlapp/user.go
```diff
@@ -60,6 +60,49 @@ func (a *User) OnCallSteps(ctx context.Context, obj *user.User) ([]escalation.St
...
+
+	if input.Email != nil {
+		usr.Email = *input.Email
+	}
+
+	if input.Role != nil {
+		usr.Role = permission.Role(*input.Role)
+	}
+
+	tx, err := a.DB.BeginTx(ctx, nil)
```
mastercactapus, line 82: Use `withContextTx` in GraphQL code

app/initstores.go
```diff
@@ -135,6 +136,13 @@ func (app *App) initStores(ctx context.Context) error {
 		return errors.Wrap(err, "init service store")
 	}
 
+	if app.BasicStore == nil {
+		app.BasicStore, err = basic.NewStore(ctx, app.db)
```
mastercactapus, line 140: Update `basic.NewProvider` to take the `Store` as an arg instead of `*sql.DB` and pass it this, so we don't initialize two.

app/app.go
```diff
@@ -84,6 +85,7 @@ type App struct {
 	AlertStore    alert.Store
 	AlertLogStore alertlog.Store
 
+	BasicStore            *basic.Store
```
mastercactapus, line 88: Maybe call the prop `AuthBasicStore` so it hints a bit more about what it is when being referenced.

mastercactapus (changes requested):

Forfold (changes requested):
Code looks good, but functionally I'm seeing field errors for username show up under the display name field instead

Forfold:
For sure:

1. Set username to something "test " (note the space at the end)
2. Fill in rest of required fields
3. Attempt submit

Screen Shot 2021-05-19 at 1 25 03 PM

app/initgraphql.go
```diff
@@ -19,6 +19,7 @@ func (app *App) initGraphQL(ctx context.Context) error {
 
 	app.graphql2 = &graphqlapp.App{
 		DB:                app.db,
+		AuthBasicStore:    *app.AuthBasicStore,
```
mastercactapus, line 22: Change `AuthBasicStore` to a pointer type since `basic.Store` is not an interface, rather than dereferencing it.

graphql2/graphqlapp/app.go
```diff
@@ -50,6 +51,7 @@ import (
 
 type App struct {
 	DB             *sql.DB
+	AuthBasicStore basic.Store
```
mastercactapus, line 54:
```suggestion
	AuthBasicStore *basic.Store
```

mastercactapus (changes requested):

Forfold (approved):
lgtm!

mastercactapus (approved):

PR #1539, opened 2021-05-13
notification: refactor interfaces

mastercactapus:

**Description:**
Refactoring the notification package with more explicit types and interfaces. Fixes issues with empty provider ID after sending.

**Additional Info:**

Provider IDs are now referred to as `externalID` when working with notification senders and a discrete type `ProviderMessageID` otherwise.
```go
// ProviderMessageID is a provider-specific identifier for a message.
type ProviderMessageID struct {
	ID       string
	Provider string
}
```

`MessageStatus` has been replaced with `Status` which omits any IDs. A new type `SendResult` is used instead.
```go
// SendResult represents the result of a sent message.
type SendResult struct {
	// ID is the GoAlert message ID.
	ID string

	// ProviderMessageID is an identifier that represents the provider-specific ID
	// of the message (e.g. Twilio SID).
	ProviderMessageID ProviderMessageID

	Status
}
```

The engine now implements the `ResultReceiver` interface instead of `Receiver`
```go
// A ResultReceiver processes notification responses.
type ResultReceiver interface {
	SetSendResult(ctx context.Context, res *SendResult) error
	// ...
```

The `UpdateStatus` method of the `Receiver` interface has been replaced with `SetMessageStatus`:
```go
type Receiver interface {
	// SetMessageStatus can be used to update the state of a message.
	SetMessageStatus(ctx context.Context, externalID string, status *Status) error
	// ...
```

The `Send` method now expects an optional `externalID`
```go
type Sender interface {
	// ...
	Send(context.Context, Message) (externalID string, status *Status, err error)
}
```

`StatusChecker` now only expects an `externalID`:
```go
type StatusChecker interface {
	Status(ctx context.Context, externalID string) (*Status, error)
}
```


notification/providermessageid.go
```diff
@@ -0,0 +1,43 @@
...
+	if p.Provider == "" || p.ID == "" {
+		return nil, nil
+	}
+
+	return p.Provider + ":" + p.ID, nil
+}
+func (p *ProviderMessageID) Scan(value interface{}) error {
+	switch v := value.(type) {
+	case string:
+		if !strings.Contains(v, ":") {
```
dctalbot, line 29:
```suggestion
		if !strings.Contains(value, ":") {
```
I think these are meant to reference `value`. Maybe rename `v` to `t`?

notification/receiver.go
```diff
@@ -0,0 +1,18 @@
...
+
+// A Receiver processes incoming messages and responses.
+type Receiver interface {
+	// SetMessageStatus can be used to update the state of a message.
+	SetMessageStatus(ctx context.Context, externalID string, status *Status) error
+
+	// Receive records a response to a previously sent message.
+	Receive(ctx context.Context, callbackID string, result Result) error
+
+	// Stop indicates a user has opted-in for notifications to this contact method.
```
dctalbot, line 13:
```suggestion
	// Start indicates a user has opted-in for notifications to this contact method.
```

notification/sender.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"github.com/pkg/errors"
+)
+
+// A Sender can send notifications.
+type Sender interface {
+	// Send should return nil error if the notification was sent successfully. It should be expected
+	// that a returned error means that the notification should be attempted again.
+	//
+	// If the sent message can have it's status tracked, a unique externalID should be returned.
```
dctalbot, line 14:
```suggestion
	// If the sent message can have its status tracked, a unique externalID should be returned.
```

notification/sender.go
```diff
@@ -0,0 +1,30 @@
...
+// A StatusChecker is an optional interface a Sender can implement that allows checking the status
+// of a previously sent message by it's externalID.
+type StatusChecker interface {
+	Status(ctx context.Context, externalID string) (*Status, error)
+}
+
+// ErrStatusUnsupported should be returned when a Status() check is not supported by the provider.
+var ErrStatusUnsupported = errors.New("status check unsupported by provider")
+
+// ReceiverSetter is an optinoal interface a Sender can implement for use with two-way interactions.
```
dctalbot, line 27:
```suggestion
// ReceiverSetter is an optional interface a Sender can implement for use with two-way interactions.
```

notification/status.go
```diff
@@ -0,0 +1,61 @@
...
+type Status struct {
+
+	// State is the current state.
+	State State
+
+	// Details can contain any additional information about the State (e.g. "ringing", "no-answer" etc..).
+	Details string
+
+	// Sequence can be used when the provider sends updates out-of order (e.g. Twilio).
+	// The Sequence number defaults to 0, and a status update is ignored unless it's
```
dctalbot, line 13:
```suggestion
	// The Sequence number defaults to 0, and a status update is ignored unless its
```

dctalbot (changes requested):

notification/providermessageid.go
mastercactapus: No, this is a Type Switch; `v` is declared on line 27 as the type-casted version of `value`. `value` is of type `interface{}` which would not work here, `v` is of type `string` within the case statement.

https://tour.golang.org/methods/16

dctalbot (approved):

KatieMSB (approved):

PR #1566, opened 2021-05-19
notification: (slack) reply in thread for alert escalations

arurao:

**Description:**
This PR sends a slack notification in a thread for any alert escalations, instead of always sending a new slack message for escalations.

**Screenshots:**
Screen Shot 2021-05-19 at 12 04 02 PM

Screen Shot 2021-05-19 at 12 04 11 PM


mastercactapus:
Since the original message already contains the alert info, and could be updated in the future, maybe we can change `text` to just be something like "Escalated."?

notification/slack/channel.go
```diff
@@ -310,7 +310,14 @@ func (s *ChannelSender) Send(ctx context.Context, msg notification.Message) (*no
...
 	case notification.Alert:
-		vals.Set("text", fmt.Sprintf("Alert: %s\n\n<%s>", t.Summary, cfg.CallbackURL("/alerts/"+strconv.Itoa(t.AlertID))))
+		if t.OriginalStatus != nil {
+			// escalated alert
+			vals.Set("thread_ts", t.OriginalStatus.ProviderMessageID)
+			// updating text to simple text since the original alert will have all alert details, and this will be a thread reply message.
+			vals.Set("text", "Escalated.")
+		} else {
+			vals.Set("text", fmt.Sprintf("Alert: %s\n\n<%s>", t.Summary, cfg.CallbackURL("/alerts/"+strconv.Itoa(t.AlertID))))
+		}
```
mastercactapus, line 320: We should change this to a short-circuit rather than if/else, then leave the "default" behavior in the base scope.
```suggestion
		if t.OriginalStatus != nil {
			// Reply in thread if we already sent a message for this alert.
			vals.Set("thread_ts", t.OriginalStatus.ProviderMessageID)
			vals.Set("text", "Escalated.")
			break
		}

		vals.Set("text", fmt.Sprintf("Alert: %s\n\n<%s>", t.Summary, cfg.CallbackURL("/alerts/"+strconv.Itoa(t.AlertID))))
```

mastercactapus (changes requested):

dctalbot (approved):

dctalbot (approved):

mastercactapus (approved):

mastercactapus (approved):

Issue #1085, opened 2020-12-01
Add Favorites button to Escalation Policies


donna-roisen:
**Is your feature request related to a problem? Please describe:**
When searching for an escalation policy, I can't see the policies that I use regularly

**Describe the solution you'd like:**
I'd like to be able to favorite the Escalation policies that I use regularly, similar to schedules and rotations. It would also be good if I create an escalation policy that it automatically is favorited for me, similar to schedules and rotations.

**Describe alternatives you've considered:**
use the same  type of pattern that is already in use for schedules and rotations

**Additional context:**

PR #1567, opened 2021-05-19
escalation policy: add user favorite support

pnengchu:

**Description:**
Users may now set favorites on Escalation Policies. This feature is replicated from Schedules and Services.

Fixes https://github.com/target/goalert/issues/1085

**Screenshots:**
image

image

**Additional Info:**
High-Level overview of changes:
* A migration was added to include a new column, 'tgt_escalation_policy_id'  for table 'user_favorites'
* The structs and schema for both Escalation Policy and Escalation Policy SearchOptions were updated with isFavorite/associated booleans for setting Favorites
* Relevant SQL queries have been updated to include escalation policies properties 
* Front-end services were addressed to include UI changes when the onClick() is initiated by the end-user. 
-minor refactor of PolicyRouter.js


escalation/policy.go
```diff
@@ -23,3 +24,8 @@ func (p Policy) Normalize() (*Policy, error) {
 
 	return &p, nil
 }
+
+// IsUserFavorite returns a boolean value based on if the escalation policy is a user favorite
```
mastercactapus, line 28: nit: I know this text was lifted from the other ones, but we can probably make it a little more succinct.

```suggestion
// IsUserFavorite returns true if this policy is a favorite of the current user.
```

escalation/search.go
```diff
@@ -26,29 +35,51 @@ type SearchOptions struct {
 
 // SearchCursor is used to indicate a position in a paginated list.
 type SearchCursor struct {
-	Name string `json:"n,omitempty"`
+	Name       string `json:"n,omitempty"`
+	IsFavorite bool   `json:"f"`
```
mastercactapus, line 39:
```suggestion
	IsFavorite bool   `json:"f,omitempty"`
```

graphql/escalation.go
```diff
@@ -148,6 +148,17 @@ func (h *Handler) escalationPolicyFields() g.Fields {
...
+			Type:        g.Boolean,
+			Description: "Indicates this esclation policy has been marked as a favorite by the user.",
+			Resolve: func(p g.ResolveParams) (interface{}, error) {
+				ep, err := getPolicy(p.Source)
+				if err != nil {
+					return nil, err
+				}
+				return ep.IsUserFavorite(), nil
+			},
+		},
```
mastercactapus, line 161: We can leave this out, the old `graphql` package is deprecated and no longer updated, it only still exists for compatibility.
```suggestion
```

mastercactapus (changes requested):

escalation/policy.go
pnengchu: addressed

escalation/search.go
pnengchu: addressed, changes made

graphql/escalation.go
pnengchu: addressed, Thanks!

web/src/app/selection/EscalationPolicySelect.js
```diff
@@ -7,6 +7,7 @@ const query = gql`
       nodes {
         id
         name
+        isFavforite
```
dctalbot, line 10:
```suggestion
        isFavorite
```

pnengchu: addressed. Good catch, thanks!

escalation/search.go
```diff
@@ -18,6 +18,15 @@ type SearchOptions struct {
 	Search string       `json:"s,omitempty"`
 	After  SearchCursor `json:"a,omitempty"`
 
+	// FavoritesUserID specifies the UserID whose favorite escalation policies want to be displayed.
+	FavoritesUserID string `json:"u,omitempty"`
+
+	// FavoritesOnly controls filtering the results to those marked as favorites by FavoritesUserID.
+	FavoritesOnly bool `json:"g,omitempty"`
+
+	// FavoritesFirst indicates that escalation policy marked as favorite (by FavoritesUserID) should be returned first (before any non-favorites).
```
dctalbot, line 27:
```suggestion
	// FavoritesFirst indicates that escalation policies marked as favorite (by FavoritesUserID) should be returned first (before any non-favorites).
```
nit

migrate/migrations/20210511113508-add-escalation-policies-favorite.sql
```diff
@@ -0,0 +1,8 @@
+-- +migrate Up
+ALTER TABLE user_favorites
+    ADD COLUMN tgt_escalation_policy_id UUID REFERENCES escalation_policies(id) ON DELETE CASCADE,
+    ADD CONSTRAINT user_favorites_user_id_tgt_escalation_policies_key UNIQUE(user_id, tgt_escalation_policy_id);
```
dctalbot, line 4:
```suggestion
    ADD CONSTRAINT user_favorites_user_id_tgt_escalation_policy_id_key UNIQUE(user_id, tgt_escalation_policy_id);
```
I would use the column name here

migrate/migrations/20210511113508-add-escalation-policies-favorite.sql
```diff
@@ -0,0 +1,8 @@
...
+-- +migrate Down
+ALTER TABLE user_favorites
+    DROP CONSTRAINT user_favorites_user_id_tgt_escalation_policies_id_key,
```
dctalbot, line 7:
```suggestion
    DROP CONSTRAINT user_favorites_user_id_tgt_escalation_policy_id_key,
```

mastercactapus (changes requested):
Looks good, I found one bit that looks missed: when creating a new EP we should set it as a favorite of the current user. See `graphql2/graphqlapp/service.go:149` as an example.

Short would be to add a `favorite` field to `CreateEscalationPolicyInput` and update the UI to pass it as `true`

dctalbot (commented):
We should write an integration test for this, but I'm OK doing that as a follow up PR. Looks great!

dctalbot (dismissed):

dctalbot (changes requested):
Not seeing the escalation policy favorited on creation - double check the input being passed into `createPolicy`

dctalbot (dismissed):

dctalbot (dismissed):

mastercactapus (changes requested):
We should add the integration test here, otherwise, there's a good chance the follow-up PR doesn't happen until something breaks.

There's already a `favorites.ts` file with all the other supported ones and a helper function to make it really easy. Let me know if you need help with this, you will just need to add a call to `check` similar to Schedules, but go to the `/services` page instead (service create has a SearchSelect for EP).

You will also need to make a small change in `support/ep.ts` to pass along the `favorite` parameter, and add the same to `ep.d.ts` in the `types` directory. (all under `cypress/`)

dctalbot (approved):

web/src/app/util/SetFavoriteButton.tsx
```diff
@@ -54,8 +54,8 @@ export function SetFavoriteButton({
       <IconButton
         aria-label={
           isFavorite
-            ? `Unset as a Favorite ${typeName}`
-            : `Set as a Favorite ${typeName}`
+            ? `Unset as a Favorite ${typeName.toLowerCase()}`
+            : `Set as a Favorite ${typeName.toLowerCase()}`
```
mastercactapus, line 58: Maybe we can use `_.startCase` so it says `Set as Favorite Escalation Policy` instead of `Set as Favorite escalationpolicy`

```suggestion
            ? `Unset as a Favorite ${_.startCase(typeName)}`
            : `Set as a Favorite ${_.startCase(typeName)}`
```

web/src/cypress/types/ep.d.ts
```diff
@@ -19,6 +19,7 @@ interface EPOptions {
   description?: string
   repeat?: number
   stepCount?: number
+  favorite?: boolean
```
mastercactapus, line 22: Add `isFavorite` as a required field to the `EP` interface above

web/src/cypress/support/ep.ts
```diff
@@ -35,6 +36,7 @@ function createEP(ep?: EPOptions): Cypress.Chainable<EP> {
         name: ep.name || 'SM EP ' + c.word({ length: 8 }),
         description: ep.description || c.sentence(),
         repeat: ep.repeat || c.integer({ min: 1, max: 5 }),
+        favorite: ep.favorite,
```
mastercactapus, line 39: Fall back to random:
```suggestion
        favorite: ep.favorite ?? c.bool(),
```

mastercactapus (changes requested):

web/src/cypress/support/ep.ts
pnengchu: addressed

web/src/cypress/types/ep.d.ts
pnengchu: addressed

web/src/app/util/SetFavoriteButton.tsx
pnengchu: addressed, applied .toLowerCase() to account for Service, Rotation, Schedule checks

dctalbot (dismissed):

dctalbot (approved):

mastercactapus (approved):

Issue #1458, opened 2021-04-23
ui: create a primary @material-ui/core theme


Forfold:
There are a lot of options we can customize at the top level, without needing to specify hex colors header variants, etc throughout the application. For instance, we can set our own breakpoint parameters since we largely only care about mobile vs. desktop sizes in changing page layout.

Creating a theme is simple enough (we have the base of one already setup [here](https://github.com/target/goalert/blob/master/web/src/app/mui.js)), though applying it and covering all our bases leftover from custom styles is what will likely take the most time.

Theme options for customization through `@material-ui/core`, with the ability to override any default component props (i.e. `Dialog` or `CardActions`) as well:
```ts
interface ThemeOptions {
  shape?: ShapeOptions
  breakpoints?: BreakpointsOptions
  direction?: Direction
  mixins?: MixinsOptions
  overrides?: Overrides
  palette?: PaletteOptions
  props?: ComponentsProps
  shadows?: Shadows
  spacing?: SpacingOptions
  transitions?: TransitionsOptions
  typography?: TypographyOptions | ((palette: Palette) => TypographyOptions)
  zIndex?: ZIndexOptions
  unstable_strictMode?: boolean
}
```

With the following palette options:
```ts
export interface PaletteOptions {
  primary?: PaletteColorOptions
  secondary?: PaletteColorOptions
  error?: PaletteColorOptions
  warning?: PaletteColorOptions
  info?: PaletteColorOptions
  success?: PaletteColorOptions
  type?: PaletteType
  tonalOffset?: PaletteTonalOffset
  contrastThreshold?: number
  common?: Partial<CommonColors>
  grey?: ColorPartial
  text?: Partial<TypeText>
  divider?: string
  action?: Partial<TypeAction>
  background?: Partial<TypeBackground>
  getContrastText?: (background: string) => string
}

export type PaletteColorOptions = SimplePaletteColorOptions | ColorPartial

export interface SimplePaletteColorOptions {
  light?: string
  main: string
  dark?: string
  contrastText?: string
}
```

We should think about:
- What default size do we want for our h1 title and breadcrumbs in the toolbar?
- What default size do we want for our h2 elements (e.g. details page card title, quick links title, subroute page titles, etc)
- What width should we set the mobile breakpoint to?
- What palette color options to use (and their values)?
- What default card elevation to use?

Out of scope:
- Configurable theme
- Multiple themes

Forfold:
[...]

PR #1571, opened 2021-05-20
ui: move breadcrumb page actions into content view

### LINKED ISSUE #1517: ui: move remaining action buttons out of toolbar
Follow-up from #1366. Details pages are complete, but breadcrumb pages are not. Actions should be moved into the containers (cards) they are modifying

**Remaining actions left to move:**
- Alert Details
  - Acknowledge
  - Close
  - Escalate
- Schedules
  - Assignment Filter
  - Overrides Filter
  - Shifts Filter
- Services
  - Alerts Ack All
  - Alerts Close All
- Users
  - Log Out Other Sessions

**Additional context:**
- Aim to remove PageActions and similar components if no longer used
- Follow-up work when this is complete could be to decide on and then move said relevant elements into the appbar (Profile icon, logout, maybe admin settings, etc)

Forfold:

**Description:**
Moves the remaining page actions out of the toolbar and into the content view. This opens up work in enhancing the toolbar for things like the profile avatar, logout button, dark/light theme button, and searching on mobile.

With the changing of some grid layouts, there was a good amount of whitespace changes so I recommend hiding that while reviewing. 🙂 

Closes #1517

**Additional Info:**
- Page Actions can't be deleted yet as we use it for the mobile searching experience.
- Primary/secondary color on some of the buttons looks odd, but will be addressed in #1458
- Converts `ScheduleOverrideList.js` and `ScheduleRouter.js` to hooks, as part of #923

**Out of Scope:**
Override List causes the filter popover to close after choosing a selection (rather than the usual process of waiting for the user to select "Done"). I spent a significant amount of time trying to compensate for this, but couldn't come up with a solid solution without rewriting QueryList and other similar components. This can be addressed in a future issue regarding layout within the content view.

**Screenshots:**
Screen Shot 2021-05-20 at 11 59 31 AM

Screen Shot 2021-05-20 at 12 01 27 PM

Screen Shot 2021-05-20 at 12 01 36 PM

Screen Shot 2021-05-20 at 12 01 46 PM

Screen Shot 2021-05-20 at 12 02 57 PM

Screen Shot 2021-05-20 at 12 03 07 PM


Forfold:
@dctalbot While they share the same general functionality, with checkboxes being paginated you'll never be able to confidently ack or close all alerts for a service- those two buttons use a different mutation: `UpdateAlertsByServiceMutation` given the new status and `ServiceID`

pnengchu (dismissed):

web/src/app/alerts/AlertsList.js
```diff
@@ -279,7 +279,9 @@ export default function AlertsList(props) {
           selectable: a.status !== 'StatusClosed',
         })}
         variables={variables}
-        filter={<AlertsListFilter serviceID={props.serviceID} />}
+        filter={
```
dctalbot, line 282: I think we should name this something like "secondaryAction" since we allow passing through arbitrary nodes? e.g. the ack/close all buttons

web/src/app/users/UserSessionList.tsx
```diff
@@ -161,6 +140,17 @@ export default function UserSessionList(
...
+          <CardActions>
+            <Button
+              color='primary'
+              variant='contained'
+              onClick={() => setEndSession('all')}
+            >
+              Log Out Other Sessions
+            </Button>
+          </CardActions>
+        )}
```
dctalbot, line 153: I think we should move this above the list into a button group to match the admin group save/reset buttons

web/src/app/services/ServiceAlerts.js
```diff
@@ -48,24 +57,30 @@ export default function ServiceAlerts(props) {
...
+    <Grid className={classes.filter} container spacing={2} alignItems='center'>
+      <Grid item>
+        <Button variant='outlined' color='primary' onClick={handleClickAckAll}>
+          Acknowledge All
+        </Button>
+      </Grid>
+      <Grid item>
+        <Button
+          variant='outlined'
+          color='secondary'
```
dctalbot, line 70:
```suggestion
          color='primary'
```
The buttons styles should match

dctalbot (changes requested):
Code looks great besides the header action issue you mentioned. We may want to separate out the header from the list components entirely, similar to what we did with the calendar toolbar. That would hopefully reduce the repetition of having to add headernote/headeraction props to each list. Otherwise, just a couple comments on design ideas

web/src/app/services/ServiceAlerts.js
Forfold: There should only be 1 primary action per page which is why this is set to secondary, but since there's already a primary button for creating an alert these should actually both be secondary actions

pnengchu (dismissed):

pnengchu (approved):

Forfold:
Thanks for the reviews helping kick off CI @pnengchu! Sorry for all the spin. @dctalbot, the Cypress integration tests should be good to go now 👍

dctalbot (approved):

pnengchu (approved):

PR #1575, opened 2021-05-21
ui/lists: fix FAB margins

### LINKED ISSUE #1550: ui: FAB margins are inconsistent
**Describe the Bug:**
There is a regression with the bottom corner FABs having different margins on different pages

**Steps to Reproduce:**
1. Go to /alerts
2. Go to /rotations
3. Observe different margins

**Expected Behavior:**
FABs use a consistent margin, ideally specified by [material design](https://material.io/components/buttons-floating-action-button)

**Observed Behavior:**
See below

**Screenshots/Stack Traces:**
Screen Shot 2021-05-18 at 7 35 43 AM
Screen Shot 2021-05-18 at 7 38 09 AM

**Application Version:**
`731a07bd869ea4bbb0c8bf6ad6ff7edf169ec0cb`

dctalbot:

**Description:**
This PR make all create FABs use the same margin as defined by material design (16dp). 

- Removes `CreateAlertFab` in favor of `CreateFAB`
- Adjust margin to 16px
- Adjust Speeddial margin to match

Fixes #1550


web/src/app/alerts/AlertsList.js
```diff
@@ -92,9 +93,10 @@ function getStatusFilter(s) {
 export default function AlertsList(props) {
   const classes = useStyles()
   const width = useWidth()
-  const isMobileScreenSize = isWidthDown('md', width)
+  const isMobileScreenSize = isWidthDown('xs', width)
```
Forfold, line 96: We're using `'sm'` in `App.js` to determine whether we're in a mobile size or not, and `'md'` in most other places. This should probably be `'sm'` to best reflect `App.js`

Forfold (changes requested):

web/src/app/alerts/AlertsList.js
Forfold: Is this a current bug in production?

Forfold: If this is only being used for the transition on the create fab, I think it'd make more sense to just rename it to something else, since it's only being used when the snackbar has the potential to overlap with its intended placement, rather than if it's mobile or not

dctalbot: Yeah, we could either extend with of the snack bar on `sm`, or use what I have at the moment

Forfold: I'm fine with what we have for the moment, the name just threw me off which is why I commented that we use sm/md normally. Probably how the bug got introduced in the first place if I had to guess. Maybe?:
```suggestion
  // used to transition fab above snackbar when snackbar width overlaps fab placement
  const isXS = isWidthDown('xs', width)
```

dctalbot: Yeah, we can clear up the naming. Might be an opportunity for us to write our own `useViewport`

```js
const {isMobile, isPhablet, isTablet, isDesktop, isWidescreen} = useViewport()
```

Would help keep things consistent

Forfold: That would be really convenient. I've been wanting to make an issue around our general layout and making our own, usable breakpoints, but don't have a concrete design yet. Def need to remember this for that!

Forfold (approved):

Forfold (approved):

KatieMSB (approved):

PR #1576, opened 2021-05-21
schedule: render calendar while loading

mastercactapus:

**Description:**
This PR removes the flicker from the calendar while loading new data. The calendar and toolbar will always be rendered, a spinner may be shown if loading takes longer than 200ms, and shows for a minimum of 1500ms (as configured elsewhere).

The toolbar nav buttons are always available allowing users to quickly navigate to future/past dates.

## Before

https://user-images.githubusercontent.com/595010/119177333-ce2ab000-ba31-11eb-8aa2-a931a8136a2c.mp4

## After

https://user-images.githubusercontent.com/595010/119177352-d2ef6400-ba31-11eb-9c72-d14fd5d6fae4.mp4


mastercactapus:
> I don't think this would resize correctly while spinning. If that's a concern, something like this could help:

@dctalbot  I wasn't able to get that to work locally (admittedly, not great with CSS), but I added a resize listener to take care of that case.

arurao (dismissed):
LGTM

pnengchu (approved):

arurao (approved):

pnengchu (approved):

PR #1594, opened 2021-06-01
overrides: lock table before duplicate check

mastercactapus:

**Description:**
It's currently possible for duplicate/overlapping overrides to be entered if they are done simultaneously. This PR resolves the issue by adding a lock to the validation trigger.

A new smoketest was added for this case, removing line 8 of the new migration should cause the test to fail.


arurao (approved):
LGTM :tada:

PR #1605, opened 2021-06-07
notificationchannel: ensure only a single copy of a type/value is created

mastercactapus:

**Description:**
Reworks how notification channels are created from an always-create method to re-using existing if possible.

### DIFF
--- a/notificationchannel/store.go
+++ b/notificationchannel/store.go
@@ -3,6 +3,10 @@ package notificationchannel
 import (
 	"context"
 	"database/sql"
+	"errors"
+	"fmt"
+
+	uuid "github.com/satori/go.uuid"
 	"github.com/target/goalert/permission"
 	"github.com/target/goalert/util"
 	"github.com/target/goalert/util/sqlutil"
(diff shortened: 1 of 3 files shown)


util/sqlutil/nulluuid.go
```diff
@@ -0,0 +1,45 @@
...
+	Valid bool
+}
+
+func (u NullUUID) Value() (driver.Value, error) {
+	if !u.Valid {
+		return nil, nil
+	}
+
+	return u.UUID.Bytes(), nil
+}
```
dctalbot, line 21: `error` is always `nil`, can we do something different here?

notificationchannel/store.go
```diff
@@ -41,27 +50,97 @@ func NewDB(ctx context.Context, db *sql.DB) (*DB, error) {
...
+		if err != nil {
+			return uuid.UUID{}, fmt.Errorf("start tx: %w", err)
+		}
+		defer tx.Rollback()
 	}
 
-	return n, nil
+	_, err = tx.StmtContext(ctx, db.lock).ExecContext(ctx)
+	if err != nil {
+		return uuid.UUID{}, fmt.Errorf("aqcuire lock: %w", err)
```
dctalbot, line 109:
```suggestion
		return uuid.UUID{}, fmt.Errorf("acquire lock: %w", err)
```

dctalbot (changes requested):

util/sqlutil/nulluuid.go
mastercactapus: We don't have an error condition, but have to match the `driver.Valuer` signature:

https://golang.org/pkg/database/sql/driver/#Valuer

dctalbot (approved):

arurao (approved):

Issue #1365, opened 2021-03-29
ui: create alert dialog stutters on close


Forfold:
**Describe the Bug:**
When closing the create alert dialog, the height of the dialog changes for a moment before closing the dialog fully, resulting in a broken looking transition.

**Potential Solutions:**
We could remove the min-height on the dialog allowing it to grow to the height needed per step, and follow the transition pattern used for the create dialogs on each list page

**Screenshots/Stack Traces:**
![create alert stutter](https://user-images.githubusercontent.com/11381794/112884708-aef05180-9084-11eb-90b9-25ac1b6bf66e.gif)

dctalbot:
This can be fixed with regular css e.g. in `elements.css` 
```css 
.create-alert-dialog {
  height: 65vh;
}
```
and in `CreateAlertDialog.js`
```js
PaperProps={{ className: 'create-alert-dialog' }}
```

Would be nice if we could do it all in JS, but not sure why Material unmounts the css before the transition occurs

PR #1617, opened 2021-06-10
ui/forms: fix dialog transitions

dctalbot:

**Description:**
This PR fixes an issue in which the component would attempt to unmount before/during the exit transition

Fixes: #1365


web/src/app/dialogs/FormDialog.js
```diff
@@ -104,16 +112,7 @@ function FormDialog(props) {
...
-      alert,
-      confirm,
-      errors,
-      loading,
-      primaryActionLabel,
-      onClose,
-      onBack,
-      onNext,
-    } = props
+    const { onBack, onNext } = props
```
mastercactapus, line 115: Is this needed?  `onBack` and `onNext` already in outer scope

web/src/app/dialogs/FormDialog.js
```diff
@@ -48,18 +47,23 @@ function FormDialog(props) {
...
     disableGutters,
     errors,
     fullScreen,
-    isUnmounting,
     loading,
     primaryActionLabel, // remove from dialogProps spread
     maxWidth,
     notices,
-    onClose,
+    onClose: onExited,
```
mastercactapus, line 66: Since this is only used in one place, let's leave it as the prop name, and remove the wrapper around `setOpen` (less indirection).

```suggestion
    onClose,
```

Further down we can do the switch

```js
<Dialog
 // ...

// only calling onClose after the exit event so that animations finish properly
onClose={()=>setOpen(false)}
onExited={onClose}
/>
```

Otherwise looking at the code below `onClose` may seem like it's the `onClose` prop rather than a thin wrapper and `onExited` being the actual `onClose` prop. If we handle things directly using original names there is no misinterpretation.

mastercactapus (changes requested):

web/src/app/dialogs/FormDialog.js
dctalbot: No, somehow missed the first time

web/src/app/dialogs/FormDialog.js
dctalbot: `onClose` is used in four other places, thus the wrapper

dctalbot: The prop really should have been named `onExited` originally. I feel it's more confusing to pass a function named `onClose` into anything other than Material's `onClose`

mastercactapus: In that case, let's call them `handleOnExited` and `handleOnClose` to denote that they are internal handlers and not the `on<event>` props passed in

mastercactapus (approved):

mastercactapus (approved):

pnengchu (approved):

PR #1628, opened 2021-06-16
ui/hooks: FormContainer

voromahery:

**Description:**
Converts Form Container into hooks 

Part of #923

### DIFF
--- a/web/src/app/forms/FormContainer.js
+++ b/web/src/app/forms/FormContainer.js
@@ -19,7 +19,7 @@ export function FormContainer(props) {
       let err
       // find first error
       _fields[field].find((validate) => {
-        err = validate(get(props.value, field))
+        err = validate(get(value, field))
         return err
       })
       if (err) err.field = field
@@ -131,4 +131,13 @@ FormContainer.propTypes = {
   // Enables functionality to remove an incoming value at it's index from
   // an array field if the new value is falsey.
   removeFalseyIdxs: p.bool,
+}
+
+FormContainer.defaultProps = {
+  errors: [],
+  value: {},
+  onChange: () => {},
+
+  mapValue: (value) => value,
+  mapOnChangeValue: (value) => value,
 }
--- a/web/src/app/forms/FormContainer.js
+++ b/web/src/app/forms/FormContainer.js
@@ -1,7 +1,5 @@
-import React, { useEffect, useState } from 'react'
-import PropTypes from 'prop-types'
-import MountWatcher from '../util/MountWatcher'
-
+import React, { useEffect, useState, useRef, useContext } from 'react'
+import p from 'prop-types'
 import { FormContext, FormContainerContext } from './context'
 import { get, set, cloneDeep } from 'lodash'
 
@@ -95,46 +93,42 @@ export function FormContainer(props) {
   }
 
   return (
-    <FormContext.Consumer>
-      <FormContainerContext.Provider
-        value={{
-          value: mapValue(value),
-          disabled: formDisabled || containerDisabled,
-          errors: persistentValidationErrors(validationErrors).concat(
-            props.errors,
-          ),
-          onChange: onChange,
-          addField: addField,
-          optionalLabels: optionalLabels,
-        }}
-      >
-        {props.children}
-      </FormContainerContext.Provider>
-    </FormContext.Consumer>
+    <FormContainerContext.Provider
+      value={{
+        value: mapValue(value),
+        disabled: formDisabled || containerDisabled,
+        errors: persistentValidationErrors.concat(props.errors),
+        onChange: onChange,
+        addField: addField,
+        optionalLabels: optionalLabels,
+      }}
+    >
+      {props.children}
+    </FormContainerContext.Provider>
   )
 }
 
 FormContainer.propTypes = {
-  value: PropTypes.object,
-  errors: PropTypes.arrayOf(
-    PropTypes.shape({
-      message: PropTypes.string.isRequired,
-      field: PropTypes.string.isRequired,
-      helpLink: PropTypes.string,
+  value: p.object,
+  errors: p.arrayOf(
+    p.shape({
+      message: p.string.isRequired,
+      field: p.string.isRequired,
+      helpLink: p.string,
     }),
   ),
 
-  onChange: PropTypes.func,
-  disabled: PropTypes.bool,
+  onChange: p.func,
+  disabled: p.bool,
 
-  mapValue: PropTypes.func,
-  mapOnChangeValue: PropTypes.func,
+  mapValue: p.func,
+  mapOnChangeValue: p.func,
 
   // If true, will render optional fields with `(optional)` appended to the label.
   // In addition, required fields will not be appended with `*`.
-  optionalLabels: PropTypes.bool,
+  optionalLabels: p.bool,
 
   // Enables functionality to remove an incoming value at it's index from
   // an array field if the new value is falsey.
-  removeFalseyIdxs: PropTypes.bool,
+  removeFalseyIdxs: p.bool,
 }
--- a/web/src/app/forms/FormContainer.js
+++ b/web/src/app/forms/FormContainer.js
@@ -1,5 +1,5 @@
-import React from 'react'
-import p from 'prop-types'
+import React, { useEffect, useState } from 'react'
+import PropTypes from 'prop-types'
 import MountWatcher from '../util/MountWatcher'
 
 import { FormContext, FormContainerContext } from './context'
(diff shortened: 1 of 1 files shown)


web/src/app/forms/FormContainer.js
```diff
@@ -1,97 +1,58 @@
...
+  const setState = useRef()
+  // const [state, setState] = useState({})
   _fields = {}
-
-  addField = (fieldName, validate) => {
-    if (!this._fields[fieldName]) {
-      this._fields[fieldName] = []
+  function addField(fieldName, validate) {
+    if (_fields[fieldName]) {
+      _fields[fieldName] = []
```
mastercactapus, line 17: `useRef` works a little differently, we can initialize it to an empty object and then reference it via the `.current` property. We can simplify it since we never update the base object.

```js
  const _fields = useRef({}).current

  function addField(fieldName, validate) {
    if (_fields[fieldName) {
      _fields[fieldName] = []
```

web/src/app/forms/FormContainer.js
```diff
@@ -117,50 +78,68 @@ export class FormContainer extends React.PureComponent {
...
-    return this.props.onChange(
+    return props.onChange(
       mapOnChangeValue(set(mapValue(oldValue), fieldName, value)),
     )
   }
 
-  render() {
-    return <FormContext.Consumer>{this.renderComponent}</FormContext.Consumer>
-  }
+  return <FormContext.Consumer>{renderComponent}</FormContext.Consumer>
```
mastercactapus, line 91: Since we're in hooks now (good catch on your last comment) we can simplify a bit of this.

1. Instead of `renderComponent` being a function, just return it's value from the parent function
2. Add `const {disabled: formDisabled, addSubmitCheck} = useContext(FormContext)` to the top of the component function
3. Move the `useEffect` for `addSubmitCheck` to the top, right under the `useContext` call

mastercactapus (changes requested):
Few more things to update. Also, you can leave the PR open -- it will auto-update when you push changes to your branch.

web/src/app/forms/FormContainer.js
```diff
@@ -1,97 +1,64 @@
...
+export function FormContainer(props) {
+  const setState = useRef().current
+  const _fields = useRef().current
+  const { value, mapValue, optionalLabels, disabled: containerDisabled } = props
+  const { disabled: formDisabled, addSubmitCheck } = useContext(FormContext)
 
-  _fields = {}
+  useEffect(() => {
+    addSubmitCheck(onSubmit)
+  }, [])
```
mastercactapus, line 19: Make sure we're returnign the value so it gets cleaned up on unmount
```suggestion
  useEffect(() => addSubmitCheck(onSubmit), [])
```

web/src/app/forms/FormContainer.js
```diff
@@ -1,97 +1,64 @@
...
-
-    mapValue: (value) => value,
-    mapOnChangeValue: (value) => value,
-  }
 
-  state = {
-    validationErrors: [],
-  }
+export function FormContainer(props) {
+  const setState = useRef().current
```
mastercactapus, line 12: This is being used for persistent validation errors
```suggestion
  const [setPersistentValidationErrors, persistentValidationErrors] = useState([])
```

web/src/app/forms/FormContainer.js
```diff
@@ -1,97 +1,64 @@
...
+      _fields[fieldName] = _fields[fieldName].filter((v) => v !== validate)
+      if (_fields[fieldName].length === 0) {
+        delete _fields[fieldName]
       }
     }
   }
 
-  onSubmit = () => {
-    const validate = (field) => {
+  function onSubmit() {
```
mastercactapus, line 35: This should be moved above the `useEffect` call

web/src/app/forms/FormContainer.js
```diff
@@ -1,97 +1,64 @@
...
       })
       if (err) err.field = field
       return err
     }
-    const validationErrors = Object.keys(this._fields)
+    const validationErrors = Object.keys(_fields)
       .map(validate)
       .filter((e) => e)
-    this.setState({ validationErrors })
+    setState({ validationErrors })
```
mastercactapus, line 49: With the above change:
```suggestion
    setPersistentValidationErrors(validationErrors)
```

web/src/app/forms/FormContainer.js
```diff
@@ -117,50 +84,55 @@ export class FormContainer extends React.PureComponent {
...
-        }}
-        onUnmount={() => {
-          this._unregister()
+  return (
+    <FormContext.Consumer>
+      <FormContainerContext.Provider
+        value={{
+          value: mapValue(value),
+          disabled: formDisabled || containerDisabled,
+          errors: validationErrors.concat(props.errors),
```
mastercactapus, line 103: Use the ones from state instead
```suggestion
          errors: persistentValidationErrors(validationErrors).concat(props.errors),
```

mastercactapus (changes requested):

web/src/app/forms/FormContainer.js
voromahery: Thank you so muck for the feedback! Could you explain to me what the difference between  `  useEffect(() => {
    addSubmitCheck(onSubmit)
  }, [])` and `useEffect(() => addSubmitCheck(onSubmit), [])` in terms of functionality. Thank you!

mastercactapus: Sure, no problem:

The first one discards the return value
```js
// these are the same

useEffect(() => { addSubmitCheck(onSubmit) }, [])

useEffect(() => {
  const cleanup = addSubmitCheck(onSubmit)

  // returns nothing/undefined
  return
}, [])
```


The second passes the return value through
```js
// these are the same
useEffect(() => addSubmitCheck(onSubmit), [])

useEffect(() => {
  const cleanup = addSubmitCheck(onSubmit)
  return cleanup
}, [])
```

Another way to think of it is `() => foo()` is the same as `() => { return foo() }`

web/src/app/forms/FormContainer.js
```diff
@@ -117,50 +84,57 @@ export class FormContainer extends React.PureComponent {
...
-
-    return (
-      <MountWatcher
-        onMount={() => {
-          this._unregister = addSubmitCheck(this.onSubmit)
-        }}
-        onUnmount={() => {
-          this._unregister()
+  return (
+    <FormContext.Consumer>
```
Forfold, line 98: With the use of `useContext` above, this wrapper shouldn't be needed

web/src/app/forms/FormContainer.js
```diff
@@ -1,97 +1,64 @@
-import React from 'react'
-import p from 'prop-types'
+import React, { useEffect, useState } from 'react'
+import PropTypes from 'prop-types'
 import MountWatcher from '../util/MountWatcher'
```
Forfold, line 3: delete unused import

web/src/app/forms/FormContainer.js
```diff
@@ -1,97 +1,64 @@
...
-      if (this._fields[fieldName].length === 0) {
-        delete this._fields[fieldName]
-      }
-    }
-  }
+export function FormContainer(props) {
+  const [persistentValidationErrors, setPersistentValidationErrors] = useState(
+    [],
+  )
+  const _fields = useRef().current
```
Forfold, line 15: missing import from react

web/src/app/forms/FormContainer.js
```diff
@@ -1,97 +1,64 @@
...
+  const { value, mapValue, optionalLabels, disabled: containerDisabled } = props
+  const { disabled: formDisabled, addSubmitCheck } = useContext(FormContext)
```
Forfold, line 17: missing import from react

web/src/app/forms/FormContainer.js
```diff
@@ -1,97 +1,64 @@
...
       .filter((e) => e)
-    this.setState({ validationErrors })
+    setPersistentValidationErrors({ validationErrors })
     if (validationErrors.length) return false
 
     return true
   }
 
-  onChange = (fieldName, e) => {
+  useEffect(() => addSubmitCheck(onSubmit), [])
```
Forfold, line 39: To replicate the `this._register` in the class component version from MountWatcher, we can return from inside useEffect to replicate unmounting. Will need to test it though:

```
useEffect(() => {
    const unregister = addSubmitCheck(onSubmit), []) // runs on mount
    
    return () => unregister() // runs on unmount
}, [])
```

web/src/app/forms/FormContainer.js
```diff
@@ -117,50 +84,57 @@ export class FormContainer extends React.PureComponent {
...
-        }}
-        onUnmount={() => {
-          this._unregister()
+  return (
+    <FormContext.Consumer>
+      <FormContainerContext.Provider
+        value={{
+          value: mapValue(value),
+          disabled: formDisabled || containerDisabled,
+          errors: persistentValidationErrors(validationErrors).concat(
```
Forfold, line 103: This is giving me undefined for `validationErrors`. Should this just be `persistentValidationErrors.concat(`?

Forfold (changes requested):

web/src/app/forms/FormContainer.js
mastercactapus: @Forfold this should work as written, see: https://github.com/target/goalert/pull/1628#discussion_r665447124

Forfold: perf! I didn't realize `addSubmitCheck` was already returning everything as needed

Forfold (changes requested):
We'll need to re-add the default props- as is the form breaks when opening some dialogs because `mapValue` is not defined. You can add `defaultProps` the same way `propTypes` was converted, at the bottom of the file just below the function component

Forfold (changes requested):
Not sure what's going on here, but every form breaks the page when I try to open it from the UI.

You can test your changes by running `make start` locally and visit the UI from `localhost:3030`. More info here: https://github.com/target/goalert/blob/master/docs/development-setup.md

Forfold:
Closing due to inactivity, feel free to re-open at any point and let us know if you would like some assistance with this one!

PR #1640, opened 2021-06-21
integrationkey: add no rows check

dctalbot:

**Description:**
Fixes https://github.com/target/goalert/issues/1709

To validate backend:
1. use GraphQL explorer to query for an integration key with an ID that doesn't exist
2. Should now get null instead of error

To validate UI:
1. Open two tabs for a particular services' integration keys
2. open a delete dialog for the same key in both
3. Confirm in one
4. Observe "No longer exists" message in dialog


mastercactapus (approved):

Forfold (approved):
Functionally test and walked through code with @pnengchu, lgtu!

mastercactapus (approved):

PR #1653, opened 2021-06-23
ui/hooks: form field

KatieMSB:

**Description:**
Converts FormField.js to use hooks.

Part of #923

### DIFF
--- a/web/src/app/forms/FormField.js
+++ b/web/src/app/forms/FormField.js
@@ -147,7 +147,9 @@ export function FormField(props) {
           {...fieldProps}
           error={checkbox ? undefined : Boolean(fieldProps.error)}
           label={formLabel ? null : fieldProps.label}
-        />
+        >
+          {fieldProps.children}
+        </Component>
         {!noError && renderFormHelperText(fieldProps.error, fieldProps.hint)}
       </FormControl>
     )
@@ -160,6 +162,9 @@ export function FormField(props) {
   )
 }
 FormField.propTypes = {
+  // pass select dropdown items as children
+  children: p.node,
+
   // one of component or render must be provided
   component: p.any,
   render: p.func,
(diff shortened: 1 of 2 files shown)


web/src/app/forms/FormField.js
```diff
@@ -9,206 +8,106 @@ import shrinkWorkaround from '../util/shrinkWorkaround'
...
-    )
+    return validate(value)
   }
 
-  renderContent(props) {
-    const { checkbox, component, formLabel, label, noError, render } =
-      this.props
+  useEffect(() => {
+    return addField(fieldName, validateField)
+  }, [])
```
mastercactapus, line 108: Can we raise this (and `validateField` it uses) up higher? Putting hooks as close to the top is helpful to avoid accidentally making them conditional (they should never be behind an `if` or early `return`) as code changes over time.

mastercactapus (changes requested):

Forfold (dismissed):
Lgtm, functionally tested as well

web/src/app/forms/FormField.js
```diff
@@ -9,206 +8,106 @@ import shrinkWorkaround from '../util/shrinkWorkaround'
...
-  validate = (value) => {
+export function FormField(props) {
+  const {
+    errors,
+    value,
+    onChange,
+    addField,
+    disabled: containerDisabled,
+    optionalLabels,
+  } = useContext(FormContainerContext)
```
Forfold, line 19: using context here makes this so much easier to understand, imo :ok_hand:

Forfold (dismissed):

Forfold (approved):

web/src/app/forms/FormField.js
```diff
@@ -9,206 +8,109 @@ import shrinkWorkaround from '../util/shrinkWorkaround'
...
-    return (
-      <FormContainerContext.Consumer>
-        {this.renderComponent}
-      </FormContainerContext.Consumer>
-    )
+  useEffect(() => {
+    // on mount
+    const unregister = () => addField(fieldName, validateField)
+    // on unmount
+    return () => unregister
```
mastercactapus, line 62: `addField` should be run on mount, and it's return value ran on unmount
```suggestion
    return addField(fieldName, validateField)
```

mastercactapus (changes requested):

Forfold (approved):

Forfold (approved):

PR #1654, opened 2021-06-23
lib: migrate to a maintained UUID library

mastercactapus:

**Description:**
Switches to a maintained UUID library.

### DIFF
--- a/user/cache.go
+++ b/user/cache.go
@@ -32,7 +32,11 @@ func (c checker) UserExistsString(idStr string) bool {
 	if idStr == "" {
 		return false
 	}
-	id, _ := uuid.Parse(idStr)
+	id, err := uuid.Parse(idStr)
+	if err != nil {
+		return false
+	}
+
 	return c.UserExistsUUID(id)
 }
 func (c checker) UserExistsUUID(id uuid.UUID) bool { _, ok := c[id]; return ok }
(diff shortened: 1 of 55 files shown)


user/cache.go
```diff
@@ -28,11 +28,12 @@ type ExistanceChecker interface {
...
 type checker map[uuid.UUID]struct{}
 
-func (c checker) UserExistsString(id string) bool {
-	if id == "" {
+func (c checker) UserExistsString(idStr string) bool {
+	if idStr == "" {
 		return false
 	}
-	return c.UserExistsUUID(uuid.FromStringOrNil(id))
+	id, _ := uuid.Parse(idStr)
```
dctalbot, line 35: This is minor, but feels off to have one error condition (empty string) return false and any others be ignored. Maybe check for all in one go?
```go
id, err := uuid.Parse(idStr)
if err != nil {
    return false
}
```

dctalbot (changes requested):

user/cache.go
mastercactapus: oh, yeah I like that much better -- prevents the possibility of an invalid UUID returning true if there was a nil user

mastercactapus:
> `graphql2/gqlgen.yml` is still specifying `github.com/satori/go.uuid.UUID`, is that needed for compatibility?

@dctalbot no, I was trying to figure out why `go generate` kept demanding a hash be in the sum file, that would be why. ty

dctalbot (approved):

m17ch (approved):
lgtm

PR #1655, opened 2021-06-23
engine: status updates refactor

arurao:

**Description:**
This PR refactors the status update functionality. This is the base work to be able to start work with two-way slack notifications module #19 

**Out of Scope:**
- Status updates for alert bundles
- Multiple contact methods with status updates for a user


mastercactapus (approved):

dctalbot (approved):

smoketest/messagebundlesms_test.go
```diff
@@ -54,5 +54,6 @@ func TestMessageBundle_SMS(t *testing.T) {
 
 	h.GraphQLQuery2(`mutation{ updateAlerts(input: {alertIDs: [1,2,3,4], newStatus: StatusClosed}){id} }`)
 
-	d1.ExpectSMS("Closed", "3 other alerts")
+	// Since with https://github.com/target/goalert/pull/1655, status updates are out of scope for alert bundles.
+	// d1.ExpectSMS("Closed", "3 other alerts")
```
dctalbot, line 58: Would it be better to delete this code?

dctalbot (changes requested):

smoketest/messagebundlesms_test.go
arurao: We definitely could, since status updates for alert bundles is out of scope for this PR. But if we implement it sooner in the future, just wanted to have this comment in there so that we could go back to testing the smoketest again. But I'm fine with either.

dctalbot (dismissed):

smoketest/messagebundlesms_test.go
mastercactapus: I think it's safe to remove -- we don't know if we will reintroduce them, or how they will work when we do, so it's a safe bet the test code may need to be re-evaluated at that time anyhow.

dctalbot (approved):

mastercactapus (approved):

PR #1659, opened 2021-06-24
schedule: require weekdayfilter be specified for time-of-day notification rules

mastercactapus:

**Description:**
Requires WeekdayFilter be set for time-of-day rules


dctalbot (dismissed):

schedule/storeoncallnotifications.go
```diff
@@ -39,6 +39,9 @@ func (store *Store) SetOnCallNotificationRules(ctx context.Context, tx *sql.Tx,
 		if r.WeekdayFilter != nil && r.Time == nil {
 			return validation.NewFieldError("Rules[%d].WeekdayFilter", "Weekday filter may only be used with Time.")
 		}
+		if r.WeekdayFilter == nil && r.Time != nil {
+			return validation.NewFieldError("Rules[%d].WeekdayFilter", "Weekday filter is required with Time.")
+		}
```
Forfold, line 44:
```suggestion
			return validation.NewFieldError("Rules[%d].WeekdayFilter", "Time is required with Weekday Filter.")
		}
		if r.WeekdayFilter == nil && r.Time != nil {
			return validation.NewFieldError("Rules[%d].WeekdayFilter", "Weekday Filter is required with Time.")
		}
```

Forfold (changes requested):
nit

schedule/storeoncallnotifications.go
mastercactapus: Time determines the mode of the rule, as per the schema and UI, backend, and the engine logic -- so the invalid field name should always be the WeekdayFilter

Forfold: But to the user, it's just one or the other- I think we should try and keep it clear and consistent in what we show.

If you don't have time, you need to fill in weekday filter, and if you don't have weekday filter, you need to fill out the time

mastercactapus: A UI user will never see this as the UI handles this directly based on `time` being null or not and always sends them to the backend together. If `time == nil` then it is an `on-change` type rule, which does not allow `weekdayFilter`. If `time != nil` then it is a `time-of-day` type rule, which requires `weekdayFilter`.

`weekdayFilter` as a requirement, is only based on if time-of-day is being used since that is its type, and thus what validation it requires. The first choice of a user, API or UI, is what type of rule they want, indicated by the presence of `time` and validation should probably follow that.

The schema documents that time drives behavior as well to remove ambiguity: 
> time indicates a time-of-day to send a message ... If time is null, a notification will be sent any time the set of on-call users changes for this schedule.
It would be confusing if you tried to create an `on-change` rule and have the API tell you you need to add a `time` field.

Though the `weekdayFilter` field still indicates that `null` is acceptable and could be made clearer (I'll push an update).

Making them dependant on each other is also a bit ambiguous because it could go either way on what requires what, so it was decided early on that the `time` field is the source of truth, and `weekdayFilter` is secondary.

Forfold (approved):

Forfold (approved):

Forfold (approved):

dctalbot (approved):

PR #1673, opened 2021-06-29
ui/schedules: use ISODates for calendar navigation

### LINKED ISSUE #1412: ui/schedules: calendar URL is unstable across timezones
**Describe the Bug:**
A URL like https://xyz.goalert.me/schedules/6633be5s-e02f-4b55-b2ed-17c084908daa?start=2021-04-01T05%3A00%3A00.000Z
will render April in America/Chicago and March in America/Los_Angeles

**Expected Behavior:**
URL displays same month regardless of user's location

**Observed Behavior:**
URL displays different month depending on user's location

**Application Version:**
`b2157d548a287d859cec4b88f87bf038fd92f315`

dctalbot:

**Description:**
This PR fixes #1412 by having the start URL query parameter use an ISODate format (as opposed to ISODateTime) since time and zone are not interesting to a user in this context.

e.g. A user in New York shares a calendar link to a user in California for Aug 2021. Both should see Aug 2021 upon clicking the link.

Fixes #1412

**Review Guide:**
The only impactful changes in this PR are switching calls from `toUTC().toISO()` to `toISODate()`
Other refactoring with no change in behavior:
- `getStartOfWeek` and `getEndOfWeek` are now timezone agnostic, and their implementation has been simplified
- `.local()` has been removed when chained onto `.fromISO()` since it is redundant (`fromISO` defaults to 'system's locale')
- `.now()` is preferred over `.local()` since it has a narrower API, but they are synonymous when no parameters are passed in
-  `useCalendarNavigation` keeps URL query param interpretation  consistent with the calendar toolbar
- `queryStart` and `queryEnd` don't need to convert to UTC since the backend handles timezone offsets

**Additional Info:**
This change is backwards compatible; luxon's `fromISO` can handle both dates and timestamps.


web/src/app/util/luxon-helpers.ts
```diff
@@ -1,33 +1,28 @@
...
-    .plus({
-      days: 6 - weekdayIndex,
-    })
-    .endOf('day')
+  // saturday
+  if (dt.weekday === 6) {
+    return dt.endOf('day')
+  }
+
+  return dt.endOf('week').minus({ day: 1 })
```
Forfold, line 27: should this be plus?

```suggestion
  return dt.endOf('week').plus({ day: 1 })
```

dctalbot: No, `dt.endOf('week')` returns the next sunday at 11:59pm, so we subtract 1 day to get saturday at 11:59pm. Also have some unit tests in luxon-helpers.test.js. Thanks for checking!

Forfold (approved):

mastercactapus (approved):

PR #1689, opened 2021-07-06
ui/hooks: WizardRouter

Voninkazo:
Hello all,
I have refactored the wizardRouter class component to a functional component.
Please have a look at it and give any feedback for things to improve.
Here is the [branch](https://github.com/Voninkazo/goalert/tree/wizard-router)

Thanks in advance!

### DIFF
--- a/web/src/app/wizard/WizardRouter.js
+++ b/web/src/app/wizard/WizardRouter.js
@@ -38,14 +38,14 @@ const mutation = gql`
   }
 `
 
-const styles = makeStyles(() => ({
+const useStyles = makeStyles(() => ({
   cardActions: {
     justifyContent: 'flex-end',
   },
 }))
 
 export default function WizardRouter() {
-  const classes = styles()
+  const classes = useStyles()
   const width = useWidth()
   const [errorMessage, setErrorMessage] = useState(null)
   const [complete, setComplete] = useState(false)
--- a/web/src/app/wizard/WizardRouter.js
+++ b/web/src/app/wizard/WizardRouter.js
@@ -1,4 +1,4 @@
-import React, {useState} from 'react'
+import React, { useState } from 'react'
 import Button from '@material-ui/core/Button'
 import Card from '@material-ui/core/Card'
 import CardContent from '@material-ui/core/CardContent'
@@ -7,8 +7,8 @@ import Dialog from '@material-ui/core/Dialog'
 import DialogContent from '@material-ui/core/DialogContent'
 import DialogContentText from '@material-ui/core/DialogContentText'
 import DialogActions from '@material-ui/core/DialogActions'
-import withStyles from '@material-ui/core/styles/withStyles'
 import { Redirect } from 'react-router-dom'
+import { makeStyles } from '@material-ui/core'
 import { DateTime } from 'luxon'
 import { fieldErrors, nonFieldErrors } from '../util/errutil'
 import WizardForm from './WizardForm'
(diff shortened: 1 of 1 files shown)


web/src/app/wizard/WizardRouter.js
```diff
@@ -45,8 +45,9 @@ const styles = {
 
 @withWidth()
```
Forfold, line 46: Instead of this decorator/higher order component (HOC) adding the width prop to this component, you can use the `useWidth` hook we have created from the util directory:

`import useWidth from '../../util/useWidth'`

and then

`const width = useWidth()`

web/src/app/wizard/WizardRouter.js
```diff
@@ -45,8 +45,9 @@ const styles = {
...
 @withStyles(styles)
```
Forfold, line 47: withStyles has also been replaced by makeStyles from material-core. You can find an example in most hooks files.

https://github.com/target/goalert/blob/master/web/src/app/services/HeartbeatMonitorList.js#L38

web/src/app/wizard/WizardRouter.js
```diff
@@ -45,8 +45,9 @@ const styles = {
...
-export default class WizardRouter extends React.PureComponent {
-  state = {
+
+export default function WizardRouter ({classes, width}) {
+ const [state, setState] = useState({
```
Forfold, line 50: When we convert from `this.state` on a Class Component to `useState` on a hooks component, we'll want to have a call to `useState` for each main key:value pair.

e.g.
```
const [errorMessage, setErrorMessage] = useState(null)
const [complete, setComplete] = useState(false)
const [redirect, setRedirect] = useState(false)
const [value, setValue] = useState(...) // everything in that value obj
```

web/src/app/wizard/WizardRouter.js
```diff
@@ -87,7 +88,35 @@ export default class WizardRouter extends React.PureComponent {
           timeZone: null,
         },
       },
-    },
+    }})
+
+
+    /*
```
Forfold, line 94:
```suggestion
  /*
```

web/src/app/wizard/WizardRouter.js
```diff
@@ -239,8 +240,8 @@ export default class WizardRouter extends React.PureComponent {
     )
   }
 
-  renderSubmittedContent() {
-    const { complete, errorMessage, value } = this.state
+  function renderSubmittedContent(state) {
```
Forfold, line 243: This is giving me the error `renderSubmittedContent' was used before it was defined`. You can bring this whole function into the component's body, and then you can delete the `state` parameter here since it will be in scope.

Forfold (changes requested):

web/src/app/wizard/WizardRouter.js
```diff
@@ -257,4 +188,59 @@ export default class WizardRouter extends React.PureComponent {
...
+              <Form
+                onSubmit={(e, isValid) => submit(e, isValid, commit)}
+                disabled={loading}
+              >
+                <CardContent>
+                  <WizardForm
+                    disabled={status.loading}
+                    errors={fieldErrors(error)}
+                    value={value}
+                    onChange={(value) => setValue({ value })}
```
Forfold, line 211: This will break until we update the reference here as well, now that we changed the state to be flat:
```suggestion
                    onChange={(value) => setValue(value)}
```

Forfold (changes requested):

web/src/app/wizard/WizardRouter.js
```diff
@@ -37,57 +38,79 @@ const mutation = gql`
   }
 `
 
-const styles = {
+const styles = makeStyles(() => ({
```
Forfold, line 41: nit:
```suggestion
const useStyles = makeStyles(() => ({
```

Forfold (approved):
Nice job- functionally tested locally and looks good!

dctalbot (approved):

PR #1690, opened 2021-07-06
ui/hooks: ScheduleCreateDialog

voromahery:

**Description:**
Converts Schedule create dialog into hooks 

Part of #923

### DIFF
--- a/web/src/app/schedules/ScheduleCreateDialog.js
+++ b/web/src/app/schedules/ScheduleCreateDialog.js
@@ -18,13 +18,11 @@ const mutation = gql`
 `
 
 export default function ScheduleCreateDialog(props) {
-  const [schedule, setSchedule] = useState({
-    value: {
-      name: '',
-      description: '',
-      timeZone: Intl.DateTimeFormat().resolvedOptions().timeZone,
-      favorite: true,
-    },
+  const [value, setValue] = useState({
+    name: '',
+    description: '',
+    timeZone: Intl.DateTimeFormat().resolvedOptions().timeZone,
+    favorite: true,
   })
 
   return <Mutation mutation={mutation}>{renderForm}</Mutation>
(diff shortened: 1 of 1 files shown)


web/src/app/schedules/ScheduleCreateDialog.js
```diff
@@ -17,36 +17,35 @@ const mutation = gql`
   }
 `
 
-export default class ScheduleCreateDialog extends React.PureComponent {
-  state = {
+export default function ScheduleCreateDialog(props) {
+  const [schedule, setSchedule] = useState({
```
Forfold, line 21: let's keep these names as `value` and `setValue`, as this would remain consistent with how we use forms across the app. Then, `name`, `description`, `timeZone`, and favorite can be listed flat for the object initialization.

Forfold (changes requested):

Forfold (changes requested):
I'm getting linting errors locally, please run `make check` and address anything that comes up. I'm also seeing the error in console
```
The prop `value.name` is marked as required in `ScheduleForm`, but its value is `undefined`.
```

when trying to use the form.

web/src/app/schedules/ScheduleCreateDialog.js
```diff
@@ -61,11 +56,12 @@ export default class ScheduleCreateDialog extends React.PureComponent {
           <ScheduleForm
             disabled={status.loading}
             errors={fieldErrors(status.error)}
-            value={this.state.value}
-            onChange={(value) => this.setState({ value })}
+            value={value}
+            onChange={(value) => setValue({ value })}
```
dctalbot, line 60:
```suggestion
            onChange={(value) => setValue(value)}
```

dctalbot (changes requested):
Looks like the form is broken

dctalbot (approved):

Forfold (approved):

PR #1698, opened 2021-07-07
Refactored Options class component to a functional component

Voninkazo:
Hello there,

The options class component is refactored to a functional component.
You're welcome to give any feedback for things to improve.
You can find the branch [here](https://github.com/Voninkazo/goalert/tree/util-options)

Thank in advance!


Forfold (changes requested):
I actually don't think this component is used anywhere anymore (in favor of the `OtherActions` component) and can be deleted.

PR #1700, opened 2021-07-07
ui/hooks: OtherActions

Voninkazo:
Hello there,

The OtherActions class component is refactored to a functional component.
Please have a look at it and make any necessary suggestions.
Find the branch [here](https://github.com/Voninkazo/goalert/tree/other-actions).

Thank you in advance!

### DIFF
--- a/web/src/app/util/OtherActions.js
+++ b/web/src/app/util/OtherActions.js
@@ -1,9 +1,8 @@
-import React from 'react'
+import React, { useState } from 'react'
 import p from 'prop-types'
 import IconButton from '@material-ui/core/IconButton'
 import { MoreHoriz as OptionsIcon } from '@material-ui/icons'
 import Hidden from '@material-ui/core/Hidden'
-
 import OtherActionsDesktop from './OtherActionsDesktop'
 import OtherActionsMobile from './OtherActionsMobile'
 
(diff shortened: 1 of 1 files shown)


web/src/app/util/OtherActions.js
```diff
@@ -16,71 +16,70 @@ const cancelable = (_fn) => {
...
-        onClick: p.func.isRequired,
-      }),
-    ).isRequired,
-    color: p.string,
-    icon: p.element,
-    placement: p.oneOf(['left', 'right']),
-  }
+export default function OtherActions(props) {
+  
+  const [state, setState] = useState({
```
Forfold, line 21: Rather than state/setState, lets do
```suggestion
  const [anchorEl, setAnchorEl] = useState(null)
```

that way we can use `anchorEl` directly without having to reference an object

web/src/app/util/OtherActions.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, {useState} from 'react'
```
Forfold, line 1: I'm noticing some linting issues here, running `yarn lint --fix` should fix the issue. This will cause `make check` to fail until then

Forfold (changes requested):

Forfold (approved):
LGTM! Approving to run tests 💯

web/src/app/util/OtherActions.js
```diff
@@ -16,71 +15,61 @@ const cancelable = (_fn) => {
...
-      p.shape({
-        label: p.string.isRequired,
-        onClick: p.func.isRequired,
-      }),
-    ).isRequired,
-    color: p.string,
-    icon: p.element,
-    placement: p.oneOf(['left', 'right']),
-  }
+export default function OtherActions({ color, icon, ref, actions, placement }) {
```
Forfold, line 18: `ref` is not a prop for `OtherActions`, which is causing `make check` to fail. You will likely have to use something like `useRef` hook to replicate `this.ref`

More info here: https://reactjs.org/docs/hooks-reference.html#useref

Forfold (changes requested):

KatieMSB (approved):

Forfold (approved):

PR #1701, opened 2021-07-07
ui/hooks: ScheduleEditDialog

voromahery:

**Description:**
Converts Schedule edit dialog into hooks 

Part of #923

### DIFF
--- a/web/src/app/schedules/ScheduleEditDialog.js
+++ b/web/src/app/schedules/ScheduleEditDialog.js
@@ -25,7 +25,7 @@ const mutation = gql`
 `
 
 export default function ScheduleEditDialog(props) {
-  const [state, setState] = useState(null)
+  const [value, setValue] = useState(null)
 
   const renderForm = (data, commit, status) => {
     return (
@@ -38,7 +38,7 @@ export default function ScheduleEditDialog(props) {
             variables: {
               input: {
                 id: props.scheduleID,
-                ...state,
+                ...value,
               },
             },
           })
@@ -48,13 +48,13 @@ export default function ScheduleEditDialog(props) {
             disabled={status.loading}
             errors={fieldErrors(status.error)}
             value={
-              state || {
+              value || {
                 name: data.name,
                 description: data.description,
                 timeZone: data.timeZone,
               }
             }
-            onChange={(value) => setState(value)}
+            onChange={(value) => setValue(value)}
           />
         }
       />
--- a/web/src/app/schedules/ScheduleEditDialog.js
+++ b/web/src/app/schedules/ScheduleEditDialog.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState } from 'react'
 import { gql } from '@apollo/client'
 import p from 'prop-types'
 import FormDialog from '../dialogs/FormDialog'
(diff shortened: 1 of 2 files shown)


web/src/app/schedules/ScheduleEditDialog.js
```diff
@@ -24,46 +24,34 @@ const mutation = gql`
   }
 `
 
-export default class ScheduleEditDialog extends React.PureComponent {
-  static propTypes = {
+export default function ScheduleEditDialog(props) {
+  ScheduleEditDialog.propTypes = {
```
dctalbot, line 28: This should be moved after the function declaration. See an [example here](https://github.com/target/goalert/blob/master/web/src/app/escalation-policies/PolicyDetails.js#L122)

web/src/app/schedules/ScheduleEditDialog.js
```diff
@@ -24,46 +24,34 @@ const mutation = gql`
...
-        variables={{ id: this.props.scheduleID }}
-        render={({ data }) => this.renderMutation(data.schedule)}
-      />
-    )
-  }
-
-  renderMutation(data) {
+  const renderMutation = (data) => {
     return (
       <Mutation mutation={mutation} onCompleted={this.props.onClose}>
```
dctalbot, line 37:
```suggestion
      <Mutation mutation={mutation} onCompleted={props.onClose}>
```
Since `props` is now a function argument

web/src/app/schedules/ScheduleEditDialog.js
```diff
@@ -24,46 +24,34 @@ const mutation = gql`
...
+export default function ScheduleEditDialog(props) {
+  ScheduleEditDialog.propTypes = {
     scheduleID: p.string.isRequired,
     onClose: p.func,
   }
 
-  state = {
-    value: null,
-  }
+  const [state, setState] = useState(null)
```
dctalbot, line 33: `useState` is not defined. We need to import it from the react package e.g. `import React, { useState } from 'react'`

web/src/app/schedules/ScheduleEditDialog.js
```diff
@@ -24,46 +24,34 @@ const mutation = gql`
...
     return (
       <Mutation mutation={mutation} onCompleted={this.props.onClose}>
-        {(...args) => this.renderForm(data, ...args)}
+        {(...args) => renderForm(data, ...args)}
       </Mutation>
     )
   }
 
-  renderForm = (data, commit, status) => {
+  const renderForm = (data, commit, status) => {
```
dctalbot, line 43: We need to move this function to be above `renderMutation` since `renderMutation` uses `renderForm` and we have a linting rule to prevent hoisting

web/src/app/schedules/ScheduleEditDialog.js
```diff
@@ -73,16 +61,24 @@ export default class ScheduleEditDialog extends React.PureComponent {
...
             value={
-              this.state.value || {
+              state || {
                 name: data.name,
                 description: data.description,
                 timeZone: data.timeZone,
               }
             }
-            onChange={(value) => this.setState({ value })}
+            onChange={(value) => setState({ value })}
```
dctalbot, line 70: We need to remove the curly bracing here since the form state is no longer keyed into an object
```js
onChange={(value) => setState( value )}
```

dctalbot (changes requested):

dctalbot (dismissed):

web/src/app/schedules/ScheduleEditDialog.js
mastercactapus: Let's name this `value` instead of `state`
```suggestion
  const [value, setValue] = useState(null)
```

mastercactapus (dismissed):

dctalbot (changes requested):
Please remove the changes to `web/src/app/services/ServiceLabelDeleteDialog.js`
These PR's should only convert one component at a time

dctalbot (approved):

Forfold (approved):

PR #1704, opened 2021-07-07
ui/hooks: ScheduleNewOverrideFAB

voromahery:

**Description:**
Converts Schedule new override FAB into hooks 

Part of #923


web/src/app/schedules/ScheduleNewOverrideFAB.js
```diff
@@ -3,30 +3,28 @@ import p from 'prop-types'
 import SpeedDial from '../util/SpeedDial'
 import { AccountSwitch, AccountMinus, AccountPlus } from 'mdi-material-ui'
 
-export default class ScheduleNewOverrideFAB extends React.PureComponent {
-  static propsTypes = {
+export default function ScheduleNewOverrideFAB(props) {
+  ScheduleNewOverrideFAB.propsTypes = {
```
dctalbot, line 7: We can not reference `ScheduleNewOverrideFAB` from within its own function declaration. For example:

```js
function test() {
  console.log(test); // undefined
}
```

Instead, we can move it to be after the function declaration
```js
export default function ScheduleNewOverrideFAB(props) {
// actions...
// return...
}


ScheduleNewOverrideFAB.propTypes = {
  onClick: p.func.isRequired,
}
```

Also, note there is a typo - it should be `propTypes`, not `propsTypes`

See example here: https://github.com/target/goalert/blob/master/web/src/app/escalation-policies/PolicyDetails.js#L122

dctalbot (changes requested):
Thanks for the PR! This looks great, just need to fix an issue with scoping for `propTypes`

dctalbot (approved):

KatieMSB (approved):

PR #1730, opened 2021-07-12
ui/hooks: ScheduleRuleForm 

voromahery:

**Description:**
Converts Schedule rule form into hooks 

Part of #923

### DIFF
--- a/web/src/app/schedules/ScheduleRuleForm.js
+++ b/web/src/app/schedules/ScheduleRuleForm.js
@@ -103,45 +103,16 @@ const query = gql`
 
 @withStyles(styles)
 @connect((state) => ({ zone: urlParamSelector(state)('tz', 'local') }))
-export default class ScheduleRuleForm extends React.PureComponent {
-  static propTypes = {
-    targetType: p.oneOf(['rotation', 'user']).isRequired,
-    targetDisabled: p.bool,
 
-    scheduleID: p.string.isRequired,
-
-    value: p.shape({
-      targetID: p.string.isRequired,
-      rules: p.arrayOf(
-        p.shape({
-          start: p.string.isRequired,
-          end: p.string.isRequired,
-
-          weekdayFilter: p.arrayOf(p.bool).isRequired,
-        }),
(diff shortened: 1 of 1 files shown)


web/src/app/schedules/ScheduleRuleForm.js
```diff
@@ -103,45 +103,16 @@ const query = gql`
 
 @withStyles(styles)
```
KatieMSB, line 104: Instead of `@withStyles`, you can use `makeStyles` to make a hook for the classes. Here's an example: https://github.com/target/goalert/blob/master/web/src/app/schedules/ScheduleShiftList.js#L54

web/src/app/schedules/ScheduleRuleForm.js
```diff
@@ -103,45 +103,16 @@ const query = gql`
...
 @connect((state) => ({ zone: urlParamSelector(state)('tz', 'local') }))
```
KatieMSB, line 105: We have a custom hook to handle this url param selector, you can import it as:
`import { useURLParam } from '../actions/hooks'`
and set the param as:
`const [zone] = useURLParam('tz', 'local') `
Exmaple here: https://github.com/target/goalert/blob/master/web/src/app/schedules/ScheduleShiftList.js#L68

KatieMSB (changes requested):

web/src/app/schedules/ScheduleRuleForm.js
```diff
@@ -100,131 +99,13 @@ const query = gql`
     }
   }
 `
+const useStyles = makeStyles(styles)
```
Forfold, line 102: lets add the styles object directly here, since `styles` isn't used anywhere else

```
const useStyles = makeStyles({
  noPadding: {
    padding: 0,
  },
  . . .
})
```

Forfold (changes requested):
looks good, though 1 small change

Forfold (approved):

KatieMSB (approved):

PR #1734, opened 2021-07-12
ui/hooks: IntegrationKeyCreateDialog

voromahery:

**Description:**
Converts IntegrationKeyCreateDialog into hooks 

Part of #923

### DIFF
--- a/web/src/app/services/IntegrationKeyCreateDialog.js
+++ b/web/src/app/services/IntegrationKeyCreateDialog.js
@@ -32,11 +32,8 @@ const query = gql`
   }
 `
 
-export default function IntegrationKeyCreateDialog() {
-  const [state, setState] = useState({
-    value: { name: '', type: 'generic' },
-    errors: [],
-  })
+export default function IntegrationKeyCreateDialog(props) {
+  const [value, setValue] = useState({ name: '', type: 'generic' })
 
   const renderDialog = (commit, status) => {
     const { loading, error } = status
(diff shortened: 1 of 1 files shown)


web/src/app/services/IntegrationKeyCreateDialog.js
```diff
@@ -32,71 +32,69 @@ const query = gql`
...
 
-export default class IntegrationKeyCreateDialog extends React.PureComponent {
-  static propTypes = {
-    serviceID: p.string.isRequired,
-    onClose: p.func,
-  }
-
-  state = {
+export default function IntegrationKeyCreateDialog() {
+  const [state, setState] = useState({
```
mastercactapus, line 36: break out `value` and `errors` to separate `useState` calls rather than one object

`const [value, setValue] = useState(`
`const [errors, setErrors] = useState([])`

mastercactapus (changes requested):

web/src/app/services/IntegrationKeyCreateDialog.js
voromahery: Is it okay to remove the **errors** entirely because it is not used in this file and may cause linting problem?

web/src/app/services/IntegrationKeyCreateDialog.js
```diff
@@ -32,71 +32,66 @@ const query = gql`
...
           })
         }}
         form={
           <IntegrationKeyForm
             errors={fieldErrors(error)}
             disabled={loading}
-            value={this.state.value}
-            onChange={(value) => this.setState({ value })}
+            value={value}
+            onChange={(value) => setValue({ value })}
```
dctalbot, line 59:
```suggestion
            onChange={(value) => setValue(value)}
```

dctalbot (changes requested):
Looks like the form is broken. You can run the test suite with `make cy-wide`

dctalbot (approved):

Forfold (approved):

PR #1739, opened 2021-07-12
slack: unit tests for on-call messages

mastercactapus:

**Description:**
Pulls the rendering logic out of the `Send` method for Slack on-call notifications into a unit-testable function and adds tests for edge cases.

### DIFF
--- a/notification/slack/oncallnotification_test.go
+++ b/notification/slack/oncallnotification_test.go
@@ -8,21 +8,27 @@ import (
 )
 
 func TestRenderOnCallNotification(t *testing.T) {
-	slackUserIDs := map[string]string{"slack.1": "slack.1.ID", "slack.2": "slack.2.ID", "slack.3": "slack.3.ID"}
+	// static user id -> slack ID mappings for testing
+	slackUserIDs := map[string]string{
+		"slack.1": "slack.1.SLACKID",
+		"slack.2": "slack.2.SLACKID",
+		"slack.3": "slack.3.SLACKID",
+	}
 	msg := notification.ScheduleOnCallUsers{
 		ScheduleName: "schedule.name",
 		ScheduleURL:  "schedule.url",
 	}
 
-	check := func(desc, expected string, users []string) {
+	// test with `msg` for the provided user ids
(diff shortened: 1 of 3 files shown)


arurao (dismissed):

notification/slack/oncallnotification_test.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+func TestRenderOnCallNotification(t *testing.T) {
+	slackUserIDs := map[string]string{"slack.1": "slack.1.ID", "slack.2": "slack.2.ID", "slack.3": "slack.3.ID"}
+	msg := notification.ScheduleOnCallUsers{
+		ScheduleName: "schedule.name",
+		ScheduleURL:  "schedule.url",
+	}
+
+	check := func(desc, expected string, users []string) {
```
Forfold, line 17: Can you add a comment briefly going over the expected input/output for this test? I think it would be useful since `users` isn't used for the test assertion, but for setting the test/check up

e.g.
```suggestion
	// `expected` represents the string in plaintext as rendered by Slack
	// `users` represents the name for each expected user link.
	// If a user is prepended with "slack.", it should render with the following format: <@slack.xxx.ID>
	check := func(desc, expected string, users []string) {
```


I wasn't aware that it would translate from `slack.x` to `<@x.ID>`.,so oculd be valuable for others to know as well

mastercactapus: There are no rules around prepending; it's just using the `map[string]string` above as an argument to the function that's being tested. I'll move it down and add a comment
```go
// renderOnCallNotificationMessage will render a message for Slack including links for the schedule and any users.
//
// If a user's ID is available in userSlackIDs, an `@` user mention will be used in place of a link to the GoAlert user's detail page.
func renderOnCallNotificationMessage(msg notification.ScheduleOnCallUsers, userSlackIDs map[string]string) string {
```

I made some of the naming more clear too.

Forfold: Gotcha, thanks for clearing it up!

Forfold (approved):

arurao (approved):

PR #1740, opened 2021-07-12
engine/schedule: fix panic on notification without assignment/rules

mastercactapus:

**Description:**
Fixes a regression where a schedule with configured on-call notification(s), but no assignments/rules would result in a panic causing the engine to skip schedule processing.

This happens because timezone information is only loaded with schedule rules, but is required for on-call notifications as well.

This PR fixes the problem by ensuring timezone information for ALL schedules is loaded before processing, rather than only those with rules associated with them.

**Steps to Reproduce/Validate**
1. Create a schedule with no rules/assignments
2. Add on-call notification
3. Observe error in logs, and other schedules will no longer update on-call users

**Out of Scope**
Future work will focus on isolating schedule processing so a single schedule won't cause the entire module to abort.

### DIFF
--- a/engine/schedulemanager/db.go
+++ b/engine/schedulemanager/db.go
@@ -21,6 +21,8 @@ type DB struct {
 	data        *sql.Stmt
 	updateData  *sql.Stmt
 
+	schedTZ *sql.Stmt
+
 	scheduleOnCallNotification *sql.Stmt
 }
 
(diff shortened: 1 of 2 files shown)


dctalbot (approved):
Functionally validated and ran tests locally

Forfold (approved):
Validated before/after, lgtm!

PR #1770, opened 2021-07-19
ui/hooks:  WizardScheduleForm

Voninkazo:

**Description:**
Refactoring a class component to a functional component.

Part of issue #923

### DIFF
--- a/web/src/app/wizard/WizardScheduleForm.js
+++ b/web/src/app/wizard/WizardScheduleForm.js
@@ -8,16 +8,17 @@ import Grid from '@material-ui/core/Grid'
 import Radio from '@material-ui/core/Radio'
 import RadioGroup from '@material-ui/core/RadioGroup'
 import Tooltip from '@material-ui/core/Tooltip'
-import withStyles from '@material-ui/core/styles/withStyles'
-import withWidth, { isWidthDown } from '@material-ui/core/withWidth'
+import { isWidthDown } from '@material-ui/core/withWidth'
 import InfoIcon from '@material-ui/icons/Info'
 import { TimeZoneSelect, UserSelect } from '../selection'
 import { FormField } from '../forms'
 import { value as valuePropType } from './propTypes'
 import { set } from 'lodash'
 import { ISODateTimePicker } from '../util/ISOPickers'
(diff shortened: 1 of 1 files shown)


web/src/app/wizard/WizardScheduleForm.js
```diff
@@ -28,98 +29,79 @@ const styles = {
...
-export default class WizardScheduleForm extends React.Component {
-  static propTypes = {
-    onChange: p.func.isRequired,
-    value: valuePropType,
-    secondary: p.bool,
+export default function WizardScheduleForm({
+  value,
+  onChange,
+  secondary,
+  forceUpdate,
```
dctalbot, line 42: This is not a prop (see prop types), so its value will be `undefined`. We may need to do some research into how [the class based forceUpdate](https://reactjs.org/docs/react-component.html#forceupdate) was being used before

dctalbot (changes requested):
Thanks for the PR! The `forceUpdate` question needs some more attention before we can approve

web/src/app/wizard/WizardScheduleForm.js
Voninkazo: Thank you for reviewing the code. I'm thinking of installing `use-force-update` dependency and import the `useForceUpdate` hook  like this [example](https://bestofreactjs.com/repo/CharlesStover-use-force-update-react-awesome-react-hooks). Is that fine?

Thanks you in advance!

dctalbot: I will take a look first. We only install dependencies as a last resort. Thanks for your patience!

web/src/app/wizard/WizardScheduleForm.js
```diff
@@ -28,94 +29,74 @@ const styles = {
...
+            />
+          </Grid>
+        </React.Fragment>
+      )
+    }
   }
 
-  getKey = () => {
-    const { secondary } = this.props
+const getKey = () => {
```
dctalbot, line 83:
```suggestion
  const getKey = () => {
```

dctalbot (approved):

PR #1779, opened 2021-07-20
ui/alerts: display pending notification notices on alert details

### LINKED ISSUE #1083: Alert Details: Display pending messages
It is not obvious to users/teams when notifications are scheduled but not yet sent.

Gray/shaded log entries at the top of the alert log view showing pending messages.
Its container might have some text (e.g. info tooltip) like: "Notifications may be delayed for a particular alert if a user has received several messages recently"
The status text could be like: "pending" or "waiting to retry"

Eventually pending messages will be logged as sent or failed.

**Additional context:**
These would be generated from `outgoing_messages` rows in the `pending` state, or `bundled` where the bundle message is in a `pending` state.

dctalbot:

**Description:**
Currently, there is little to no visibility from the UI to indicate throttling is taking place. This PR render notices on the Alert Details screen, listing the destinations that have been waiting to receive an alert for more than 1 minute.

Fixes #1083 

**Screenshots:**
Single:
Screen Shot 2021-07-20 at 1 45 35 PM

Multiple:
Screen Shot 2021-07-20 at 1 53 02 PM
Screen Shot 2021-07-20 at 1 53 07 PM

Notices displayed when alerts are pending

```graphql
type Alert {
  ...
  pendingNotifications: [AlertPendingNotification!]!
}

type AlertPendingNotification {
  destination: String!
}
```

**Example for how to review:**
1. Set up Slack integration with a channel
2. Create alert for said channel and escalate multiple times in a row
3. View alert details


alert/store.go
```diff
@@ -255,9 +260,63 @@ func NewDB(ctx context.Context, db *sql.DB, logDB alertlog.Store) (*DB, error) {
...
+	}
+
+	err = validate.UUID("ServiceID", serviceID)
+	if err != nil {
+		return nil, err
+	}
+
+	rows, err := db.findPendingNotifications.QueryContext(ctx, alertID, serviceID)
+	if errors.Is(err, sql.ErrNoRows) {
+		err = nil
```
Forfold, line 295: Can we return here since we know there are no rows?

```suggestion
		return nil, nil
```

alert/store.go
```diff
@@ -255,9 +260,63 @@ func NewDB(ctx context.Context, db *sql.DB, logDB alertlog.Store) (*DB, error) {
...
+			COALESCE(u.name, nc.name) AS dest_name
+			FROM outgoing_messages om
+			LEFT JOIN user_contact_methods ucm 
+			ON om.user_id = ucm.user_id
+			LEFT JOIN notification_channels nc 
+			ON nc.id = om.channel_id
+			LEFT JOIN users u
+			ON u.id = om.user_id
+			WHERE om.last_status='pending'
+			AND (now() - om.created_at) > INTERVAL '1' MINUTE
```
Forfold, line 276: I think it would be useful to show this right away since the log is empty until a notification is queued up to send. I was a little confused when testing on the alert details page and nothing was showing up in the log or as a notice at first- I expected to either see that I was being rate-limited, or that my message was being sent

Screen Shot 2021-07-21 at 12 13 00 PM

graphql2/graphqlapp/alert.go
```diff
@@ -375,6 +375,28 @@ func (a *Alert) RecentEvents(ctx context.Context, obj *alert.Alert, opts *graphq
 	return conn, err
 }
 
+// PendingNotifications returns a list of notifications that are waiting to be sent for some reason such as throttling
+func (a *Alert) PendingNotifications(ctx context.Context, obj *alert.Alert) ([]graphql2.AlertPendingNotification, error) {
```
Forfold, line 379: Can we use `alert` as the name instead of `obj`?

web/src/app/alerts/components/AlertDetails.js
```diff
@@ -423,8 +424,18 @@ function AlertDetails(props) {
...
+    type: 'WARNING',
+    message: `Notification Pending for ${n.destination}`,
+    details:
+      'One or more notifications are waiting to be sent. Contact methods may be rate limited on a per-service basis.',
+  }))
+
   return (
-    <Grid container spacing={2}>
+    <Grid container spacing={2} justify='center'>
+      <Grid item xs={12} lg={9}>
```
Forfold, line 436: It doesn't look like `lg={9}` is doing anything here

```suggestion
      <Grid item xs={12}>
```

web/src/app/alerts/components/AlertDetails.js
```diff
@@ -423,8 +424,18 @@ function AlertDetails(props) {
 
   const { data: alert } = props
 
+  const notices = alert.pendingNotifications.map((n) => ({
+    type: 'WARNING',
+    message: `Notification Pending for ${n.destination}`,
+    details:
+      'One or more notifications are waiting to be sent. Contact methods may be rate limited on a per-service basis.',
```
Forfold, line 431: I think we can simplify this a bit. My reasoning comes from the writing guide: https://material.io/design/communication/writing.html#principles

We can probably leave out that one or more notifications are waiting since it's mainly repeating the title.

```suggestion
    message: 'Notification Pending',
    details: `${n.destination} is being temporarily rate-limited.`,
```

Forfold (changes requested):

alert/store.go
dctalbot: Got it, I think we added this threshold originally to give some time for non-throttled alerts to go out. @mastercactapus any thoughts on this? Maybe we can lower to 30s

alert/store.go
dctalbot: Yeah, I wanted it to be an empty slice as opposed to nil since the number of rows has been determined at this point. Maybe better to do something like this
```go
if err != nil && !errors.Is(err, sql.ErrNoRows) {
```

graphql2/graphqlapp/alert.go
dctalbot: No, alert is a package name

web/src/app/alerts/components/AlertDetails.js
dctalbot: Good call, I agree we should cut the redundancy. Not sure we can say definitively that rate-limiting is the cause though. Maybe something like this?

```js
{
message: `Notification Pending for ${n.destination}`,
details: `Contact methods may be rate limited`
}
```

alert/store.go
mastercactapus: I'd say 10-15s would be the lowest, 30s would be reasonable.

web/src/app/alerts/components/AlertDetails.js
dctalbot: It sets the width of the notice
Screen Shot 2021-07-21 at 3 25 47 PM

alert/store.go
dctalbot: Hmm, still thinking on this one. It probably is better to just short circuit like you suggested

web/src/app/alerts/components/AlertDetails.js
mastercactapus: "pending" messages do not mean rate-limiting necessarily. "waiting to be sent" is about as accurate as it can be -- slow network, service disruption, retries, engine not running, and any number of other things can contribute to a message being in a pending state

alert/store.go
dctalbot: Ended up going with your suggestion since we still expose an empty array to the client e.g. `pendingNotifications: []` thanks!

web/src/app/alerts/components/AlertDetails.js
Forfold:
```suggestion
    message: `Notification Pending for ${n.destination}`,
    details:
      'One or more notifications are waiting to be sent.',
```

web/src/app/alerts/components/AlertDetails.js
Forfold: Are notices being rendered outside of our App's container? I feel like this component should adhere to the same rules we apply to all other grid items here. Setting to `9` feels like an arbitrary value and could lead to it being misaligned in the future, since the other grid items are all set to `12`.

web/src/app/alerts/components/AlertDetails.js
dctalbot: I think it's important to call out "rate-limited" or "throttled" if we have reason to think that is a likely cause

web/src/app/alerts/components/AlertDetails.js
dctalbot: Interesting, it's within the same container but the existing alert details use [this css](https://github.com/target/goalert/blob/master/web/src/app/styles/materialStyles.js#L3-L9) instead of  the `xs={12} lg={9}` props. I'll do that again here to prevent a data sync issue, but it's nearing tech debt territory :/

dctalbot: These cards have 75% width within the app container's 75% width

Forfold: So it looks like it's happening because we're nesting the notification panel inside of a grid, which is already a grid- I've come across layout issues like this in the past when you have multiple grids trying to work together. I can do some investigating on this today. We shouldn't need to use `lg={9}` when the other cards do not

web/src/app/alerts/components/AlertDetails.js
mastercactapus: We could add something to give a little extra context, as long as it's accurate and doesn't imply the reason is known:

Something like:
> This could be due to rate-limiting, processing, or network delays.

The vast majority of users will never see this warning unless something **other than** rate-limiting is happening (e.g., network delays, prioritization for non-first-alerts, service disruption, etc...). Those that are rate-limiting are a smaller subset of configurations, and thus the same users/services each time (e.g., a service that alerts frequently, or too-aggressive notification settings).

In any case, since rate-limiting has a small inference that the user did something wrong or should configure differently, we should avoid any wording that implies the user is being rate-limited unless it's 100% accurate -- and until we refactor the message-sending code with time-to-send & reason, we do not have that information available.

The goal here is only to inform the user that there is a message ready, but not yet in the alert log, so they know the policy is processing as it should.

dctalbot: Thanks for the insight, I went ahead and updated the wording: 
```
message: `Notification Pending for ${n.destination}`,
details: 'One or more notifications are waiting to be sent. This could be due to rate-limiting, processing, or network delays.',
```

Forfold (approved):

alert/store.go
```diff
@@ -255,9 +260,63 @@ func NewDB(ctx context.Context, db *sql.DB, logDB alertlog.Store) (*DB, error) {
 			FROM services
 			WHERE id = $1
 		`),
+
+		findPendingNotifications: p(`
+			SELECT
+			COALESCE(ucm.type::text, nc.type::text) AS dest_type,
```
mastercactapus, line 266: These should be scanned separately, something like email would be in both in the future.

`engine/message/db.go:401` for an example

alert/store.go
```diff
@@ -255,9 +260,63 @@ func NewDB(ctx context.Context, db *sql.DB, logDB alertlog.Store) (*DB, error) {
...
+	rows, err := db.findPendingNotifications.QueryContext(ctx, alertID, serviceID)
+	if errors.Is(err, sql.ErrNoRows) {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	result := make([]AlertPendingNotification, 0)
```
mastercactapus, line 302: don't rely on/depend/differenciate between `[]slice{}` and `nil`

If absolutely necessary use `result := []AlertPendingNotification{}` instead of `make` if an empty slice. However, code at the store level should just return nil and not guarantee a slice.

If there's a requirement to return an empty slice somewhere else, then use
```
if len(result) == 0 {
  return []type{}, nil
}
```

To keep the intent clear -- since in Go []slice{} and nil are used interchangeably, move the conversion to the very end of the chain -- and only if necessary. In this case, we should allow `nil` all the way to the UI and handle it there if necessary. As written nil is already being returned for all acked/closed alerts below

```suggestion
	var result []AlertPendingNotification
```

alert/store.go
```diff
@@ -255,9 +260,63 @@ func NewDB(ctx context.Context, db *sql.DB, logDB alertlog.Store) (*DB, error) {
...
+	for rows.Next() {
+		var destType string
+		var destName string
```
mastercactapus, line 305: group same-type variables for scanning to reduce noise
```suggestion
		var destType, destName string
```

graphql2/graphqlapp/alert.go
```diff
@@ -375,6 +375,28 @@ func (a *Alert) RecentEvents(ctx context.Context, obj *alert.Alert, opts *graphq
...
+	var result []graphql2.AlertPendingNotification
+
+	if obj.Status != alert.StatusTriggered {
+		return result, nil
+	}
+
+	rows, err := a.AlertStore.FindPendingNotifications(ctx, obj.ID, obj.ServiceID)
```
mastercactapus, line 386: Name this something other than rows, it implies it's a `sql.Rows` which would need to be closed. something like `pending` or `notifications` or just `p` since it's only used for a few lines that are close by without scrolling

alert/store.go
```diff
@@ -255,9 +260,63 @@ func NewDB(ctx context.Context, db *sql.DB, logDB alertlog.Store) (*DB, error) {
...
+			FROM outgoing_messages om
+			LEFT JOIN user_contact_methods ucm 
+			ON om.user_id = ucm.user_id
+			LEFT JOIN notification_channels nc 
+			ON nc.id = om.channel_id
+			LEFT JOIN users u
+			ON u.id = om.user_id
+			WHERE om.last_status='pending'
+			AND (now() - om.created_at) > INTERVAL '15 SECONDS'
+			AND (om.alert_id = $1 OR (om.message_type = 'alert_notification_bundle' AND om.service_id = $2));
```
mastercactapus, line 277: Let's format this a little differently:
```sql
SELECT
	foo,
	bar
FROM baz b
LEFT JOIN qux q ON b.id = q.id
LEFT JOIN quux qu ON b.id = qu.id
WHERE
	b.id = 1 AND
	b.id = 2 AND
	b.id = 3
```

mastercactapus (changes requested):

alert/store.go
```diff
@@ -55,6 +55,9 @@ type Manager interface {
 	// ServiceInfo will return the name of the given service ID as well as the current number
 	// of unacknowledged alerts.
 	ServiceInfo(ctx context.Context, serviceID string) (string, int, error)
+
+	// FindPendingNotifications will return destination info for alerts that are waiting to be sent
+	FindPendingNotifications(ctx context.Context, alertID int, serviceID string) ([]AlertPendingNotification, error)
```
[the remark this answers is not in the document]
dctalbot: Sure, we can do that. The graphql bits should remain in the `alert` package since this is on the `Alert` entity, right?

alert/store.go
dctalbot:
```sql
WITH svc_id AS (
	SELECT service_id FROM alerts WHERE id = 79672
),

contact_methods AS (
	SELECT 
		cm.type::text AS dest_type,
		u.name AS dest_name,
		om.last_status,
		om.service_id,
		om.created_at
	FROM outgoing_messages om
	JOIN users u ON om.user_id = u.id
	JOIN user_contact_methods cm ON u.id = cm.user_id
),

noti_channels AS (
	SELECT 
		nc.type::text AS dest_type,
		nc.name AS dest_name,
		om.last_status,
		om.service_id,
		om.created_at
	FROM outgoing_messages om
	JOIN notification_channels nc ON om.channel_id = nc.id
)

SELECT dest_type, dest_name FROM (
	SELECT * FROM contact_methods
	UNION ALL
	SELECT * FROM noti_channels
) AS result

WHERE result.last_status='pending' 
AND (now() - result.created_at) > INTERVAL '15 SECONDS'
AND result.service_id = (select service_id from svc_id);
```

Is something like this what you had in mind?

alert/store.go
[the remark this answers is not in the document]
mastercactapus: Do you mean `alert.go` in graphql2? if so then yes

alert/store.go
mastercactapus: Not exactly, I was only referring to the columns on that line -- separate them in the select instead of using `coalesce` for the type info

Forfold (dismissed):
LGTM! 15 seconds seems to work well

mastercactapus (approved):

mastercactapus (approved):

arurao (approved):

PR #1785, opened 2021-07-21
notification: slack status updates

arurao:

**Description:**
This PR adds the ability to receive status updates in slack for alerts. 
All updates are currently as in-thread replies, with the exception of escalations being broadcasted to the channel if it is the second notification being sent out to that channel.
This PR also adjusts the rate limits so that only types like sms, voice and email follow the specified rate limits and status update specific rate limits. Slack will not be based off of these limits, to ensure a quick end user experience.

**Screenshots:**
Screen Shot 2021-08-05 at 4 20 27 PM


notification/alertstatus.go
```diff
@@ -5,6 +5,14 @@ type AlertStatus struct {
...
 	AlertID    int
 	LogEntry   string
+
+	// Summary of the alert that this status is in regards to.
+	Summary string
+	// Details of the alert that this status is in regards to.
+	Details string
+
+	// OriginalStatus is the status of the first Alert notification to this Dest for this AlertID.
+	OriginalStatus *SendResult
```
mastercactapus, line 15: Make this a value instead of a pointer since it's always required & guaranteed for status. Eliminates the need for unnecessary nil checks elsewhere.

** will need to update a few other places for the type change (compile errors)

```suggestion
	OriginalStatus SendResult
```

mastercactapus (changes requested):

mastercactapus (dismissed):

notification/slack/channel.go
```diff
@@ -321,12 +321,16 @@ func (s *ChannelSender) Send(ctx context.Context, msg notification.Message) (str
 		if t.OriginalStatus != nil {
 			// Reply in thread if we already sent a message for this alert.
 			vals.Set("thread_ts", t.OriginalStatus.ProviderMessageID.ExternalID)
-			vals.Set("text", "Escalated.")
+			// only if escalation results in second notification to channel
```
dctalbot, line 324:
```suggestion
```
This comment is a bit confusing. Is it restating line 322?

notification/alertstatus.go
```diff
@@ -5,6 +5,14 @@ type AlertStatus struct {
 	CallbackID string
 	AlertID    int
 	LogEntry   string
+
+	// Summary of the alert that this status is in regards to.
+	Summary string
```
dctalbot, line 10: I would be interested in understanding more about the tradeoffs re using an embedded Alert vs. repeating these fields. It seems like this pattern is easier right now, but it's not clear to me that it's a better design

dctalbot (changes requested):

Forfold:
Is there something we can say instead of just "Notified."? I have a feeling that this will be a confusing message to many users since it doesn't refer to why they are being notified within the message.

notification/alertstatus.go
```diff
@@ -5,6 +5,17 @@ type AlertStatus struct {
...
+	// Details of the alert that this status is in regards to.
+	Details string
+
+	// OriginalStatus is the status of the first Alert notification to this Dest for this AlertID.
+	OriginalStatus SendResult
+
+	// Status is the most recent status of the Alert.
+	Status string
```
mastercactapus, line 18: can this be alert.Status instead of a string?

notification/slack/channel.go
```diff
@@ -321,12 +321,16 @@ func (s *ChannelSender) Send(ctx context.Context, msg notification.Message) (str
 		if t.OriginalStatus != nil {
 			// Reply in thread if we already sent a message for this alert.
 			vals.Set("thread_ts", t.OriginalStatus.ProviderMessageID.ExternalID)
-			vals.Set("text", "Escalated.")
+			vals.Set("text", "Broadcasting to channel, due to repeat notification.")
```
dctalbot, line 324:
```suggestion
			vals.Set("text", "Broadcasting to channel due to repeat notification.")
```
https://style.mla.org/splices/

mastercactapus (changes requested):

notification/alertstatus.go
arurao: That was my initial thought too, but that leads to an import cycle. So to keep things simple, I thought of having it as a string instead.

mastercactapus (changes requested):

arurao:
Do not merge before https://github.com/target/goalert/pull/1833. Since that PR refactors some bits that help in avoiding any circular imports in this PR.

dctalbot (changes requested):

dctalbot (approved):

mastercactapus (approved):

PR #1791, opened 2021-07-23
ui/hooks: WizardForm

voromahery:

**Description:**
Converts this file into hooks

Part of #923


web/src/app/wizard/WizardForm.js
```diff
@@ -12,198 +12,195 @@ import Typography from '@material-ui/core/Typography'
...
-    onChange: p.func.isRequired,
-    value: valuePropType,
-    errors: p.arrayOf(
-      p.shape({
-        message: p.string.isRequired,
-      }),
-    ),
-  }
+export default function WizardForm(props) {
+  const { onChange, value, width } = props
```
dctalbot, line 31:
```suggestion
  const { onChange, value } = props
  const width = useWidth()
```
We should import `useWidth` since we aren't using the decorator anymore. See [example here](https://github.com/target/goalert/blob/master/web/src/app/alerts/AlertsList.js#L95)

web/src/app/wizard/WizardForm.js
```diff
@@ -12,198 +12,195 @@ import Typography from '@material-ui/core/Typography'
...
+  const forceUpdate = useState()[1]
```
dctalbot, line 32:
```suggestion
  const forceUpdate = useState()[1]
```
Not sure about this syntax, but we need to review the usage of `forceUpdate` regardless. I will take a look some time and let you know how to proceed!

dctalbot (changes requested):

web/src/app/wizard/WizardForm.js
voromahery: Thanks!

web/src/app/wizard/WizardForm.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState } from 'react'
```
dctalbot, line 1:
```suggestion
import React from 'react'
```

web/src/app/wizard/WizardForm.js
```diff
@@ -12,199 +12,198 @@ import Typography from '@material-ui/core/Typography'
...
-    errors: p.arrayOf(
-      p.shape({
-        message: p.string.isRequired,
-      }),
-    ),
-  }
+export default function WizardForm(props) {
+  const { onChange, value } = props
+  const width = useWidth()
+  const forceUpdate = useState()[1]
```
dctalbot, line 34:
```suggestion
```

web/src/app/wizard/WizardForm.js
```diff
@@ -12,199 +12,198 @@ import Typography from '@material-ui/core/Typography'
...
+  const classes = useStyles()
 
   handleSecondaryScheduleToggle = (e) => {
```
dctalbot, line 37:
```suggestion
  const handleSecondaryScheduleToggle = (e) => {
```

dctalbot (changes requested):
Thanks, looks like there are a few errors here. I've suggested some fixes!

voromahery:
> Thanks, looks like there are a few errors here. I've suggested some fixes!

Hello,

Could you give me the step on how to run this code locally so that I can track all the errors, please?

dctalbot (approved):

Forfold (approved):

PR #1796, opened 2021-07-26
ui/temporary schedules: adjust add button placement and header text

Forfold:

**Description:**
Users have reported that the temporary schedule form can be confusing- especially when they only need to add one user as the add and submit buttons appear to do the same thing.

It was suggested that the submit button be disabled until at least one shift has been added, but currently we can't allow this as an empty Temporary Schedule is an acceptable way to block off time where nobody is on-call (for whatever reason).

**Screenshots:**
Screen Shot 2021-07-26 at 12 30 49 PM

- Add shift fab removed from the middle
- Add shift text button added under shift fields
- Step 2 text adjusted

**Additional Info:**
The button renders as dark grey as that is what our secondary color in the theme is currently set to. I chose to use the secondary color to avoid adding tech debt for when we transition to use a proper theme.

### DIFF
--- a/web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
+++ b/web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
@@ -184,7 +184,7 @@ export default function TempSchedAddShiftsStep({
           </Grid>
           <Grid item>
             <DialogContentText className={classes.contentText}>
-              The schedule will be exactly as configured on this step for the
+              The schedule will be exactly as configured here for the
               entire duration (ignoring all assignments and overrides).
             </DialogContentText>
           </Grid>
--- a/web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
+++ b/web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
@@ -6,6 +6,7 @@ import {
   Typography,
   makeStyles,
 } from '@material-ui/core'
+import ArrowRightAltIcon from '@material-ui/icons/ArrowRightAlt'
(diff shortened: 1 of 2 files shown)


web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
```diff
@@ -191,8 +184,8 @@ export default function TempSchedAddShiftsStep({
           </Grid>
           <Grid item>
             <DialogContentText className={classes.contentText}>
-              The schedule will be exactly as configured here for the entire
-              duration (ignoring all rules/overrides).
+              The schedule will be exactly as configured on this step for the
```
dctalbot, line 187:
```suggestion
              The schedule will be exactly as configured here for the
```
When editing, there is only one "step", but I'm OK with either of these

dctalbot (approved):

web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
Forfold: Good point, I'll make sure to account for that

Forfold: We could probably just leave that wording out honestly: "The schedule will be exactly as configured for the duration blah blah.." seems to read just as well

Forfold: Eh, here works. I suppose people could get confused by what instance of the schedule they are configuring without the adverb

dctalbot (approved):

mastercactapus (approved):

PR #1803, opened 2021-07-28
config/general: add application name field

KatieMSB:

**Description:**
Adds Application Name in the config under the Admin Page and uses it as the title of the application.


app/inithttp.go
```diff
@@ -251,7 +251,7 @@ func (app *App) initHTTP(ctx context.Context) error {
 	mux.HandleFunc("/health", app.healthCheck)
 	mux.HandleFunc("/health/engine", app.engineStatus)
 
-	webH, err := web.NewHandler(app.cfg.UIURL, app.cfg.HTTPPrefix)
+	webH, err := web.NewHandler(app.cfg.UIURL, app.cfg.HTTPPrefix, app.ConfigStore.Config().General.ApplicationName)
```
mastercactapus, line 254: This needs to come from context on-demand as inithttp is only run once.

We'll need to alter how we handle index.html and the graphql playground so we pull this from the request context -- otherwise, config changes won't take effect until instances are manually restarted.

config/config.go
```diff
@@ -19,6 +19,7 @@ type Config struct {
 	fallbackURL string
 
 	General struct {
+		ApplicationName              string `public:"true" info:"Sets the name used throughout the application."`
```
mastercactapus, line 22: We should include a bit about how this is used
```suggestion
		ApplicationName              string `public:"true" info:"Override the name used in messaging and page titles. If unset 'GoAlert' will be used."`
```

smoketest/harness/harness.go
```diff
@@ -236,6 +236,8 @@ func (h *Harness) Start() {
 	h.t.Helper()
 
 	var cfg config.Config
+	cfg.General.ApplicationName = "GoAlert"
```
mastercactapus, line 239: Instead of setting this, we should add a `Config.ApplicationName()` method that handles the default value

mastercactapus (changes requested):

config/config.go
```diff
@@ -19,6 +19,7 @@ type Config struct {
 	fallbackURL string
 
 	General struct {
+		ApplicationName              string `public:"true" info:"Override the name used in messaging and page titles. If unset 'GoAlert' will be used."`
```
dctalbot, line 22:
```suggestion
		ApplicationName              string `public:"true" info:"The name used in messaging and page titles. Defaults to \"GoAlert\".`
```

config/config.go
```diff
@@ -307,6 +308,14 @@ func (cfg Config) ValidReferer(reqURL, ref string) bool {
 	return false
 }
 
+// ApplicationName will return the General.ApplicationName and set to 'GoAlert' if previously empty
```
dctalbot, line 311:
```suggestion
// ApplicationName will return the General.ApplicationName
```
We shouldn't leak implementation details here

web/src/app/main/components/ToolbarTitle.js
```diff
@@ -94,9 +95,12 @@ NameLoader.propTypes = {
 function ToolbarTitle() {
   const width = useWidth()
   const classes = useStyles()
+  const [applicationName] = useConfigValue('General.ApplicationName')
 
   const renderTitle = (title) => {
-    document.title = `GoAlert - ${title}`
+    document.title = `${
+      applicationName === '' ? 'GoAlert' : applicationName
+    } - ${title}`
```
dctalbot, line 103:
```suggestion
    document.title = `${applicationName} - ${title}`
```
Can we rely on the backend default? If so, better not to repeat logic in https://github.com/target/goalert/pull/1803/files#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R312

config/config.go
```diff
@@ -19,6 +19,7 @@ type Config struct {
 	fallbackURL string
 
 	General struct {
+		ApplicationName              string `public:"true" info:"The name used in messaging and page titles. Defaults to \"GoAlert\".`
```
dctalbot, line 22:
```suggestion
		ApplicationName              string `public:"true" info:"The name used in messaging and page titles. Defaults to \"GoAlert\"."`
```
Agh, missed the closing quote before.

dctalbot (dismissed):

dctalbot (dismissed):

config/config.go
```diff
@@ -307,6 +308,14 @@ func (cfg Config) ValidReferer(reqURL, ref string) bool {
 	return false
 }
 
+// ApplicationName will return the General.ApplicationName
+func (cfg Config) ApplicationName() string {
+	if cfg.General.ApplicationName == "" {
+		cfg.General.ApplicationName = "GoAlert"
```
mastercactapus, line 314: Just return instead of assigning to the struct
```suggestion
		return "GoAlert"
```

config/store.go
```diff
@@ -315,6 +315,9 @@ func (s *Store) updateConfigTx(ctx context.Context, tx *sql.Tx, fn func(Config)
 func (s *Store) Config() Config {
 	s.mx.RLock()
 	cfg := s.rawCfg
+
+	cfg.General.ApplicationName = cfg.ApplicationName()
```
mastercactapus, line 319: Don't save/overwrite the value, if the *saved* data is empty, just leave it that way.
```suggestion
```

graphql2/graphqlapp/app.go
```diff
@@ -85,31 +85,27 @@ type App struct {
...
-			log.Log(req.Context(), err)
-			http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
-		}
-	}))
+
+	err := permission.LimitCheckAny(req.Context())
+	if err != nil {
+		http.Error(w, http.StatusText(http.StatusUnauthorized), http.StatusUnauthorized)
+		return
+	}
```
mastercactapus, line 102: Do this check first

graphql2/graphqlapp/app.go
```diff
@@ -85,31 +85,27 @@ type App struct {
...
-func (a *App) PlayHandler() http.Handler {
+func (a *App) PlayHandler(w http.ResponseWriter, req *http.Request) {
 	var data struct {
-		Version     string
-		PackageName string
+		ApplicationName string
+		Version         string
+		PackageName     string
 	}
+	data.ApplicationName = a.ConfigStore.Config().ApplicationName()
```
mastercactapus, line 94: Use config from the current request context

graphql2/graphqlapp/slack.go
```diff
@@ -116,6 +112,7 @@ func (q *Query) GenerateSlackAppManifest(ctx context.Context) (string, error) {
 	}
 	var t bytes.Buffer
 	cfg := config.FromContext(ctx)
+	// sets ApplicationName and CallbackURL from config
```
mastercactapus, line 115: I think this can be left out, nothing is being set here, we're just passing config as the template context
```suggestion
```

smoketest/harness/harness.go
```diff
@@ -236,6 +236,7 @@ func (h *Harness) Start() {
 	h.t.Helper()
 
 	var cfg config.Config
+	cfg.ApplicationName()
```
mastercactapus, line 239: This is a `noop` ApplicationName shouldn't update the struct, and it's not a pointer receiver anyway so the change would be thrown out.
```suggestion
```

web/handler.go
```diff
@@ -67,19 +69,23 @@ func NewHandler(urlStr, prefix string) (http.Handler, error) {
...
+
+		var buf bytes.Buffer
+		err := indexTmpl.Execute(&buf, renderData{
+			ApplicationName: cfg.ApplicationName(),
+			Prefix:          prefix,
+			ExtraScripts:    extraScripts,
+		})
+		if err != nil {
+			log.Log(req.Context(), err)
+			http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
```
mastercactapus, line 83: This needs to return, we can only reply once to a single request.

```suggestion
		if errutil.HTTPError(ctx, w, err) {
			return
```

web/src/app/main/components/ToolbarTitle.js
```diff
@@ -94,9 +95,10 @@ NameLoader.propTypes = {
 function ToolbarTitle() {
   const width = useWidth()
   const classes = useStyles()
+  const [applicationName] = useConfigValue('General.ApplicationName')
```
mastercactapus, line 98: Fall back to the value from `index.html` if empty; maybe add it as a global var like public path.

mastercactapus (changes requested):

graphql2/graphqlapp/app.go
```diff
@@ -85,31 +85,30 @@ type App struct {
...
-		err := playTmpl.Execute(w, data)
-		if err != nil {
-			log.Log(req.Context(), err)
-			http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
-		}
-	}))
+
+	err = playTmpl.Execute(w, data)
+	if err != nil {
+		log.Log(req.Context(), err)
```
mastercactapus, line 109: use `errutil.HTTPError` for this and the one above so we have one place we're handling status codes & logging for HTTP errors

config/config.go
```diff
@@ -346,6 +355,10 @@ func (cfg Config) Validate() error {
 		)
 	}
 
+	if cfg.General.ApplicationName != "" {
+		err = validate.ASCII("General.ApplicationName", cfg.General.ApplicationName, 0, 32)
```
mastercactapus, line 359: Use `validate.Many(err, validate.ASCII(...` so we don't ignore/erase previous errors

web/src/app/main/components/ToolbarTitle.js
```diff
@@ -94,9 +95,10 @@ NameLoader.propTypes = {
...
 
   const renderTitle = (title) => {
-    document.title = `GoAlert - ${title}`
+    document.title = `${applicationName || document.title} - ${title}`
```
mastercactapus, line 101: This is going to be recursive if the config value is empty; we should fall back to the title on page-load rather than the current title. Maybe add `applicationName` next to `pathPrefix` in `index.html`

mastercactapus (changes requested):

mastercactapus (approved):

dctalbot (approved):

PR #1809, opened 2021-07-29
slack: generate manifest for a new slack app from admin settings ui

Forfold:

**Description:**
This PR adds a query option to generate a Slack manifest for external users to easily create their own GoAlert Slack Application for integration into GoAlert. Admins can copy the generated manifest and create the appropriate `.yml` file required by Slack to create an app.

**Screenshots:**
Screen Shot 2021-07-29 at 1 23 33 PM

Screen Shot 2021-07-29 at 1 23 11 PM

- Action button added at the bottom of Slack settings from Admin Config.
- Added word-wrap to the markdown container to prevent overflow.

- Added query `generateSlackManifest` to the GraphQL schema.
- Set primary action buttons in `CardActions` to `variant='contained'` and `color='primary'`. This prop was previously unused.


pnengchu (dismissed):

web/src/app/admin/AdminConfig.tsx
```diff
@@ -229,25 +231,26 @@ export default function AdminConfig(): JSX.Element {
...
-                        endAdornment: (
-                          <InputAdornment position='end'>
-                            <CopyText value={h.value} placement='left' />
-                          </InputAdornment>
-                        ),
-                      }}
-                      fullWidth
-                    />
-                  ))}
               </Form>
```
dctalbot, line 234: The `TextField` below has to be within a `Form`

web/src/app/admin/SlackActions.tsx
```diff
@@ -0,0 +1,118 @@
...
+            color='primary'
+            endIcon={<OpenInNewIcon />}
+            component='a'
+            href='https://api.slack.com/apps'
+            target='_blank'
+            rel='noopener noreferrer'
+            data-cy='configure-in-slack'
+          >
+            Configure in Slack
+          </Button>
```
dctalbot, line 113: We should use `component={AppLink}` here e.g. https://github.com/target/goalert/blob/master/web/src/app/alerts/CreateAlertDialog/CreateAlertDialog.js#L103

web/src/app/admin/SlackActions.tsx
```diff
@@ -0,0 +1,118 @@
...
+          {renderContent()}
+          <DialogContentText>
+            Learn more about manifests{' '}
+            <a
+              href='https://api.slack.com/reference/manifests'
+              target='_blank'
+              rel='noopener noreferrer'
+            >
+              here
+            </a>
```
dctalbot, line 94: We should use an `AppLink` with the `newTab` prop

dctalbot (changes requested):

web/src/app/admin/AdminConfig.tsx
Forfold: Why's that?

dctalbot: I thought for sure input elements would have to be within a form element but apparently that isn't strictly true according to the html spec

Forfold: It's also only being used to show hints via the copy text component, so there's nothing relevant to the Form to submit here

pnengchu (approved):

mastercactapus (approved):

dctalbot (approved):

PR #1818, opened 2021-08-02
ui/tech debt: update CopyText interface

Forfold:

**Description:**
The `CopyText` component has a `textOnly` prop. After using this component for some time, I have been under the assumption that this prop was used to display the component without the icon (i.e. only the title text). However, this is _not_ the case and has to do with if the user is copying a URL or not so it can be properly encoded from the `href` event target. This small discrepancy caused me to lose a good chunk of time, so I propose we refresh the API a tad, to avoid something like this in the future.

- `textOnly` -> `asURL`, logic flipped


Forfold:
`urlEncoded` could also be an option for the name

pnengchu (approved):

dctalbot (approved):

PR #1825, opened 2021-08-04
ui/admin: prevent default form submit on sms test form

dctalbot:

**Description:**
This PR fixes an issue where a `button` within a `form` submits the form by default

**Additional Info:**
This may be browser specific; reviewers may validate in Firefox and Chrome

### DIFF
--- a/web/src/app/admin/AdminSMSSend.tsx
+++ b/web/src/app/admin/AdminSMSSend.tsx
@@ -105,6 +105,7 @@ export default function AdminSMSSend(): JSX.Element {
                 send()
               }}
               loading={sendStatus.loading}
+              noSubmit
             />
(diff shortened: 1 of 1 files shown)


mastercactapus (approved):

Forfold:
I'm confused what the issue is here? I'm not noticing any off behavior in Firefox or Chrome when filling out the form/sending a test. Removing this from being a submit button would remove accessibility from the form

Forfold:
Moving the onClick to the form onSubmit would be the correct way to solve this I believe.

Forfold:
I think `noSubmit` still has its place, but it would be in the instance that you wouldn't want to submit with your mouse clicking a button as well as with a keyboard.

Forfold:
Created issue #1839 for tracking

PR #1833, opened 2021-08-05
twilio/sms: don't reply to unknown numbers

mastercactapus:

**Description:**
This PR makes updates to ensure reply SMS messages aren't sent to disabled or unknown numbers.

Additionally, destination type conversion code has been moved out of the independent type packages (`contactmethod` and `notificationchannel`) into the `notification` package. Since `DestType` depends on both and it made for confusing logic & import cycles.


notification/twilio/sms.go
```diff
@@ -244,6 +244,13 @@ func (s *SMS) ServeMessage(w http.ResponseWriter, req *http.Request) {
 		}
 
 		if isPassive {
+			valid, err := s.r.IsKnownDest(ctx, from)
+			if err != nil {
+				log.Log(ctx, fmt.Errorf("check if known SMS number: %w", err))
+			} else if !valid {
+				// don't respond if the number is not known
+				return
+			}
```
dctalbot, line 253: `if err && !valid` won't return, is that intended?

mastercactapus: Yes, but we can discuss if that should be the case. In the case of a DB/network issue, it will fall back to sending the reply and only blocks it on a sucessful negative check.

dctalbot: Makes sense, thanks for confirming. I think this behavior is best

dctalbot (approved):

dctalbot (approved):

m17ch (approved):
lgtm 👍

PR #1837, opened 2021-08-10
ui/schedules: remove timezone toggle from temporary schedules wizard

dctalbot:

**Description:**
This PR removes the timezone toggle from the temporary schedules forms in favor of having users always configure dates/times in the timezone of the corresponding Schedule. 

If a user's system time is in the same zone as the relevant Schedule, no time zone signifiers will be displayed. However, if they differ, form fields' hints will display the value in local time along with the local zone abbreviation e.g. `11:00 PM CDT`.

**Screenshots:**
When Schedule in local zone:
Screen Shot 2021-08-10 at 11 32 48 AM
Screen Shot 2021-08-10 at 11 33 08 AM

When Schedule in alternate zone:
Screen Shot 2021-08-10 at 11 33 31 AM
Screen Shot 2021-08-10 at 1 34 45 PM

**Additional Info:**
This PR also removes a `stepText` prop from a couple of the stepper components.


pnengchu (approved):

m17ch (approved):
lgtm 👍

PR #1843, opened 2021-08-11
ui/hooks:SearchUpdater

voromahery:

**Description:**
Converts this file into hooks 

Part of #923


web/src/app/util/AppBarSearchContainer.js
```diff
@@ -76,34 +76,31 @@ export class SearchProvider extends React.PureComponent {
...
-  _mounted = false
-
-  componentDidMount() {
-    this._mounted = true
-    this.props.trackMount(true)
-    this.props.setActions(this.props.children)
-  }
+function SearchUpdater(props) {
+  const { trackMount, setActions, children } = props
+  let _mounted = false
```
dctalbot, line 81: This should use `useRef`

dctalbot (changes requested):
Thanks for the PR! Please see docs regarding how to convert class instance variables https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables

web/src/app/util/AppBarSearchContainer.js
```diff
@@ -76,34 +76,26 @@ export class SearchProvider extends React.PureComponent {
...
-  }
+function SearchUpdater(props) {
+  const { trackMount, setActions, children } = props
+  const mounted = useRef(false)
 
-  _mounted = false
+  useEffect(() => {
+    mounted.current = true
+    trackMount(true)
+    setActions(children)
```
dctalbot, line 86:
```suggestion
    setActions(children)
    return () => {
      mounted.current = false
      trackMount(false)
      setActions(null)
    }
```
We must convert `componentWillUnmount` by returning a cleanup function

web/src/app/util/AppBarSearchContainer.js
```diff
@@ -76,34 +76,26 @@ export class SearchProvider extends React.PureComponent {
...
+  }, [])
 
-  componentDidMount() {
-    this._mounted = true
-    this.props.trackMount(true)
-    this.props.setActions(this.props.children)
+  if (mounted) {
```
dctalbot, line 89:
```suggestion
  if (mounted.current) {
```

dctalbot (changes requested):

dctalbot (approved):
Nice work, thanks for the PR!

PR #1848, opened 2021-08-11
engine/statusmanager: fix performance and behavior issues

mastercactapus:
**Description:**
This PR fixes a number of issues/regressions with the status update code:
1. performance issue when querying for updates
2. sending messages after CM disabled, or status updates turned off
3. sending updates for events that are very old
4. updates went to users who initiated the action

The existing smoketest was updated to validate 4. A new smoketest was added to validate 2 & 3.
For the first issue, it's harder to reproduce but the query was rewritten to use subqueries to avoid scanning
the alerts table during the join resulting in 100x performance improvement in certain workloads.

### DIFF
--- a/engine/statusupdatemanager/update.go
+++ b/engine/statusupdatemanager/update.go
@@ -85,12 +85,16 @@ func (db *DB) update(ctx context.Context) error {
 
 		var logUserID sql.NullString
 		err = tx.StmtContext(ctx, db.latestLogEntry).QueryRowContext(ctx, alertID, event).Scan(&logID, &logUserID)
+		if errors.Is(err, sql.ErrNoRows) {
+			err = nil
+			logID = 0
+		}
 		if err != nil {
 			return fmt.Errorf("lookup latest log entry of '%s' for alert #%d: %w", event, alertID, err)
 		}
 
-		// Only insert message if the user is not the same as the log event user.
-		if !userID.Valid || userID.String != logUserID.String {
+		// Only insert message if the user is not the same as the log event user and we have a recent log entry.
+		if logID > 0 && (!userID.Valid || userID.String != logUserID.String) {
 			_, err = tx.StmtContext(ctx, db.insertMessage).ExecContext(ctx, uuid.New(), chanID, cmID, userID, alertID, logID)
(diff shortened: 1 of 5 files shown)


smoketest/statusupdatescancel_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+// TestStatusUpdatesCancel checks that status updates are unsubscribed when updates, or the contact method, are disabled.
+func TestStatusUpdatesCancel(t *testing.T) {
+	t.Parallel()
+
+	sql := `
+	insert into users (id, name, email, role) 
+	values 
+		({{uuid "user"}}, 'bob', 'joe@test.com', 'admin');
+	insert into user_contact_methods (id, user_id, name, type, value)
```
dctalbot, line 21: Can we be explicit about setting the value of `disabled`? Improves readability + robustness of test

engine/statusupdatemanager/update.go
```diff
@@ -83,21 +97,29 @@ func (db *DB) update(ctx context.Context) error {
...
+		}
 		if err != nil {
 			return fmt.Errorf("lookup latest log entry of '%s' for alert #%d: %w", event, alertID, err)
 		}
 
-		_, err = tx.StmtContext(ctx, db.insertMessage).ExecContext(ctx, uuid.New(), chanID, cmID, userID, alertID, logID)
-		if err != nil {
-			return fmt.Errorf("insert status update message for id=%d: %w", id, err)
+		// Only insert message if the user is not the same as the log event user and we have a recent log entry.
+		if logID > 0 && (!userID.Valid || userID.String != logUserID.String) {
```
dctalbot, line 111: At first glance, it seems odd that we would insert messages for invalid user ID's. Can you confirm that's what we want here?

dctalbot (changes requested):

engine/statusupdatemanager/update.go
mastercactapus: It is, we won't have a valid user ID for a notification channel

dctalbot (approved):

arurao (approved):

PR #1849, opened 2021-08-11
ui: use real from number in dialogs

mastercactapus:
**Description:**
This PR updates the Test and Verify dialogs, as well as the admin Send-SMS tool, to use the From value returned by Twilio when sending messages.

This is in preparation for SID support when the FromNumber config value may be a SID and multiple numbers could be used when delivering messages.

A new field, `formattedSrcValue` was added to the `NotificationState` type in the GraphQL schema.


m17ch (dismissed):

arurao (dismissed):
LGTM

arurao (approved):

m17ch (approved):

PR #1859, opened 2021-08-13
ui/hooks: URLErrorDialog

Voninkazo:

**Description:**
Converting a class component to a functional component.

Part of #923

### DIFF
--- a/web/src/app/main/URLErrorDialog.js
+++ b/web/src/app/main/URLErrorDialog.js
@@ -3,8 +3,13 @@ import { Dialog, DialogContent, DialogActions } from '@material-ui/core'
 import DialogTitleWrapper from '../dialogs/components/DialogTitleWrapper'
 import DialogContentError from '../dialogs/components/DialogContentError'
 import LoadingButton from '../loading/components/LoadingButton'
+import { useURLParam } from '../actions'
 
-export default function URLErrorDialog({ errorMessage, errorTitle, onClose }) {
+export default function URLErrorDialog({ onClose }) {
+  const errorMessage = useURLParam('errorMessage')
+  const errorTitle = useURLParam('errorTitle')
+
+  const open = Boolean(errorMessage) || Boolean(errorTitle)
   const open = Boolean(errorMessage) || Boolean(errorTitle)
 
   return (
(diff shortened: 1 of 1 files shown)


web/src/app/main/URLErrorDialog.js
```diff
@@ -1,58 +1,37 @@
...
 
-@connect(
-  (state) => ({
-    errorMessage: urlParamSelector(state)('errorMessage'),
-    errorTitle: urlParamSelector(state)('errorTitle'),
-  }),
-  (dispatch) => ({
-    resetError: () => dispatch(resetURLParams('errorMessage', 'errorTitle')),
-  }),
-)
```
dctalbot, line 18: These should be converted to use `useURLParam` and `useResetURLParams`. Please see https://github.com/target/goalert/blob/d4db3547ae6e41b24aa9827afd447e2695d6a893/web/src/app/alerts/components/AlertsListFilter.js#L48-L51 for example

web/src/app/main/URLErrorDialog.js
```diff
@@ -1,58 +1,37 @@
...
-export default class URLErrorDialog extends React.Component {
-  handleDialogClose = () => {
-    this.props.resetError()
-  }
+export default function URLErrorDialog({ errorMessage, errorTitle, onClose }) {
```
dctalbot, line 7:
```suggestion
export default function URLErrorDialog({ onClose }) {
```
Once the URLParam hooks are implemented, these will no longer be injected as props

dctalbot (changes requested):

web/src/app/main/URLErrorDialog.js
Voninkazo: Thank you for the useful feedback. But I wonder if we sill need the 
`handleDialogClose = () => {
    this.props.resetError()
  }`
function anymore because it's used nowhere in the body component. And the same with the dispatch function `resetError` whis is being called inside that function

dctalbot: Interesting question! It looks like you uncovered a bug. We should replace `this.onClose` with  `handleDialogClose`  since `this.onClose` is not defined. We'll have to make that fix twice - once for the `Dialog` `onClose` prop and once for `LoadingButton` `onClick` prop

dctalbot: Hi @Voninkazo! Just checking in to see if you had any follow-up questions for this

PR #1860, opened 2021-08-16
search: add indexes and use word-search

mastercactapus:
**Description:**
This PR updates search behavior to use a word-based approach instead of a substring match. Doing so allows index-based searching which is a lot more performant (especially when searching through alert summaries).

Additionally, the alert query was broken up into two steps, one to get a list of services that match the query, and a second purely for the alerts with matching queries or service names. This substantially improved query time and memory use in Postgres.


m17ch (dismissed):
lgtm

m17ch (dismissed):

m17ch (approved):

arurao (approved):
LGTM

PR #1862, opened 2021-08-16
slack: 2-way ack/escalate/close on alert msg

Forfold:

**Description:**

Screen Shot 2021-08-16 at 2 37 47 PM

**How to test:**
You'll need to leverage our Sendit tool locally to test this feature:
1. `$ make regendb`
2. `$ make start`
3. Open localhost:3030 and login as an admin
4. Configure sendit if not already
5. Create a Slackbot at api.slack.com (or use one of the GoAlert Sandbox ones from our testing workspace)
    - Under "Basic Information", copy the **Client ID**, **Client Secret**, and **Signing Secret** to your GoAlert admin config
    - Under "Interactivity & Shortcuts", enable and set your request URL to: `https://sendit.dev.target.goalert.me/YOUR_HTTP_PREFIX/api/v2/slack/action`
    - Under "Oath & Permissions", copy the **Bot User OAuth Token** to your GoAlert admin config
        - Set a Redirect URL to `https://sendit.dev.target.goalert.me/YOUR_HTTP_PREFIX/api/v2/slack/auth`
6. Save GoAlert admin config
7. Install and invite your Slack bot to the channel you would like to test in
8. Back in GoAlert (still on localhost), go to Escalation Policies and create an EP where **repeat is set to 0**.
9. Create a step on the EP with your slack channel added
10. Create a service, selecting our new EP
11. Stop GoAlert and restart with Sendit
12. Add `auth_subjects` entry for your user.
  - `provider_id`: `"slack:WORKSPACE_ID"` (can be copied from URL when opening workspace within a browser: https://app.slack.com/client/WORKSPACE_ID)
  - `subject_id`: `"SLACK_USER_ID"` (can be copied from slack profile -> more ... -> copy member ID)
13. Begin testing by creating alerts for the service you just created

**Additional Info:**
- OAth tokens do not expire, so we determine if a user is valid if they have an entry within `auth_subjects`. Entries within this table are considered fully authorized/linked with their respective GoAlert user.
- Need to think about an icon for the bot, and if/what colors and emojis we want to use for the buttons.


alert/log/store.go
```diff
@@ -260,6 +265,20 @@ func txWrap(ctx context.Context, tx *sql.Tx, stmt *sql.Stmt) *sql.Stmt {
...
 	return tx.StmtContext(ctx, stmt)
 }
+
+func (db *DB) FindNCByValue(ctx context.Context, tx *sql.Tx, value string) (string, string, error) {
+	err := permission.LimitCheckAny(ctx, permission.All)
+	if err != nil {
+		return "", "", err
+	}
+	var channelID, name string
+	err = txWrap(ctx, tx, db.lookupNCIdNameByValue).QueryRowContext(ctx, value).Scan(&channelID, &name)
```
mastercactapus, line 275: Value isn't validated before being passed to the DB

alert/log/store.go
```diff
@@ -293,9 +312,9 @@ func (db *DB) logAny(ctx context.Context, tx *sql.Tx, insertStmt *sql.Stmt, id i
 			r.subject._type = SubjectTypeChannel
 			var ncType notificationchannel.Type
 			var name string
-			err = txWrap(ctx, tx, db.lookupNCTypeName).QueryRowContext(ctx, src.ID).Scan(&ncType, &name)
+			err = db.lookupNCTypeName.QueryRowContext(ctx, src.ID).Scan(&ncType, &name)
```
mastercactapus, line 315: needs to be wrapped within the transaction

app/inithttp.go
```diff
@@ -165,6 +166,13 @@ func (app *App) initHTTP(ctx context.Context) error {
 		UserStore:           app.UserStore,
 	})
 
+	slack := slack.NewHandler(slack.Config{
+		AlertStore:    app.AlertStore,
+		AlertLogStore: app.AlertLogStore,
+		UserStore:     *app.UserStore,
+		AuthHandler:   *app.AuthHandler,
+	})
```
mastercactapus, line 174: Slack shouldn't be accessing these stores, it should handle signature validation itself and pass info to the engine for responses where things can be authenticated.

app/initslack.go
```diff
@@ -10,8 +10,10 @@ import (
 func (app *App) initSlack(ctx context.Context) error {
 	var err error
 	app.slackChan, err = slack.NewChannelSender(ctx, slack.Config{
-		BaseURL:   app.cfg.SlackBaseURL,
-		UserStore: app.UserStore,
+		BaseURL:       app.cfg.SlackBaseURL,
+		AlertStore:    app.AlertStore,
+		AlertLogStore: app.AlertLogStore,
+		UserStore:     *app.UserStore,
```
mastercactapus, line 16: Same as above, the Slack provider shouldn't be talking to the DB directly

auth/handler.go
```diff
@@ -182,6 +182,18 @@ func (h *Handler) EndAllUserSessionsTx(ctx context.Context, tx *sql.Tx) error {
...
+	var userID string
+	err := h.userLookup.QueryRowContext(ctx, providerID, subjectID).Scan(&userID)
+	if errors.Is(err, sql.ErrNoRows) {
+		err = nil
+	}
+	if err != nil {
+		return "", err
+	}
+	return userID, nil
+}
```
mastercactapus, line 195: This has no input validation and should be handled in the engine in a similar way to how we use a phone number to map to a user

notification/slack/actions.go
```diff
@@ -0,0 +1,179 @@
...
+// Payload represents the relevant payload information sent from Slack
+type Payload struct {
+	ResponseURL string `json:"response_url"`
+	Actions     []Action
+	Channel     slack.Channel
+}
+
+// Action represents the information given from an action event within Slack
+// e.g. clicking to acknowledge an alert from slack
+type Action struct {
```
mastercactapus, line 33: Is it necessary to export these as public?

notification/slack/actions.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// NewHandler creates a new Handler, registering API routes using chi.
+func NewHandler(c Config) *Handler {
+	return &Handler{c: c}
+}
+
+func httpErr(w http.ResponseWriter, err error) error {
+	http.Error(w, http.StatusText(http.StatusMethodNotAllowed), http.StatusMethodNotAllowed)
+	panic(err)
```
mastercactapus, line 47: We should not be panicking/crashing the application here

notification/slack/actions.go
```diff
@@ -0,0 +1,179 @@
...
+	if err != nil {
+		return httpErr(w, err)
+	}
+	req.Body.Close()
+	req.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+
+	_, err = sv.Write(body)
+	if err != nil {
+		return httpErr(w, err)
+	}
```
mastercactapus, line 75: Use `io.TeeReader` to update `sv` as we read the body. Also we can remove the `req.Body.Close` since it's redundant to the `defer` above.

```suggestion
	body, err := ioutil.ReadAll(io.TeeReader(req.Body, sv))
	if err != nil {
		return httpErr(w, err)
	}
	req.Body = ioutil.NopCloser(bytes.NewBuffer(body))
```

notification/slack/actions.go
```diff
@@ -0,0 +1,179 @@
...
+		return err
+	}
+	var actionErr error
+	switch action.ActionID {
+	case "ack":
+		actionErr = h.c.AlertStore.UpdateStatus(ctx, alertID, alert.StatusActive)
+	case "esc":
+		actionErr = h.c.AlertStore.Escalate(ctx, alertID)
+	case "close":
+		actionErr = h.c.AlertStore.UpdateStatus(ctx, alertID, alert.StatusClosed)
```
mastercactapus, line 106: This needs to go to our receiver and not the DB directly, we'll probably need to update it's signature to contain a status update return value and add a version that accepts a subject id

notification/slack/actions.go
```diff
@@ -0,0 +1,179 @@
...
+			if err != nil {
+				fmt.Println("error finding user")
+				httpErr(w, err)
+			}
+
+			// send Unauthorized message if user is not linked
+			if userID == "" {
+				_, err := api.PostEphemeralContext(ctx, payload.Channel.ID, payload.User.ID, needsAuthMsgOpt())
+				if err != nil {
+					fmt.Println("error posting ephemeral auth msg")
```
mastercactapus, line 163: use the log package instead of `fmt` for reporting errors. Using `errutil.HTTPError` will handle this

notification/slack/channel.go
```diff
@@ -312,6 +313,9 @@ func (s *ChannelSender) loadChannels(ctx context.Context) ([]Channel, error) {
 
 func (s *ChannelSender) Send(ctx context.Context, msg notification.Message) (*notification.SentMessage, error) {
 	cfg := config.FromContext(ctx)
+	var api = slack.New(cfg.Slack.AccessToken)
+
+	var a alert.Alert
```
mastercactapus, line 318: Define this only where needed

notification/slack/channel.go
```diff
@@ -327,30 +331,52 @@ func (s *ChannelSender) Send(ctx context.Context, msg notification.Message) (*no
...
+		_, ts, _, err := api.UpdateMessageContext(ctx, msg.Destination().Value, t.Dest.Value, msgOpt...)
+		if err != nil {
+			return nil, err
+		}
+		vals.Set("thread_ts", t.OriginalStatus.ProviderMessageID.ExternalID)
+		// todo, post to thread
+		return &notification.SentMessage{
+			ExternalID: ts,
+			State:      notification.StateDelivered,
+			SrcValue:   "",
```
mastercactapus, line 365: SrcValue should be the bot user ID

notification/slack/channel.go
```diff
@@ -327,30 +331,52 @@ func (s *ChannelSender) Send(ctx context.Context, msg notification.Message) (*no
...
+		}, nil
 	case notification.AlertBundle:
+		vals.Set("token", cfg.Slack.AccessToken)
```
mastercactapus, line 368: Why is this being set here and not below?

notification/slack/channel.go
```diff
@@ -327,30 +331,52 @@ func (s *ChannelSender) Send(ctx context.Context, msg notification.Message) (*no
...
-	default:
-		return nil, errors.Errorf("unsupported message type: %T", t)
+		return post(s, vals)
 	}
-	vals.Set("token", cfg.Slack.AccessToken)
 
+	return nil, errors.Errorf("unsupported message type: %T", msg.Type())
+}
+
+func post(s *ChannelSender, vals url.Values) (*notification.SentMessage, error) {
```
mastercactapus, line 379: Should we just use the api library for this? we should also call it `chatPostMessage` or something more clear

notification/slack/messageoptions.go
```diff
@@ -0,0 +1,95 @@
...
+
+	alertID := strconv.Itoa(a.ID)
+
+	if a.Status == alert.StatusTriggered {
+		actions = slack.NewActionBlock("", ackButton(alertID), escButton(alertID), closeButton(alertID), openLinkButton(url))
+	} else if a.Status == alert.StatusActive {
+		actions = slack.NewActionBlock("", escButton(alertID), closeButton(alertID), openLinkButton(url))
+	} else {
+		actions = slack.NewActionBlock("", openLinkButton(url))
+	}
```
mastercactapus, line 76: Use a switch here to indicate selection instead of ordered operations.

notification/slack/messageoptions.go
```diff
@@ -0,0 +1,95 @@
...
+// 	lastStatusText := slack.NewTextBlockObject("plain_text", lastStatus, true, true)
+// 	return slack.NewContextBlock("", []slack.MixedElement{lastStatusText}...)
+// }
+
+func needsAuthMsgOpt() slack.MsgOption {
+	msg := slack.NewTextBlockObject("plain_text", "Unauthorized. Please link your GoAlert account to continue", false, false)
+	return slack.MsgOptionBlocks(slack.NewSectionBlock(msg, nil, nil))
+}
+
+func CraftAlertMessage(a alert.Alert, url, responseURL string) []slack.MsgOption {
```
mastercactapus, line 64: This needs the callbackID to be referenced by the response

notification/slack/messageoptions.go
```diff
@@ -0,0 +1,95 @@
...
+	txt := slack.NewTextBlockObject("plain_text", "Close :ballot_box_with_check:", true, false)
+	return slack.NewButtonBlockElement("close", alertID, txt)
+}
+
+func openLinkButton(url string) *slack.ButtonBlockElement {
+	txt := slack.NewTextBlockObject("plain_text", "Open in GoAlert :link:", true, false)
+	s := slack.NewButtonBlockElement("openLink", "", txt)
+	s.URL = url
+	return s
+}
```
mastercactapus, line 52: This seems really noisy, can we make it a bit more generic instead of a bunch of 2 line functions?
Here's an example of what I'm thinking:

```go
// one helper func

type buttonType int

const (
	buttonTypeAck buttonType = iota
	buttonTypeEsc
	buttonTypeClose
	buttonTypeOpenLink
)

func makeButtons(callbackID, url string) func(buttonType) *slack.ButtonBlockElement {
	return func(typ buttonType) *slack.ButtonBlockElement {
		var action, text, setURL string
		switch typ {
		case buttonTypeAck:
			action = "ack"
			text = "Acknowledge :eyes:"
		case buttonTypeEsc:
			action = "esc"
			text = "Escalate :arrow_up:"
		case buttonTypeClose:
			action = "close"
			text = "Close :ballot_box_with_check:"
		case buttonTypeOpenLink:
			action = "openLink"
			text = "Open in GoAlert :link:"
			setURL = url
		default:
			panic("button type not supported")
		}

		el := slack.NewButtonBlockElement(
			action,
			callbackID,
			slack.NewTextBlockObject("plain_text", text, true, false),
		)
		if setURL != "" {
			el.URL = setURL
		}
		return el
	}
}

// and then inside the render func
	button := makeButtons(callbackID, url)
	switch a.Status {
	case alert.StatusTriggered:
		actions = slack.NewActionBlock("",
			button(buttonTypeAck),
			button(buttonTypeEsc),
			button(buttonTypeClose),
			button(buttonTypeOpenLink),
		)
	case alert.StatusActive:
		actions = slack.NewActionBlock("",
			button(buttonTypeEsc),
			button(buttonTypeClose),
			button(buttonTypeOpenLink),
		)
	case alert.StatusClosed:
		actions = slack.NewActionBlock("",
			button(buttonTypeOpenLink),
		)
	}
```
---
Two switches with only the relevant diffs -- helper method just switches the text and actions of the buttons, and the render method just switches which buttons we're including in the message. A type enum makes it clear what button types we're supporting/using.

user/store.go
```diff
@@ -571,6 +571,20 @@ func (s *Store) FindAllAuthSubjectsForUser(ctx context.Context, userID string) (
...
+			result = sub
+		}
+		return nil
+	}, userID)
+	if err != nil {
+		return nil, err
+	}
+	return &result, nil
+}
+
```
mastercactapus, line 587: I think this is redundant, `AuthSubjectsFunc` already supports filtering by providerID and user

mastercactapus (changes requested):

notification/slack/channel.go
Forfold: this is still a todo

notification/slack/messageoptions.go
Forfold: yeah I can definitely give that a shot

user/store.go
Forfold: I wanted to have a helper function that only returned 1 user so it'd be easier to use within other functions without needing to loop through the result

app/inithttp.go
Forfold: where in the engine would the logic be handled from the request?

notification/slack/actions.go
Forfold: yep 👍  i have these in for local testing atm

notification/slack/channel.go
Forfold: it doesn't seem like i have access to the bot ID from the result using the library. what is this used for? iirc we didn't need this value for anything on the first pass.

notification/slack/channel.go
Forfold: i left this here until i wrap up the rest of the logic so all send routes still work. but yeah i can essentially do that to send a text Slack message for bundling and on-call user notifications

user/store.go
Forfold: changing this to return a User object from a given subjectID

notification/slack/actions.go
Forfold: looks like these aren't needed anymore- removed

notification/slack/channel.go
Forfold: completed- all actions should use the slack-go lib now

alert/log/store.go
```diff
@@ -22,6 +22,7 @@ import (
 type Store interface {
 	FindOne(ctx context.Context, logID int) (*Entry, error)
 	FindAll(ctx context.Context, alertID int) ([]Entry, error)
+	FindNCByValue(ctx context.Context, tx *sql.Tx, value string) (string, string, error)
```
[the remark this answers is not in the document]
Forfold: removed- shouldn't be needed anymore

notification/slack/actions.go
```diff
@@ -0,0 +1,129 @@
...
+			actionType = notification.ResultAcknowledge
+		case "esc":
+			actionType = notification.ResultEscalate
+		case "close":
+			actionType = notification.ResultResolve
+		default:
+			errutil.HTTPError(ctx, w, errors.New("unknown action"))
+			return
+		}
+		a, err := h.r.ReceiveFor(ctx, "callbackID", "slack:"+payload.Team.ID, payload.User.ID, actionType)
```
arurao, line 116: Should one of these statements be removed?

arurao (changes requested):

Forfold:
Being implemented elsewhere...

PR #1864, opened 2021-08-17
users: add user favorite support

### LINKED ISSUE #1086: Add Favorites button to Users
Today, I have to search for the users I work with regularly and cannot group or favorite them

I would like to be able to favorite the users that I work with regularly so I can quickly find them if needed

not sure what else I would do

**Additional context:**
this would work ideally in the same way as favorites for rotations and schedules!

pnengchu:

**Description:**
End Users may now set favorites on other users for a custom experience. This feature is replicated from Escalation Policies and similar functionalities that involve setting user favorites from the UI. 

Fixes #1086

**Screenshots:**
Screen Shot 2021-08-17 at 2 35 07 PM
Screen Shot 2021-08-17 at 2 34 53 PM


spencerpauly (dismissed):

migrate/migrations/20210817132557-add-user-favorite.sql
```diff
@@ -0,0 +1,8 @@
+-- +migrate Up
+ALTER TABLE user_favorites
+    ADD COLUMN tgt_user_id UUID REFERENCES users(id) ON DELETE CASCADE,
+    ADD CONSTRAINT user_favorites_user_id_tgt_user_id_key UNIQUE(user_id, tgt_user_id);
+-- +migrate Down
+ALTER TABLE user_favorites
+    DROP CONSTRAINT user_favories_user_id_tgt_user_id_key, 
+    DROP COLUMN tgt_user_id;
```
mastercactapus, line 8: nit: end with newline
```suggestion
    DROP COLUMN tgt_user_id;

```

web/src/cypress/support/util.ts
```diff
@@ -162,3 +162,16 @@ export function testScreen(
     describe(screenName(), () => fn(screen()))
   })
 }
+
+export function setFavorites(ids: Array<string>): Promise<unknown> {
```
mastercactapus, line 166: Since this is user-specific it also shouldn't be in `util` but in the `users` support file with a corresponding name like `setFavoriteUser`.

Lastly, just have it handle a single user, return Cypress.Chainable and call it from `.filter(...).forEach( setFavoriteUser )`

Cypress will already handle chaining the events and does not actually support/use promises properly. Also treat it as a private helper and don't export it since we're really not adding support outside of `createManyUsers`

mastercactapus (changes requested):

migrate/migrations/20210817132557-add-user-favorite.sql
pnengchu: addressed.

web/src/cypress/support/util.ts
pnengchu: addressed.

spencerpauly (dismissed):

migrate/migrations/20210817132557-add-user-favorite.sql
```diff
@@ -0,0 +1,9 @@
+-- +migrate Up
+ALTER TABLE user_favorites
+    ADD COLUMN tgt_user_id UUID REFERENCES users(id) ON DELETE CASCADE,
+    ADD CONSTRAINT user_favorites_user_id_tgt_user_id_key UNIQUE(user_id, tgt_user_id);
+-- +migrate Down
+ALTER TABLE user_favorites
+    DROP CONSTRAINT user_favories_user_id_tgt_user_id_key,
```
KatieMSB, line 7:
```suggestion
    DROP CONSTRAINT user_favorites_user_id_tgt_user_id_key, 
```

KatieMSB (changes requested):

migrate/migrations/20210817132557-add-user-favorite.sql
pnengchu: addressed

spencerpauly (approved):

user/search.go
```diff
@@ -51,20 +64,41 @@ var searchTemplate = template.Must(template.New("search").Funcs(search.Helpers()
...
+	return "lower(usr.name)"
+}
+
+func (opts renderData) SearchStr() string {
+	if opts.Search == "" {
+		return ""
+	}
+
+	return "%" + search.Escape(opts.Search) + "%"
+}
```
mastercactapus, line 100: This is no longer used

mastercactapus (changes requested):

user/search.go
pnengchu: addressed

user/store.go
```diff
@@ -104,9 +106,11 @@ func NewStore(ctx context.Context, db *sql.DB) (*Store, error) {
...
-				id, name, email, avatar_url, role, alert_status_log_contact_method_id
-			FROM users
-			WHERE id = $1
+				u.id, u.name, u.email, u.avatar_url, u.role, u.alert_status_log_contact_method_id, fav is distinct from null
+			FROM users u
+			LEFT JOIN user_favorites fav ON
+				fav.tgt_user_id = u.id AND fav.user_id = $2
+			WHERE u.id = $1
 		`),
 		findOneForUpdate: p.P(`
```
KatieMSB, line 115: Since `findOneForUpdate` doesn't return the user favorite status, the row returned from the query should be scanned into a user object directly with `row.Scan()` instead of `u.scanFrom()`.

user/store.go
```diff
@@ -104,9 +106,11 @@ func NewStore(ctx context.Context, db *sql.DB) (*Store, error) {
 
 		findOne: p.P(`
 			SELECT
-				id, name, email, avatar_url, role, alert_status_log_contact_method_id
-			FROM users
-			WHERE id = $1
+				u.id, u.name, u.email, u.avatar_url, u.role, u.alert_status_log_contact_method_id, fav is distinct from null
+			FROM users u
+			LEFT JOIN user_favorites fav ON
+				fav.tgt_user_id = u.id AND fav.user_id = $2
```
KatieMSB, line 112: `userID` will need to be passed when the `findOne` statement is queried, but might need separate returns where `findOne` and `findOneForUpdate` are queried since `findOneForUpdate` doesn't use `userID`.

KatieMSB (changes requested):

user/store.go
pnengchu: Addressed, fixed bug by implementing `row.Scan()` if `forUpdate` returns true. Previous behavior observed prompted an error when editing a single user role to Admin.

user/store.go
pnengchu: Addressed, fixed bug by passing additional `userID` into `findOne`. Previous behavior observed was a failure to render _all_ user profiles with individual avatars.

mastercactapus (dismissed):

PR #1881, opened 2021-09-01
notification: use application name in messaging

mastercactapus:
**Description:**
This PR continues the work from #1803 and uses the application name in notification providers.


KatieMSB (dismissed):

notification/email/sender.go
```diff
@@ -50,19 +53,19 @@ func (s *Sender) Send(ctx context.Context, msg notification.Message) (*notificat
 	var subject string
 	switch m := msg.(type) {
 	case notification.Test:
-		subject = "GoAlert: Test Message"
+		subject = "Test Message"
 		e.Body.Title = "Test Message"
 		e.Body.Intros = []string{"This is a test message from GoAlert."}
```
dctalbot, line 58:
```suggestion
		e.Body.Intros = []string{"This is a test message from " + cfg.ApplicationName() + "."}
```
Should these be substituted?

dctalbot (changes requested):

notification/email/sender.go
mastercactapus: Hmm. maybe it should just be `This is a test message.` since the from is already defined elsewhere with the name.

mastercactapus: We're already setting the product name too; so it's redundant

dctalbot: We set it as the "from" name, right? If so, either works for me. I don't find it overly redundant.

KatieMSB (approved):

dctalbot (approved):

PR #1891, opened 2021-09-08
ui/schedules: Temp sched no coverage on page checkbox

spencerpauly:

**Description:**

This is a different solution from the one in this draft PR: https://github.com/target/goalert/pull/1870

**Out of Scope:**

**Screenshots:**

Screen Shot 2021-09-08 at 11 31 02 AM

User now must select "allow no coverage" checkbox if they'd like to be able to create a temp schedule with coverage gaps in it.

**Additional Info:**
Previous solution done in this draft PR: https://github.com/target/goalert/pull/1870


web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
```diff
@@ -211,21 +234,48 @@ export default function TempSchedAddShiftsStep({
...
+                    name='isAwareOfNoCoverage'
+                  />
+                }
+              />
+            </Alert>
+          )}
         </Grid>
+        {/* <Grid item>
+          
+        </Grid> */}
```
dctalbot, line 278:
```suggestion
```

web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
```diff
@@ -46,6 +60,11 @@ type AddShiftsStepProps = {
 
   scheduleID: string
   edit?: boolean
+
+  isAllowingNoCoverage: boolean
+  setIsAllowingNoCoverage: (isAllowing: boolean) => void
+  isShowingNoCoverageWarning: boolean
+  hasNoCoverageGaps: boolean
```
dctalbot, line 67: I think we've got one too many props than is necessary here. We should update interface to something like

```
  showCoverageGapsWarning?: boolean
  coverageGapsAllowed?: boolean
  setCoverageGapsAllowed: (val: boolean) => void
```

and pass through `showCoverageGapsWarning={isShowingNoCoverageWarning && hasNoCoverageGaps}` from the parent

web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
```diff
@@ -211,21 +234,48 @@ export default function TempSchedAddShiftsStep({
...
+              </FormHelperText>
+              <FormControlLabel
+                label='Allow gaps in coverage'
+                labelPlacement='end'
+                control={
+                  <Checkbox
+                    data-cy='no-coverage-checkbox'
+                    checked={isAllowingNoCoverage}
+                    onChange={(e) => setIsAllowingNoCoverage(e.target.checked)}
+                    name='isAwareOfNoCoverage'
```
dctalbot, line 269:
```suggestion
                    name='allowCoverageGaps'
```

dctalbot (changes requested):

web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
spencerpauly: Agreed. Did in #3ce4da3.

web/src/app/schedules/temp-sched/TempSchedDialog.tsx
```diff
@@ -82,6 +83,25 @@ export default function TempSchedDialog({
     },
   })
 
+  const [isAllowingNoCoverage, setIsAllowingNoCoverage] = useState(false)
+  const [isShowingNoCoverageWarning, setIsShowingNoCoverageWarning] =
+    useState(false)
+
+  const hasNoCoverageGaps =
```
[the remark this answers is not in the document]
spencerpauly: Changed in 3ce4da3

dctalbot (dismissed):

web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
```diff
@@ -211,20 +232,44 @@ export default function TempSchedAddShiftsStep({
...
+              />
+            </div>
           </div>
+          {showCoverageGapsWarning && (
+            <Alert severity='error' className={classes.noCoverageError}>
+              <AlertTitle>Gaps in coverage</AlertTitle>
+              <FormHelperText>
+                There are gaps in coverage. During these gaps nobody will
+                receive alerts. If you still want to proceed, check the box and
+                then click Retry.
```
Forfold, line 257: user could be making actions with something other than a mouse

```suggestion
                then retry.
```

web/src/app/schedules/temp-sched/TempSchedDialog.tsx
```diff
@@ -82,6 +83,25 @@ export default function TempSchedDialog({
     },
   })
 
+  const [isAllowingNoCoverage, setIsAllowingNoCoverage] = useState(false)
+  const [isShowingNoCoverageWarning, setIsShowingNoCoverageWarning] =
```
Forfold, line 87: What are your thoughts on using names like `allowNoCoverage, setAllowNoCoverage` along with `hasCoverage, setHasCoverage`. Removes the double neg for what's the current coverage is to try and help

web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
```diff
@@ -46,6 +60,10 @@ type AddShiftsStepProps = {
 
   scheduleID: string
   edit?: boolean
+
+  showCoverageGapsWarning?: boolean
```
Forfold, line 64: rather than being dependent on a prop to show the warning, can the list determine this itself? looks like it only needs a call to `getCoverageGapItems` from `'./shiftsListUtil'`

Forfold (changes requested):

web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
```diff
@@ -211,20 +232,44 @@ export default function TempSchedAddShiftsStep({
...
+                }}
+                edit={edit}
+              />
+            </div>
           </div>
+          {showCoverageGapsWarning && (
+            <Alert severity='error' className={classes.noCoverageError}>
+              <AlertTitle>Gaps in coverage</AlertTitle>
+              <FormHelperText>
+                There are gaps in coverage. During these gaps nobody will
```
dctalbot, line 255:
```suggestion
                There are gaps in coverage. During these gaps nobody on this schedule will
```

web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
dctalbot: I wonder if we can clarify that this impacts the associated schedule too - not just the temporary schedule.

There are gaps in coverage. During these gaps nobody will receive alerts. This effectively disables <scheduleName> during these times. To proceed anyway, check the box and then click Retry.

web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
spencerpauly: Yep with the props I think the child could determine for itself if there's gaps. It just needs the shifts and start and end. I like that change because it avoids having repeat data in the props for `tempSchedAddShiftsStep` when it can easily be derived.

web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
Forfold: True, we could also add in the sched name to the second sentence:

```
There are gaps in coverage. During these gaps, nobody on <schedule_name> will
receive incoming alerts. If you still want to proceed, check the box and retry.
```

dctalbot: I like that, only thing thing I would change is I'd remove "incoming" since it's unclear which direction that implies (are alerts "incoming" or "outgoing"). Easier to omit altogether IMO.

web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
spencerpauly: Just submitted a new version. Let me know if that resolves this.

web/src/app/schedules/temp-sched/TempSchedDialog.tsx
spencerpauly: Removed double negative from naming

web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
spencerpauly: I just put "this schedule" for the schedule name. If I wanted to fetch the schedule name I could use a context from the scheduleDetails or make another query in this component to get the name from the sched id. I'm a little weary to have the dialog reliant on a parent context though since it's a dialog. I'd be interested to hear your thoughts on this.

Forfold (dismissed):

web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
dctalbot: If we did this, we would use a new query (single source of truth), but I think "this schedule" is good for this PR 👍

Forfold: I think "this schedule" works well. It puts more emphasis on the page they're on, where-as I could see them second guessing themselves for a moment if we put the full schedule name in, thinking "am I configuring the right thing?".

web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
```diff
@@ -102,10 +122,14 @@ export default function TempSchedAddShiftsStep({
...
   value,
   edit,
+  coverageGapsAllowed,
+  setCoverageGapsAllowed,
+  isShowingCoverageGapsWarning,
 }: AddShiftsStepProps): JSX.Element {
   const classes = useStyles()
   const [shift, setShift] = useState(null as Shift | null)
   const [submitted, setSubmitted] = useState(false)
+  const { zone } = useScheduleTZ(scheduleID)
```
dctalbot, line 132:
```suggestion
  const { zone, q } = useScheduleTZ(scheduleID)
```
We need to handle the loading state `q.loading`. See below for one solution

web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
```diff
@@ -159,6 +183,10 @@ export default function TempSchedAddShiftsStep({
     })
     setSubmitted(false)
   }
+  const schedInterval = parseInterval({ start: start, end: end })
```
dctalbot, line 186:
```suggestion
  const schedInterval = parseInterval({ start: start, end: end }, zone)
```

web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
```diff
@@ -159,6 +183,10 @@ export default function TempSchedAddShiftsStep({
...
+
+  const hasCoverageGaps =
+    getCoverageGapItems(schedInterval, value, zone).length > 0
```
dctalbot, line 189:
```suggestion
  const hasCoverageGaps = (() => {
  if (q.loading) return false
  const schedInterval = parseInterval({ start: start, end: end }, zone)
  return getCoverageGapItems(schedInterval, value, zone).length > 0
  })()
```

One trick that could work well here is invoking an anonymous function to set `hasCoverageGaps`

web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
dctalbot: Or delete this and use suggestion below

web/src/app/schedules/temp-sched/TempSchedDialog.tsx
```diff
@@ -82,6 +83,25 @@ export default function TempSchedDialog({
     },
   })
 
+  const [shouldAllowNoCoverage, setShouldAllowNoCoverage] = useState(false)
+  const [isShowingCoverageGapsWarning, setIsShowingCoverageGapsWarning] =
+    useState(false)
+
+  const hasCoverageGaps =
+    getCoverageGapItems(schedInterval, value.shifts, zone).length > 0
```
dctalbot, line 91:
```suggestion
    !q.loading && getCoverageGapItems(schedInterval, value.shifts, zone).length > 0
```
`zone` may not be ready if `q.loading`

dctalbot (changes requested):
Apologies for not catching it the first time, but we need to handle the loading state of the time zone from `useScheduleTZ`

web/src/app/schedules/temp-sched/TempSchedAddShiftsStep.tsx
dctalbot: Yeah, just want to make clear that the original schedule's assignments won't be active during a gap in a temporary schedule's assignments

dctalbot (approved):

Forfold (approved):

PR #1899, opened 2021-09-14
admin: allow Twilio messaging service SID as "from number"

dctalbot:

**Description:**
This PR adds support for using Twilio's messaging service SID to send messages. 

- When entering a phone number, users will not have a plus-sign adornment, but the plus sign will be auto-prepended on phonenumber-only fields. Regardless, a plus sign is required to be in the input field for phone inputs.
- Certain fields, like the admin send tool "From" field, will allow with a phone number or an SID.

React:
- The `TelTextField` component now accepts an array of `inputTypes` supporting `'tel'` and/or `'sid'` ('shortcode'/'short' would be a reasonable addition in the future). This value defaults to phonenumber-only (`['tel']`).


config/config.go
```diff
@@ -94,7 +94,7 @@ type Config struct {
 
 		AccountSID string
 		AuthToken  string `password:"true" info:"The primary Auth Token for Twilio. Must be primary (not secondary) for request valiation."`
-		FromNumber string `public:"true" info:"The Twilio number to use for outgoing notifications."`
+		FromNumber string `displayName:"From Number or Messenger SID" public:"true" info:"The Twilio number or Messenger SID to use for outgoing notifications."`
```
m17ch, line 97: I think we need to tweak all references:

`Messenger SID` -> `Twilio Messaging Service SID`

dctalbot: I think the S stands for service

mastercactapus (approved):

m17ch (approved):
lgtm 👍

PR #1901, opened 2021-09-15
ui/schedules: display override events on schedule calendar

spencerpauly:

**Description:**
Currently, the calendar only shows the resulting shifts from an override. This PR adds an override item to the calendar that shows the details of the override as-well. This lets you edit these details straight from the calendar. You can change the details of the override such as who is affected & for what duration. You can also remove overrides straight from the calendar.

**Screenshots:**

Image of new override event on calendar.
Screen Shot 2021-09-27 at 3 44 13 PM

Tooltip when calendar item is selected:
Screen Shot 2021-09-27 at 3 43 56 PM

- The pencil icon takes you to the edit override dialog
- The trash can takes you to the delete override dialog

Users will now be able to see overrides on the schedule calendar, and perform edit and delete operations for current/future overrides via the tooltip.

### DIFF
--- a/web/src/app/schedules/calendar/ScheduleCalendarEventWrapper.js
+++ b/web/src/app/schedules/calendar/ScheduleCalendarEventWrapper.js
@@ -169,35 +169,44 @@ export default function ScheduleCalendarEventWrapper({ children, event }) {
     if (DateTime.fromJSDate(event.end) <= DateTime.utc()) return null
     if (event.tempSched) return renderTempSchedButtons()
     if (event.fixed) return null
-    if (event.isOverride) return renderOverrideButtons(event)
+    if (event.isOverride) return renderOverrideButtons()
 
     return renderShiftButtons()
   }
 
-  function renderText() {
-    if (event.isOverride) {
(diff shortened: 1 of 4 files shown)


web/src/app/schedules/calendar/ScheduleCalendarEventWrapper.js
```diff
@@ -129,8 +167,46 @@ export default function ScheduleCalendarEventWrapper({ children, event }) {
...
+      <Grid item xs={12}>
+        <Typography variant='body2'>
+          {getDesc(event.override.addUser, event.override.removeUser)}
+        </Typography>
+      </Grid>
+    )
+  }
+
+  function renderDescription() {
+    if (event.isOverride) return renderOverrideDescription()
```
dctalbot, line 207: We can move this check the top of `renderOverrideDescription` e.g. `if (!event.isOverride) return null`.
That way we can remove this function, which is just serving as an if statement

Forfold (approved):

dctalbot (approved):

PR #1905, opened 2021-09-16
admin: add display names for config values

KatieMSB:

**Description:**
Adds support for `ConfigValue` and `ConfigHint` values to have `displayNames` that may be different or change from the existing `id` fields. 

GraphQL:
- `ConfigValue` and `ConfigHint` now support a `displayName` field. This value may change over time, whereas the respective `id` fields will not.


web/src/app/admin/AdminConfig.tsx
```diff
@@ -224,7 +216,7 @@ export default function AdminConfig(): JSX.Element {
                     )
                     .map((f: ConfigValue) => ({
                       id: f.id,
-                      label: formatHeading(_.last(f.id.split('.'))),
+                      displayName: f.displayName,
```
dctalbot, line 219: We should be able to remove this `map()` now

devtools/configparams/run.go
```diff
@@ -207,6 +208,36 @@ func printType(prefix string, v reflect.Type, details string, public, pass bool)
 		panic(fmt.Sprintf("not implemented for type %T", v.Kind()))
 	}
 
-	f = append(f, field{ID: key, Type: typ, Desc: details, Value: value, Public: public, Password: pass})
+	f = append(f, field{ID: key, DisplayName: printDisplayName(key, displayName), Type: typ, Desc: details, Value: value, Public: public, Password: pass})
 	return f
 }
+
+func printDisplayName(key, displayName string) string {
```
dctalbot, line 215: Nit: I think it would be a better design to do `if displayName != ""` short circuit at the top of this function. Then, the implementation of `format` be can brough into the main body of `printDisplayName`

e.g. 
```
if displayName != "" { 
    return displayName
}

var formattedGroups []string
str := strings.Split(key, ".")[1]
//pattern groups words in string by capital letters
pattern := regexp.MustCompile("([A-Z]*)([A-Z][^A-Z]+|$)")
for _, subStr := range pattern.FindAllStringSubmatch(str, -1) {
	if subStr[1] != "" {
		formattedGroups = append(formattedGroups, subStr[1])
	}
	if subStr[2] != "" {
		formattedGroups = append(formattedGroups, subStr[2])
	}
}
formattedStr := strings.Join(formattedGroups, " ")

formattedStr = strings.Replace(formattedStr, "Two Way", "Two-Way", -1)
formattedStr = strings.Replace(formattedStr, "Enable V 1 Graph QL", "Enable V1 GraphQL", -1)
formattedStr = strings.Replace(formattedStr, "R Ls", "RLs", -1) // fix usages of `URLs`

return formattedStr
}
```

graphql2/mapconfig.go
```diff
@@ -13,103 +13,103 @@ import (
...
+		{ID: "General.DisableSMSLinks", DisplayName: "Disable SMS Links", Type: ConfigTypeBoolean, Description: "If set, SMS messages will not contain a URL pointing to GoAlert.", Value: fmt.Sprintf("%t", cfg.General.DisableSMSLinks)},
+		{ID: "General.DisableLabelCreation", DisplayName: "Disable Label Creation", Type: ConfigTypeBoolean, Description: "Disables the ability to create new labels for services.", Value: fmt.Sprintf("%t", cfg.General.DisableLabelCreation)},
+		{ID: "General.DisableCalendarSubscriptions", DisplayName: "Disable Calendar Subscriptions", Type: ConfigTypeBoolean, Description: "If set, disables all active calendar subscriptions as well as the ability to create new calendar subscriptions.", Value: fmt.Sprintf("%t", cfg.General.DisableCalendarSubscriptions)},
+		{ID: "General.EnableV1GraphQL", DisplayName: "Enable V1 Graph QL", Type: ConfigTypeBoolean, Description: "Enables the deprecated /v1/graphql endpoint (replaced by /api/graphql).", Value: fmt.Sprintf("%t", cfg.General.EnableV1GraphQL)},
```
dctalbot, line 36:
```suggestion
		{ID: "General.EnableV1GraphQL", DisplayName: "Enable V1 GraphQL", Type: ConfigTypeBoolean, Description: "Enables the deprecated /v1/graphql endpoint (replaced by /api/graphql).", Value: fmt.Sprintf("%t", cfg.General.EnableV1GraphQL)},
```

dctalbot (changes requested):

dctalbot (approved):

PR #1912, opened 2021-09-20
ui/temporary schedules: design enhancements

Forfold:
**Description:**
Updates the dialog for Temporary Schedules per new design decisions with the aim to improve the general user experience when creating or editing a temporary schedule.

- Form Fields altered to use `variant='outline'` (can revert if undesirable)
- Dialog is no longer fullscreen
- Stepper is no longer used. Start and End fields have been added above the fields to add a shift
- Default start times are now the next Monday through Friday
- Coverage warning moved to the top of the shift list
  - This change ensures the warning is seen right away on mobile or desktop.
- Retains shift duration when adding shifts
- Adds a notice to the top of each day if it is out of the set start/end range

- "Retry" vs. "Submit" text for `LoadingButton` is now determined based off a true attempt count, rather than the number of errors sent from props.

**Out of Scope:**
- Fix transitions
- Clicking on "No Coverage" warning opens Add Shift panel, sets start time to beginning of the gap in coverage

**Screenshots:**
Screen Shot 2021-09-20 at 12 49 21 PM
Screen Shot 2021-09-20 at 12 49 31 PM
Screen Shot 2021-09-23 at 10 22 41 AM


web/src/app/schedules/temp-sched/TempSchedDialog.tsx
```diff
@@ -34,30 +63,40 @@ export default function TempSchedDialog({
...
+  const [allowNoCoverage, setAllowNoCoverage] = useState(false)
+  const [hasSubmitted, setHasSubmitted] = useState(false)
 
   useEffect(() => {
     // set default start, end times when zone is ready
     if (!value.start && !value.end && !q.loading && zone) {
-      const nextSunday = getNextWeekday(7, DateTime.now(), zone)
-      const followingSunday = nextSunday.plus({ week: 1 })
+      const nextMonday = getNextWeekday(1, DateTime.now(), zone)
+      const nextFriday = nextMonday.plus({ days: 5 })
```
dctalbot, line 84:
```suggestion
      const nextFriday = nextMonday.plus({ days: 4 })
```

Forfold: This would set it to end at the end of Thursday, rather than including Friday in the default times

dctalbot: Ah yeah, I think I understand. In that case,maybe rename to `thruFriday` since the value is 12:00 on Saturday. Or maybe a comment. Otherwise, looks like a code smell IMO

spencerpauly (approved):

web/src/app/schedules/temp-sched/shiftsListUtil.ts
```diff
@@ -49,6 +49,72 @@ export function getSubheaderItems(
...
+        sortedShifts.reduce(
+          (result, candidate) =>
+            candidate.end > result.end ? candidate : result,
+          sortedShifts[0],
+        )?.end,
+      )
+    : null
+
+  // timezone is loaded in after initial render, check if sched interval is valid
+  if (!schedInterval.start || !schedInterval.end) {
```
dctalbot, line 72:
```suggestion
  if (zone === '') {
```
I think this would also work, probably clearer

web/src/app/schedules/temp-sched/shiftsListUtil.ts
```diff
@@ -49,6 +49,72 @@ export function getSubheaderItems(
...
+    : null
+  const lastShiftEnd = firstShiftStart
+    ? DateTime.fromISO(
+        sortedShifts.reduce(
+          (result, candidate) =>
+            candidate.end > result.end ? candidate : result,
+          sortedShifts[0],
+        )?.end,
+      )
+    : null
```
dctalbot, line 69: We can avoid doing a `sortBy` and `reduce` using the strategy here: https://github.com/target/goalert/pull/1912/files#diff-60811a7d064a778258ca3292d7d6b6ca8dd915deea69291f2abb52cb53886568R28-R35

web/src/app/schedules/temp-sched/TempSchedAddNewShift.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+  function handleAddShift(): void {
+    if (fieldErrors(true).length) {
+      setSubmitted(true)
+      return
+    }
+    if (!shift) return // ts sanity check
+
+    onChange(mergeShifts(value.shifts.concat(shift)))
+    const end = DateTime.fromISO(shift.end)
```
dctalbot, line 129:
```suggestion
    const end = DateTime.fromISO(shift.end, { zone })
```

web/src/app/schedules/temp-sched/TempSchedAddNewShift.tsx
```diff
@@ -0,0 +1,270 @@
...
+    const diff = end.diff(DateTime.fromISO(shift.start))
```
dctalbot, line 130:
```suggestion
    const diff = end.diff(DateTime.fromISO(shift.start, { zone }))
```

web/src/app/schedules/temp-sched/TempSchedAddNewShift.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+  // set start equal to the temporary schedule's start
+  // can't this do on mount since the step renderer puts everyone on the DOM at once
+  useEffect(() => {
+    setShift({
+      start: value.start,
+      end: DateTime.fromISO(value.start).plus({ hours: 8 }).toISO(),
+      userID: '',
+    })
+  }, [value.start])
```
dctalbot, line 94:
```suggestion
  useEffect(() => {
    if (zone === '') return
    setShift({
      start: value.start,
      end: DateTime.fromISO(value.start, { zone }).plus({ hours: 8 }).toISO(),
      userID: '',
    })
  }, [value.start, zone])
```

web/src/app/schedules/temp-sched/TempSchedAddNewShift.tsx
dctalbot: Every call to `DateTime.fromISO` needs to include the zone info, there are a few more instanced of this below

dctalbot (changes requested):
Looking great! I wonder if we should just turn off transitions for the time being. It seems like we've never gotten it quite right and might just need some focus in a dedicated PR

web/src/app/schedules/temp-sched/shiftsListUtil.ts
dctalbot: I don't think that link really works - see lines 28 through 35 of this file

web/src/app/schedules/temp-sched/TempSchedAddNewShift.tsx
Forfold: Thanks! Easy to forget that sort of thing

web/src/app/schedules/temp-sched/TempSchedAddNewShift.tsx
```diff
@@ -0,0 +1,272 @@
...
+              <FormField
+                fullWidth
+                component={ISODateTimePicker}
+                label='Shift Start'
+                name='shift-start'
+                fieldName='start'
+                min={edit ? value.start : now}
+                mapOnChangeValue={(value: string, formValue: Value) => {
+                  if (!manualEntry) {
+                    const diff = DateTime.fromISO(value).diff(
```
dctalbot, line 180:
```suggestion
                    const diff = DateTime.fromISO(value, { zone }).diff(
```

web/src/app/schedules/temp-sched/TempSchedAddNewShift.tsx
```diff
@@ -0,0 +1,272 @@
...
+                      DateTime.fromISO(formValue.start),
```
dctalbot, line 181:
```suggestion
                      DateTime.fromISO(formValue.start, { zone }),
```

web/src/app/schedules/temp-sched/TempSchedAddNewShift.tsx
```diff
@@ -0,0 +1,272 @@
...
+                    )
+                    formValue.end = DateTime.fromISO(formValue.end)
```
dctalbot, line 183:
```suggestion
                    formValue.end = DateTime.fromISO(formValue.end, { zone })
```

web/src/app/schedules/temp-sched/TempSchedAddNewShift.tsx
```diff
@@ -0,0 +1,272 @@
...
+                        </ClickableText>
+                      </div>
+                    </React.Fragment>
+                  }
+                  timeZone={zone}
+                  disabled={q.loading}
+                />
+              ) : (
+                <FormField
+                  fullWidth
```
dctalbot, line 222:
```suggestion
                  fullWidth
                  disabled={q.loading}
```

web/src/app/schedules/temp-sched/TempSchedAddNewShift.tsx
```diff
@@ -0,0 +1,272 @@
...
+                  component={NumberField}
+                  label='Shift Duration (hours)'
+                  name='shift-end'
+                  fieldName='end'
+                  float
+                  // value held in form input
+                  mapValue={(nextVal: string, formValue: Value) => {
+                    const nextValDT = DateTime.fromISO(nextVal)
```
dctalbot, line 230:
```suggestion
                    const nextValDT = DateTime.fromISO(nextVal, { zone })
```

web/src/app/schedules/temp-sched/TempSchedAddNewShift.tsx
```diff
@@ -0,0 +1,272 @@
...
+                    if (!formValue || !nextValDT.isValid) return ''
+                    return nextValDT
+                      .diff(DateTime.fromISO(formValue.start), 'hours')
```
dctalbot, line 233:
```suggestion
                      .diff(DateTime.fromISO(formValue.start, { zone }), 'hours')
```

web/src/app/schedules/temp-sched/TempSchedAddNewShift.tsx
```diff
@@ -0,0 +1,272 @@
...
+                      .hours.toString()
+                  }}
+                  // value held in state
+                  mapOnChangeValue={(nextVal: string, formValue: Value) => {
+                    if (!nextVal) return ''
+                    return DateTime.fromISO(formValue.start)
```
dctalbot, line 239:
```suggestion
                    return DateTime.fromISO(formValue.start, { zone })
```

dctalbot (changes requested):

web/src/app/schedules/temp-sched/TempSchedAddNewShift.tsx
Forfold: I'll do all of this here- but I haven't really _done_ any of this work. Things like this could probably be handled in a separate PR that fixes specific logic, since this body of work is intended mainly for the visual tweaks

Forfold:
@dctalbot Interesting, I wonder if that's a conflict in splitting it up at midnight

web/src/app/schedules/temp-sched/TempSchedAddNewShift.tsx
dctalbot: Ah yeah, definitely could be split it out. Was hard to tell in the file diff since it was all in a new file. Changes should be good though, thanks!

dctalbot (dismissed):

Forfold:
Working on fixing some tests atm

dctalbot (approved):

dctalbot (approved):

spencerpauly (approved):

dctalbot (changes requested):
Seeing an issue when editing and "fast-forwarding" the temp sched start time. I suggest disabling the start/end fields for now.

https://user-images.githubusercontent.com/17692467/135143871-72019ee8-a3dc-464a-ad89-a53d2b53a0bd.mov

dctalbot (approved):

spencerpauly (approved):

PR #1929, opened 2021-09-28
twilio: Messaging Service SID support/fix for SMS messages

mastercactapus:
**Description:**
Reverts #1922 and re-implements as an SMS-only feature as Messaging Service does not handle voice call pooling.


dctalbot (approved):
functionally validated, looks good!

m17ch (approved):
lgtm

PR #1956, opened 2021-10-06
ui/temp-sched: Fix temp schedule shift transitions

spencerpauly:

**Description:**
Temporary schedule has some bugs involving the transitions. This PR seeks to fix those. Here's the list of actions that created weird transitions:
- [x] Invalid Start/End Warning
    - List will disappear and warning will transition in and out
- [x] Changing Start Date
    - All shifts will transition in and out for new date added
- [x] Adding shift in between existing shifts
    - Some or all shifts will transition again
- [x] Delete shift in between existing shifts
    - Some or all shifts will transition again
- [x] Sometimes list items transition out below the schedule list

**Out of Scope:**
We noticed a small bug when setting the start/end dates of the calendar that caused the page to freeze. Fix is coming in a different PR.

**Screenshots:**

Fixed animations in temp schedule

**Additional Info:**

### DIFF
--- a/web/src/app/lists/FlatList.tsx
+++ b/web/src/app/lists/FlatList.tsx
@@ -39,14 +39,18 @@ const useStyles = makeStyles({
     backgroundColor: lightGrey,
   },
   slideEnter: {
+    maxHeight: '0px',
+    opacity: 0,
     transform: 'translateX(-100%)',
   },
   slideEnterActive: {
+    maxHeight: '60px',
+    opacity: 1,
     transform: 'translateX(0%)',
-    transition: 'transform 500ms',
+    transition: 'all 500ms',
   },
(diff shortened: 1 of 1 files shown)


web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
```diff
@@ -121,7 +120,7 @@ export default function TempSchedShiftsList({
 
           return {
             scrollIntoView: true,
-            id: s.start + s.userID,
+            id: s.userID + index.toString(),
```
KatieMSB, line 123:
```suggestion
            id: s.start + s.userID + index.toString(),
```
We'll want to add the start time back in otherwise we might get multiple list items with the same id.

PR #1958, opened 2021-10-06
ui/temp-sched: Enforce date bounds

KatieMSB:

**Description:**
Enforces the start/end time bounds for temporary schedules to help prevent slow downs in the ui from having bounds too far apart.

### DIFF
--- a/web/src/app/schedules/temp-sched/TempSchedDialog.tsx
+++ b/web/src/app/schedules/temp-sched/TempSchedDialog.tsx
@@ -216,6 +216,7 @@ export default function TempSchedDialog({
                   name='start'
                   label='Schedule Start'
                   min={now}
+                  max={DateTime.fromISO(now).plus({ year: 1 }).toISO()}
                   validate={() => validate()}
                   timeZone={zone}
                   disabled={q.loading || edit}
@@ -230,6 +231,7 @@ export default function TempSchedDialog({
                   name='end'
                   label='Schedule End'
                   min={value.start}
+                  max={DateTime.fromISO(value.start).plus({ month: 1 }).toISO()}
                   validate={() => validate()}
                   timeZone={zone}
                   disabled={q.loading || edit}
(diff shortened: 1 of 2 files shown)


mastercactapus (dismissed):

web/src/app/schedules/temp-sched/TempSchedDialog.tsx
```diff
@@ -216,6 +216,7 @@ export default function TempSchedDialog({
                   name='start'
                   label='Schedule Start'
                   min={now}
+                  max={DateTime.fromISO(now).plus({ year: 1 }).toISO()}
```
dctalbot, line 219:
```suggestion
                  max={DateTime.fromISO(now, { zone }).plus({ year: 1 }).toISO()}
```

web/src/app/schedules/temp-sched/TempSchedDialog.tsx
```diff
@@ -230,6 +231,7 @@ export default function TempSchedDialog({
                   name='end'
                   label='Schedule End'
                   min={value.start}
+                  max={DateTime.fromISO(value.start).plus({ month: 3 }).toISO()}
```
dctalbot, line 234:
```suggestion
                  max={DateTime.fromISO(value.start, { zone }).plus({ month: 3 }).toISO()}
```

web/src/app/util/ISOPickers.js
```diff
@@ -62,26 +62,38 @@ function useISOPicker(
     setInputValue(value && dtValue ? dtValue.toFormat(format) : '')
   }, [value, zone])
 
+  // sets min and max if set
+  const inputProps = otherProps?.inputProps ?? {}
+  if (min) inputProps.min = DateTime.fromISO(min).toFormat(format)
```
dctalbot, line 67: I think we need to be careful with time zones here. Basically whenever we use `fromISO` we should pass in a `zone` option, otherwise it defaults to the system zone and uses the wrong time. But I see that isn't always the case in this file, so it's hard to understand without actually digging in. There is also some converting between UTC which further complicates it.

This file would be a good candidate to refactor in typescript, so maybe resolving these discrepancies can wait until then. It should work for 99.99% of use cases :)

dctalbot (changes requested):

PR #1968, opened 2021-10-12
deps: replace withWidth calls in @material-ui

Forfold:
This PR is a prerequisite to #1946 `deps: upgrade @material-ui to @mui v5`. Material has fully deprecated the use of `withWidth` with the new theme structure, so I migrated to use the new recommended solution in all spots, which is to use `useMediaQuery`.

Info here: https://mui.com/components/use-media-query/#migrating-from-withwidth


mastercactapus (approved):

web/src/app/details/DetailsPage.tsx
```diff
@@ -196,9 +191,7 @@ export default function DetailsPage(p: DetailsPageProps): JSX.Element {
       {/* Primary Page Content */}
       {p.pageContent && (
         <Grid
-          className={
-            isWidthDown('sm', width) ? classes.smPageBottom : undefined
-          }
+          className={!isMobile ? classes.smPageBottom : undefined}
```
dctalbot, line 194:
```suggestion
          className={isMobile ? classes.smPageBottom : undefined}
```

PR #1971, opened 2021-10-13
temp schedules: no coverage notice interaction

Forfold:
**Description:**
This PR adds the ability to click on a "No Coverage" notice within the Temporary Schedules Dialog to set the shift's start and end times in the form.

e.g.
1. 24hr No Coverage on Oct 14th
2. Clicking this would set the start to `October 14th, 12:00 AM`, and the duration to `24 hours`

- Coverage Notices within the Temporary Schedule Dialog are now clickable buttons
- Clicking a coverage notice will also cause the shift form to expand open, if closed

### DIFF
--- a/web/src/app/schedules/temp-sched/TempSchedAddNewShift.tsx
+++ b/web/src/app/schedules/temp-sched/TempSchedAddNewShift.tsx
@@ -27,10 +27,8 @@ type AddShiftsStepProps = {
   edit?: boolean
   showForm: boolean
   setShowForm: (showForm: boolean) => void
-  // rename to handleCoverageGap everywhere
-  handleShowForm: (gapCoverage: Interval) => void
-  shift: Shift|null,
-  setShift: (shift: Shift)=> void
+  shift: Shift | null
+  setShift: (shift: Shift) => void
 }
 
 type DTShift = {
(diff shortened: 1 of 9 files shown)


web/src/app/schedules/temp-sched/TempSchedAddNewShift.tsx
```diff
@@ -227,13 +239,11 @@ export default function TempSchedAddNewShift({
...
                   mapValue={(nextVal: string, formValue: Value) => {
                     const nextValDT = DateTime.fromISO(nextVal, { zone })
-                    if (!formValue || !nextValDT.isValid) return ''
-                    return nextValDT
-                      .diff(
-                        DateTime.fromISO(formValue.start, { zone }),
-                        'hours',
-                      )
-                      .hours.toString()
+                    const formValDT = DateTime.fromISO(formValue?.start, {
```
spencerpauly, line 242:
```suggestion
                    const formValDT = DateTime.fromISO(formValue?.start ?? '', {
```

I'm not sure if luxon needs you to handle the undefined case here but it's being handled like this in other places.

spencerpauly (approved):

KatieMSB (approved):

PR #1975, opened 2021-10-14
dev/mockslack: track/assert thread and broadcast replies

mastercactapus:
**Description:**
This PR updates the mockslack server and smoketest harness to allow for testing thread and broadcast replies for the Slack notifier.

### DIFF
--- a/devtools/mockslack/chatpostmessage.go
+++ b/devtools/mockslack/chatpostmessage.go
@@ -14,8 +14,8 @@ type ChatPostMessageOptions struct {
 
 	AsUser bool
 
-	ThreadTS       string
-	BroadcastReply bool
+	ThreadTS  string
+	Broadcast bool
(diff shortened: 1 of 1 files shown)


Forfold (dismissed):

devtools/mockslack/chatpostmessage.go
```diff
@@ -14,7 +14,8 @@ type ChatPostMessageOptions struct {
 
 	AsUser bool
 
-	ThreadTS string
+	ThreadTS       string
+	BroadcastReply bool
```
dctalbot, line 18:
```suggestion
	Broadcast bool
```
Can we maintain parity as we do with `ThreadTS`?

dctalbot (approved):

Forfold (approved):

PR #1993, opened 2021-10-26
ui/temp-sched: display historical shifts

KatieMSB:

**Description:**
Historical shifts are now displayed when editing an existing temporary schedule as disabled list items above active shifts.


m17ch (dismissed):
LGTM 👍

web/src/app/schedules/temp-sched/TempSchedAddNewShift.tsx
```diff
@@ -184,7 +184,7 @@ export default function TempSchedAddNewShift({
                 name='shift-start'
                 fieldName='start'
                 min={value.start}
-                max={value.end}
+                max={DateTime.fromISO(value.end).plus({ year: 1 }).toISO()}
```
Forfold, line 187: Do we have `zone` available for these calls?

```suggestion
                max={DateTime.fromISO(value.end, { zone }).plus({ year: 1 }).toISO()}
```

web/src/app/schedules/temp-sched/TempSchedShiftsList.tsx
```diff
@@ -147,23 +143,26 @@ export default function TempSchedShiftsList({
...
                       <Error color='error' />
                     </Tooltip>
                   )}
-                  <IconButton
-                    aria-label='delete shift'
-                    onClick={() => onRemove(s)}
-                  >
-                    <Delete />
-                  </IconButton>
+                  {isHistoricShift ? null : (
```
Forfold, line 158: This could also be written as
```
{!isHistoricShift && (
  <IconButton
    aria-label='delete shift'
    onClick={() => onRemove(s)}
  >
    <Delete />
  </IconButton>
)}
```

Forfold (changes requested):
One thing I noticed is I set the time to these two shifts to 9:06am, but as time passed it was changing the start time in the list:

Screen Shot 2021-11-08 at 9 14 18 AM

I also noticed that the list doesn't self-update historic shifts -> disabled as time passes when the dialog is left open- I was originally thinking we could just set the poll interval to every 60s, but I don't think we can do that since we'd want it to update when creating too. So maybe a useEffect that updates once a minute? Might be overkill though

Forfold (approved):
LGTM, times are now staying set:

**8:00am**
Screen Shot 2021-11-12 at 8 01 12 AM

**8:02am**
Screen Shot 2021-11-12 at 8 01 19 AM

m17ch (approved):

PR #2004, opened 2021-11-01
notification/slack: update existing messages with status

mastercactapus:
**Description:**
This PR updates the Slack notifier to include a color bar with alert messages and update the existing message instead of posting status updates in-thread.

Unacked alerts will have a red bar and details included:
unacked

Acked alerts will reflect the latest log entry, a yellow bar, and retain details:
acked

When escalated, the alert summary and link is broadcast to the channel:
escalated

When closed, details of the alert is hidden, bar turns green, and the link remains with the last log entry:
closed

### DIFF
--- a/alert/store.go
+++ b/alert/store.go
@@ -615,7 +615,7 @@ func (db *DB) CreateOrUpdateTx(ctx context.Context, tx *sql.Tx, a *Alert) (*Aler
 		var oldStatus Status
 		err = tx.Stmt(db.createUpdAck).
 			QueryRowContext(ctx, n.ServiceID, n.DedupKey()).
-			Scan(&n.ID, &n.Summary, &n.Details, &n.CreatedAt, &oldStatus)
+			Scan(&n.ID, &n.Summary, &n.Details, &oldStatus, &n.CreatedAt)
 		if oldStatus != n.Status {
 			logType = alertlog.TypeAcknowledged
 		}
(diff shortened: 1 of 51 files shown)


Forfold (approved):
As a personal preference, I think it could look a bit cleaner without the nested quote line for the details- LGTM!

mastercactapus:
@Forfold  Maybe we can try tweaking it during the two-way interaction work; we'll be changing the rendering logic for that into its final state. I added it so it would stand out since there's no pre-processing for markdown -> slack markdown, but I agree the double color-bar isn't ideal.

Forfold:
@mastercactapus That sounds good! Yeah I definitely don't think something like that should hold us back, so it has my approval :)

Forfold:
It could be neat to have a button to show the full details (rendered with `mkdwn`) as a modal within Slack

dctalbot (changes requested):
Code looks good, but seeing some odd behavior...

I created an alert with a summary of "test", and got the following -

Native notification on mac is missing preview:
Screen Shot 2021-11-02 at 2 03 46 PM

There is a stray greater than symbol in the message body:
Screen Shot 2021-11-02 at 2 06 20 PM

It only goes away if I close the alert.

dctalbot (approved):

Forfold (approved):

PR #2014, opened 2021-11-04
notification/slack: add interactivity to alert notifications

mastercactapus:
**Description:**
Adds two-way interactivity as an option for Slack notifications.

**Additional Info:**
Summary of changes:
- make alert log store support both contact methods and notification channels for ack/close actions
- new http route `/api/v2/slack/message-action` added
- new config option `Slack.InteractiveMessages` to opt-in
- action support in the mockslack server
- new `ReceiveSubject` method that can map a provider/subject ID to a GoAlert user for notification channel responses
- remove `alert` as a dependency for the `notification` package by using a new AlertState type instead of alert.Status
- updated Send() method in Slack ChannelSender to add action blocks
- added support for Ephemeral messages and actions and assertions in test harness/mockslack
- added slack interaction smoketest
- added FindOneBySubject to user.Store
- added FieldErrorf to validation package to support formatting error messages cleaner

### DIFF
--- a/alert/log/store.go
+++ b/alert/log/store.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/target/goalert/integrationkey"
+	"github.com/target/goalert/notification"
 	"github.com/target/goalert/notificationchannel"
 	"github.com/target/goalert/permission"
 	"github.com/target/goalert/user/contactmethod"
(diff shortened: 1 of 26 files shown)


Forfold (approved):

arurao (approved):
LGTM :tada:

PR #2029, opened 2021-11-15
ui/schedules/calendar: Increase override query page size for calendar

### LINKED ISSUE #1987: ui/schedules/calendar: missing calendar events for override
**Describe the Bug:**
When a schedule has a large number of overrides, the override events on the calendar stop rendering, perhaps due to pagination of the query.

**Expected Behavior:**
Every override is represented on the Schedule Calendar.

**Observed Behavior:**
Not every override is represented on the Schedule Calendar.

**Application Version:**
`784cb94d36e73fdfb48c7d7d61040d8b5916f328`

spencerpauly:

**Description:**
Work-around for #1987 by raising overrides fetched per-page from 15 (default) to 149. 

**Out of scope:**
If there are more than 149 overrides for a schedule (unlikely, but still a limit to keep in mind), they won't all be queried. Would need a more permanent fix like #2028 to handle that.

### DIFF
--- a/web/src/app/schedules/calendar/ScheduleCalendarQuery.tsx
+++ b/web/src/app/schedules/calendar/ScheduleCalendarQuery.tsx
@@ -13,7 +13,10 @@ const query = gql`
     $start: ISOTimestamp!
     $end: ISOTimestamp!
   ) {
-    userOverrides(input: { scheduleID: $id, start: $start, end: $end }) {
+    userOverrides(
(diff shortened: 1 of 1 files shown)


spencerpauly:
I'm going to add lots of overrides to the calendar and test this quick, then I'll make it a full PR.

dctalbot (approved):

Forfold (approved):

PR #2032, opened 2021-11-17
ui/lists: refactor paginated list

### LINKED ISSUE #1606: ui: override list filter container state does not persist on change
**Describe the Bug:**
In a recent refactor (#1571), there is a known regression for the override list in which the filter popover closes on change

**Steps to Reproduce:**
1. Go to override list
2. Use filter

**Expected Behavior:**
Filter remains open until a blur event, or user clicks "Done"

**Observed Behavior:**
Filter closes on change

**Application Version:**
`fe36cc1ae6bf177695e9aa81fa7ad6d42dbd373e`

**Additional Context:**
See #1571

KatieMSB:

**Description:**
Refactors `QueryList.tsx`, `PaginatedList.tsx`, and `ControlledPaginatedList.tsx` to render the list action header and header note outside of `PaginatedList.tsx`.

Fixes #1606


web/src/app/lists/PageControls.tsx
```diff
@@ -0,0 +1,127 @@
...
+export const PageControlsContextProvider = (props: {
+  children: ReactNode
+}): JSX.Element => {
+  const [page, setPage] = useState<number>(0)
+  return (
+    <PageControlsContext.Provider value={{ page, setPage }}>
+      {props.children}
+    </PageControlsContext.Provider>
+  )
+}
```
dctalbot, line 42: Since this is only used once, it doesn't need to be a reusable component. It would be better to in-line `PageControlsContext.Provider` the only place it's used IMO

web/src/app/lists/QueryList.tsx
```diff
@@ -101,6 +107,7 @@ export default function QueryList(props: QueryListProps): JSX.Element {
     ...listProps
   } = props
   const { input, ...vars } = variables
+  const { page } = useContext(PageControlsContext)
```
dctalbot, line 110: `useContext` must have a Provider as an ancestor in the component tree. Otherwise, it won't be aware of any state change events.

> The current context value is determined by the value prop of the nearest <MyContext.Provider> above the calling component in the tree. 

https://reactjs.org/docs/hooks-reference.html#usecontext

dctalbot (changes requested):
Seeing some odd behavior if you page forward, perform a search, and clear the search bar. The user will be left with an empty list, and paging backwards will show more empty lists. This is likely because the context consumer is not a child component of the context provider. 

More broadly, I'm not confident Context is the right tool to reach for here. Generally, Context is used as a store pub/sub for props that would otherwise need to passed down through many deeply nested components e.g. a theme/setTheme model. Curious if you explored any alternatives?

web/src/app/lists/PaginatedList.tsx
```diff
@@ -417,16 +268,8 @@ export function PaginatedList(props: PaginatedListProps): JSX.Element {
...
-        <Card>
-          {cardHeader}
-          {infiniteScroll ? renderAsInfiniteScroll() : renderList()}
-        </Card>
-      </Grid>
-      {!infiniteScroll && (
-        <PageControls onBack={onBack} onNext={onNext} isLoading={isLoading} />
-      )}
-    </Grid>
+    <React.Fragment>
```
dctalbot, line 271: Fragment isn't needed here

web/src/app/lists/QueryList.tsx
```diff
@@ -101,6 +110,7 @@ export default function QueryList(props: QueryListProps): JSX.Element {
     ...listProps
   } = props
   const { input, ...vars } = variables
+  const [page, setPage] = useState(0)
```
dctalbot, line 113: Any reason this state can't be stored within PaginatedList?

dctalbot (changes requested):
As far as I can tell, the particular bug in question can be solved by extracting the list header to be rendered outside of PaginatedList, which is what this PR does, but all of the state refactoring to use the context api isn't necessary, and it really complicates the data flow IMO

web/src/app/lists/PaginatedList.tsx
```diff
@@ -161,16 +81,11 @@ function LoadingItem(props: { dense?: boolean }): JSX.Element {
...
-  // header elements will be displayed at the top of the list.
-  headerNote?: string // left-aligned
-  headerAction?: JSX.Element // right-aligned
-
   items: PaginatedListItemProps[] | CheckboxItemsProps[]
   itemsPerPage?: number
 
+  pageCount?: number
+
   isLoading?: boolean
```
Forfold, line 89: How should we be using this? I see `isLoading` as both a propType and being used in the context but not sure if it should just be one or the other

Forfold: It looks like we are passing down `loading` from Apollo a few components, but then use the context instead inside the page controls. I don't see anywhere that `isLoading` gets updated in the context though so maybe we should use the prop instead and remove `isLoading` from the context?

KatieMSB:
Reverted changes that pulled out the PageControls from the PaginatedList since it wasn't needed for the main bug fix.

web/src/app/lists/PaginatedList.tsx
```diff
@@ -82,7 +81,7 @@ function PageControls(props: {
       alignItems='center'
       className={classes.controls}
     >
-      <Grid item>
+      <CardActions>
```
Forfold, line 84: I think we want to keep the controls outside of the card, unless that was a design decision.

current impl:

Screen Shot 2022-02-04 at 12 07 23 PM

incoming change:

Screen Shot 2022-02-04 at 12 06 57 PM

KatieMSB: Right, I forgot that's why the PageControls were pulled out of the PaginatedList file. I switched it back and cleaned up the context so it's just using state now.

dctalbot (dismissed):

dctalbot (dismissed):

Forfold (commented):
It looks like there's an issue when I go forward a page, then back a page, it appends the results to the dataset instead of showing just that page

dctalbot (approved):

Forfold (approved):

PR #2036, opened 2021-11-22
ui/schedules: Convert ScheduleDetails component to typescript

spencerpauly:

**Description:**
First of a series of PR's to convert some of the remaining JS files in the schedules section to typescript. These changes will be helpful so we can have strong typings in the future if we choose to clean up some of the timezone code.

**Out of Scope:**
Still need to convert: `scheduleCalendarOverrideDialog.js`, `scheduleCalendarEventWrapper.js`, `ScheduleCalendar.js` and any connected components.

**Screenshots:**

No user-facing changes.


web/src/app/schedules/ScheduleDetails.tsx
```diff
@@ -34,28 +34,60 @@ const query = gql`
...
-export const ScheduleCalendarContext = React.createContext({
-  onNewTempSched: () => {},
-  onEditTempSched: () => {},
-  onDeleteTempSched: () => {},
-  // ts files infer function signature, need parameter list
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  setOverrideDialog: (overrideVal) => {},
-})
+// todo - move this interface to more generic file when those are converted to typescript
+interface ScheduleCalendarOverrideDialogProps {
```
dctalbot, line 38:
```suggestion
interface OverrideDialog {
```
Instead of introducing something like this in a half-way state and supporting it with a comment, I'd recommend keeping it simple. A subsequent PR can take care of the work you anticipate.

web/src/app/schedules/ScheduleDetails.tsx
```diff
@@ -34,28 +34,60 @@ const query = gql`
...
+  removeUserReadOnly: boolean
+  defaultValue: {
+    addUserID?: string
+    removeUserID?: string
+    start: string
+    end: string
+  }
+}
+
+interface ScheduleCalendarContextProps {
```
dctalbot, line 49:
```suggestion
interface ScheduleCalendarContext {
```
We should reserve the Props suffix for components IMO

web/src/app/schedules/ScheduleDetails.tsx
```diff
@@ -34,28 +34,60 @@ const query = gql`
...
+
+export default function ScheduleDetails({
+  scheduleID,
+}: ScheduleDetailsProps): JSX.Element {
   const [showEdit, setShowEdit] = useState(false)
   const [showDelete, setShowDelete] = useState(false)
-  const [configTempSchedule, setConfigTempSchedule] = useState(null)
+  const [configTempSchedule, setConfigTempSchedule] = useState<
+    TempSchedValue | null | undefined
+  >()
```
dctalbot, line 77:
```suggestion
  >(null)
```
Is there a reason we are no longer initializing with `null`?

web/src/app/schedules/ScheduleDetails.tsx
```diff
@@ -89,10 +121,10 @@ export default function ScheduleDetails({ scheduleID }) {
           onClose={() => setShowDelete(false)}
         />
       )}
-      {configTempSchedule && (
+      {configTempSchedule !== undefined && (
         <TempSchedDialog
-          value={configTempSchedule === true ? null : configTempSchedule}
-          onClose={() => setConfigTempSchedule(null)}
+          value={configTempSchedule}
+          onClose={() => setConfigTempSchedule(undefined)}
```
dctalbot, line 127:
```suggestion
          onClose={() => setConfigTempSchedule(null)}
```
Same as above, we generally try to avoid defining a value as `undefined`

dctalbot (changes requested):

web/src/app/schedules/ScheduleDetails.tsx
spencerpauly: Agreed. Good catch.

web/src/app/schedules/ScheduleDetails.tsx
spencerpauly: Yep. So this is one of the only functional changes I made. Before we were setting this to variable to type:
null = don't show tempSched window
true = show tempSched window, but with no passed in config (aka. it's a brand new temp schedule)
TempSchedValue = show tempSched, with a passed in config (aka. we're editing a current temp schedule)

So basically this variable is holding 2 pieces of information: 
A. Should we show the temp schedule window?, and 
B. What should we pass in as the current temp schedule?

And this was odd to convert to typescript because this useState would be of type <TempSchedValue | null | true>. Kind of a weird typing.

So instead it's now of type:
undefined = don't show tempSched window
null = show tempSched window, but with no passed in config (aka. it's a brand new temp schedule)
TempSchedValue = show tempSched, with a passed in config (aka. we're editing a current temp schedule)

It's still confusing though, I agree. Alternatively we could split this into 2 pieces of state: a `shouldShowTempSchedDialog` and an `activeTempSched`, or (and this would probably be the cleanest) we could make a generic custom hook that manages this kind of thing.

API could be like: { isOpen, onOpen, onClose } = useCustomDialog();

where the onOpen looks like (roughly):

```

function useCustomDialog<T>() {
  const [isOpen, setIsOpen] = useState();
  const [value, setValue] = useState<T>();

  const onOpen = (newValue) => {
    setValue(newValue);
    setIsOpen(true);
  }

  const onClose = () => {
    setValue(null);
    setIsOpen(false);
  }

  return {isOpen, value, onOpen, onClose}
}
```

dctalbot: I see, thanks for the explanation. Agreed, this was/is not most intuitive design. We could probably use an  empty object `{}` as an "empty config" for new temp schedules. That would keep the type as `TempSchedValue | null` or maybe `Partial<TempSchedValue> | null`, but not sure if any other refactoring would be required. 

I'm good to merge your changes here though, since nothing has fundamentally changed.

dctalbot (approved):

KatieMSB (approved):

PR #2040, opened 2021-11-23
ui/temp-sched: auto update historical shifts in list

KatieMSB:

**Description:**
Shifts in an existing temporary schedule now auto updates to historical shifts in the list if the shift ends while a user is editing the schedule.


spencerpauly (approved):
Looks good. If we ever need to reuse this component, add more complex interaction, or add a GraphQL query to this component or a child of it, we'll just have to remember that this re-renders every minute in some scenarios.

I think it works here though because (a) It's low in the component tree, within a dialog, and also not a component that's really meant to be reusable (at least today). 

Also (b) it only re-renders every minute. So likely shouldn't cause issues even though it possibly wastes some re-renders.

Forfold (dismissed):

Forfold (approved):

spencerpauly (approved):

PR #2047, opened 2021-12-01
ui/schedules: Convert schedule calendar components to TS

spencerpauly:

**Description:**
Converted the `ScheduleCalendar` and `ScheduleCalendarEventWrapper` to typescript and did any other tweaks needed to let that happen.

**Out of Scope:**
This could use a refactor to strip away some unnecessary data. I decided to keep functional changes minimal though and will possibly submit another PR for refactoring.

**Screenshots:**

None (hopefully)


web/src/app/schedules/calendar/ScheduleCalendar.tsx
```diff
@@ -268,14 +336,15 @@ function ScheduleCalendar(props) {
               fontFamily: theme.typography.body2.fontFamily,
               fontSize: theme.typography.body2.fontSize,
             }}
-            tooltipAccessor={() => null}
+            tooltipAccessor={() => ''}
```
spencerpauly, line 339: I can't find good docs for what this param is needed for, but it usually takes return type of string so I'm doing this. Should we remove this parameter or have it return something else?

web/src/app/schedules/calendar/ScheduleCalendar.tsx
```diff
@@ -123,81 +182,90 @@ function ScheduleCalendar(props) {
     )
   }
 
-  const getCalEvents = (shifts, _tempScheds, userOverrides) => {
-    const tempSchedules = _tempScheds.map((sched) => ({
-      start: sched.start,
-      end: sched.end,
+  const getCalEvents = (
```
spencerpauly, line 185: From what I can tell, an event is "Fixed' if it's a temp schedule event or temp schedule shift event. I just removed the terminology of "fixed" and made it more clear, but the term "fixed" may be being used in react-big-calendar. I couldn't find any docs mentioning it though so I'm assuming that's a term we're using outside of react-big-calendar.

web/src/app/schedules/calendar/ScheduleCalendar.tsx
dctalbot: https://jquense.github.io/react-big-calendar/examples/index.html#prop-tooltipAccessor string should be fine

web/src/app/schedules/calendar/ScheduleCalendar.tsx
dctalbot: Temporary Schedules as a feature used to be called Fixed Schedules, so this is leftover from that. Nice catch 👍🏻

web/src/app/schedules/calendar/ScheduleCalendar.tsx
```diff
@@ -123,81 +182,90 @@ function ScheduleCalendar(props) {
...
+    shifts: OnCallShift[],
+    _tempScheds: TemporarySchedule[],
+    userOverrides: UserOverride[],
+  ): ScheduleCalendarEvent[] => {
+    const tempSchedules: TempSchedEvent[] = _tempScheds.map((sched) => ({
+      type: 'tempSched',
+      start: new Date(sched.start),
+      end: new Date(sched.end),
```
dctalbot, line 193: I would be more comfortable approving this PR if it didn't change the implementation here. The native Date objects are hard to work with and reason about since they are not timezone-aware. If we are going to refactor, I think luxon's `DateTime` would be the way to go, but for this PR, the scope should be limited to introducing types.

web/src/app/schedules/calendar/ScheduleCalendarEventWrapper.tsx
```diff
@@ -34,35 +41,43 @@ const useStyles = makeStyles({
...
   }
 
-  function handleCloseShiftInfo() {
+  function handleCloseShiftInfo(): void {
     setAnchorEl(null)
   }
 
-  function handleKeyDown(event) {
-    const code = event.key
+  function handleKeyDown(_event: KeyboardEvent): void {
```
dctalbot, line 73: Same as above

web/src/app/schedules/calendar/ScheduleCalendarEventWrapper.tsx
```diff
@@ -71,52 +86,52 @@ export default function ScheduleCalendarEventWrapper({ children, event }) {
...
-        removeUserID: event.userID,
+        removeUserID: _event.userID,
       },
     })
   }
 
-  function renderTempSchedButtons() {
-    if (DateTime.fromISO(event.end) <= DateTime.utc()) {
+  function renderTempSchedButtons(
+    _event: TempSchedEvent | TempSchedShiftEvent,
```
dctalbot, line 95: Same as above

dctalbot (changes requested):
I really like the way you reasoned about the different calendar event types. I think most of the refactors here are in the right direction, but I'd try to limit them in PR's that are focused on types for the sake of review, especially when it comes to handling date/time. Left a few suggestions, let me know what you think!

web/src/app/schedules/calendar/ScheduleCalendar.tsx
spencerpauly: Yeah so that `new Date()` code here was in the original JavaScript version of these files. You can find it in line 192 of the original. The old PR essentially converted all these shifts to one format then converted them again into a generic format at the end of the getCalEvents function. When I converted it to typescript, it made more sense not to have that intermediary conversion step, and instead just to convert everything to the output format right away. Because of that I moved the new Date code from the end of the getCalEvents function to instead being in the mapping for each of the events individually.

I agree, I actually backtracked some changes to try and make this PR have less changes compared to it's JavaScript version. If you want I can backtrack some more and bring back the original format of the getCalEvents function then make a follow-up PR to bring back cleaner typings. Let me know if that would make it easier!

Forfold: I think the idea was to keep everything as an ISOTimeStamp since it's ultimately just a string that can be parsed easily everywhere, but the calendar has (or at least had at one point) a hard requirement to use Javascript Date objects so we would convert when we want to render that component

Forfold: Checking out what the Cal is doing in master, I see what you're saying about the string vs date thing. I didn't realize we were setting it as a string then converting it to a Date all within the same function, with nothing really happening in between, so that seems fine to me. I've made a couple updates though (since I know you're on a new team now) so we can try and get this merged soon

Forfold (dismissed):

Forfold (approved):

dctalbot (commented):
Circling back to what I was stuck on the first time reviewing - there is currently a bug in master on line 80:
`    if (DateTime.fromISO(event.end) <= DateTime.utc()) {`

but `event.end` is a native Date obj. 

This is solved (intentionally or not) in this PR on line 167:
`    if (DateTime.fromJSDate(calEvent.end) <= DateTime.utc()) {`

web/src/app/schedules/calendar/ScheduleCalendar.tsx
dctalbot: Got it, this makes sense to me, thanks for the explanation!

dctalbot (approved):

PR #2052, opened 2021-12-07
graphql: add debugMessages query

mastercactapus:
**Description:**
Adds a `debugMessages` query that can be used by an admin to retrieve a list of recent messages from the `outgoing_messages` table.

### DIFF
--- a/graphql2/graphqlapp/query.go
+++ b/graphql2/graphqlapp/query.go
@@ -134,7 +134,7 @@ func (a *Query) DebugMessages(ctx context.Context, input *graphql2.DebugMessages
 		if m.UserName != "" {
 			msg.UserName = &m.UserName
 		}
-		if m.Status.SrcValue != "" && !m.Dest.Type.IsUserCM() {
+		if m.Status.SrcValue != "" && m.Dest.Type.IsUserCM() {
 			src, err := a.formatDest(ctx, notification.Dest{Type: m.Dest.Type, Value: m.Status.SrcValue})
 			if err != nil {
 				return nil, fmt.Errorf("format src: %w", err)
--- a/graphql2/graphqlapp/query.go
+++ b/graphql2/graphqlapp/query.go
@@ -41,15 +41,14 @@ func (a *App) formatNC(ctx context.Context, id string) (string, error) {
 
 	return fmt.Sprintf("%s (%s)", n.Name, typeName), nil
 }
-
-func (a *DebugMessage) Destination(ctx context.Context, obj *notification.RecentMessage) (string, error) {
-	if !obj.Dest.Type.IsUserCM() {
-		return (*App)(a).formatNC(ctx, obj.Dest.ID)
+func (a *DebugMessage) formatDest(ctx context.Context, dst notification.Dest) (string, error) {
+	if !dst.Type.IsUserCM() {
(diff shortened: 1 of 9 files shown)


graphql2/graphqlapp/query.go
```diff
@@ -2,17 +2,124 @@ package graphqlapp
...
+		str.WriteString(obj.Status.Details)
+	}
+	return str.String(), nil
+}
+func (a *DebugMessage) Source(ctx context.Context, obj *notification.RecentMessage) (string, error) {
+	if obj.Status.SrcValue == "" {
+		return "", nil
+	}
+
+	return notification.Dest{Type: obj.Dest.Type, Value: obj.Status.SrcValue}.String(), nil
```
dctalbot, line 101: I don't think we should rely on the destination interface for formatting the source value. I sent a test message and got the following response:

Screen Shot 2021-12-07 at 4 25 22 PM

I would expect `source` to be the Twilio FromNumber

notification/status.go
```diff
@@ -70,3 +72,38 @@ const (
...
+		*s = StateSending
+	case "sent":
+		*s = StateSent
+	case "delivered":
+		*s = StateDelivered
+	case "failed":
+		*s = StateFailedPerm
+	case "stale":
+		*s = StateFailedTemp
+	default:
```
dctalbot, line 90: Is this switch missing the `bundled` case?

dctalbot (changes requested):

graphql2/graphqlapp/query.go
mastercactapus: Good catch, we should use the same formatting as destination and keep it as an empty string if value is blank.

notification/status.go
mastercactapus: Ooh that's a good call, we may need to add a new state. Normally "bundled" messages aren't processed at all, but they would still be returned here.

dctalbot (dismissed):

graphql2/graphqlapp/query.go
```diff
@@ -2,18 +2,161 @@ package graphqlapp
...
+			Status:      msgStatus(m.Status),
+			Destination: dest,
+		}
+		if m.UserID != "" {
+			msg.UserID = &m.UserID
+		}
+		if m.UserName != "" {
+			msg.UserName = &m.UserName
+		}
+		if m.Status.SrcValue != "" && !m.Dest.Type.IsUserCM() {
```
dctalbot, line 137:
```suggestion
		if m.Status.SrcValue != "" && m.Dest.Type.IsUserCM() {
```
We want source values for sms, correct?

dctalbot (changes requested):

dctalbot (approved):

arurao:
While functionally validating I am seeing some alertIDs being returned as null. I don't think `outgoing_messages` table allows has any null values in the alertID column. Is this valid then?

Screen Shot 2021-12-13 at 10 39 37 PM

mastercactapus:
@arurao only alert notifications and bundles have an alert ID, test, validation, on-call, etc.. don't. Alert status updates *should* but don't due to an unrelated bug

arurao (approved):

PR #2055, opened 2021-12-09
alert details: adjust escalation policy card details

### LINKED ISSUE #932: ui/alert logs: adjust time between escalations column
**Describe the Bug:**
After an escalation policy has fully processed an alert through all steps, the table from the alert details shows all values in the "Time Between Escalations" column as `—`. This should likely change to something more meaningful.

**Steps to Reproduce:**
1. Create an alert
2. Visit the alert's details
3. Wait for the alert to finish escalating (or manually escalate)
4. See results as shown below: 

**Screenshots/Stack Traces:**
image

**Possible Solutions:**
- Rename column title to "Escalation Info"
- Change `—` to the timestamp of when that step last escalated
- If a step is not currently active or has not been escalated yet, display "Awaiting escalation" or similar
- Truncate countdown text, in favor of terse wording, using exact seconds if < 1 minute remaining
    - i.e. `4m`, `1h`, or `48s`
- Display text after the countdown string, displaying as "4m until next escalation"
- Display countdown text of fully escalated steps as "Fully escalated" or similar (i.e. a step that fulfills the EP's repeat count)

Forfold:
**Description:**
- Removes the "Time Between Escalations" column on alert details
- Adds a "Last Escalated" note above the EP table
- Adds a "Next Escalation" note above the EP table
- Adds the option to use a `SlackLink` in a similar capacity as the other `AppLink`s; however, it redirects to the channel within the Slack app.

Closes #932

**Screenshots:**
Screen Shot 2021-12-09 at 9 00 37 AM


dctalbot (commented):
Looks great besides this peculiar issue in which the countdown stops counting down when there is an escalate event

https://user-images.githubusercontent.com/17692467/146261325-fadaf04d-b25a-44f0-ac88-e657b7a68844.mov

Forfold:
@dctalbot I don't think this is an issue, it's just the short amount of time while waiting for the next update from the network. We could change it to something else but then it could potentially feel awkward with a flicker of text.

Forfold:
Gotcha, I only noticed it hang on the "0" seconds in the video for a second at first glance, which is what I thought you were referring to. I didn't experience that issue myself so I'll do some additional testing.

Forfold:
@dctalbot The problem was that when the `<Countdown />` component finished counting down once, it flipped a bool to true and would stop counting down. Adding the `overtime` prop overrides this but causes the component to oddly count _up_ if completed, so I instead changed this to "Escalating..." while waiting on the network, since before it was just hanging at 0 until the next poll/request comes in which would normally be fine.

KatieMSB (approved):

PR #2061, opened 2021-12-20
ui: admin message logs

Forfold:
This PR adds a new route to the admin pages, `/admin/logs` to display Outgoing Logs based on the schema changes introduced in #2052.

The first 1,000 results are being pulled in and are paginated via a load more button when scrolling to the bottom of the list for this first pass. We are also utilizing a new dependency, Fuse.js, for fuzzy searching since the data is not filtered within the GraphQL layer.

Additional work on the layout can be optimized in a future release.

It also adds the ability to use an optionally render `AppLink` with an icon.

**Known issues / Next Up:**
* [x] Needs test coverage
* [x] Discuss search keys
* [ ] Discuss infinite loading (load size, functionality, etc)
* [ ] Mobile optimization needed still (for filter/search area)
* [ ] Make logs list virtualized
* [ ] Possibly add additional filters dropdown
* [ ] Possibly narrow card list max-width
* [ ] Fix ISOPicker

**Screenshots:**

Screen Shot 2021-12-22 at 10 16 38 AM

Screen Shot 2021-12-22 at 10 16 51 AM


mastercactapus:
Doesn't seem to work when there are a bunch of sent messages:
image

mastercactapus:
Thoughts on search that I think are needed for some of the common use cases:
- partial contact method values (e.g. email domain, phone number fragment, substrings in general)
- contact method type in search bar (e.g. search for "SMS")
- total count (e.g. "showing 50 of 259 results" "showing 100 of 1000+ results")

Suggestions/ideas:
- maybe break up terms by comma? (e.g. `sms + pending` to find pending SMS)
- icons for message types instead of just text (or instead)

---

Here are some of the questions we'd like to be able to answer with this tool:
- How many pending messages are there?
- How many alert notifications sent in the last hour?
- How many SMS messages per service, etc..
- Was a phone call sent between 3 and 4pm to this user?
- When was the last message to this contact method/phone number?
- What's the provider-specific ID for this message? (e.g., slack ID, twilio ID)
- What's the GoAlert message ID for this message?

I think most of these are covered already, just placing for reference

web/src/app/util/AppLink.tsx
```diff
@@ -1,15 +1,17 @@
 import React, { forwardRef, ForwardRefRenderFunction } from 'react'
 import { Link, LinkProps, useLocation } from 'react-router-dom'
 import joinURL from './joinURL'
+import { OpenInNew } from '@mui/icons-material'
 
 export interface AppLinkProps extends LinkProps {
   to: string
   newTab?: boolean
+  icon?: boolean
```
dctalbot, line 9: This feels like a mix of concerns, I'd recommend keeping the logic for this outside of this component e.g. 
```
<div><AppLink /><Icon /></div>
```
would be more composable

web/src/app/admin/admin-outgoing-logs/OutgoingLogCard.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  let statusStyles
+  const s = status.toLowerCase()
+  if (s.includes('deliver')) statusStyles = statusDict.success
+  if (s.includes('fail')) statusStyles = statusDict.error
+  if (s.includes('temp')) statusStyles = statusDict.warning
+  if (s.includes('pend')) statusStyles = statusDict.info
+
+  console.log('selected: ', selected)
+
```
dctalbot, line 70:
```suggestion
```

dctalbot (changes requested):

web/src/app/util/AppLink.tsx
Forfold: The idea was to have the icon be in a single element for when the mouse pointer changes to a hand. We can probably do one of these but I feel like it would be an API enhancement overall.

Forfold:
@mastercactapus 

- [x] partial contact method values (e.g. email domain, phone number fragment, substrings in general)
- [x] contact method type in search bar (e.g. search for "SMS")
- [ ] total count (e.g. "showing 50 of 259 results" "showing 100 of 1000+ results")

We were able to handle searching by type and partials by setting "ignoreLocation" to true with Fuse. I'm not sure if Fuse allows for searching by comma-separated keys. @spencerpauly any ideas/are you familiar?

mastercactapus:
@Forfold we can talk about the search bits and see if we want to go that route for now. In that case, we'd split the search terms ourselves and implement AND/OR logic as needed I think -- but might be better to leave it out for the initial merge

spencerpauly:
@mastercactapus @Forfold If we want additional search power for things such as AND, OR or NOT operators we could use the extended search mode in fuse.js. Here's the docs for it: https://fusejs.io/examples.html#extended-search. I'm testing it out right now and it seems to work well.

web/src/app/admin/admin-outgoing-logs/OutgoingLogCard.tsx
```diff
@@ -0,0 +1,115 @@
...
+    },
+    info: {
+      backgroundColor: '#E8F4FD',
+      color: '#0E3C61',
+    },
+  }
+
+  let statusStyles
+  const s = status.toLowerCase()
+  if (s.includes('deliver')) statusStyles = statusDict.success
```
mastercactapus, line 64: This can include `sent` which is also a final-state -- delivery notification isn't always supported

web/src/app/admin/admin-outgoing-logs/OutgoingLogsControls.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+export default function OutgoingLogsControls(p: Props): JSX.Element {
+  const classes = useStyles()
+
+  const [start, setStart] = useURLParam<string>('start', '')
+  const [end, setEnd] = useURLParam<string>('end', '')
+  const [key, setKey] = useState(0)
+  const resetDateRange = useResetURLParams('start', 'end')
+  const [limit] = useURLParam<string>('limit', '1')
+  const _limit = parseInt(limit, 10)
```
mastercactapus, line 34: We should invert this, the convention is to use the `_` prefix as temporary meaning "don't use" and the non-prefixed in the rest of the function body.

Better yet, passing `1` instead of `'1'` as the default value should cause `useURLParam` to return an int without parsing. It will convert the parameter to whatever type is passed in as default.

web/src/app/admin/admin-outgoing-logs/OutgoingLogsControls.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  const resetFilters = (): void => {
+    resetDateRange()
+    setKey(key + 1)
```
mastercactapus, line 38: What's the purpose of `setKey` here? Forcing remount due to a bug?

web/src/app/admin/admin-outgoing-logs/OutgoingLogsControls.tsx
```diff
@@ -0,0 +1,98 @@
...
+        justifyContent='flex-end'
+        alignItems='flex-end'
+        sx={{ width: 'fit-content' }}
+      >
+        <Grid item>
+          <Search />
+        </Grid>
+        <Grid item>
+          <Typography color='textSecondary'>
+            Showing {_limit * LOAD_AMOUNT} of {p.totalCount} results
```
mastercactapus, line 92: If `p.totalCount` is 1000 we should display `1000+` or `at least 1000` since there are more messages that will not be returned by the server (we set the limit to 1000 in the parent).

web/src/app/admin/admin-outgoing-logs/OutgoingLogsList.tsx
```diff
@@ -0,0 +1,119 @@
...
+      ignoreLocation: true,
+    },
+  })
+
+  let filteredResults = results.slice() // copy results array
+  filteredResults = filteredResults.filter((result) => {
+    if (!start && !end) return true
+
+    const startDT = DateTime.fromISO(start)
+    let endDT = DateTime.fromISO(end)
```
mastercactapus, line 41: We should move this stuff outside/before the loop so we're not parsing them for each item.

web/src/app/admin/admin-outgoing-logs/OutgoingLogsList.tsx
```diff
@@ -0,0 +1,119 @@
...
+    const createdAtDT = DateTime.fromISO(result.item.createdAt)
+
+    if (start && !end) {
+      endDT = DateTime.now()
```
mastercactapus, line 45: We should update the logic to ignore end time if it's not set instead of using `DateTime.now()` -- this is going to potentially filter items based on the user's clock & clock drift when the input field is blank. There should be no filtering on end time if the value is empty/null.

web/src/app/admin/admin-outgoing-logs/OutgoingLogsList.tsx
```diff
@@ -0,0 +1,119 @@
...
+  onSelect: (debugMessage: DebugMessage) => void
+}
+
+export default function OutgoingLogsList(props: Props): JSX.Element {
+  const { debugMessages = [], selectedLog, onSelect } = props
+
+  const [searchTerm] = useURLParam('search', '')
+  const [start] = useURLParam('start', '')
+  const [end] = useURLParam('end', '')
+  const [limit, setLimit] = useURLParam<string>('limit', '1')
```
mastercactapus, line 24: Same as elsewhere, remove the `_limit` and pass an integer as the default value to have the type always be an integer.

Secondly, I think this might make sense to be the total count -- as a URL parameter, I think it's going to make more sense to a user to see `limit=100` meaning 100 results than `limit=2` meaning 100 results.

We should change it elsewhere to keep consistent

```suggestion
  const [limit, setLimit] = useURLParam<string>('limit', LOAD_AMOUNT)
```

web/src/app/admin/admin-outgoing-logs/OutgoingLogsList.tsx
```diff
@@ -0,0 +1,119 @@
...
+  }, [searchTerm, start, end])
+
+  // set search within fuse on search change
+  useEffect(() => {
+    setSearch(searchTerm)
+  }, [searchTerm])
+
+  // what appends stuff to results
+  function onNext(): void {
+    setLimit((_limit + 1).toString())
```
mastercactapus, line 69:
```suggestion
    setLimit(limit + LOAD_AMOUNT)
```

With the above change

web/src/app/admin/admin-outgoing-logs/OutgoingLogsList.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  return (
+    <Box
+      display='flex'
+      flexDirection='column'
+      alignItems='stretch'
+      width='full'
+    >
+      {filteredResults
+        .slice(0, _limit * LOAD_AMOUNT)
```
mastercactapus, line 80: With the above
```suggestion
        .slice(0, limit)
```

web/src/app/admin/admin-outgoing-logs/OutgoingLogsList.tsx
```diff
@@ -0,0 +1,119 @@
...
+        .map(({ item: debugMessage }) => (
+          <OutgoingLogCard
+            key={debugMessage.id}
+            debugMessage={debugMessage}
+            selected={selectedLog?.id === debugMessage.id}
+            onSelect={() => onSelect(debugMessage)}
+          />
+        ))}
+      {_limit * LOAD_AMOUNT < filteredResults.length ? (
```
mastercactapus, line 89: Same
```suggestion
      {limit < filteredResults.length ? (
```

web/src/app/admin/admin-outgoing-logs/hooks.ts
```diff
@@ -0,0 +1,70 @@
...
+  }, [search, fuse, data, keys, options])
+
+  useEffect(() => {
+    async function set(): Promise<void> {
+      if (fuse.current) {
+        setFuseResults(await fuse.current.search(search))
+      }
+    }
+    set()
+  }, [search, fuse, data])
```
mastercactapus, line 58: async/await doesn't seem necessary here, we're not returning the result of `set()` or using it recursively. We also need to handle cancellations or we'll get problems with unmount at the wrong time or race conditions with results being called out-of-order.

```suggestion
  useEffect(() => {
    if (!fuse.current) return

    let callback = setFuseResults
    fuse.current.search(search).then(results => callback(results))
    
    // reset callback before unmount/update
    return () => {callback = ()=>{}}
  }, [search, fuse, data])
```

mastercactapus (changes requested):

web/src/app/admin/admin-outgoing-logs/OutgoingLogsControls.tsx
spencerpauly: Fixed in ed10afd

web/src/app/admin/admin-outgoing-logs/OutgoingLogsControls.tsx
spencerpauly: I went in and inverted that. I'm getting an error using a number type for the useURLParam. Looks fixable but likely a bug in the hook.

Screen Shot 2021-12-28 at 11 28 28 AM
m

web/src/app/admin/admin-outgoing-logs/OutgoingLogsControls.tsx
spencerpauly: Yep, this is to get around a bug. The ISODateTimePicker doesn't seem to update when you change it's value. This was a temporary fix to allow changing the ISODateTimePicker's value as if it were a controlled component by just completely re-rendering it. I tried seeing if I could find where the issue is in the ISO picker but it wasn't an easy fix so I added this for now.

web/src/app/admin/admin-outgoing-logs/OutgoingLogsControls.tsx
dctalbot: This is a bug I found when refactoring the routing -- solved in #2067. Currently, we don't handle numeric parameters.

Forfold: Yeah, it was only returning a string when I implemented which is why I had everything converting to strings at the moment. As for the `_limit` i did that so i didn't have to rename the urlParam to be different, but makes sense to use the standard convention anyways.

web/src/app/admin/admin-outgoing-logs/OutgoingLogsControls.tsx
mastercactapus: I see that seems reasonable. Throw a comment in with a brief explanation for why it's necessary so it doesn't get removed later

web/src/app/admin/admin-outgoing-logs/OutgoingLogCard.tsx
spencerpauly: Added 👍

web/src/app/admin/admin-outgoing-logs/OutgoingLogsControls.tsx
spencerpauly: Added 👍

web/src/app/admin/AdminRouter.js
```diff
@@ -21,6 +22,8 @@ function AdminRouter() {
       <Route exact path='/admin/config' component={AdminConfig} />
       <Route exact path='/admin/limits' component={AdminLimits} />
       <Route exact path='/admin/toolbox' component={AdminToolbox} />
+      <Route exact path='/admin/logs' component={AdminOutgoingLogs} />
```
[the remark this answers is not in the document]
Forfold: Sounds good. I also went in and renamed the files starting with files `OutgoingLog` to `DebugMessage` instead so they match the frontend API

web/src/app/util/AppLink.tsx
dctalbot: Got it, I think that would be better handled using css. I had a go at it this morning on the `message-logs-frontend-newtab` branch, would love to hear your thoughts

Forfold: I disagree, but sure.

Forfold: Pushed up a solution using CSS.

dctalbot (dismissed):

web/src/app/admin/admin-message-logs/DebugMessagesList.tsx
```diff
@@ -0,0 +1,122 @@
...
+    const createdAtDT = DateTime.fromISO(result.item.createdAt)
+    if (startDT && startDT > createdAtDT) return false
+    if (endDT && endDT < createdAtDT) return false
+    return true
+  })
+
+  // reset page load amount when filters change
+  useEffect(() => {
+    onResetLoadMore()
+  }, [searchTerm, start, end])
```
mastercactapus, line 70: This seems off, this component is triggering an on-change for fields it does not control (i.e., start and end are never changed here) also, it's being called on-render instead of on the change event. It makes it hard to follow the control flow, causes multiple renders for a single change, and can become a loop if `onResetLoadMore` changes any values.

We should change this to only be called on a change event. In this case though, there are no events being triggered here, only passed in, and `useFuse` should be updated to accept the search term instead of needing to be kept in-sync with calls to `setSearch`

web/src/app/admin/admin-message-logs/useFuse.ts
```diff
@@ -0,0 +1,66 @@
...
+    showResultsWhenNoSearchTerm && search === ''
+      ? data.map((data, i) => ({
+          item: data,
+          score: 1,
+          refIndex: i,
+        }))
+      : fuseResults
+
+  return { results, search, setSearch }
+}
```
mastercactapus, line 66: We should make this stateless so we don't need to call `setSearch` all the time by just passing in the query directly. If we need to make it stateful, then let's do that 100% inside `useFuse` so that everything can be clean outside of here.

```suggestion
export function useFuse<T>({
  data,
  keys,
  search = DEFAULT_QUERY,
  options = {},
}: FuseParams<T>): FuseResults<T> {
  const { showResultsWhenNoSearchTerm, ...fuseOptions } = options
  const [fuseResults, setFuseResults] = useState<Fuse.FuseResult<T>[]>([])
  const fuse = useRef<Fuse<T>>()

  useEffect(() => {
    if (!data) return

    fuse.current = new Fuse(data, {
      ...defaultOptions,
      ...fuseOptions,
      keys,
    })
  }, [search, fuse, data, keys, options])

  useEffect(() => {
    if (!fuse.current) return
    setFuseResults(fuse.current.search(search))
  }, [search, fuse, data])

  const results =
    showResultsWhenNoSearchTerm && search === ''
      ? data.map((data, i) => ({
          item: data,
          score: 1,
          refIndex: i,
        }))
      : fuseResults

  return results
}
```

web/src/app/admin/admin-message-logs/AdminDebugMessagesLayout.tsx
```diff
@@ -0,0 +1,130 @@
...
+    minWidth: 250,
+  },
+}))
+
+export default function AdminDebugMessagesLayout(): JSX.Element {
+  const classes = useStyles()
+
+  // all data is fetched on page load, but the number of logs rendered is limited
+  const [numRendered, setNumRendered] = useState(LOAD_AMOUNT)
+  const [selectedLog, setSelectedLog] = useState<DebugMessage | null>(null)
```
mastercactapus, line 71: Would it be possible to take the URL params here and filter the items we get from data? It seems like we're getting the data here, passing it down, and then asking for events & info back from filtering up the tree causing multiple renders and making the data flow more complicated than necessary.

Maybe if we use fuse here, and just pass the total/filtered/showing counts down we can bypass that. For change events we can pass event handlers to `DebugMessageControls`

Maybe something like this to decouple things a bit:
```jsx
<DebugMessageControls search={search} onSearchChange={setSearch} />
```

Then `DebugMessagesList` can just render the messages we pass to it and have a couple of props for "load more" events but all the processing/control will happen here in the component in charge of getting the data.

mastercactapus (changes requested):

web/src/app/admin/admin-message-logs/AdminDebugMessagesLayout.tsx
spencerpauly: I looked into this when making the PR, but moving the hooks up to the parent component would be difficult since that parent renders conditionally depending on network data. The solution would be to add another component in-between, so the structure would look like this:
```
             AdminDebugMessagesLayout (Network requests & conditionally render page)
                                                               |
                AdminDebugMessages (Hold state and do all the filtering/limiting/sorting)
                         /                                     |                                \
DebugMessageControls  DebugMessagesList   DebugMessageDetails (Mostly stateless children components)
```
I decided against it just because it wasn't totally necessary and we could get this PR out faster. We did plan to make a V2 PR though to clean up some things, and we can make this change in there.

Forfold: The idea in the current implementation was that the controls handles all filtering/searching/controlling what is shown in the list, and the layout would set the page layout based on loading/not. I can see how having to call the URL param hooks more than once can get cumbersome, but overall I think it makes sense rather than doing everything inside of the Layout file and bogging it down- since it would feel awkward to pass down the search/setSearch as props when we can just grab it with the `useURLParam` hook wherever we need it. I agree though that maybe a HOC in as a middle layer can control it all instead, if there is an issue with the page re-rendering too many times. That would bring the setSearch logic closer to the filtering logic as well.

Forfold: Layout (queries for data then render loading/error/data)
↪ DebugMessages (controls all the data to filter based on URL Params, as well as the selected log)
↪ Controls (setters passed down to buttons) ↔ List (renders data) ↔ Details (renders more data based on selected log)

mastercactapus (approved):

dctalbot (dismissed):

mastercactapus (approved):

dctalbot (approved):

PR #2067, opened 2021-12-22
ui: refactor `useURLParam` and `useResetURLParams`

dctalbot:

**Description:**
This PR continues our effort to reduce our dependency on redux for client side routing by:
- refactoring `useURLParam` to use the `react-router` API's directly
- writing unit tests for _getting_ a URL param and sanitizing param values pre-_setting_ them
- refactoring `useURLParam` to share logic in `setURL`, and to not clear all params by default if an empty list is passed (too much magic, unrealistic use case)

**API changes:**
- Introduce a new `useURLParams` hook that accepts an object of name-defaultValue pairs, and returns a `[params, setParams]` array. This prevents issues that crop up when invoking `setParams` synchronously in a single render frame. This case is further handled by aborting and logging an error to the console.
- An example of this issue and its solution is provided in this PR in the `schedules/calendar` package.
- `useURLParam` now uses `useURLParams` under the hood.

Bug fixes:
- `sanitizeParam` did not handle numeric values correctly, and it didn't trim string values when given a `string[]`
- a few usages of `useURLParam` were not passing in the required `defaultValue` argument

### DIFF
--- a/web/src/app/actions/hooks.ts
+++ b/web/src/app/actions/hooks.ts
@@ -25,28 +25,28 @@ export function sanitizeURLParam(value: Value): string | string[] {
   }
 }
 
-// getParamValues converts each value from string|string[]
-// into the desired type based on the respective default value.
+// getParamValues converts each URL search param into the
+// desired type based on its respective default value.
 export function getParamValues<T extends Record<string, Value>>(
   location: Location,
-  params: T,
+  params: T, // <name, default> pairs
 ): T {
   const result = {} as Record<string, Value>
   const q = new URLSearchParams(location.search)
 
-  for (const [k, defaultv] of Object.entries(params)) {
-    if (!q.has(k)) {
-      result[k] = defaultv
-    } else if (Array.isArray(defaultv)) {
-      result[k] = q.getAll(k)
-    } else if (typeof defaultv === 'boolean') {
-      result[k] = q.get(k) === '1'
-    } else if (typeof defaultv === 'string') {
-      result[k] = q.get(k) || ''
-    } else if (typeof defaultv === 'number') {
-      result[k] = +(q.get(k) as string)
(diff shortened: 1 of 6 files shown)


pnengchu (approved):

web/src/app/actions/hooks.ts
```diff
@@ -1,46 +1,111 @@
...
+  if (!url.has(name)) return defaultValue
+  if (Array.isArray(defaultValue)) return url.getAll(name) as T
+  if (typeof defaultValue === 'boolean') return (url.get(name) === '1') as T
+  if (typeof defaultValue === 'string') return (url.get(name) || '') as T
+  if (typeof defaultValue === 'number') return +(url.get(name) as string) as T
+  return defaultValue
+}
+
+// setURL will replace the latest browser history entry with the provided config.
+function setURL(
```
mastercactapus, line 44: nit: can we call this `setParams` or similar?

Also, order the args by importance/change frequency: history, location, params
history is our fixed API for updating, we use location as a reference, and params is the real argument we're using to update the search string.

web/src/app/actions/hooks.ts
```diff
@@ -1,46 +1,111 @@
...
-
-  function setValue(newValue: T): void {
-    if (window.location.pathname !== urlPath) {
-      warn(
-        'useURLParam was called to set a parameter, but location.pathname has changed, aborting',
-      )
-      return
+  const location = useLocation()
+  const history = useHistory()
+  const url = new URLSearchParams(location.search)
```
mastercactapus, line 70: nit: should probably call this `params` here and elsewhere since it's not the full URL

mastercactapus (changes requested):

mastercactapus (approved):

mastercactapus (approved):

PR #2068, opened 2021-12-23
graphql: batch notification channel requests

mastercactapus:
**Description:**
Causes notification channel lookups to be batched by a new dataloader. Fixes issue in #2061 where too many logs resulted in access denied due to a recursive lookup.

### DIFF
--- a/dataloader/ncloader.go
+++ b/dataloader/ncloader.go
@@ -35,7 +35,7 @@ func (l *NCLoader) FetchOne(ctx context.Context, id uuid.UUID) (*notificationcha
 		return nil, err
 	}
 	if v == nil {
-		return nil, err
+		return nil, nil
 	}
 	return v.(*notificationchannel.Channel), nil
 }
--- a/dataloader/ncloader.go
+++ b/dataloader/ncloader.go
@@ -1,0 +1,53 @@
+package dataloader
+
+import (
+	"context"
+	"time"
+
+	"github.com/target/goalert/notificationchannel"
+)
+
+// NCLoader will load notification channels from postgres.
+type NCLoader struct {
+	*loader
+	store notificationchannel.Store
+}
+
+// NewNCLoader will create a new CMLoader using the provided store for fetch operations.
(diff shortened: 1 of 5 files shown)


dataloader/ncloader.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/target/goalert/notificationchannel"
+)
+
+// NCLoader will load notification channels from postgres.
+type NCLoader struct {
+	*loader
+	store notificationchannel.Store
+}
+
+// NewNCLoader will create a new CMLoader using the provided store for fetch operations.
```
dctalbot, line 17:
```suggestion
// NewNCLoader will create a new NCLoader using the provided store for fetch operations.
```

dataloader/ncloader.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// FetchOne will fetch a single record from the store, batching requests to the store.
+func (l *NCLoader) FetchOne(ctx context.Context, id uuid.UUID) (*notificationchannel.Channel, error) {
+	v, err := l.loader.FetchOne(ctx, id.String())
+	if err != nil {
+		return nil, err
+	}
+	if v == nil {
+		return nil, err
```
dctalbot, line 38:
```suggestion
		return nil, nil
```

dctalbot (approved):

notificationchannel/store.go
```diff
@@ -16,6 +17,7 @@ import (
 type Store interface {
 	FindAll(context.Context) ([]Channel, error)
 	FindOne(context.Context, uuid.UUID) (*Channel, error)
+	FindMany(context.Context, []string) ([]Channel, error)
```
Forfold, line 20: Should this take an array of uuids instead, or is string sufficient?

mastercactapus: I think string is okay for now, our validation and `sqlutil` stuff expects strings when dealing with multiple (same as `DeleteManyTx`) eventually we should bring everything to UUID, but for now I followed the existing convention in this package to keep it focused on the fix until we can refactor the rest.

Forfold (approved):

PR #2070, opened 2021-12-27
Bump Go dependencies

mastercactapus:
**Description:**
Updates direct Go dependencies.

**Notes:**
- Add missing makefile target dependency
- replaced deprecated `grpc.WithInsecure()` usage with `insecure.NewCredentials()`
- removed invalid deprecated flag in gql schema
- removed unused internal code (`internal/*` and `sqlhighlight.go`)
- removed unnecessary conditional in `carrier.go` (linter update catch)
- updated protoc version
- replace jwt deprecated `StandardClaims` usage with `RegisteredClaims`


pnengchu (approved):

graphql2/schema.graphql
```diff
@@ -503,7 +503,7 @@ input ScheduleTargetInput {
 }
 
 input ScheduleRuleInput {
-  id: ID @deprecated(reason: "ignored")
+  id: ID
```
Forfold, line 506: Are we un-deprecating this or should we instead now remove the field?

mastercactapus: It's not currently merged into the official spec (https://github.com/graphql/graphql-spec/pull/805) and gets flagged as invalid with the latest gqlparser where they added validation https://github.com/99designs/gqlgen/issues/1636

Since it caused the build to fail, currently not allowed, and the field is ignored (rather than having some legacy behavior) I just removed it from the schema. Open to alternatives, but it seemed very low risk

Forfold (approved):

PR #2074, opened 2021-12-28
notification: add alert summary to status updates

### LINKED ISSUE #2056: Add option to include original summary/details in alert status updates
Our network engineers complained that they can't tell which alert closed at a glance when they have numerous active alerts. I'd like to ask that you provide the option to include the summary and/or details from the original alert in subsequent status updates. I suggested they could follow the link in the SMS to the GoAlert web UI and see which alert was closed, but they said this was too inconvenient to do for each alert when there are many alerts.

Forfold:
**Description:**
This PR adds the alert summary to status update messages. Status updates show the alert ID, but do not show the summary or provide a link to the alert, which can make the investigation of an alert cumbersome.

The current alert SMS template is conditionally rendering alert, alert bundle, and alert status messages, so I've made three explicit templates that can be leveraged when setting the type on the `alertSMS` struct before rendering. This should help with maintenance should we want to change any of the templates again.

Closes #2056


notification/twilio/alertsms.go
```diff
@@ -20,30 +21,39 @@ import (
...
+
 {{- if .Link }}
+{{.Link}}
+{{end}}
+
+{{- if .Code}}
+Reply '{{.Code}}a' to ack, '{{.Code}}c' to close.{{end}}`))
+
+// bundleTempl uses the Count, Body, Link, and Code to render a message
+var bundleTempl = template.Must(template.New("alertSMS").Parse(`Svc '{{.Body}}': {{.Count}} unacked alert{{if gt .Count 1}}s{{end}}
```
mastercactapus, line 44: we should change the template name for error messages

notification/twilio/alertsms.go
```diff
@@ -125,7 +152,7 @@ func (a alertSMS) Render(maxLen int) (string, error) {
 		}
 		a.Body = strings.TrimSpace(a.Body[:newBodyLen])
```
mastercactapus, line 153: We're going to need to have an extra/alternate version of this that starts trimming the summary for status messages before resorting to trimming the log entry

mastercactapus (changes requested):

notification/twilio/alertsms.go
Forfold: Would that be something like
```
newSummaryLen := len(a.Summary) - (buf.Len() - maxLen)
if newSummaryLen <= 0 {
	return "", errors.New("message too long to include summary")
}
a.Summary = strings.TrimSpace(a.Summary[:newSummaryLen])
```
within the same if statement before attempting to execute a second time? Or should we have a new if-statement entirely with it's own re-attempt?

mastercactapus: I'd say since today we have no summary, we'd trim the summary down until it's empty, and if we needed to keep going, then trim the body. We can do that with two separate count values.

Additionally:
- Status template should be updated so that if the summary is empty we omit the colon & space
- Status template needs to handle when the `Count` value is non-zero.

notification/twilio/alertsms.go
```diff
@@ -19,31 +20,29 @@ import (
...
+Reply '{{.Code}}a' to ack, '{{.Code}}c' to close.{{end}}`))
+
+var bundleTempl = template.Must(template.New("alertBundleSMS").Parse(`Svc '{{.ServiceName}}': {{.Count}} unacked alert{{if gt .Count 1}}s{{end}}
 
-Reply '{{.Code}}a{{if .Count}}a{{end}}' to ack{{if .Count}} all{{end}}, '{{.Code}}c{{if .Count}}c{{end}}' to close{{if .Count}} all{{end}}.
-{{- end}}`,
-))
+{{- if .Link }}
+
+	{{.Link}}
```
Forfold, line 37: A lot of the tests require the messages to have extra returns (`\n`) or tabs (`\t`). Do we want to retain this?

mastercactapus (dismissed):

notification/twilio/alertsms.go
dctalbot: I'd say yes, I like the spacing personally

Forfold: 😢

dctalbot (changes requested):
Unable to run `make start`, looks like there are some unintended changes to go.mod? Got it to work using `git checkout master -- go.mod`

Also, the spacing is a bit odd now that I see it, I like the separation with the newlines, but the indentation isn't necessary IMO. If we do want to change, could go in a new issue potentially. For reference, here is my UX for an integration I named "web int":


Screen Shot 2022-02-08 at 9 54 38 AM

Forfold:
@dctalbot what version of Go are you on? I had to update to 1.17.

Forfold:
I see, well it's a result of `make generate` (see last commits) so it should probably go in

dctalbot (approved):

mastercactapus (approved):

PR #2080, opened 2022-01-03
dev: add runproc tool for dev use cases

mastercactapus:
**Description:**
This PR adds the `runproc` tool which is a simpler alternative to runjson that uses a Procfile-style input instead of JSON. This is intended to make it easier for devs to add extra commands (e.g. ngrok or sendit) and maintain. Additionally the output uses colored prefix and original output which should be a better experience.

It also handles shutdown better/faster as well as the startup sequence lowering the amount of time a dev needs to wait.

image

**Out of Scope**

Only `make start` is using the new tool currently. `start-prod` and the cypress test commands are untouched and will be updated in further PRs with additional testing & validation.


devtools/runproc/signals.go
```diff
@@ -0,0 +1,5 @@
+package main
```
dctalbot, line 1: Curious why this needs its own file?

devtools/runproc/signals_unix.go
```diff
@@ -0,0 +1,9 @@
+package main
+
+import (
+	"golang.org/x/sys/unix"
+)
+
+func init() {
+	shutdownSignals = append(shutdownSignals, unix.SIGTERM)
```
dctalbot, line 8: Why not in-line this with `shutdownSignals` declaration?

dctalbot (approved):
This looks awesome 🤩

devtools/runproc/signals.go
mastercactapus: platform differences. `signals.go` builds on all platforms and has the base declarations, `signals_unix.go` only builds on unix-systems and includes `unix.SIGTERM` which isn't defined on things like windows (and would fail to compile otherwise)

m17ch (approved):
lgtm 👍  
functionally validated

PR #2085, opened 2022-01-06
ui/services: Alert metrics dashboard

dctalbot:

**Description:**
This PR is the first iteration of the Alert Metrics dashboard feature. It provides:
- a new Service Details quick link page that displays metrics for a given service
- a graph of alert counts, grouped by date in the user's system timezone (shows 100 alerts max). 
- a table of alert details
- a date range filter to view the past 1, 2, 3, or 4 weeks of alerts.
- a CSV export button

**Screenshots:**
Screen Shot 2022-01-10 at 12 12 15 PM


web/src/app/selection/QuerySelect.js
```diff
@@ -130,6 +130,8 @@ export const querySelectPropTypes = {
   multiple: p.bool,
   name: p.string,
   placeholder: p.string,
+
+  label: p.string,
```
Forfold, line 134: Is this needed? It looks like it's not being used within the file since it's bunched in with the `otherProps` spread

web/src/app/services/AlertMetrics/AlertMetricsCSV.tsx
```diff
@@ -0,0 +1,70 @@
...
+})
+
+export default function AlertMetricsCSV(
+  props: AlertMetricsCSVProps,
+): JSX.Element {
+  const classes = useStyles()
+  const zoneAbbr = DateTime.local().toFormat('ZZZZ')
+  // Note: the data object is ordered
+  const data = props.alerts.map((a) => ({
+    [`createdAt (${zoneAbbr})`]: DateTime.fromISO(a.createdAt).toLocal().toSQL({
```
Forfold, line 34: So this one is interesting, it shows the full value in when i click on a cell but just the `minutes:seconds.milliseconds` in the cell preview. Is there a way around that?

Screen Shot 2022-01-17 at 11 38 37 AM

Forfold (changes requested):

web/src/app/selection/QuerySelect.js
dctalbot: It's left over from when we had the ServiceSelect filter, so it's needed for that, but not for this PR. I'll remove

web/src/app/services/AlertMetrics/AlertMetricsCSV.tsx
dctalbot: That would be an Excel setting, and might depend on the locale/system settings. I think you can double click the cell to expand, or find this formatting drop down:
Screen Shot 2022-01-18 at 11 03 33 AM

Behind the scenes, the raw format we give it is `YYYY-MM-DD HH:MM:SS`

Forfold (approved):
LGTM! Loads well with a clean console  🎉  Functionally tested exporting a CSV of alerts as well.

arurao (approved):
LGTM. Functionally validated. Nice work!

PR #2086, opened 2022-01-06
ui/cypress: admin message logs

Forfold:
This PR adds cypress tests for the admin outgoing message logs PR (#2061). I have the base of this PR to merge into that branch: `message-logs-frontend`, though we can change it to master if that gets merged in first.

Out of scope:
- Load more test w/ support function to create many logs at once
- Asserting destination (phone numbers are given with a different format from DB, notification channels are evaluated differently)
- Provider ID/Source info
- It would be nice to have a one-to-one API to create a DebugMessage and return a DebugMessage. Would need to do work around message type + status and picking the correct strings if one is not provided in options. e.g. if `type: 'failed'` is given, find a good way return `c.pickOne([possible failed status message strings])` for the status.


arurao (approved):
LGTM

.github/dependabot.yml
```diff
@@ -11,3 +11,8 @@ updates:
       # ignore all patch updates
       - dependency-name: '*'
         update-types: ['version-update:semver-patch']
+
+      # @date-io/luxon@1 requires luxon@1
+      - dependency-name: 'luxon'
+      # luxon@1 requires @types/luxon@1
+      - dependency-name: '@types/luxon'
```
mastercactapus, line 18: Merge issue? I see there are Makefile changes too

Forfold: Ah I thought I fixed it with the revert. Yeah

Forfold: Not sure why it's all showing up since I merged `message-logs-frontend` into this branch. I reverted the accidental master merge

Forfold: Should be fixed now. I was trying to get the changes from #2088 into this so it could pass on CI

PR #2091, opened 2022-01-14
dev: switch remaining targets to runproc

mastercactapus:
**Description:**
Switches remaining make targets from `runjson` to `runproc` and add support for supplementary git-ignored Proc-files with an added `.env` extension.

Affected make targets to validate:
- `make start-prod`
- `make cy-wide`
- `make cy-wide-prod`
- `make cy-wide-prod-run`
- `make cy-mobile`
- `make cy-mobile-prod`
- `make cy-mobile-prod-run`


mastercactapus:
Yeah, explicit may be better. What about `-l Procfile.local` and avoid `.env` entirely so it doesn't get confused? `-l` files can be treated as optional (and won't fail if it doesn't exist) rather than having an auto-magic extension that needs to be understood. Plus we can use the same `.local` file for `make start` and `make start-prod` that way for things like ngrok/sendit

mastercactapus:
yes, instead of `-e .env` because this is intended to be a supplementary Procfile with additional commands and not for environmental variables

or any other number of tools you want to start on your local machine along with `make start` -- maybe a grafana server, or gRPC plugin(s)

The contents of the file would be something like:
```Procfile
ngrok: ngrok http 3030
grpc1: go run ./my-tool

@watch-file=./my-plugin-server.cfg
plugin: ./my-plugin-server
```

dctalbot (approved):
LGTM, really like how the cypress output turned out

dctalbot (dismissed):

dctalbot (approved):

m17ch (approved):

PR #2093, opened 2022-01-18
Support for `application/json` in incoming webhook

### LINKED ISSUE #2092: Supporte 'application/json' content type in incoming webhook
incoming webhook should also support `application/json` type instead of just `application/form-data`

Submit json format data should be accepted.

pandey-adarsh147:

**Description:**
Incoming webhook will now accept `application/json` content type
{"summary":"", "details":"", "action":""}

https://github.com/target/goalert/issues/2092

Usage: `Fixes #2092`

### DIFF
--- a/genericapi/handler.go
+++ b/genericapi/handler.go
@@ -92,11 +92,7 @@ func (h *Handler) ServeCreateAlert(w http.ResponseWriter, r *http.Request) {
 
 	summary := r.FormValue("summary")
 	details := r.FormValue("details")
-
-	status := alert.StatusTriggered
-	if r.FormValue("action") == "close" {
-		status = alert.StatusClosed
-	}
+	action := r.FormValue("action")
 
 	contentType := r.Header.Get("Content-type")
 	if contentType == "application/json" {
@@ -107,10 +103,12 @@ func (h *Handler) ServeCreateAlert(w http.ResponseWriter, r *http.Request) {
 
 		summary = alertMessage.Summary
 		details = alertMessage.Details
-		action := alertMessage.Action
-		if action == "close" {
-			status = alert.StatusClosed
-		}
+		action = alertMessage.Action
+	}
+
+	status := alert.StatusTriggered
+	if action == "close" {
+		status = alert.StatusClosed
 	}
 
 	summary = validate.SanitizeText(summary, alert.MaxSummaryLength)
--- a/genericapi/handler.go
+++ b/genericapi/handler.go
@@ -2,6 +2,8 @@ package genericapi
 
 import (
 	"database/sql"
+	"encoding/json"
+	"io/ioutil"
 	"net/http"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


genericapi/handler.go
```diff
@@ -84,15 +92,28 @@ func (h *Handler) ServeCreateAlert(w http.ResponseWriter, r *http.Request) {
 
 	summary := r.FormValue("summary")
 	details := r.FormValue("details")
+	action := r.FormValue("action")
 
-	summary = validate.SanitizeText(summary, alert.MaxSummaryLength)
-	details = validate.SanitizeText(details, alert.MaxDetailsLength)
+	contentType := r.Header.Get("Content-type")
+	if contentType == "application/json" {
```
mastercactapus, line 98: Let's use the ParseMediaType helper to handle any variations/encoding info:
```suggestion
	ct, _, err = mime.ParseMediaType(r.Header.Get("Content-Type"))
	if err != nil {
		http.Error(w, err.Error(), http.StatusUnsupportedMediaType)
		return
	}
	
	if ct == "application/json" {
```

genericapi/handler.go
```diff
@@ -84,15 +92,28 @@ func (h *Handler) ServeCreateAlert(w http.ResponseWriter, r *http.Request) {
...
+		requestBoday, _ := ioutil.ReadAll(r.Body)
+
+		var alertMessage AlertMessage
```
mastercactapus, line 101: For this, let's do an inline struct since we don't need it elsewhere. Also we can ignore the `json` tags since Go parses JSON ignoring case already. Lastly, let's make the fields pointers so we can fallback to query parameters for any missing fields (this follows the current behavior of `FormValue`)

```suggestion
		var b struct {
			Summary, Details, Action *string
		}
```

genericapi/handler.go
```diff
@@ -84,15 +92,28 @@ func (h *Handler) ServeCreateAlert(w http.ResponseWriter, r *http.Request) {
...
+		json.Unmarshal(requestBoday, &alertMessage)
+
+		summary = alertMessage.Summary
+		details = alertMessage.Details
+		action = alertMessage.Action
```
mastercactapus, line 106: Now for these we can do a nil-check and ignore any missing values
```suggestion
		if b.Summary != nil {
			summary = *b.Summary
		}
		if b.Details != nil {
			details = *b.Details
		}
		if b.Action != nil {
			action = *b.Action
		}
```

package.json
```diff
@@ -42,5 +42,6 @@
     "stylelint": "14.2.0",
     "stylelint-config-standard": "23.0.0",
     "typescript": "4.4.3"
-  }
+  },
+  "version": "0.0.0"
```
mastercactapus, line 46: Unrelated editor change committed by mistake?
```suggestion
  }
```

mastercactapus (changes requested):
Few changes requested, in summary:
- check for errors and return appropriate 4xx response code
- Ensure we fallback to query parameters for missing fields
- use mime helper for parsing/normalizing Content-Type header

mastercactapus:
@pandey-adarsh147 is this something you still wanted to get in? I may be able to wrap it up for you too, but you'll need to allow edits from maintainers on your end for me to push changes.

PR #2096, opened 2022-01-19
timeutil/duration: add `ParseISODuration`

dctalbot:

**Description:**
This PR introduces a utility for parsing ISO duration strings into an object. The supported formats are `PnYnMnDTnHnMnS` and `PnW`. `P` (for Period) is always first, and `nX` represents `n` units of `X`. `T` is a separator between the nominal date components and the accurate time components.

References include this [ISO working paper](https://www.loc.gov/standards/datetime/iso-tc154-wg5_n0038_iso_wd_8601-1_2016-02-16.pdf) and [RFC3339](https://datatracker.ietf.org/doc/html/rfc3339#appendix-A).

Additional context is provided in code comments.

### DIFF
--- a/util/timeutil/duration.go
+++ b/util/timeutil/duration.go
@@ -8,6 +8,9 @@ import (
 	"github.com/pkg/errors"
 )
 
+// ISODuration contains the components of an ISO duration string.
+// The time components are combined into seconds, and the weeks component
+// is interpreted as a shorthand for 7 days.
 type ISODuration struct {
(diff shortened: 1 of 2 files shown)


mastercactapus (changes requested):
Can we add a `String()` method for parity? It will help with things like debugging/logging in the future and if we ever need to serialize it back.

mastercactapus (approved):

KatieMSB (approved):

PR #2099, opened 2022-01-20
graphqlapp/alert: introduce AlertMetrics query

dctalbot:

**Description:**
This PR introduces a new graphql endpoint for query alert metrics data. The schema for this is provided below.

```graphql
alertMetrics(input: AlertMetricsOptions!): [AlertDataPoint!]
```


graphql2/graphqlapp/alert.go
```diff
@@ -177,6 +179,85 @@ func (q *Query) mergeFavorites(ctx context.Context, svcs []string) ([]string, er
 	return svcs, nil
 }
 
+// splitRangeByDuration splits the timeframe given between since and until by the duration provided.
+// Each segment is then transformed into an AlertDataPoint based on the given alerts.
+// The given alerts are required to be sorted by their CreatedAt field.
+func splitRangeByDuration(since, until time.Time, dur timeutil.ISODuration, alerts []alert.Alert) (result []graphql2.AlertDataPoint) {
+	if since.After(until) {
```
spencerpauly, line 186: When I test the query with a `since` date that's after the `until` date the alertMetrics query returns null. I think that makes sense, but I could also see it making sense to returning an empty array. Any preference there?

graphql2/graphqlapp/alert.go
```diff
@@ -177,6 +179,85 @@ func (q *Query) mergeFavorites(ctx context.Context, svcs []string) ([]string, er
...
+		for i < len(alerts) && alerts[i].CreatedAt.Before(upperBound) {
+			next.AlertCount++
+			i++
+		}
+		result = append(result, next)
+		ts, upperBound = upperBound, upperBound.AddDate(dur.Years, dur.Months, dur.Days).Add(dur.TimePart)
+		if upperBound.After(until) {
+			upperBound = until
+		}
+	}
```
spencerpauly, line 212: A lot of the logic makes sense, but this piece took me a little bit to figure out. So it seems you make a period, then fill in the alert count for that period, then make the next period, then fill in the alert count, etc etc.

Could this maybe be done more clearly by making all the periods first in one pass, then coming back and filling in all the data after? Especially since I'm sure we'll be adding to this in the future if we want to add more metrics.

graphql2/graphqlapp/alert.go
dctalbot: Great catch, nil is favored in Go, but it should be serialized to an empty array at the graphql layer

graphql2/graphqlapp/alert.go
dctalbot: Sure, thanks for the feedback! We can try that. If we ever need to eke out some performance, this should be faster, but generally we favor clean code over perf

graphql2/graphqlapp/alert.go
```diff
@@ -177,6 +179,95 @@ func (q *Query) mergeFavorites(ctx context.Context, svcs []string) ([]string, er
...
+			Valid: true,
+		},
+		Limit: 100,
+		Sort:  alert.SortModeDateIDReverse,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	period, err := timeutil.ParseISODuration(*opts.Period)
```
mastercactapus, line 254: We should validate length

```go
err := validate.ISODuration("period", period, ISODuration{TimePart: 24*time.Hour}, ISODuration{Months: 1})
if err != nil {
	return nil, err
}
```

graphql2/graphqlapp/alert.go
```diff
@@ -177,6 +179,95 @@ func (q *Query) mergeFavorites(ctx context.Context, svcs []string) ([]string, er
...
+	if err != nil {
+		return nil, err
+	}
+
+	_24hr, err := time.ParseDuration("24h")
```
mastercactapus, line 259: Use constant expressions for duration `24*time.Hour`

graphql2/graphqlapp/alert_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/target/goalert/alert"
+	"github.com/target/goalert/util/timeutil"
+)
+
+func TestSplitRangeByDurationAlertCounts(t *testing.T) {
+
+	loc, err := time.LoadLocation("America/Chicago")
```
mastercactapus, line 15: If the split func uses the timezone (currently it looks like `since` arg drives this) we should add a test for that explicitly or convert to UTC to avoid it.

graphql2/schema.graphql
```diff
@@ -22,6 +22,9 @@ type Query {
   # Returns a paginated list of alerts.
   alerts(input: AlertSearchOptions): AlertConnection!
 
+  # Returns an array of alert metric data points
+  alertMetrics(input: AlertMetricsOptions!): [AlertDataPoint!]
```
mastercactapus, line 26: Avoid `null` by schema definition.
```suggestion
  alertMetrics(input: AlertMetricsOptions!): [AlertDataPoint!]!
```

graphql2/schema.graphql
```diff
@@ -863,6 +880,8 @@ enum AlertSearchSort {
   dateIDReverse
 }
 
+scalar ISODuration
```
mastercactapus, line 883: Let's add a comment here since it's a type

validation/validate/timeframe.go
```diff
@@ -0,0 +1,15 @@
+package validate
+
+import (
+	"time"
+
+	"github.com/target/goalert/validation"
+)
+
+func TimeFrame(fname string, since time.Time, until time.Time, timeFrame time.Duration) error {
```
mastercactapus, line 9: Add a comment for new public validation functions.

Also:
- Anything to be lifted from `schedule/validate.go`? We're calling this `TimeRange` there
- For this use, should it be validated based on an ISODuration instead?

spencerpauly (approved):

graphql2/graphqlapp/alert.go
```diff
@@ -177,6 +179,95 @@ func (q *Query) mergeFavorites(ctx context.Context, svcs []string) ([]string, er
...
+		upperBound := until
+		if ptindex != len(result)-1 {
+			upperBound = result[ptindex+1].Timestamp
+		}
+
+		for i < len(alerts) && alerts[i].CreatedAt.Before(upperBound) {
+			result[ptindex].AlertCount++
+			i++
+		}
+	}
```
mastercactapus, line 222: If we sort the alerts, can we just walk through, trimming the slice?

Something like:
```go
	countAlerts:=func(ts time.Time) int {
		var count int
		for i,a:=range alerts {
			if !a.CreatedAt.Before(ts) {
				alerts = alerts[i:]
				return count
			}
			if !a.CreatedAt.Before(until) {
				alerts=nil
				return count
			}

			count ++
		}

		alerts = nil
		return count
	}

	ts:=since
	// trim alerts before since
	countAlerts(ts)
	for ts.Before(until) {
		result=append(result, graphql2.AlertDataPoint{
			Timestamp: ts,
			AlertCount: countAlerts(ts),
		})
		ts=ts.AddDate(dur.Years, dur.Months, dur.Days).Add(dur.TimePart)
	}
```

mastercactapus (changes requested):

graphql2/graphqlapp/alert.go
```diff
@@ -177,6 +179,95 @@ func (q *Query) mergeFavorites(ctx context.Context, svcs []string) ([]string, er
 	return svcs, nil
 }
 
+// splitRangeByDuration splits the timeframe given between since and until by the duration provided.
+// Each segment is then transformed into an AlertDataPoint based on the given alerts.
+// The given alerts are required to be sorted by their CreatedAt field.
+func splitRangeByDuration(since, until time.Time, dur timeutil.ISODuration, alerts []alert.Alert) (result []graphql2.AlertDataPoint) {
+	// serialized as empty array
+	result = []graphql2.AlertDataPoint{}
```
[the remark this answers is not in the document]
dctalbot: This doesn't seem to work, the response is still `null`

graphql2/graphqlapp/alert.go
dctalbot: Oh nice, trimming the slice like that is a neat trick. We originally had something closer to this w/o the trimming and decided it was too expensive, but if this works, it should keep it effectively O(n)

dctalbot: Committed a new version, lmk what you think

graphql2/graphqlapp/alert_test.go
dctalbot: Went with the latter

validation/validate/timeframe.go
KatieMSB: It looks like `validate/duration.go` can actually work for this instead of having a new function, and for this case using time.Duration might be easier since there isn't a good way to compare ISODurations right now.

graphql2/graphqlapp/alert.go
```diff
@@ -177,6 +179,88 @@ func (q *Query) mergeFavorites(ctx context.Context, svcs []string) ([]string, er
...
+	err = validate.Many(
+		validate.Range("ServiceIDs", len(opts.FilterByServiceID), 1, 1),
+		validate.TimeFrame("Timeframe", opts.Since, opts.Until, 1200*time.Hour),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	opts.Since = opts.Since.UTC()
+	opts.Until = opts.Until.UTC()
```
mastercactapus, line 236: I think the only issue is the timezone with the `splitRangeByDuration` helper -- can we move this to the function call, or maybe it better to have the function handle it internally?

We can comment on the helper that it performs the split in UTC -- if we ever need a timezone-aware version we can make one that takes a `time.Location` argument. If we fix it internally to the helper, the caller of the func doesn't need to be aware that if they forget to call `.UTC()` they could get unexpected results.

graphql2/graphqlapp/alert.go
mastercactapus: We could probably do ISODuration validation by UTC comparison

```go
func (dur ISODuration) Duration() time.Duration {
  t := time.Time{}.UTC()
  return t.AddDate(dur.Years, dur.Months, dur.Days).Add(dur.TimePart).Sub(t)
}
```

or at least use that in `GT` or `Equal` comparison methods.

mastercactapus (changes requested):

graphql2/graphqlapp/alert.go
spencerpauly: I think this looks good too. I especially like this part:

```
result=append(result, graphql2.AlertDataPoint{
			Timestamp: ts,
			AlertCount: countAlerts(ts),
		})
```

I think that's very clear. It returns an object with these fields and if a field needs to be calculated we pull that out like the countalerts function so this loop keeps it's clarity. Works for me +1.

mastercactapus (approved):

m17ch (approved):
LGTM 👍

PR #2110, opened 2022-01-26
dev: only rm build and node_modules on `make clean`

mastercactapus:
**Description:**
Updates `make clean` target to only cleanup build output and node_modules, leaving things like `Procfile.local` or config backups intact

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -199,7 +199,7 @@ resetdb: config.json.bak
 	go run ./devtools/resetdb --no-migrate
 
 clean:
-	rm -rf bin node_modules web/src/node_modules web/src/cypress/videos web/src/cypress/screenshots web/src/build/static
+	rm -rf bin node_modules web/src/node_modules web/src/build/static
 
 build-docker: bin/goalert bin/mockslack
 
--- a/Makefile
+++ b/Makefile
@@ -199,7 +199,7 @@ resetdb: config.json.bak
 	go run ./devtools/resetdb --no-migrate
 
 clean:
-	git clean -xdf
+	rm -rf bin node_modules web/src/node_modules web/src/cypress/videos web/src/cypress/screenshots web/src/build/static
 
 build-docker: bin/goalert bin/mockslack
 


dctalbot (approved):

Makefile
```diff
@@ -199,7 +199,7 @@ resetdb: config.json.bak
 	go run ./devtools/resetdb --no-migrate
 
 clean:
-	git clean -xdf
+	rm -rf bin node_modules web/src/node_modules web/src/cypress/videos web/src/cypress/screenshots web/src/build/static
```
Forfold, line 202: I think there could be some other things too. I noticed that I have a `yarn-error.log` file on my end locally, for instance, which isn't in the repo

mastercactapus: Hmm, maybe we should discuss the purpose of the target because I see your point, but that's part of the nature of this change -- to ignore things that don't affect the build output and not wipe away anything we don't know about.

As originally implemented, `make clean` just did the equiv. of a fresh git-clone, that may or may not include losing your local config.
This PR is intended to make it behave more like the normal convention of a `make clean` which is to remove executables and object files (so that the next build will compile everything from scratch). Effectively cleaning out any potentially stale/cached output.

Something like `yarn-error.log` doesn't have any impact on the build output and isn't really a "known" file (though is a side-effect of yarn).
That being said, maybe removing cypress videos/screenshots should be excluded too?

At the end of the day the primary problem I think that needs to be solved by `make clean` is fresh build output. No possibility of an old webpack build, modified or outdated node module, or including a GoAlert binary with stale build flags. 

That way something like `make clean release` will give every developer the same exact output regardless of their local state (assuming no uncommitted changes)

What do you think?

Forfold: Replied in Slack, but sounds good to me 👍  I think it could be worth removing the Cypress bits for the same reason then. Never know if someone wants to hold onto that information I suppose

Forfold (approved):

dctalbot (approved):

PR #2111, opened 2022-01-26
services/alert-metrics: add pagination in datagrid

KatieMSB:

**Description:**

Usage: `Fixes #<issue number>`, or `Fixes (paste link of issue)`.

**Out of Scope:**

**Screenshots:**

**Additional Info:**


web/src/app/services/AlertMetrics/AlertMetricsTable.tsx
```diff
@@ -29,11 +29,22 @@ const useStyles = makeStyles(() => ({
...
   const classes = useStyles()
-  const { alerts } = props
+  const [page, setPage] = useState(0)
+  const [rowCount, setRowCount] = useState(0)
+
+  useEffect(() => {
+    if (props.alerts.length) {
+      setRowCount(props.alerts.length)
+    }
+  }, [page, props.alerts])
```
dctalbot, line 45: Instead of this, we can give `props.alerts` a default value of `[]` when destructuring 
e.g. `const { alerts = [], loading } = props`

Then, pass `rowCount={alerts.length}` to `DataGrid`

web/src/app/services/AlertMetrics/AlertMetricsTable.tsx
```diff
@@ -123,14 +138,20 @@ export default function AlertMetricsTable(
     <Grid container className={classes.tableContent}>
       <Grid item xs={12}>
         <DataGrid
-          rows={alerts}
-          loading={props.loading}
+          rows={getRows()}
```
dctalbot, line 141:
```suggestion
          rows={rows}
```

web/src/app/services/AlertMetrics/AlertMetricsTable.tsx
```diff
@@ -123,14 +138,20 @@ export default function AlertMetricsTable(
...
           components={{
             ExportIcon: DownloadIcon,
             Toolbar: CustomToolbar,
           }}
+          onPageChange={(newPage: number) => setPage(newPage)}
+          page={page}
+          paginationMode='server'
+          pageSize={10}
+          rowsPerPageOptions={[10]}
+          rowCount={rowCount}
```
dctalbot, line 154:
```suggestion
          rowCount={alerts.length}
```

dctalbot (changes requested):
Now that we have pagination, it would be nice to make the table a bit taller so it doesn't have a scrollbar for whatever the default row density is.

It looks like polling is turned off in the `useEffect` but is never turned on again. Is that intentional? We can pass `pollInterval: 0` instead of making it stateful.

Also, looks like there is a deprecation method with the `useQuery` method. 

Screen Shot 2022-02-14 at 11 15 51 AM

We should add a merge policy in apollo.js instead

KatieMSB:
Closing to deprioritize

PR #2167, opened 2022-02-15
Migrate playground to graphiql

### LINKED ISSUE #2160: Migrate graphql playground to graphiQL
GraphQL playground is [being retired](https://github.com/graphql/graphql-playground/issues/1143) in favor of graphiQL

Replace graphql Playground with graphiQL.

**Additional context:**
- We should also implement an SRI check on the resource, a security mechanism that is missing today.
- It's also possible to bundle the dependency client-side now that we are running everything in one container, but IMO we should only do so if we can split it into its own chunk so it doesn't bloat the main UI.
- This would be a big change for existing users, so maybe some communication is needed there.

dctalbot:

**Description:**
This PR migrates the graphql playground to use graphiql and closes #2160


KatieMSB:
Looks like the cypress tests will also need to be updated now that the ui has changed a bit.

mastercactapus (approved):

KatieMSB (approved):

KatieMSB (approved):

PR #2171, opened 2022-02-16
dev: refactor calsub to use GORM

mastercactapus:
**Description:**
This PR replaces most of the store & raw SQL code in the calsub package with hooks and adds GORM.
- `*gorm.DB` is passed through context, allowing transactions and Debug() to be passed down as well
- catch-all for GORM calls requiring some level of authentication for any db access
- App now has it's own adjustable time offset `App.Now()` behaves like `time.Now()` but should be in-sync with the DB
- `permission.UserContext` will no longer auth empty strings


mastercactapus:
Will break this into smaller PRs, starting with #2180

Issue #2106, opened 2022-01-25
sendit: log requests in output


mastercactapus:
**What problem would you like to solve? Please describe:**
The send-it tool is silent when proxying, would be nice to log requests to make it easier to find things like 404s, 500s, etc...

**Describe the solution you'd like:**
Simple log of path & response code for proxied requests.
Stretch-goal, color coded codes:
- 2xx = white/default theme
- 4xx = yellow
- 5xx = red
- other = light blue?

**Describe alternatives you've considered:**
GoAlert already logs requests, but as a proxy tool, extra attention from send-it would be helpful.

dctalbot:
We should also consider logging requests in goalert core during development

PR #2172, opened 2022-02-16
dev: Fix make start output

### LINKED ISSUE #2156: dev: fix cy-wide-prod-run and cy-mobile-prod-run output
When using `cy-wide-prod-run` and `cy-mobile-prod-run` make outputs a lot of noise in the console that makes it hard to follow the test progress.

The `Entering directory` and `Leaving directory` don't need to be shown. There are also a lot of SMTP errors logged that may not be necessary.

image

### LINKED ISSUE #2157: dev: fix make start exit on Go build error
Currently a syntax error/build failure for the GoAlert binary results in `make start` exiting.

We should fix this (maybe adjust the Procfile?) so that it retries after outputting the error but keeps running.

image

mastercactapus:
**Description:**
This PR fixes some noise issues in the log output.
- Engine cycle is no longer logged by default
- After 5 sec, warnings will be logged with an error at the end of the cycle
- An error will be logged each minute if a cycle is stuck that long
- "Build Failed" will be logged and retried after 3sec on failure with `make start` instead of exiting
- Prometheus is changed to `warn` log level
- Cypress tests should no longer spam make messages

Fixes #2156 
Fixes #2157 

**Out of Scope**
Request logging has been left out and can be addressed separately (or with #2106 )

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,9 +78,9 @@ cy-wide-prod: web/src/build/static/app.js cypress
 cy-mobile-prod: web/src/build/static/app.js cypress
 	CONTAINER_TOOL=$(CONTAINER_TOOL) CYPRESS_viewportWidth=375 CYPRESS_viewportHeight=667 CY_ACTION=$(CY_ACTION) go run ./devtools/runproc -f Procfile.cypress.prod
 cy-wide-prod-run: web/src/build/static/app.js cypress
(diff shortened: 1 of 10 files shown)


dctalbot (approved):

arurao (approved):
LGTM

Issue #1068, opened 2020-11-24
ui: global settings dialog


Forfold:
As we continue to iterate over GoAlert, we have begun to offer our users a number of opt-in customizations and features that are scattered throughout the application. The goal I envision for this enhancement is to gather those items and have them live in a single dialog (fullscreen on small screens), initially accessible from the sidebar.

These _current_ items include:
- Alert Status Updates
- Calendar subscriptions
  - This can be a multi-select, with each subscriptions being a chip.
  - Chips should be removable, which would delete the calendar subscription.
  - Creating a new subscription from within the settings dialog would be out of scope.
- Give Feedback
- Log Out

_Future_ items could include:
- Update avatar image
- Toggle dark mode / set theme
- Display times in X tz
- Set/toggle timestamp format
- Update basic auth password (see #1247)

I think enhancements such as these could be introduced via a `user_preferences` column of type `jsonb` on the `users` table.

**Out of scope:**
I would like to expand on this feature after this first implementation. My thoughts are that we can ultimately move the profile Avatar and settings button from the sidebar and into the top toolbar.

From here, we would move any of the context menus from that area in toolbar and have them rendered above the DetailsPage header card. I believe this will give more visibility into the options available to the user for their current context they are working in.

Forfold:
Pulling this into the project board to start working on an initial design

Forfold:
Accepted Design (changes listed below):

<img width="1260" alt="Screen Shot 2022-03-03 at 11 30 09 AM" src="https://user-images.githubusercontent.com/11381794/156638849-654c3b6e-b3fc-469c-9437-a77688498546.png">

- admin icon to use Wrench
- wizard icon to use Wizard Hat
- user avatar shown in toolbar
- profile link moved to settings popover
- logout and feedback moved to settings popover
- theme mode moved to settings popover

PR #2174, opened 2022-02-17
ui: add dark mode toggle to navbar

### LINKED ISSUE #1694: add dark mode toggle
Extending off of #1458, while we are discussing the various sizes and colors of different aspects of our theme, we can certainly turn on the default material-ui dark mode theme (as a toggle) since this is supported out of the box, and adjust any out-of-place elements as necessary. Then, when we get to the rest of the theme settings in #1458, the bulk of that PR should hopefully remain in the mui theme file.

I'm thinking we can use local storage to store the setting, that way each device has its own preference set (i.e. if we used the DB, a user may have light mode on their PC and dark mode on their phone). We can also use a media query to make our best guess as to what the user currently has their device set to. Though if we don't know, setting to `light` may be best as that is the closest to the current palette.

- Both default `@material-ui/core` dark/light themes enabled
- Toolbar toggle to switch between themes

N/A

**Additional context:**
It may be wise to settle on a primary/secondary color for dark and light themes before fully merging- we don't want to change app branding too often if we can help it.

Forfold:
Closes #1694

Introduces a button group for dark mode to the user profile, defaulting to the system preference, saving to local storage. The primary grey color for light mode has been retained, with MUI defaults being leveraged for dark mode.

Component changes to accommodate dark mode:
- Adds `ThemePicker` component to render option buttons from profile
- Adds `themeConfig.ts` file containing a `ThemeContext`, and a `ThemeProvider` implementing the theme config, removing the old config within `mui.ts`
- Alt white GoAlert logo when dark mode is active
- Fixes search bar background when dark mode is active
- Fixes a reference to the old theme obj in `ScheduleCalendar`, now that theme has moved
- Fixes `AlertListFilter` icon to use proper contrast color when in dark mode
  - (Out of scope: default IconButton color to `secondary`)
- Fixes the highlighted item text color in `FlatList`
- Fixes `FlatList` list item subheader background color always being white (now transparent to match bg, no difference to light mode)
- Fixes icon when in dark mode for `OtherActions` in `RotationUserList` if the user is active
- Adds a white svg icon that the theme switches between for the `Slack` and `SlackBW` icons
- Fixes type error from theme imports (after theme implementation changed) in several files
- Removes light grey color for cancel buttons, in favor of the default primary color
- Fixes error messages text color in dark mode

Out of Scope:
- Choosing primary and secondary colors for light & dark mode
- Changing the hard-coded colors like `FlatList` highlighted items, or `SetFavoriteButton`'s heart color to better fit both light & dark mode palettes
- Transitioning button and similar components to use `secondary` colors
- Creating a settings dialog to improve user experience (see: #1068)

Screenshots:

Screen Shot 2022-02-21 at 11 30 02 AM

Screen Shot 2022-02-21 at 11 30 17 AM


dctalbot (approved):
Next up should include:
- Handling error screens (e.g. 404 sad face)
- Handling secondary dialog buttons (e.g. cancel, close)

Forfold:
Good callout on the error screens- that's a good addition to add to our draft issue

mastercactapus:
@dctalbot Seems like something that got missed but caught in review. Hard to see, but hopefully an easy fix:
image

Secondary buttons aren't as bad, but that also looks like it's not intentional:

image


image

I think the secondary is supposed to have blue text matching the other button (I could be wrong)
image

---

We should fix things like that before merging, anyone with a dark system theme will get the dark mode by default so we should make sure it's finished, at least the simple stuff.

@Forfold Adjusting color for form dialogs and error components shouldn't require major changes, correct?

Forfold:
The blue text is primary @mastercactapus, so probably not the best option for the cancel buttons. We have secondary as `grey` in dark mode for the purposes of consistency, but I can remove that in favor of the default (purple) if we want.

mastercactapus:
@Forfold Yeah, that's what I noticed in the MUI dialog examples:
https://mui.com/components/dialogs/

Since that's what we do in light mode I assumed dark mode would work the same, in any case, my concern is just the readability. Whatever you think is best

Forfold:
Well we don't actually have a primary or secondary color for light mode. We just have grey. So I was assuming that it should be secondary, but yeah since material.io (https://material.io/components/dialogs#anatomy) shows both buttons using primary too we can go with that

mastercactapus:
@Forfold sounds good I took a look locally, much better. What about the components in `Errors.js` IIRC those are just SVG and text, can we lighten those in dark mode?

As before, whatever you think is best, just looking for some better contrast

Forfold:
Yep, working through the error component right now

mastercactapus (approved):

dctalbot (approved):

PR #2178, opened 2022-02-17
slack: limit details to 3000 characters

### LINKED ISSUE #2175: Slack bug - invalid_attachments
Issue to track troubleshooting errors that have been manifesting in logs:

```
{
AlertID: XXX
AuthSystemComponent: "SendMessage"
DestType: "DestTypeSlackChannel"
ProviderName: "Slack-Channel"
ProviderType: 3
Trigger: "INTERVAL"
error: "send notification: invalid_attachments"
level: "error"
msg: ""
}
```

Perhaps invalid payload (maybe empty summary/details)?

mastercactapus:
**Description:**
Limits the length of details in the original slack message to 3000 characters.

This is done with a new helper `RenderSize` (and `RenderSizeN` for some alert messages) that will use a render function and assert the length is within the limit. If it is not, it performs a binary search to trim the input until it fulfills the requirement.

Fixes #2175

### DIFF
--- a/notification/slack/channel.go
+++ b/notification/slack/channel.go
@@ -278,8 +278,16 @@ func alertMsgOption(ctx context.Context, callbackID string, id int, summary, det
 		details = ""
 	}
 	if details != "" {
+		escaped := slackutilsx.EscapeMessage(details)
+		if len(escaped) > 3000 {
+			newDetailsLen := len(details) - (len(escaped) - 3000)
+			if newDetailsLen < 0 {
+				newDetailsLen = 0
+			}
+			escaped = slackutilsx.EscapeMessage(details[:newDetailsLen])
+		}
 		blocks = append(blocks, slack.NewSectionBlock(
-			slack.NewTextBlockObject("mrkdwn", slackutilsx.EscapeMessage(details), false, false), nil, nil),
+			slack.NewTextBlockObject("mrkdwn", escaped, false, false), nil, nil),
 		)
 	}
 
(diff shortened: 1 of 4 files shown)


m17ch (dismissed):

notification/slack/channel.go
```diff
@@ -278,8 +278,16 @@ func alertMsgOption(ctx context.Context, callbackID string, id int, summary, det
 		details = ""
 	}
 	if details != "" {
+		escaped := slackutilsx.EscapeMessage(details)
+		if len(escaped) > 3000 {
+			newDetailsLen := len(details) - (len(escaped) - 3000)
+			if newDetailsLen < 0 {
+				newDetailsLen = 0
+			}
+			escaped = slackutilsx.EscapeMessage(details[:newDetailsLen])
```
dctalbot, line 287: if len(details) == 5000 and len(escaped) == 4000, then newDetailsLen == 4000. Contrived example, but logic seems off here?

mastercactapus: Yeah, it's not perfect -- best effort really. It's the same logic we apply to SMS messages -- we trim the input by the amount the output is off. But now that you mention it, maybe I'll do some quick testing to see if it can be abused to still break -- unlike SMS the output in this case will always be larger than the input.

mastercactapus: ^ same or larger

dctalbot (dismissed):

dctalbot (approved):

dctalbot (approved):

PR #2179, opened 2022-02-17
ui/tech debt: OtherActions

Forfold:
Cleans up the implementation of the Icon and IconButton in `OtherActions`. If an icon was provided as a prop, it wouldn't be rendered with an `IconButton` as it was only being added to the JSX from the `defaultProps` section.

Also fixes the color of an IconButton for the `AlertsListFilter` to match the checkbox OtherActions within the page, for consistency.

### DIFF
--- a/web/src/app/util/OtherActions.js
+++ b/web/src/app/util/OtherActions.js
@@ -15,7 +15,7 @@ const cancelable = (_fn) => {
   return cFn
 }
 
-export default function OtherActions({ color, Icon, actions, placement }) {
+export default function OtherActions({ color, IconComponent, actions, placement }) {
   const [anchorEl, setAnchorEl] = useState(null)
   const onClose = cancelable(() => setAnchorEl(null))
   const ref = useRef(null)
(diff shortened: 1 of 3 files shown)


dctalbot (dismissed):

web/src/app/util/OtherActions.js
```diff
@@ -15,25 +15,28 @@ const cancelable = (_fn) => {
   return cFn
 }
 
-export default function OtherActions({ color, icon, actions, placement }) {
+export default function OtherActions({ color, Icon, actions, placement }) {
```
mastercactapus, line 18: We should probably call this `IconComponent` now

web/src/app/util/OtherActions.js
```diff
@@ -62,15 +65,11 @@ OtherActions.propTypes = {
     }),
   ).isRequired,
   color: p.string,
-  icon: p.element,
+  Icon: p.object,
```
mastercactapus, line 68:
```suggestion
  Icon: p.elementType,
```

mastercactapus (changes requested):

dctalbot (approved):

dctalbot (approved):

dctalbot (approved):

PR #2182, opened 2022-02-21
cypress: upgrade to version 9.5.0

mastercactapus:
**Description:**
Upgrades Cypress to the latest version (`9.5.0`).

### DIFF
--- a/devtools/ci/dockerfiles/cypress-env/Dockerfile
+++ b/devtools/ci/dockerfiles/cypress-env/Dockerfile
@@ -1,4 +1,4 @@
-FROM docker.io/cypress/included:8.7.0
+FROM docker.io/cypress/included:9.5.0
 ENTRYPOINT [ "/bin/sh" ]
 RUN apt-get update && apt-get install -y --no-install-recommends postgresql postgresql-contrib && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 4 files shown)


dctalbot (approved):

dctalbot (approved):

web/src/cypress/support/page-action.ts
```diff
@@ -1,8 +1,19 @@
...
+function pageAction(s: string): Cypress.Chainable {
   return cy
     .get('[data-cy=app-bar]')
     .find('button[data-cy=other-actions]')
     .menu(s)
 }
 
 Cypress.Commands.add('pageAction', pageAction)
+
+export {}
```
Forfold, line 19: What are these empty exports used for?

mastercactapus: To force typescript into module mode so that they can all be written the same way, in the same scope:
https://www.typescriptlang.org/docs/handbook/2/modules.html#non-modules

Forfold: Cool, and I imagine it's fine not exporting the functions themselves now since they're added onto `cy` anyways. Thanks!

Forfold (approved):

PR #2190, opened 2022-02-22
cypress: patch flaky test

dctalbot:

**Description:**
This PR adds adds a check to wait for the quick links to be present before navigating to/from on-call notifications


Forfold (approved):
Do you think it's a scroll issue? I don't see it loading in async and it's odd that it keeps happening to just this one link

Edit: Looks like it's conditionally rendered based on the `slackEnabled` flag- so the race condition's there

mastercactapus (approved):

PR #2198, opened 2022-02-24
darkmode: fix generic errors and red-on-black for fav and links

mastercactapus:
**Description:**
Fixes a few dark mode color issues missed in review:
- GenericError was rendering black-on-black, updated to match the rest of the error components
- Replace red-on-black with primary color for both fav buttons and links

More info:
https://accessibility.psu.edu/color/colorvisiondetails/

---
Old vs. New
image image

Simulated examples:

Protanopia (no red)
image image

Deuteranopia (no green)
image image

Tritanopia (no blue)
image image

Achromatopsia (no color)
image image

Contrast Loss
image image

### DIFF
--- a/web/src/app/util/SetFavoriteButton.tsx
+++ b/web/src/app/util/SetFavoriteButton.tsx
@@ -5,6 +5,7 @@ import NotFavoriteIcon from '@mui/icons-material/FavoriteBorder'
 import Tooltip from '@mui/material/Tooltip'
 import makeStyles from '@mui/styles/makeStyles'
 import _ from 'lodash'
+import { Theme } from '@mui/material'
 
 interface SetFavoriteButtonProps {
   typeName: 'rotation' | 'service' | 'schedule' | 'escalationPolicy' | 'user'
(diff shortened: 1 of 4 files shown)


web/src/app/util/SetFavoriteButton.tsx
```diff
@@ -13,14 +14,17 @@ interface SetFavoriteButtonProps {
   loading?: boolean
 }
 
-const useStyles = makeStyles({
+const useStyles = makeStyles((theme: Theme) => ({
   favorited: {
-    color: 'rgb(205, 24, 49)',
+    color:
+      theme.palette.mode === 'dark'
+        ? theme.palette.primary.main
```
Forfold, line 21: I think we should maintain keeping the heart red (just what makes sense to me). Could we use `theme.palette.error` for an accessible red? Or maybe we could use a star icon instead and use `theme.palette.warn` for an accessible yellow on both light and dark modes.

mastercactapus: I'm open to that if it works -- firefox has a way to simulate under the Accessibility tab in the dev tools -- if you want to give it a try and post the results

Otherwise, bear in mind this is only interim and not meant to be a decision on permanent colors, just something that resolves current state while we discuss and decide next steps

Forfold (approved):

dctalbot (approved):

PR #2212, opened 2022-03-03
ui: add settings popover

### LINKED ISSUE #1068: ui: global settings dialog

Forfold:
Closes #1068.

Adds a Current User Avatar to the top app bar. On click, it will open a popover with the following options:
- Manage Profile
- Theme Mode Picker
- Feedback (if enabled)
- Logout

Also,
- Removes logout, feedback, and profile from the navbar
- Removes the theme mode picker from the profile
- Sets the admin nav icon to a wrench
- Sets the wizard nav icon to a wizard hat
- Removes dividers from nav list

Screen Shot 2022-03-03 at 2 10 00 PM

Screen Shot 2022-03-04 at 8 59 21 AM


web/src/app/main/components/UserSettingsPopover.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  const [anchorEl, setAnchorEl] = React.useState<HTMLButtonElement | null>(null)
+
+  const handleClick = (event: React.MouseEvent<HTMLButtonElement>): void => {
+    setAnchorEl(event.currentTarget)
+  }
+
+  const handleClose = (): void => {
+    setAnchorEl(null)
+  }
```
mastercactapus, line 31: These don't do a lot and are only used once anyway -- can we just inline them to remove the indirection?

web/src/app/main/components/UserSettingsPopover.tsx
```diff
@@ -0,0 +1,101 @@
...
+          </Grid>
+        </Grid>
+      </Popover>
+
+      {/* {renderSidebarLink(LogoutIcon, '/api/v2/identity/logout', 'Logout', {
+        onClick: (e) => {
+          e.preventDefault()
+          logout()
+        },
+      })} */}
```
mastercactapus, line 98: Can this be deleted?

mastercactapus (changes requested):

mastercactapus (dismissed):

KatieMSB (approved):

mastercactapus (dismissed):

KatieMSB (approved):

web/src/cypress/integration/profile.ts
```diff
@@ -73,6 +73,71 @@ function testProfile(): void {
...
+
+      cy.get('[data-cy="manage-profile"]')
+        .find('button')
+        .contains('Dark')
+        .click()
+      cy.get('div[id="app-root"]').should(
+        'have.css',
+        'background-color',
+        'rgb(18, 18, 18)',
+      )
```
mastercactapus, line 107: These seem fragile and not helpful if say we change the mode as the test would break even if it's behaving correctly.

Could we just test that the color changes? (E.g. read current, click button, assert a change, click previous button, assert same as before)

mastercactapus (changes requested):

web/src/cypress/integration/profile.ts
Forfold: I guess so, we can do that

Forfold: Ill have to see if Cypress can do that effectively.

Forfold: Resolved in most recent commit

mastercactapus (dismissed):

mastercactapus (approved):

KatieMSB (dismissed):

KatieMSB (approved):

mastercactapus (approved):

PR #2219, opened 2022-03-07
dev: fix resetdb

dctalbot:

**Description:**
This PR fixes a regression in which the alerts table's alerts_id_seq sequence was not being updated after the COPY command. Without this, a developer cannot readily create alerts locally.

### DIFF
--- a/devtools/resetdb/main.go
+++ b/devtools/resetdb/main.go
@@ -277,6 +277,10 @@ func fillDB(ctx context.Context, url string) error {
 	dt.Wait()
 	_, err = pool.Exec(ctx, "alter table alerts enable trigger all")
 	must(err)
+
+	// fix sequences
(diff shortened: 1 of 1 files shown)


devtools/resetdb/main.go
```diff
@@ -277,6 +277,10 @@ func fillDB(ctx context.Context, url string) error {
 	dt.Wait()
 	_, err = pool.Exec(ctx, "alter table alerts enable trigger all")
 	must(err)
+
+	// fix sequences
+	_, err = pool.Exec(ctx, "SELECT pg_catalog.setval('public.alerts_id_seq', (select count(*) from public.alerts), true);")
```
mastercactapus, line 282: Shouldn't this grab the max + 1? That way future changes don't assume sequential ordering

dctalbot: Sure, seems unlikely though. I suppose we could also account for negative values, so would be max(max_id, 1)

mastercactapus (approved):

arurao (approved):
LGTM

PR #2235, opened 2022-03-10
db: add startup check for use_next_db

mastercactapus:
**Description:**
Adds a sanity check at startup to refuse to connect to a DB that has been explicitly marked as stale by a switchover.

**Review Tips:**
Set the `current_state` column to `use_next_db` in the `switchover_state` table to validate the application refuses to start.

You can also change the query to be invalid inside app.go to ensure other DB errors also result in it failing to start.

Lastly, the app should start and work normally as long as `use_next_db` is not set.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -128,6 +128,30 @@ type App struct {
 
 // NewApp constructs a new App and binds the listening socket.
 func NewApp(c Config, db *sql.DB) (*App, error) {
+	var err error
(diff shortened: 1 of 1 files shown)


mastercactapus:
@dctalbot My bad, added info to the PR description

dctalbot (approved):

KatieMSB (approved):

Issue #2214, opened 2022-03-04
ui/theme: Semantic/Component Updates


Forfold:
Need to update several components to maintain consistency and accessibility across light and dark modes.

- [x] outgoing message logs text field type: use outline, be consistent with search textfield (#2322)
- [x] schedule calendar event colors: (#2247)
    - All shifts: primary (no matter the type)
    - Override/Temp schedule header shift: secondary
- [x] calendar outline in dark mode: dark grey (#2247)
- [x] red badge color in ep form: secondary (#2243)
- [x] link color (text displayed with underline): secondary (#2229)
- [x] MUI Link component in our AppLink instead of a raw anchor tag (#2229)
- [x] copy text button color: primary (#2239)
- [x] open in new tab icon color: primary (#2239)
- [x] IconButton default color: primary (#2239)
- [x] highlighted item color: mark user as 'selected' with ListItem, use secondary (#2238)
- [x] loading items in FlatList: use skeleton from @mui (#2221)

---

from material design guidelines:

primary colors
- top app bar
- contained, text, and outlined buttons

secondary colors (optional, use sparingly to accent select parts of UI)
- FAB buttons
- Selection controls (sliders, switches, toggles, etc)
- Highlighting selected text
- Progress bars
- Links
- Headlines

---

Palettes to test
- New primary color: `blueGrey[400]` (#78909c)
  - light: #a7c0cd
  - dark: #4b636e
- New secondary color: `red[800]` (#c62828) - concern here with having red links, may not be easy to stay WCAG compliant
  - light: #ff5f52
  - dark: #8e0000

PR #2238, opened 2022-03-10
ui/lists: use theme for highlighted items

dctalbot:

**Description:**
This PR uses the theme for highlighted and being-dragged list items. We can use the `selected` prop for highlight, and I decided to go with the `theme.palette.background.default` for the being-dragged.

Components of interest are: 
- user session list
- rotation user list
- ep step list

Part of #2214

**Screenshots:**
Before:
Screen Shot 2022-03-10 at 3 20 16 PM

After:
Screen Shot 2022-03-10 at 3 15 21 PM


Forfold (changes requested):
Highlighted items show slightly lighter than I'd hope they'd turn out in light mode with the color PR merged. I wonder if there's a way to use the same color as the highlighting used in the sidebar? It's just a bit darker. I also wonder if it would be beneficial to change the secondary item text from grey to something darker, or the same as the primary text for highlighted items

Screen Shot 2022-03-14 at 9 32 40 AM

Forfold:
this could be an opportunity to use the tertiary color from the m3 palette, since the primary and secondary are both somewhat similar

mastercactapus (approved):

Forfold (approved):
pulled this down to functionally test. I think the color we have set now will work because we have it against the white background of the cards

PR #2244, opened 2022-03-15
ui: fix console issues for escalation policy step form and autocomplete components

Forfold:

**Description:**
- Addresses a console error where `tabIndex` was expecting a number but was receiving strings
- Addresses a console error where `activeStep` was receiving null for closed, but was expecting a number (`-1`)
- Addresses a console warning in Autocomplete where it's expecting the selected value to be present within the options array


web/src/app/selection/MaterialSelect.tsx
```diff
@@ -142,11 +145,12 @@ export default function MaterialSelect(
       data-cy='material-select'
       data-cy-ready={!isLoading}
       classes={customCSS}
-      {...multi}
+      multiple={multiple}
       value={value}
       inputValue={inputValue}
       disableClearable={required}
       disabled={disabled}
+      filterSelectedOptions
```
dctalbot, line 153: this changes the UX in "non-multiple" mode - is that intentional?

Forfold: yeah, this was intentional because the only time the selected option would show in the dropdown as well is if you were to pick one of the first five options without ever searching anything. so i opted for the consistency between multiple/single selects since it helped get around the bugs a bit easier

dctalbot (approved):

web/src/app/selection/MaterialSelect.tsx
mastercactapus: What bugs are we talking about? With single selection it feels off with things like this:
image

It also means if you explicitly search for a user, they will be omitted if they happen to be selected:

image

It's not an issue for multi because the chip displays, but in single select mode there is no indicator

Forfold: gotcha, the issue was that it expects the value to be in the options array at all times, so without this prop we'll need to add it back to the array and have it show even when searching for a different string. it's been a bit so i'll look into it again today

mastercactapus: I remember that; we had to do some fancy stuff with memorization to keep the references intact IIRC

Forfold: Screen Shot 2022-04-25 at 9 35 11 AM

here's the bug @mastercactapus- I'll work towards a fix this afternoon

Forfold: should be resolved now 👍

dctalbot (approved):
functionally validated

mastercactapus (approved):

PR #2246, opened 2022-03-15
dev: replace webpack-dev-server with watch

mastercactapus:
**Description:**
This PR replaces the use of `webpack-dev-server` and the required proxy with a method that will serve files from the local filesystem or in memory. For localdev, webpack is run in watch mode and a new tool [live.js](https://livejs.com) polls for changes to UI files.

This is in prep for the esbuild work on the graphql explorer.

**Review Tips**
Everything should work as it did before, though `make start` may start a little faster. If you already have code in `web/src/build` it will be served while the dev build runs in the background instead of getting a hang with "wait until bundle complete" warning.

Changes to `index.html` or `app.js` should result in a page reload as they always have.


Procfile
```diff
@@ -1,10 +1,10 @@
...
 @watch-file=./bin/goalert
-goalert: ./bin/goalert -l=localhost:3030 --ui-url=http://localhost:3035 --db-url=postgres://goalert@localhost:5432/goalert?sslmode=disable --listen-sysapi=localhost:1234 --listen-prometheus=localhost:2112
+goalert: ./bin/goalert -l=localhost:3030 --ui-dir=web/src/build --db-url=postgres://goalert@localhost:5432/goalert?sslmode=disable --listen-sysapi=localhost:1234 --listen-prometheus=localhost:2112
 
 smtp: go run github.com/mailhog/MailHog -ui-bind-addr=localhost:8025 -api-bind-addr=localhost:8025 -smtp-bind-addr=localhost:1025 | grep -v KEEPALIVE
 prom: bin/tools/prometheus --log.level=warn --config.file=devtools/prometheus/prometheus.yml --storage.tsdb.path=bin/prom-data/ --web.listen-address=localhost:9090
 
 @watch-file=./web/src/webpack.config.js
-ui: yarn workspace goalert-web webpack serve --config ./webpack.config.js
+ui: yarn workspace goalert-web webpack --config ./webpack.config.js --watch
```
dctalbot, line 10:
```suggestion
ui: yarn workspace goalert-web webpack --config ./webpack.config.js
```
We should move this config option to the config file (`watch: true`) and also use this option to ignore `node_modules` - https://webpack.js.org/configuration/watch/#watchoptionsignored

Procfile.cypress
```diff
@@ -1,10 +1,10 @@
 build: while true; do make -qs bin/goalert || make bin/goalert || (echo '\033[0;31mBuild Failure'; sleep 3); sleep 0.1; done
 
 @watch-file=./bin/goalert
-goalert: go run ./devtools/waitfor postgres://postgres@localhost:5433 && go run ./devtools/procwrap -test=localhost:3042 bin/goalert -l=localhost:3042 --ui-url=http://localhost:3045 --db-url=postgres://postgres@localhost:5433 --slack-base-url=http://localhost:3040/slack --log-errors-only
+goalert: go run ./devtools/waitfor postgres://postgres@localhost:5433 && go run ./devtools/procwrap -test=localhost:3042 bin/goalert -l=localhost:3042 --ui-dir=web/src/build --db-url=postgres://postgres@localhost:5433 --slack-base-url=http://localhost:3040/slack --log-errors-only
 
 @watch-file=./web/src/webpack.config.js
-ui: yarn workspace goalert-web webpack serve --config ./webpack.config.js --port=3045
+ui: yarn workspace goalert-web webpack --config ./webpack.config.js --watch
```
dctalbot, line 7:
```suggestion
ui: yarn workspace goalert-web webpack --config ./webpack.config.js
```
Same as above

dctalbot (changes requested):

Procfile
mastercactapus: I agree with ignoring `node_modules` -- I would have assumed that was the default since `node_modules` can be so heavy.

As for the flag -- I think that may be a little surprising and would prefer to keep it explicit. `webpack --config ./file-name.js` will build and exit with prod and cypress config files. Rather than having dev behave differently, I think having it called out in the command is clearer.

That being said there's a good chance esbuild will be replacing this soon anyhow, so it's not too impactful.

dctalbot (approved):

arurao (approved):
LGTM

PR #2255, opened 2022-03-17
ui/cypress: introduce cypress-axe for a11y testing

Forfold:
**Description:**
Adds the package `cypress-axe` (and it's required dependency `axe-core`), introducing a suite of tests scanning each main route (excluding details pages and sub pages so far). Any issues found regardless of severity will result in a test failure. There were several issues initially found, I've addressed 3 which are resolved outlined below, and marked the rest for todo:

- Navbar: `<ul>` descendents must be `<li>` elements. We were rendering `<a>` tags, so I wrapped these with `<li>`
- Removes the `role='navigation'` attribute from the navbar list, as it is an invalid role. The `<nav>` parent rendered above it is sufficient
- Alerts list checkbox control and list items: a `<label>` element is required. Since the intent is insinuated visually, we can hide the label but keep it available for screen readers (using clipping, as readers may skip over content with `display: 'none'` or `visibility: 'hidden'`). See https://www.w3.org/WAI/tutorials/forms/labels/
- Addresses `FormField` not using `htmlFor` when using the `formLabel` prop to render a `<FormLabel>` (fixing an issue in the Wizard)
- Fixes an issue in Admin Config where accordion elements were using the same value for different attributes
- Fixes an issue in Admin System Limits where labels and fields were not associated properly
- Addresses the reset time range button on the Admin Message Logs list not having a reader accessible label

**Out of Scope:**
- A11y testing on details pages and subpages
- A11y testing individual dialogs and popovers (I think for this, we can inject axe and use `cy.checkA11y()` in the individual tests that already exist, to save time in CI)

**Screenshots:**

Screen Shot 2022-05-02 at 10 40 00 AM


web/src/app/admin/AdminConfig.tsx
```diff
@@ -180,7 +180,7 @@ export default function AdminConfig(): JSX.Element {
           >
             <AccordionSummary
               aria-expanded={section === groupID}
-              aria-controls={`accordion-sect-${groupID}`}
+              aria-controls={`accordion-section-summary-${groupID}`}
```
mastercactapus, line 183: Doesn't this need to match the `id=` below?
```suggestion
              aria-controls={`accordion-section-details-${groupID}`}
```

web/src/cypress/support/a11y.ts
```diff
@@ -0,0 +1,21 @@
...
+      /** Test the accessibility of the current state of the page */
+      validateA11y: typeof validateA11y
+    }
+  }
+}
+
+// no selector provided will result in the entire page being checked
+// inject should be true once per page, after the cy.visit call
+function validateA11y(selector = 'main[id="content"]', inject = true): void {
+  if (inject) cy.injectAxe()
```
mastercactapus, line 13: Will this break if `validateA11y` is called multiple times on different components but the same page/test?

web/src/cypress/integration/a11y.ts
```diff
@@ -0,0 +1,72 @@
...
+      route: string,
+      waitForPageLoad: () => void,
+    ): void {
+      it(testName, () => {
+        cy.visit(route)
+        cy.injectAxe()
+        waitForPageLoad()
+        cy.checkA11y(undefined, {
+          includedImpacts: ['critical'], // only report and assert for critical impact items
+        })
```
mastercactapus, line 30: What is the default ruleset, and do we need to specify it here?

mastercactapus (changes requested):

web/src/cypress/support/a11y.ts
Forfold: it will if inject = true for all of them. that's a good callout, i think I'll set inject to `false` by default and the first time it's called it'll just need to set to true instead.

i originally had it this way since we call it just once in a bunch of tests more often than calling it multiple times on the same page for a given test

web/src/app/admin/AdminConfig.tsx
Forfold: it looks like the ID below isn't even needed (and is throwing duplicate ID errors), because MUI automatically sets the `id` fields properly if you provide `aria-controls`:

Screen Shot 2022-04-25 at 1 13 30 PM

Screen Shot 2022-04-25 at 1 14 44 PM

web/src/cypress/integration/a11y.ts
Forfold: the default ruleset has everything turned on, including:

- wcag2a
- wcag2aa
- wcag2aaa
- wcag21a
- wcag21aa
- wcag21aaa
- best-practice
- wcag***
- ACT
- section508
- section508.*.*
- experimental
- cat.*

from our discussion though we only want to include WCAG for now:
- wcag2a
- wcag2aa
- wcag2aaa
- wcag21a
- wcag21aa
- wcag21aaa
- best-practice
- wcag***

I'll make these changes 👍

web/src/cypress/support/a11y.ts
mastercactapus: Is there a global variable or something we can detect to make that automatic?

mastercactapus (dismissed):

mastercactapus (dismissed):

m17ch (approved):
lgtm

m17ch (approved):

m17ch (dismissed):

web/src/cypress/support/a11y.ts
```diff
@@ -0,0 +1,45 @@
...
+  })
+})
+
+// no selector provided will result in the entire page being validated
+function validateA11y(selector = 'main[id="content"]'): void {
+  cy.window().then((win: Cypress.AUTWindow & { _axeInjected?: boolean }) => {
+    if (win._axeInjected) return
+    win._axeInjected = true
+    return cy.injectAxe()
+  })
```
mastercactapus, line 27: Should we combine these? put the `_axeInjected` check in `injectAxe` and then always call it from `validateA11y`, or just put the full contents in `validateA11y`

Forfold: yeah that's a good point- originally with us calling the lib version I had it separate but if we're going to grab the source ourselves we might as well combine them. 👍

m17ch (approved):

Forfold:
Not sure how to resolve CI not seeing axe-core.. It works locally >.<

Forfold:
Closing for now- out of date, different testing strat in CI, and we may migrate to playwright over time. We can cherry-pick the accessibility fixes as needed.

Issue #2122, opened 2022-02-01
engine/metricsmanager: aggregate daily metrics for alerts


mastercactapus:
Blocked by #2121 

calculate/store daily metrics from the alert_metrics table to allow efficient lookup of yearly metrics

# daily_alert_metrics table

| id  | date | service_id | alert_count | avg_tta | avg_ttc | escalated_alerts_count=0 |
| --- | ---- | ---------- | ----------- | ------- | ------- | ------------------------ |

index on (service_id, date) -- look for postgres spacial date index type

PR #2258, opened 2022-03-21
engine/metrics: populate closed_at field in alert_metrics

dctalbot:

**Description:**
This PR populates a `closed_at` column in `alert_metrics` that will help with the implementation of #2122 

It also refactors the operation to work from oldest to newest, maintaining a `(timestamp, id)` cursor in the engine's state. To regenerate `alert_metrics` e.g. to catch any that were missed, one only needs to reset the engine state.

**How to validate:**
1. `make regendb` and `make start`
2. For each engine cycle, observe `alert_metrics` being populated (500 at a time) and the `engine_processing_version.state` value incrementing in a forward fashion
3. Wait for step 2 to finish, and observe that count(alert_metrics) == count(alerts where status = 'closed')
4. Create and close an alert
5. Wait two minutes before observing the new row in `alert_metrics` and the new `engine_processing_version.state` value

Optionally:

1. Delete some number of rows from `alert_metrics`
2. Clear `engine_processing_version.state` and observe the deleted rows regenerate

### DIFF
--- a/engine/metricsmanager/update.go
+++ b/engine/metricsmanager/update.go
@@ -11,7 +11,7 @@ import (
 )
 
 type State struct {
-	V1 struct {
+	V2 struct {
 		NextAlertID int
 	}
 }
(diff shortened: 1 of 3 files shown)


engine/metricsmanager/update.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 type State struct {
-	V1 struct {
+	V2 struct {
```
mastercactapus, line 14: The struct/meaning hasn't changed, no need to increment the version

migrate/migrations/20220317153438-alert-metrics-add-closed-at.sql
```diff
@@ -0,0 +1,18 @@
+-- +migrate Up
+
+UPDATE engine_processing_versions
+SET "version" = 2
+WHERE type_id = 'metrics';
+
+ALTER TABLE alert_metrics ADD COLUMN closed_at TIMESTAMP WITH TIME ZONE;
```
mastercactapus, line 7: We should make this not-null and truncate the table so we recalculate existing data

mastercactapus (changes requested):
Were we going to change this to do a sliding time window instead of going by alert ID?

m17ch (dismissed):
lgtm 👍

m17ch (approved):

mastercactapus (approved):

PR #2259, opened 2022-03-22
ui/alerts: fix log events order

dctalbot:

**Description:**
This PR fixes an issue in which alert log events would be shuffled at random upon Apollo polling. The solution orders the log entries by timestamp in desc order. We also query the `id` fields for improved cacheing.

To validate, create an alert with many event log entries and observe the list for a minute to make sure polling does not shuffle the list items.


Forfold (approved):
I wasn't able to replicate the alerts reordering in master, but the logic LGTM/works here 👍

Issue #2123, opened 2022-02-01
graphql: connect alertMetrics query to new data source


mastercactapus:
Use tables from #2121 and #2122 for the `alertMetrics` query.

- Past data should be pulled from the daily table, current/recent counts/averages should be calculated from the alert_metrics table.
- make TTA/TTC/Escalation count available in schema

**Out of Scope**
- No UI updates/changes, should be transparent to users

dctalbot:
Proposed schema:

```graphql
type AlertDataPoint {
  timestamp: ISOTimestamp!
  alertCount: Int!
  avgTimeToAck: ISODuration
  avgTimetoClose: ISODuration!
  escalatedCount: Int!
}
```
`avgTimeToAck` will omit `null` values from the calculation

PR #2274, opened 2022-03-31
graphql/alert: use new data source for alertMetrics

KatieMSB:

**Description:**
This pr updates current `AlertMetrics()` endpoint to pull data from the alert_metrics table instead of the alerts table.

1. Adds new `DataPoint` type in alerts.
2. Adds new `alertmetrics` package with Search function.
3. Updates `AlertMetrics()` endpoint to use new `AlertMetricsStore.Search`.

Part of #2123 

**Out of Scope**
To be done in separate prs:
- Update schema and calculate new metrics, i.e avgTimeToAck, avgTimeToClose, escalatedCount.
- Pull past data from new daily_alert_metrics table table.


alert/alertmetrics/search.go
```diff
@@ -0,0 +1,104 @@
...
+
+	// UpperBound will only include alert metrics that were created before the provided time.
+	UpperBound time.Time `json:"b,omitempty"`
+}
+
+var searchTemplate = template.Must(template.New("alert-metrics-search").Funcs(search.Helpers()).Parse(`
+	SELECT
+		am.alert_id,
+		am.service_id,
+		a.created_at
```
dctalbot, line 33: I think we want to use the closed_at time here

alert/alertmetrics/search.go
```diff
@@ -0,0 +1,104 @@
...
+	FROM alert_metrics am
+	JOIN alerts a
+	ON a.id = am.alert_id
```
dctalbot, line 36: `closed_at` is now a column in `alert_metrics`, so we can avoid doing a join here

alert/alertmetrics/search.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/target/goalert/validation/validate"
+)
+
+// SearchOptions contains criteria for filtering and sorting alert metrics.
+type SearchOptions struct {
+	// ServiceIDs, if specified, will restrict alert metrics to those with a matching ServiceID on IDs, if valid.
+	ServiceIDs []string `json:"v,omitempty"`
+
+	// LowerBound will omit any alert metrics created any time before the provided time.
+	LowerBound time.Time `json:"n,omitempty"`
```
dctalbot, line 23: Nit: I'd suggest using `Since` and `Until`

dctalbot (changes requested):

alert/datapoint.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"time"
+)
+
+// A DataPoint is a recording of an Alert metric.
+type DataPoint struct {
+	AlertID     int                  `json:"alert_id"`
+	ServiceID   string               `json:"service_id"`
+	Timestamp   time.Time            `json:"timestamp"`
```
dctalbot, line 11:
```suggestion
	ClosedAt   time.Time            `json:"closed_at"`
```
timestamp is used in the grqphql layer since the granularity is configurable, but here we know it's actually the closed_at field of a record in alert_metrics

graphql2/graphqlapp/alert.go
```diff
@@ -191,7 +192,7 @@ func (q *Query) mergeFavorites(ctx context.Context, svcs []string) ([]string, er
 
 // splitRangeByDuration maps each interval of r to an AlertDataPoint based on the given alerts.
```
dctalbot, line 193:
```suggestion
// splitRangeByDuration maps each interval of r to an AlertDataPoint based on the given alert metrics.
```

graphql2/graphqlapp/alert.go
```diff
@@ -191,7 +192,7 @@ func (q *Query) mergeFavorites(ctx context.Context, svcs []string) ([]string, er
...
 // The given alerts are required to be sorted by their CreatedAt field.
```
dctalbot, line 194:
```suggestion
// The given metrics are required to be sorted by their ClosedAt field.
```

alert/alertmetrics/search.go
```diff
@@ -0,0 +1,101 @@
...
+	WHERE true
+	{{if .ServiceIDs}}
+		AND service_id = any(:services)
+	{{end}}
+	{{ if not .Until.IsZero }}
+		AND closed_at < :until
+	{{ end }}
+	{{ if not .Since.IsZero }}
+		AND closed_at >= :since
+	{{ end }}
```
dctalbot, line 44:
```suggestion
	{{ end }}
	ORDER BY closed_at
```
For compatibility with `splitRangeByDuration`

dctalbot (changes requested):

alert/alertmetrics/search.go
```diff
@@ -0,0 +1,102 @@
...
+		alert_id,
+		service_id,
+		closed_at
+	FROM alert_metrics
+	WHERE true
+	{{if .ServiceIDs}}
+		AND service_id = any(:services)
+	{{end}}
+	{{ if not .Until.IsZero }}
+		AND closed_at < :until
```
dctalbot, line 40:
```suggestion
		AND (date(timezone('UTC'::text, closed_at))) < :until
```

alert/alertmetrics/search.go
```diff
@@ -0,0 +1,102 @@
...
+	{{ end }}
+	{{ if not .Since.IsZero }}
+		AND closed_at >= :since
```
dctalbot, line 43:
```suggestion
		AND (date(timezone('UTC'::text, closed_at))) >= :since
```

alert/alertmetrics/search.go
```diff
@@ -0,0 +1,102 @@
...
+	{{ end }}
+	ORDER BY closed_at
```
dctalbot, line 45:
```suggestion
	ORDER BY (date(timezone('UTC'::text, closed_at)))
```

dctalbot (changes requested):
Postgres requires some verbosity `(date(timezone('UTC'::text, closed_at)))` in order to actually leverage the index on date fields. You can test the query by prepending EXPLAIN ANALYZE e.g.

```sql
explain analyze	SELECT
		alert_id,
		service_id,
		closed_at
	FROM alert_metrics
	WHERE true
		AND service_id = any('{}')
		AND (date(timezone('UTC'::text, closed_at))) < '2022-03-29'
		AND (date(timezone('UTC'::text, closed_at))) >= '2022-03-29'
		ORDER BY (date(timezone('UTC'::text, closed_at)));
```

and the output should note that an index scan is being used as opposed to a seq scan.

It's possible there is a cleaner way (maybe a special index type?), but I don't know of one

dctalbot (approved):

dctalbot (dismissed):

web/src/cypress/support/alert.ts
```diff
@@ -246,7 +247,28 @@ function closeAlert(id: number): Cypress.Chainable<void> {
...
+      (select timestamp from alert_logs where alert_id = a.id and event = 'closed' order by timestamp limit 1)
+    from alerts a
+    where a.id = any('{${alertIDs}}') and a.status='closed' and a.service_id is not null on conflict do nothing`
+
+  for (let i = 0; i < alerts.length; i++) {
+    alertIDs.push(Number(alerts[i].id))
+  }
+
+  return cy.sql(query)
+}
```
mastercactapus, line 268: Can we create & close alerts and rely on the engine to generate the metrics instead? The table schema can change over time, and the cypress tests should be user-interaction-level wherever possible.

alert/datapoint.go
```diff
@@ -0,0 +1,12 @@
...
+import (
+	"time"
+)
+
+// A MetricRecord is a recording of an Alert metric.
+type MetricRecord struct {
+	AlertID   int       `json:"alert_id"`
+	ServiceID string    `json:"service_id"`
+	ClosedAt  time.Time `json:"closed_at"`
+}
```
mastercactapus, line 12: Can this go in the `alertmetrics` package as a `Record` type instead? Also, the `json` tags can be removed

mastercactapus (changes requested):

web/src/cypress/support/alert.ts
KatieMSB: It can take up to two minutes for an alert to get pulled into the alert metrics table after it's closed, so this is trying to work around that time delay. We could add a delay to the cypress check instead.

web/src/cypress/integration/services.ts
```diff
@@ -658,7 +658,8 @@ function testServices(screen: ScreenFormat): void {
         .should('contain', closedAlert.summary)
         .should('not.contain', openAlert.summary)
 
-      cy.get('path[name="Alert Count"]')
+      // timeout to wait for alerts to populate in alert metrics table
+      cy.get('path[name="Alert Count"]', { timeout: 120000 })
```
dctalbot, line 662: I think we can do better than this -- we shouldn't bottleneck tests for a non-critical feature

dctalbot (changes requested):

web/src/cypress/support/alert.ts
dctalbot: I think seeding the database makes for sense for this feature. We can test the engine churn with a backend unit test

web/src/cypress/integration/services.ts
mastercactapus: Try `cy.task('engine:trigger')` before this line, which should cause and wait for an engine cycle and thus metrics update before making the assertion.

KatieMSB: When looking for closed alerts we're checking the timestamps that are before the current time minus two minutes so even with triggering the engine there's still a two minute wait.
https://github.com/target/goalert/blob/1b3b4000a62e8249c82f2f3649f5949405cd55bb/engine/metricsmanager/db.go#L46-L47

mastercactapus: ah, I misread -- thought it was 12s for slowness, not 2min for the buffer.

for this to work we need to either insert alert log entries with custom timestamps or have the ability to move the DB clock like we do for smoketests.

dctalbot (approved):

mastercactapus (approved):

mastercactapus (approved):

dctalbot (approved):

PR #2304, opened 2022-04-18
devtools/pgdump-lite: exclude tables option

mastercactapus:
**Description:**
Adds a skip/exclude tables option as `-T`, similar to the exclude tables option in `pg_dump`.

### DIFF
--- a/smoketest/harness/harness.go
+++ b/smoketest/harness/harness.go
@@ -361,6 +361,7 @@ func (h *Harness) modifyDBOffset(d time.Duration) {
 
 	h.setDBOffset(h.delayOffset)
 }
+
 func (h *Harness) setDBOffset(d time.Duration) {
 	h.mx.Lock()
 	defer h.mx.Unlock()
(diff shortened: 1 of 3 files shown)


smoketest/harness/harness.go
```diff
@@ -579,7 +580,7 @@ func (h *Harness) dumpDB() {
 	if err != nil {
 		h.t.Fatalf("failed to get abs dump path: %v", err)
 	}
-	os.MkdirAll(filepath.Dir(file), 0755)
+	os.MkdirAll(filepath.Dir(file), 0o755)
```
dctalbot, line 583:
```suggestion
	os.MkdirAll(filepath.Dir(file), 0755)
```
Is this a typo?

mastercactapus: It's an intentional change; the form was added in Go 1.13 to make it more obvious it is an octal literal and not decimal.

https://github.com/golang/go/issues/33363

m17ch (approved):

PR #2305, opened 2022-04-18
sqltrace: remove package in favor of native interfaces

mastercactapus:
**Description:**
This PR removes the `sqltrace` package in favor of an explicit `sqldrv` package that handles the retry logic. The listener code has also been updated to use the built-in methods for accessing raw DB connections without needing a separate connection pool.

The existing tracing code isn't used/maintained, so the only important bit was the retry logic. This is part of the cleanup work for the switchover refactor.

**Review**
Add a second instance of goalert to `Procfile.local` (create if it doesn't exist)
```Procfile
goalert2: ./bin/goalert -l=localhost:3032 --ui-dir=web/src/build --db-url=postgres://goalert@localhost
```
- Test normal application operation (smoketest & cypress tests should be sufficient).
- Ensure listener code works (update config, observe reload in goalert2)


dctalbot (approved):
functionally validated

KatieMSB (approved):

PR #2320, opened 2022-04-19
fix: resolve psql test failing

tony-tvu:

**Description:**
psql-lite binary wasn't being accessed correctly as a dependency for new contributors, so this fixes that.

### DIFF
--- a/web/src/cypress/support/sql.ts
+++ b/web/src/cypress/support/sql.ts
@@ -11,7 +11,7 @@ function sql(query: string): Cypress.Chainable {
   const dbURL =
     Cypress.env('DB_URL') || 'postgres://goalert@localhost:5432?sslmode=disable'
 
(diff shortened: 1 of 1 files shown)


mastercactapus (approved):

dctalbot (approved):

Forfold:
Looks like this is failing in CI. Do we need to update our concourse worker's goalert image tag for the run-tests task?

mastercactapus:
The relative path isn't going to work for CI, I didn't think about that when I suggested it -- instead, maybe we go with the `psql-lite` as it was originally, but we can update the PATH variable for localdev to include the `bin` directory from the repo.

tony-tvu:
RESOLVED: .zshrc was overriding the path. This was an issue on the local machine.

PR #2332, opened 2022-04-25
dev: use esbuild for Cypress tests

mastercactapus:
**Description:**
This PR switches Cypress tests to use esbuild. Cypress now always runs from `bin/build/integration` instead of `web/src` ensuring identical structure and code to the way they are run on CI.

- webpack is no longer a dependency
- use imports instead of `cy.fixture()`
- tests are kept separate, instead of bundled into an `all.js` giant one

**Out of Scope**
`jest` config will need to be updated separately away from babel before removing those dependencies

**Review**
- Cypress targets: `cy-wide` `cy-wide-prod` `cy-wide-prod-run` & mobile
- prefix `make cy-wide GOALERT_HTTP_PREFIX=/foobar`

This PR only affects the cypress code so a pass in CI along with local dev commands should cover validation.


Forfold (approved):
lgtm! functionally validated with the mentioned commands for desktop and mobile as well

KatieMSB (dismissed):
Looks good, functionally validated

mastercactapus:
@dctalbot, yeah, I figure the dep cleaning would be separate. I only pulled the main ones here to ensure nothing was missed. IIRC jest might use babel & config still, so that might need some changes for that PR

dctalbot (approved):

KatieMSB (approved):

Issue #1711, opened 2021-07-07
ui/lists: `FlatList` loading state


dctalbot:
**What problem would you like to solve? Please describe:**
Currently, there is no built in loading state for the `FlatList` component. The precedent set in paginated lists is a gray skeleton placeholder for each item based on their `itemsPerPage` props. FlatList does not have this prop and so another solution may be better.

**Describe the solution you'd like:**
Based on https://material.io/components/progress-indicators, a linear progress indicators looks promising for this use case

**Describe alternatives you've considered:**
A central spinner, or skeletons

**Additional context:**
https://material-ui.com/api/linear-progress/#linearprogress-api
https://material-ui.com/components/skeleton/#skeleton

Forfold:
It would be nice if we could adapt what we do here as an example for the rest of the placeholder loading elements in the app- the original skeleton design was our best effort way back when but I think could definitely be improved this time around

arurao:
Implemented this in `PaginatedList`, move it down to `FlatList`

Forfold:
don't show the skeleton component unless loading for >1s to avoid flickering

leyasalazar:
Tony solved it May 2, 2022

PR #2355, opened 2022-05-02
ui/lists: add loading skeletons to FlatList

tony-tvu:

**Description:**
- This PR adds loading skeletons to FlatList items. 
- FlatList will render skeletons immediately but will disappear as soon as data is finished fetching as long as it is less than DEFAULT_LOAD_DELAY_MS (200 ms). If it is still fetching AFTER 200ms, then the skeletons will continue to display for DEFAULT_LOAD_WAIT_MS (1500 ms). The skeletons will only disappears once it is finished fetching after this point.
- Most instances of <FlatList /> have been updated to pass `fetching` or `loading` as a prop for FlatList’s isLoading prop
- Components that are on pages that have a loading spinner for multiple list queries (ex. ServiceDetails and UserDetails) were not updated to have a loading prop passed into FlatList. Only pages where a single FlatList is rendered has had these updates. 

Pages where new FlatList implementations are located and can be tested:
- /escalation-policies/{serviceID}/services
- /rotations/{serviceID}
- /schedules/{serviceID}/escalation-policies
- /schedules/{serviceID}/assignments
- /schedules/{serviceID}/shifts
- /schedules/{serviceID}/on-call-notifications
- /services/{serviceID}/heartbeat-monitors
- /services/{serviceID}/integration-keys
- /services/{serviceID}/labels
- /users/{userID}/schedule-calendar-subscriptions
- /users/{userID}/sessions

https://github.com/target/goalert/issues/1711

**Screenshots:**
**Without Network Throttling:**

without network throttling

**With Network Throttling:**

with network throttling


Forfold (dismissed):
lgtm!

web/src/app/lists/FlatList.tsx
```diff
@@ -376,9 +405,27 @@ export default function FlatList({
...
+  function renderSkeletons(): ReactElement | ReactElement[] {
+    const renderedSkeletons: ReactElement[] = []
+    if (displayLoading) {
+      while (items.length > renderedSkeletons.length) {
+        renderedSkeletons.push(
+          <LoadingItem key={'list_' + renderedSkeletons.length} />,
+        )
+      }
+    }
+    return renderedSkeletons
```
mastercactapus, line 417: I think we can simplify this a bit
```suggestion
    if (!displayLoading) return null

    return items.map((_, idx) => <LoadingItem key={'list_' + idx} />)
```

web/src/app/lists/FlatList.tsx
```diff
@@ -376,9 +405,27 @@ export default function FlatList({
...
+  }
+
   // renderList handles rendering the list container as well as any
   // header elements provided
   function renderList(): JSX.Element {
+    let listContent
+    if (isLoading) listContent = renderSkeletons()
+    else if (items.length <= 0) listContent = renderEmptyMessage()
```
mastercactapus, line 425: nit:
```suggestion
    else if (!items.length) listContent = renderEmptyMessage()
```

web/src/app/lists/FlatList.tsx
```diff
@@ -162,8 +181,18 @@ export default function FlatList({
...
+  isLoading,
   ...listProps
 }: FlatListProps): JSX.Element {
+  const [displayLoading, setDisplayLoading] = useState(false)
+
+  useEffect(() => {
+    // show loading skeletons after 1 second to avoid load flickering
+    setTimeout(() => {
+      setDisplayLoading(true)
+    }, 1000)
```
mastercactapus, line 193: This could get unmounted during the load and needs to be cleaned up
```suggestion
    // show loading skeletons after 1 second to avoid load flickering
    const t = setTimeout(() => {
      setDisplayLoading(true)
    }, 1000)
    
    return () => clearTimeout(t)
```

mastercactapus (changes requested):
The behavior should be a little different, in `config/index.ts` we have two values defined for spinner behavior:

```js
export const DEFAULT_SPIN_DELAY_MS = 200
export const DEFAULT_SPIN_WAIT_MS = 1500
```

Intended behavior is to not show the spinner/skeleton until 200ms have passed, and if we show it, keep it displayed for at least 1500ms, even if we load before then.

As written, the initial delay in this PR is 1000ms without retaining the loading state so there still may be a flicker.

web/src/app/lists/FlatList.tsx
```diff
@@ -162,8 +181,20 @@ export default function FlatList({
...
+  const [displayLoading, setDisplayLoading] = useState(false)
+
+  useEffect(() => {
+    // show loading skeletons after 1 second to avoid load flickering
+    const t = setTimeout(() => {
+      setDisplayLoading(true)
+    }, 1000)
+
+    return () => clearTimeout(t)
+  }, [])
```
Forfold, line 196: refer to `Spinner.tsx`'s useEffect.

you could make this a helper function that both components import

tony-tvu:
> 

This comment has been addressed in latest commits. PR summary has also been edited with latest updates and new GIFs for throttled and un-throttled behavior.

tony-tvu:
> The behavior should be a little different, in `config/index.ts` we have two values defined for spinner behavior:
> 
> ```js
> export const DEFAULT_SPIN_DELAY_MS = 200
> export const DEFAULT_SPIN_WAIT_MS = 1500
> ```
> 
> Intended behavior is to not show the spinner/skeleton until 200ms have passed, and if we show it, keep it displayed for at least 1500ms, even if we load before then.
> 
> As written, the initial delay in this PR is 1000ms without retaining the loading state so there still may be a flicker.

This comment has been addressed in latest commits. PR summary has also been edited with latest updates and new GIFs for throttled and un-throttled behavior.

PR #2361, opened 2022-05-04
cypress: fix lost input in search.js

mastercactapus:
**Description:**
This PR fixes issues where the list page search box may drop typed characters while searching.

### DIFF
--- a/web/src/app/util/Search.js
+++ b/web/src/app/util/Search.js
@@ -107,6 +107,19 @@ export default function Search(props) {
   function renderMobile() {
     return (
       <React.Fragment>
+        <AppBarSearchContainer>
+          <IconButton
+            aria-label='Search'
+            data-cy='open-search'
+            onClick={() => setShowMobile(true)}
+            size='large'
+            sx={(theme) => ({
+              color: theme.palette.mode === 'light' ? 'inherit' : undefined,
+            })}
+          >
+            <SearchIcon />
+          </IconButton>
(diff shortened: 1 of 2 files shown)


m17ch (dismissed):

web/src/app/util/Search.js
```diff
@@ -51,21 +51,26 @@ const useStyles = makeStyles((theme) => {
  */
 export default function Search(props) {
   const [searchParam, setSearchParam] = useURLParam('search', '')
+  // track the last value so we know if it changed externally
+  // or from a local event so we don't lose typed characters.
+  const [searchValue, setSearchValue] = useState(searchParam)
```
Forfold, line 56: is this a duplicate instantiation of line 59? `const [search, setSearch] = useState(searchParam)`

mastercactapus: No, but this should have a more clear name as to its purpose. This state variable tracks the previous param value so we can distinguish the difference between the URL changing (e.g. from a filter popover) and updating the URL because the user typed in the search field directly.

mastercactapus: Basically this component will call `setSearchParam` and `setSearchValue` together, but external changes will only result in `searchParam` changing so that we can use `prevParamValue !== searchParam` to determine an external change.

Forfold (approved):

Forfold (approved):

arurao (approved):

Issue #2318, opened 2022-04-19
ui: convert javascript files to typescript


Forfold:
convert remaining `.jsx` files to `.tsx` and `.js` files to `.ts`

remaining files:
(tbd)

PR #2364, opened 2022-05-04
ui/alerts: convert AlertDetails file to typescript

tony-tvu:

**Description:**
Part of an ongoing effort to convert javascript files to typescript.


web/src/app/alerts/components/AlertDetails.tsx
```diff
@@ -208,10 +240,10 @@ export default function AlertDetails(props) {
     return 'None'
   }
 
-  function renderEscalationPolicySteps() {
+  function renderEscalationPolicySteps(): ReactNode {
```
Forfold, line 243: I think this one can be `JSX.Element`. I generally use that type when returning components, and I use the `ReactNode` return type if there's a possibility that it'll return null

web/src/app/alerts/components/AlertDetails.tsx
```diff
@@ -252,7 +285,7 @@ export default function AlertDetails(props) {
     })
   }
 
-  function renderAlertDetails() {
+  function renderAlertDetails(): JSX.Element | null {
```
Forfold, line 288:
```suggestion
  function renderAlertDetails(): ReactNode {
```

web/src/app/styles/materialStyles.ts
```diff
@@ -1,4 +1,7 @@
-export const styles = (theme) => ({
+import { Theme } from '@mui/material/styles'
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const styles = (theme: Theme): any => ({
```
Forfold, line 4: I think this can be `CSSProperties` imported from `react`

```
import { CSSProperties } from 'react'

export const styles = (theme: Theme): CSSProperties => ({
```

Forfold, line 4: I think this can be `CSSProperties` imported from `react`

```
import { CSSProperties } from 'react'

export const styles = (theme: Theme): CSSProperties => ({
```

Forfold (approved):

web/src/app/alerts/components/AlertDetails.tsx
```diff
@@ -28,22 +29,50 @@ import {
...
-    epHeader: {
-      paddingBottom: 8,
-    },
-  }
-})
+interface MenuOption {
+  icon: ReactElement
+  label: string
+  handleOnClick: () => void
+}
```
Forfold, line 55: we can import this from `@mui` instead, here's what they have defined:

```
export type Action = {
  label: string // primary button text, use for a tooltip if secondary action
  handleOnClick: MouseEventHandler<HTMLButtonElement>

  icon?: JSX.Element // if true, adds a start icon to a button with text
}
```

web/src/app/alerts/components/AlertDetails.tsx
```diff
@@ -28,22 +29,50 @@ import {
...
+
+interface EscalationPolicyProps {
```
Forfold, line 57: What do you think about renaming this to Info or something similar? Since it's being used mainly for our render functions instead of being passed to an EP component

```suggestion
interface EscalationPolicyInfo {
```

web/src/app/alerts/components/AlertDetails.tsx
```diff
@@ -189,14 +221,14 @@ export default function AlertDetails(props) {
...
               : ''
-            const minTxt = parseInt(minutes)
-              ? `${minutes} minute${parseInt(minutes) === 1 ? '' : 's'} `
+            const minTxt = parseInt(`${minutes}`)
+              ? `${minutes} minute${parseInt(`${minutes}`) === 1 ? '' : 's'} `
               : ''
             const secTxt = `${seconds} second${
-              parseInt(seconds) === 1 ? '' : 's'
+              parseInt(`${seconds}`) === 1 ? '' : 's'
             }`
```
Forfold, line 232: it looks like all of these are being returned as numbers from `Countdown` so I think we can forego the `parseInt` calls altogether

```
const hourTxt = hours ? `${hours} hour${hours === 1 ? '' : 's'} ` : ''
const minTxt = minutes ? `${minutes} minute${minutes === 1 ? '' : 's'} ` : ''
const secTxt = `${seconds} second${seconds === 1 ? '' : 's'}`
```

Forfold (approved):

KatieMSB (approved):

PR #2375, opened 2022-05-09
rotation: maintain active user when deleting a user on the rotation

### LINKED ISSUE #2373: rotations: delete user results in rotation advance
**Describe the Bug:**
Deleting a user from a rotation can result in the rotation advancing before it should.

**Steps to Reproduce:**
1. Create a rotation with 3 users, A, B, and C
2. Set the 2nd user as active (user B)
3. Delete the first user from GoAlert (user A)
4. Observe user C becomes the active user

**Expected Behavior:**
Deleting a user from GoAlert should have the same effect as removing a user from a rotation: advance if the removed user is the active user, no change otherwise.

**Observed Behavior:**
The rotation advances when a user is deleted.

**Application Version:**
master as of 451e805d7befff9b3997c21e9245680e9eb26c63

mastercactapus:
**Description:**
When deleting a user as an admin, if that user was a member of a rotation, the active index would stay the same, even if deleting the user shortened the list of participants.

Fixes #2373

### DIFF
--- a/smoketest/rotationdeluser_test.go
+++ b/smoketest/rotationdeluser_test.go
@@ -1,0 +1,63 @@
+package smoketest
+
+import (
+	"encoding/json"
+	"fmt"
+	"testing"
+
+	"github.com/target/goalert/smoketest/harness"
+)
+
+// RotationDelUser tests that rotations preserve the active user when a user is deleted.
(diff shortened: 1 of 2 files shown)


KatieMSB (approved):

Forfold (approved):
lgtm and functionally tested

PR #2376, opened 2022-05-09
engine/metricsmanager: fix daily alert metrics out of sync

### LINKED ISSUE #2368: engine/alertmetrics: error when processing daily metrics
**Describe the Bug:**
It seems that daily metrics calculation can get into a state where it gets stuck failing on a duplicate value.

**Steps to Reproduce:**
1. Run latest version (as of this issue) with previous data

**Expected Behavior:**
No errors

**Observed Behavior:**
Continual error in logs each engine cycle:
```
Engine.MetricsManager: insert daily metrics: ERROR: duplicate key value violates unique constraint \"daily_alert_metrics_service_id_date_key\" (SQLSTATE 23505)
```

**Application Version:**
As of git commit 0f38a157f41a009b665cb6b55e3bfdb7e545746e

**Additional Context:**
It seems that the engine module state can get out-of-sync from existing data and the conflict isn't handled properly.

KatieMSB:

**Description:**
This PR updates the metrics manager to do nothing if data gets out of sync on inserting to daily alert metrics table.

Fixes #2368

### DIFF
--- a/engine/metricsmanager/db.go
+++ b/engine/metricsmanager/db.go
@@ -86,7 +86,8 @@ func NewDB(ctx context.Context, db *sql.DB) (*DB, error) {
 				count(*) filter (where escalated=true)
 			from alert_metrics
 			where (date(timezone('UTC'::text, closed_at))) = $1
-			group by service_id;
+			group by service_id
+			on conflict do nothing
 		`),
 	}, p.Err
 }


Forfold (approved):
lgtm, but long term should we look into handling this in some way? maybe falling back to the more recent calculation of the daily metrics?

KatieMSB:
> lgtm, but long term should we look into handling this in some way? maybe falling back and updating to the more recent calculations for the daily metrics?

I think we should be good without having to update, in this case the metrics are pulled from the day before the current day so I don't think they'll be changing.

mastercactapus (approved):

Issue #2370, opened 2022-05-06
Create Alert Dialog cannot be completed


phucnv282:
**Steps to Reproduce:**
Create a new Alert on UI

**Expected Behavior:**
The new alert was created successfully and Create Alert Dialog closes

**Observed Behavior:**
Create Alert Dialog always retry after confirming

**Screenshots/Stack Traces:**
![Screenshot from 2022-05-06 14-29-43](https://user-images.githubusercontent.com/33126800/167087853-d9202fbb-123a-490d-9512-a23d4c828330.png)

**Application Version:**
`
$ latest commit
...
`

mastercactapus:
This seemed to be an issue with apollo which should now be fixed in master as we're no longer seeing it after #2356 was merged; @phucnv282 can you confirm?

mastercactapus:
Closing as we are no longer able to reproduce in master after the fixes in #2356 were merged. Reopen if it is still an issue

PR #2381, opened 2022-05-10
bugfix/deps: fix dialogs not closing after submitting

Forfold:
We noticed in #2370 that it's currently not possible to have the `<FormDialog />` component close automatically after creating an alert. After some investigation I found [this issue](https://github.com/apollographql/apollo-client/issues/9602) that outlines a similar problem. One of the Apollo maintainers suggested updating to latest first which has some fixes for loading states. This resolved the issue for me locally.

### DIFF
--- a/web/src/package.json
+++ b/web/src/package.json
@@ -23,7 +23,7 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "@apollo/client": "3.5.7",
+    "@apollo/client": "3.6.2",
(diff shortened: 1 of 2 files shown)


Forfold:
Looks like some types are failing with the upgrade, looking into that now

Forfold:
There's a dependabot PR for this upgrade that I didn't see, so let's go with that PR instead (#2356)

PR #2386, opened 2022-05-11
ui/users: convert UserNotificationRuleForm file to typescript

tony-tvu:

**Description:**

### DIFF
--- a/web/src/app/forms/FormField.js
+++ b/web/src/app/forms/FormField.js
@@ -226,4 +226,6 @@ FormField.propTypes = {
   type: p.string,
   select: p.bool,
   timeZone: p.string,
+
+  userID: p.string,
 }
--- a/web/src/app/users/UserNotificationRuleForm.tsx
+++ b/web/src/app/users/UserNotificationRuleForm.tsx
@@ -3,22 +3,18 @@ import Grid from '@mui/material/Grid'
 import TextField from '@mui/material/TextField'
 import { FormContainer, FormField } from '../forms'
 import UserContactMethodSelect from './UserContactMethodSelect'
+import { UserNotificationRule } from '../../schema'
 
 interface UserNotificationRuleFormProps {
   userID: string
 
-  value: Value
+  value: UserNotificationRule
 
   errors: Error[]
 
   onChange: () => void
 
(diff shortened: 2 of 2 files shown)


web/src/app/users/UserNotificationRuleForm.tsx
```diff
@@ -1,11 +1,34 @@
...
 import Grid from '@mui/material/Grid'
 import TextField from '@mui/material/TextField'
 import { FormContainer, FormField } from '../forms'
 import UserContactMethodSelect from './UserContactMethodSelect'
 
-export default function UserNotificationRuleForm(props) {
+interface UserNotificationRuleFormProps {
+  userID: string
+
+  value: Value
```
KatieMSB, line 10: We can use the `UserNotificationRule` type here from the `schema.d.ts` instead of defining a new type

KatieMSB (changes requested):

KatieMSB (approved):

web/src/app/forms/FormField.js
```diff
@@ -226,4 +226,6 @@ FormField.propTypes = {
   type: p.string,
   select: p.bool,
   timeZone: p.string,
+
+  userID: p.string,
```
Forfold, line 230: not sure if we should include this here, `<FormField />` can be used pretty generally by a number of components, for example a timezone select dropdown that wouldn't have a userID on hand. which component was complaining that you needed to add this?

Forfold (changes requested):

web/src/app/forms/FormField.js
tony-tvu: This is the error I get when I remove `userID: p.string,` from the FormField props:
Screen Shot 2022-05-11 at 3 24 18 PM

Forfold: Gotcha, looks like it's being passed through to `UserContactMethodSelect` and isn't actually being used here (the line where `...otherFieldProps` is instantiated).

I don't think we can avoid needing to define it so it's probably fine to make it optional

```
userID?: p.string,
```

Forfold (approved):

PR #2388, opened 2022-05-11
calendar subscription: increase calculated shifts from 1 month to 1 year

### LINKED ISSUE #1459: Increase duration of calendar subscription
Currently, the calendar subscriptions only go out for ~30 days. I'm on call every 8 weeks and can't see when my next shift is until I enter that 30 day threshold.

Increase calendar subscriptions from 1 month to 6 or 12 months ideally

Check GoAlert application directly, but this is inconvenient

m17ch:

**Description:**
This PR increases the calendar subscription window of calculated shifts from 1 month to 1 year.

Fixes #1459 

This change will not require any action from any users that have existing subscriptions created.  The next time a calendar agent fetches from subscription endpoint, data for the full year will be pulled. 

**Notes:**
An analysis of impact using a worst case scenario for number of shifts (hourly rotation with 2 users, on call 24x7) resulted in a payload of 1.2 MB for 1 year (from 102 KB for 1 month).  No noticeable detrimental effect on query execution times or request duration on calendar subscription endpoint.  Application logs were reviewed and real-world payload sizes were substantially smaller than test case above, so the increase in window duration should not be impactful.

### DIFF
--- a/calsub/http.go
+++ b/calsub/http.go
@@ -30,7 +30,7 @@ func (s *Store) ServeICalData(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	shifts, err := s.oc.HistoryBySchedule(ctx, cs.ScheduleID, n, n.AddDate(0, 1, 0))
+	shifts, err := s.oc.HistoryBySchedule(ctx, cs.ScheduleID, n, n.AddDate(1, 0, 0))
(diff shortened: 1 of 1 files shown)


Forfold (approved):
lgtm, and thanks for taking the time to thoroughly review any potential impacts!

mastercactapus (approved):

PR #2391, opened 2022-05-12
ui: fix slow typing in admin config fields

Forfold:
In the admin config, typing is very non-performant presumably because we are re-rendering the form on every keystroke change.

Letting the fields use local state and debouncing setting the value for the mutation solves this issue.

I'm opening this as a draft because I'm currently seeing an issue where on page load the last field is rendering like it has changes to save. Clicking reset causes it to go away (until you refresh the page again). Pic shown in details below.

<details>

</details>


Forfold:
probably would be better to separate this into forms and/or memoize each section

PR #2393, opened 2022-05-12
ui: allow choosing theme color with native input

Forfold:
This PR adds a color picker to the Theme Mode (now renamed to Appearance) ButtonGroup in the Settings Popover, so users can choose whichever color they prefer as their source for GoAlert's palette. The color's hex # is saved in local storage and works for both light and dark modes.

The main primary theme has been changed to leverage `@material/material-color-utilities` for calculating the proper hex colors. This is the backend for the tool we used as a team to pick the hard-coded color values currently in master.

**Screenshots:**

Screen Shot 2022-05-12 at 2 26 12 PM

Screen Shot 2022-05-12 at 2 26 00 PM


mastercactapus (approved):

web/src/app/theme/ThemePicker.tsx
```diff
@@ -98,15 +98,22 @@ export default function ThemePicker(): JSX.Element {
...
                 onChange={(e) => {
-                  setSourceColor(e.target.value)
+                  if (e.target.value !== sourceColor) {
+                    setSourceColor(e.target.value)
+                  }
+                }}
+                onInput={(e) => {
+                  if (e.currentTarget.value !== sourceColor) {
+                    setSourceColor(e.currentTarget.value)
+                  }
```
mastercactapus, line 108: Can we trigger a `change` event in the test instead of binding to `onInput`? Also, why the need for the `!==` check?

Forfold: Through testing I found that I needed to have `onInput` as well in order to properly set the value. When I was attempting to trigger a change event nothing was happening even when I was setting force to true. I added the if statement just in case both a trigger and a change event were called simultaneously, mostly as a sanity check

mastercactapus (dismissed):

mastercactapus (approved):

tony-tvu (approved):
LGTM

Issue #2412, opened 2022-05-25
ui: drag and drop broken in strict mode


Forfold:
`react-beautiful-dnd` is no longer a maintained package and as a result is not fully supported in React 18. Production is thankfully safe, but drag and drop completely breaks in Strict Mode.

Find an alternative library + implement.

- FlatList
- EscalationPolicySteps refactor to remove the use of drag and drop (business logic decision). The idea is that escalation policies are set up once and that's generally it. Usually we see step one as the primary, step 2 as a secondary schedule or person (if applicable), and step 3 is usually a manager or team lead. After this is in place there aren't many opportunities to require drag and drop.

PR #2413, opened 2022-05-25
ui: fix drag and drop in FlatList while in strict mode

Forfold:
This PR fixes part of #2412 the drag and drop functionality when in Strict Mode by replacing the current DND library we use (`react-beautiful-dnd`) with `react-dnd`, as `react-beautiful-dnd` is no longer being maintained.

Also does some cleaning up in `FlatList`, simplifying some function calls and organizing the functions to make a bit more sense overall.

out of scope:
- removing drag and drop from `PolicyStepsCard`
- removing `react-beautiful-dnd` and `@types/react-beautiful-dnd` packages afterwards


tony-tvu (approved):
Functionally tested :white_check_mark:

tony-tvu (approved):

web/src/app/lists/DraggableListItem.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const style = {
+    transform: CSS.Transform.toString(transform),
+    transition,
+    backgroundColor: isDragging ? theme.palette.background.default : 'inherit',
+    cursor: 'pointer',
+    zIndex: isDragging ? 9001 : 1,
+  }
+
+  return (
+    <div ref={setNodeRef} style={style} {...attributes} {...listeners}>
```
KatieMSB, line 85: Looks like the listeners here are blocking the event for the other actions button for the rotation users list.

KatieMSB (changes requested):

Forfold:
group review: reverse the states as in always show edit controls, toggle button hides them, and then shows drag handles. move edit toggle to the left/in line with avatars

Forfold:
merged latest from master and wrapped up flipping the state of the toggle from the outcome of the group review, so this should be good to go again! @KatieMSB @tony-tvu

Forfold:
good catch @tony-tvu! i think that's a focus issue- once focused on the list item you should be able to tab again to focus on the drag handle. when i migrated it to FlatList i forgot about that

tony-tvu (approved):
lgtm!

Forfold:
whoops, needed to update the tests too @tony-tvu my bad! should be good to go now. running a full make check on my end to make sure the rest passes

web/src/app/lists/FlatList.tsx
```diff
@@ -154,23 +133,58 @@ export default function FlatList({
...
+  const sensors = useSensors(
+    useSensor(PointerSensor),
+    useSensor(KeyboardSensor, {
+      coordinateGetter: sortableKeyboardCoordinates,
+    }),
+  )
+  const [dndItems, setDndItems] = useState(
+    // use IDs to sort, fallback to index
+    items.map((i, idx) => (i.id ? i.id : idx.toString())),
+  )
```
KatieMSB, line 151:
```suggestion
  const [dndItems, setDndItems] = useState(
    // use IDs to sort, fallback to index
    items.map((i, idx) => (i.id ? i.id : idx.toString())),
  )
  
  useEffect(() => {
    setDndItems(items.map((i, idx) => (i.id ? i.id : idx.toString())))
  }, [items])
```

The `dndItems` can get out of sync with the list `items` if users are removed from the rotation, adding a useEffect to keep updating on change seems to fix it

KatieMSB (changes requested):

KatieMSB (approved):

tony-tvu (approved):

PR #2415, opened 2022-05-26
ui: migrate from react-router to wouter

mastercactapus:
**Description:**
Replaces the use of `react-router-dom` with `wouter`

This work is part of the CI stability work and resolves lost-event issues by switching libraries.

See: https://github.com/remix-run/react-router/issues/7460
and: https://github.com/remix-run/react-router/issues/8809

- The `/profile` pages and links continue to function, but redirect to `/user/<id>`. This wasn't strictly necessary but was part of the routing cleanup. See more below.
- Admin pages now have a prefix of `Admin:` followed by the page name.
- Alert links will now be redirected to `/services/:serviceID/alerts/:alertID`
- As long as the alert ID is valid, the URL will be "repaired"
- Breadcrumbs will now reflect the full URL path

**Additional Info:**
- Following `inithttp.go` for the backend, all routes are now registered in the `AppRoutes.tsx` file instead of scattered around the application.
- The router is no longer used to pass data to components, they now expect props directly (removes use of `useParams`)
- `Redirect` replaces `Navigate` (same behavior)
- `const [, navigate] = useLocation()` replaces `const navigate = useNavigate()` (same behavior)
- List pages now have their own components
- Most `*Router` components have been removed, the exception being the `WizardRouter` which isn't a router
- `const urlKey = useURLKey()` replaces `const { key: urlKey } = useLocation()`
- Mobile app bar auto-closes when the URL key changes, rather than manually attempting to track clicks
- `SideBarDrawerList` and `NavSubMenu` have been replaced with a rewritten component called `NavBar` filled with standardized `NavBarLink` and `NavBarSubLink` components. Styling has been left unchanged.
- `ToolbarAction` is now fed a list of routes directly from `AppRoutes.tsx` so it no longer needs to be kept in sync manually.
- `ToolbarTitle` has been reworked to fix some edge cases, like updating the document title at the right time and resolving some race conditions. It also has been generalized slightly to remove _some_ special cases.
- `refetchAll` has been added to the window to be called from tests when necessary

image
image


m17ch (approved):
LGTM 👍

KatieMSB (approved):

PR #2421, opened 2022-05-31
http: add --public-url flag to replace various URL validation approaches

### LINKED ISSUE #2359: http: strict url handling
Currently, URLs used for callbacks and validation (e.g., Twilio or auth) are a combination of using `General.PublicURL` and trying to rebuild the incoming URL from HTTP headers like host, forwarded scheme, and referer.

We should switch to using the public URL for validation and URL generation, with the added ability to override/set the Public URL by a command-line flag, falling back to the DB config value.

This will prevent issues with reverse proxies sending the wrong headers and causing hard to debug problems. We should also update the OIDC code to always issue a redirect to the Public URL instead of relying on the referer & whitelist. With the ability to set a per-instance public URL it will still be possible to achieve what was needed with the whitelist.

Work's been started in the branch `strict-url-validation` using just the Public URL via command line flag (taking precedence over `General.PublicURL`) and adding deprecation support to show it as deprecated in the UI

mastercactapus:
**Description:**
This PR adds `--public-url` as a flag to replace the global `General.PublicURL` config and `--http-prefix` flag. When in use: request validation, auth redirects, and referer validation is all done based on the `--public-url` value. For support on multiple URLs, multiple instances of GoAlert can be run for each `--public-url`.

Closes #2359 

**Screenshots:**
Deprecated config values are now called out in the UI:
image


Procfile.prod
```diff
@@ -1,7 +1,7 @@
 build: while true; do make -qs bin/goalert BUNDLE=1 || make bin/goalert BUNDLE=1 || (echo '\033[0;31mBuild Failure'; sleep 3); sleep 0.1; done
 
 @watch-file=./bin/goalert
-goalert: ./bin/goalert -l=localhost:3030 --db-url=postgres://goalert@localhost --listen-sysapi=localhost:1234 --listen-prometheus=localhost:2112
+goalert: ./bin/goalert -l=localhost:3030 --db-url=postgres://goalert@localhost --listen-sysapi=localhost:1234 --listen-prometheus=localhost:2112 --public-url=http://localhost:3030$HTTP_PREFIX
```
Forfold, line 4: looking at the other procfiles, this is the only one with `localhost:3030` to match the `-l` flag. just making sure that's intended

app/cmd.go
```diff
@@ -657,6 +659,17 @@ func getConfig(ctx context.Context) (Config, error) {
 		UIDir: viper.GetString("ui-dir"),
 	}
 
+	if cfg.PublicURL != "" {
+		u, err := url.Parse(cfg.PublicURL)
+		if err != nil {
+			return cfg, errors.Wrap(err, "parse public url")
+		}
+		if cfg.HTTPPrefix != "" {
+			return cfg, errors.New("public-url and http-prefix cannot be used together")
```
Forfold, line 668: is this because we are using the HTTP Prefix as an env variable when we build the public URL, if set?

auth/cookies.go
```diff
@@ -12,18 +16,28 @@ func SetCookie(w http.ResponseWriter, req *http.Request, name, value string) {
...
 // SetCookieAge behaves like SetCookie but also sets the MaxAge.
 func SetCookieAge(w http.ResponseWriter, req *http.Request, name, value string, age time.Duration) {
+	cfg := config.FromContext(req.Context())
+	u, err := url.Parse(cfg.PublicURL())
+	if err != nil {
+		panic(err)
+	}
+	if u.Path == "/" {
+		u.Path = "/"
+	}
```
Forfold, line 26: this seems redundant?

auth/handler.go
```diff
@@ -208,16 +208,35 @@ func (h *Handler) FindAllUserSessions(ctx context.Context, userID string) ([]Use
...
-// ServeLogout will clear the current session cookie and end the session (if any).
+// ServeLogout will clear the current session cookie and end the session(s) (if any).
 func (h *Handler) ServeLogout(w http.ResponseWriter, req *http.Request) {
-	h.setSessionCookie(w, req, "")
+	ClearCookie(w, req, CookieName)
+	var sessionIDs []string
+	for _, c := range req.Cookies() {
+		if c.Name != CookieName && c.Name != v1CookieName {
+			continue
+		}
```
Forfold, line 218: should this instead be
```suggestion
		if c.Name != CookieName || c.Name != v1CookieName {
			continue
		}
```

web/src/app/admin/AdminSection.tsx
```diff
@@ -86,8 +86,19 @@ export default function AdminSection(props: AdminSectionProps): JSX.Element {
...
-                secondary={f.description}
+                primary={(f.deprecated ? '* ' : '') + f.label}
+                secondary={
+                  f.deprecated ? (
+                    <React.Fragment>
+                      <b>Deprecated:</b> {f.deprecated}
+                      <br />
+                      <br />
+                      {f.description}
+                    </React.Fragment>
```
Forfold, line 97: I haven't verified how this looks, but
```suggestion
                    <React.Fragment>
                      <Alert severity="warning" sx={{ pb: 1 }}>
                        <AlertTitle>Deprecated</AlertTitle>
                        {f.deprecated}
                      </Alert>
                      {f.description}
                    </React.Fragment>
```

app/cmd.go
mastercactapus: The path of the `--public-url` flag is used _AS_ the prefix, and so if you specified both it would be ambiguous which one should take precedence so we have it bail out if we detect both.

Since `--public-url` is a new flag, this shouldn't cause existing setups to fail, and anyone migrating to the new flag will get an error if they forget to unset the old one (if they are using it). Safe upgrade path, with a fail-to-start scenario if misconfigured.

auth/cookies.go
Forfold: hahaha sounds good 🙂

auth/handler.go
Forfold: gotcha, in my head i read it as maybe needing to be "if the current version cookie name isn't present, or the old version isn't present, continue/skip processing"

web/src/app/admin/AdminSection.tsx
Forfold: Nice! Yeah I like that a lot, the only thing is maybe adding padding around it, and if we wanted to keep it skinnier, we could omit using the `AlertTitle` and just do `<Alert severity='warning'>Deprecated: {f.deprecated}</Alert>`

mastercactapus: image

mastercactapus: I think a good follow-up would be to have the deprecated ones hidden by default with maybe a checkbox somewhere "Show deprecated fields". For now, some of these fields are still used if the flag isn't set, so we can keep that out of scope.

KatieMSB (approved):

Forfold (approved):

PR #2433, opened 2022-06-07
ui/rotations: convert RotationAddUserDialog file to ts

tony-tvu:

**Description:**

### DIFF
--- a/web/src/app/rotations/RotationAddUserDialog.tsx
+++ b/web/src/app/rotations/RotationAddUserDialog.tsx
@@ -12,6 +12,10 @@ interface RotationAddUserDialogProps {
   onClose: () => void
 }
 
+interface Value {
+  users: string[]
+}
+
 const mutation = gql`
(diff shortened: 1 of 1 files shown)


web/src/app/rotations/RotationAddUserDialog.tsx
```diff
@@ -32,6 +45,10 @@ const RotationAddUserDialog = (props) => {
     onCompleted: onClose,
   })
 
+  if (error) {
+    return <GenericError error={error.message} />
+  }
```
Forfold, line 50: I believe formdialog handles the error so we should be able to leave this out, but then update line 61 to be `nonFieldErrors` instead of `fieldErrors`. The field errors should already be handled by the form container, so we want to make sure any other type of errors get handled by the form dialog

Forfold (changes requested):

Forfold (approved):

KatieMSB (approved):

PR #2438, opened 2022-06-08
dev: upgrade time circuits and add a flux capacitor (engine cycle monitor and pgmocktime)

mastercactapus:
**Description:**
This PR allows for the manipulation of space and time within the confines of a Postgres database.

It is now possible to move time forward, back, alter its speed, reverse its flow, and even go back to a specific point in the future. Engine cycles can be triggered using 2-stages for more efficient test runs (grab ID, trigger cycle, wait for ID).

This work also resolves a long-standing issue where the DB time code for smoketests changed the engine's perception of time, but failed to account for DEFAULT values in the tables. It is now possible to control things like alert-log timestamps to the millisecond, by freezing altering the passage of time.

- There is a new endpoint `/health/engine/cycle` that will return the next engine cycle ID
- The `/health/engine` endpoint now accepts an optional `id=` parameter
- The `pgmocktime` tool was added that has both an API and CLI interface
- Engine cycle triggering/monitoring was refactored into its own struct
- Time and engine tasks have been moved into discrete cypress helper methods (instead of `.task(...)`)
- refactored cypress engine trigger code to wait for a full cycle cleanly (mirroring backend tests)

**Screenshots:**
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,7 @@ goalert-client.key: system.ca.pem plugin.ca.key plugin.ca.pem
 goalert-client.ca.pem: system.ca.pem plugin.ca.key plugin.ca.pem
 	go run ./cmd/goalert gen-cert client
 
-cypress: bin/goalert bin/psql-lite node_modules web/src/schema.d.ts
+cypress: bin/goalert bin/psql-lite bin/pgmocktime node_modules web/src/schema.d.ts
 	yarn cypress install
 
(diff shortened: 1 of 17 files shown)


tony-tvu (approved):

tony-tvu:
Great work on this @mastercactapus and @KatieMSB :brain: !!

Forfold (approved):
lgtm!

Issue #2436, opened 2022-06-07
alert metrics: data parity 


Forfold:
- [x] Add `closedBefore` and `notClosedBefore` to `Alert` query
  - Allows `alerts` query to match graph data
- [x] Add `metrics` prop to Alert type
  - use dataloader
  - join each alert against `alert_metrics` table
- [ ] Load 100 alerts per page for up to a year

PR #2440, opened 2022-06-08
alerts: add notClosedBefore and closedBefore to query parameters

tony-tvu:

**Description:**
This PR adds `closedBefore` and `notClosedBefore` to Alert query params. The alert searchTemplate SQL query has also been updated to account for the new parameters when they are passed in. 

https://github.com/target/goalert/issues/2436


web/src/app/services/AlertMetrics/AlertMetrics.tsx
```diff
@@ -83,6 +83,8 @@ export default function AlertMetrics({
         notCreatedBefore: since.toISO(),
         createdBefore: until.toISO(),
         filterByStatus: ['StatusClosed'],
+        notClosedBefore: since.toISO(),
+        closedBefore: until.toISO(),
```
Forfold, line 87: Do we need to include this in a separate `useQuery` call since we're also setting `createdBefore` and `notCreatedBefore` here? cc @KatieMSB

KatieMSB: Good point, we probably actually don't need to set `createdBefore` and `notCreatedBefore` anymore since we only care about closed alerts for metrics.

alert/search.go
```diff
@@ -134,6 +140,12 @@ var searchTemplate = template.Must(template.New("alert-search").Funcs(search.Hel
 			{{ end }}
 		)
 	{{ end }}
+	{{ if not .ClosedBefore.IsZero }}
+		AND (select timestamp from alert_logs where alert_id = a.id and event = 'closed') < :closedBeforeTime
```
mastercactapus, line 144: This should probably be `and exists (select 1 from ...)` (`1` is just a convention for an unused placeholder). Also, since this will largely be used with metrics, I think we should join against the `alert_metrics` table instead of logs so that we don't have an intermediate state where an alert is returned from this filter before metrics are recorded. It should be a cheaper index as well.

alert/search.go
```diff
@@ -134,6 +140,12 @@ var searchTemplate = template.Must(template.New("alert-search").Funcs(search.Hel
...
+	{{ end }}
+	{{ if not .NotClosedBefore.IsZero }}
+		AND (select timestamp from alert_logs where alert_id = a.id and event = 'closed') > :notClosedBeforeTime
```
mastercactapus, line 147: Same ^

mastercactapus (changes requested):

mastercactapus (approved):

PR #2443, opened 2022-06-09
dataloader: convert to generics

mastercactapus:
**Description:**
This PR replaces the individual copy-paste `dataloader` code with generics added in Go 1.18.

This should make it easier to add new data loaders as needed for the GraphQL layer.


tony-tvu (dismissed):

graphql2/graphqlapp/dataloaders.go
```diff
@@ -114,17 +117,16 @@ func (app *App) FindOneCM(ctx context.Context, id string) (*contactmethod.Contac
 
 // FindOneNC will return a single notification channel for the given id, using the contexts dataloader if enabled.
 func (app *App) FindOneNC(ctx context.Context, id uuid.UUID) (*notificationchannel.Channel, error) {
-	loader, ok := ctx.Value(dataLoaderKeyNC).(*dataloader.NCLoader)
+	loader, ok := ctx.Value(dataLoaderKeyNC).(*dataloader.Loader[string, notificationchannel.Channel])
```
Forfold, line 120: would it be worth adding a `NewStoreLoaderUUID()` for these?
```suggestion
	loader, ok := ctx.Value(dataLoaderKeyNC).(*dataloader.Loader[uuid.UUID, notificationchannel.Channel])
```

Forfold (dismissed):

graphql2/graphqlapp/dataloaders.go
mastercactapus: Not yet; globally, I think we should switch to UUID types instead of strings for all of our UUID fields, but that gets a little complicated.

I'd say that since the store methods use strings today, we should leave that out of the scope of this PR. To make that change, we'll need to update the `FindMany` methods to take UUIDs in, the struct ID fields, and the graphql input types to match. Could be another issue like class->hooks, interface->struct, js->ts, etc... though.

Forfold: Gotcha, that makes sense!

Forfold (approved):

tony-tvu (approved):

PR #2444, opened 2022-06-09
graphql: add metrics prop to Alert type

mastercactapus:
**Description:**
This PR adds a dataloader and a `metrics` prop to the `Alert` type.

**Out of Scope**
Only the `escalated` value is supported, `timeToAck` and `timeToClose` will be handled in a separate PR.


tony-tvu (approved):

Forfold (approved):
lgtm!

PR #2465, opened 2022-06-23
alertmetrics: process metrics calculation in background

### LINKED ISSUE #2436: alert metrics: data parity

mastercactapus:
**Description:**
This PR moves graph and CSV processing to a web worker, when supported.

- alerts query limit increased to 1000
- metrics page now goes back to up to 1 year, and the graph aggregation can be switched between daily, weekly, and monthly
- a new `useWorker` hook can be used to process the result of a hook in a web worker, with graceful degradation
- graph and table data now only extends to the start of the current interval*
- a new `worker.js` file is now built for web workers

*Note: This means alerts from this week/month/day (dep. on current interval) will not be included in the graph or table since the counts and times wouldn't be a valid comparison since the final data point isn't complete. If the current month is June, then 1 year of data will display LAST June through May -- 12 full months, skipping the current partial month.

Closes #2436


tony-tvu (approved):
Functionally works as expected and code is readable.

KatieMSB (approved):

PR #2481, opened 2022-06-29
ui/admin: add message logs notification count line graph

### LINKED ISSUE #2460: ui/message logs: add notification count by phone number
When looking through message logs either for an issue for a customer or for debugging purposes, we need a better way to filter through and see user data.

As of today, the auth limit is quickly exceeded, and there is no way to see an overview of the data shown.

- Show line graph of counts by day on the route for message logs
- The graph only displays if you are searching for a number
- The page title reads "Message Logs" instead of "message-logs"

**Additional Context:**
 - Omitting `status_bundled` from the current message_logs query should be sufficient
- Keep auth limit in mind (use sane default limits for the search query)

**Next Up / Out Of Scope:**
1. Filter by sent/delivered (goalert to the provider), pending, and failed (all active by default)
2. Have options to breakdown graph interval by day, hour, or minute buckets
3. Select by user
4. Select by service
5. Show currently throttled users
    - Sublist? Maybe as a sidebar on message logs
6. Paginate/search functionality (will solve exceeded auth limit issues)

Forfold:
Part of #2460, related to #2600

- Updated the top controls to better fit the graph
- Uses the first and last dates of the currently fetched message logs for the initial graph display
- Prioritizes using the start/end times for the graph display, if set in the filter controls up top
- Shows line graph of counts by day on the route for message logs
- The graph can be hidden under an accordion
- The page title reads "Message Logs" instead of "message-logs"
- Notifications that get bundled are omitted from showing (as nothing is actually sent to the user for these log entries)
- Can change the interval of the graph to go by daily, hourly, or by the minute. Default: daily

Screen Shot 2022-08-08 at 12 57 52 PM


tony-tvu (approved):
I was noodling around with the Autosizer recommendation from [this thread](https://github.com/recharts/recharts/issues/172) with no success. Glad to see you got it to work!

tony-tvu (dismissed):

tony-tvu (approved):

web/src/app/admin/admin-message-logs/AdminDebugMessagesLayout.tsx
```diff
@@ -77,50 +62,84 @@ export default function AdminDebugMessagesLayout(): JSX.Element {
...
   const startDT = params.start ? DateTime.fromISO(params.start) : null
   const endDT = params.end ? DateTime.fromISO(params.end) : null
 
-  const filteredResults = results.filter((result) => {
-    const createdAtDT = DateTime.fromISO(result.item.createdAt)
-    if (startDT && startDT > createdAtDT) return false
-    if (endDT && endDT < createdAtDT) return false
-    return true
-  })
+  const filteredData: DebugMessage[] = data?.debugMessages
```
mastercactapus, line 77: Can we move the filtering and bucketing into a separate hook and maybe pass it through the web worker?

graphql2/graphqlapp/query.go
```diff
@@ -173,6 +173,11 @@ func (a *Query) DebugMessages(ctx context.Context, input *graphql2.DebugMessages
 			return nil, fmt.Errorf("format dest: %w", err)
 		}
 
+		// notifications that end up bundled are omitted
+		if m.MessageType == notification.MessageTypeAlert && m.LastStatus == notification.StateBundled {
```
mastercactapus, line 177: This should be handled in the DB query so the limit applies **after** they are filtered out

web/src/app/admin/admin-message-logs/AdminDebugMessagesLayout.tsx
```diff
@@ -77,50 +62,84 @@ export default function AdminDebugMessagesLayout(): JSX.Element {
...
+  let ivl: Interval | null = null
+  if (s && e && hasData) {
+    ivl = Interval.fromDateTimes(s, e)
+  }
 
-  if (error) return <GenericError error={error.message} />
-  if (loading && !data) return <Spinner />
+  const intervalType = 'daily'
+  const graphData = ivl
+    ? ivl.splitBy({ days: 1 }).map((i) => {
```
mastercactapus, line 124: Can we add a field to make this configurable? Daily won't be super helpful for determining if throttling is happening; we'll want hour-by-hour and even minute-by-minute

I want to be careful of scope, but I think that field is necessary to get what we need from the graph.

The alert metrics code has an example of the separate hook, bucketing, and worker code -- let me know you'd like additional help/context with this as it's relatively new in the repo

mastercactapus (changes requested):

Forfold:
no longer blocked, but will need updating to work with the pagination changes merged from #2644

Forfold:
should be ready for review again 👌

mastercactapus (changes requested):
We need to fix the page rendering hang; it freezes up while loading data on the graph several times and isn't usable.

Here are a few things we can do:
- See if we can calculate the graph data more efficiently
- Run through the `useWorker` helper to put the graph calculation code in the background and out of the render loop. This will also allow caching data, so we're not re-calculating everything on every page render.
- Moving the data/page fetching stuff into the worker as well (would help ALL graphs)

---

As for optimizing, doing a quick profile I noticed right away most of our time is in this recursive filter:
image

image

In that case, for each data point, we filter through the entire list of data, parsing the time from the ISO string for every item. And this is happening on every render call, even if the data doesn't change.

Since they are normalized ISO strings, we *could* probably do a string compare, and also, if we walk through the list, segmenting as we go, we can avoid going through the list more than once. Lastly, the worker helper should handle caching the result so that each render will be quick and the data will filter in as it gets processed.

mastercactapus:
I started taking a look at this. Though it's a bit big, I've gotten pulled away on a few occasions, which has made it difficult to finish the review. Going to focus on functional validation first.

If there are some pieces you can break out into "pre-flight" PR(s), we can prioritize getting those in immediately to help shrink it down.

web/src/app/admin/admin-message-logs/AdminMessageLogsControls.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+export default function AdminMessageLogsControls(): JSX.Element {
+  const [params, setParams] = useURLParams({
+    search: '',
+    start: DateTime.now().minus({ hours: 8 }).toISO(),
+    end: DateTime.now().toISO(),
+  })
+
+  const [filterByUser, setFilterByUser] = useState('')
+  const [filterByService, setFilterByService] = useState('')
```
tony-tvu, line 20: It looks like changing the search value and/or filtering by user/service is not re-triggering `useMessageLogs()` to get filtered logs. I believe you mentioned that this story is just adding the graph, so should we disable the search/filtering capabilities for now and address that in a separate story?

tony-tvu (changes requested):

Forfold:
@tony-tvu After looking at this again, yeah the user and service selects are in local state so nothing ever happens. I'll remove them for this PR so it's easier to review.

web/src/app/admin/admin-message-logs/AdminMessageLogsControls.tsx
Forfold: @tony-tvu After looking at this again, yeah the user and service selects are in local state so nothing ever happens. I'll remove them for this PR so it's easier to review.

Forfold: Removed 👍

mastercactapus:
I started looking at this again; part of what has made it hard to review is the changes mixed with moving/renaming files, which makes it more difficult to follow what is being changed.

It might be worth separating the reorganization from the refactoring/feature changes, assuming this didn't require a complete rewrite.

Forfold:
@mastercactapus sure, I'll rename the files in a separate PR. There was some refactoring, but it's been awhile so the diff will be good for all of us to see that way.

web/src/app/admin/admin-message-logs/useMessageLogs.ts
```diff
@@ -0,0 +1,141 @@
...
+        q.data.messageLogs.nodes,
+        q.data.messageLogs.pageInfo.hasNextPage,
+        q.data.messageLogs.pageInfo.endCursor,
+        undefined,
+      ]
+    }
+
+    const throttledSetLogs = _.throttle(
+      (logs, loading) => {
+        setLogs(logs) // todo: additional sorting needed?
```
tony-tvu, line 100: It may be helpful to sort by createdAt here.

tony-tvu (dismissed):
Functionally working as expected. Just left a couple comments.

web/src/app/admin/admin-message-logs/useMessageLogs.ts
Forfold: The idea is that it should always come back sorted from the backend, but I'll revisit this to be sure so we don't put any extra load on the front end since it's already doing it's best to get everything loaded

Forfold: Sounds good 👍 I'll make sure we sort on the backend to avoid any extra strain on the browser while it's loading in data.

tony-tvu (approved):
LGTM

web/src/app/admin/admin-message-logs/AdminMessageLogsGraph.tsx
```diff
@@ -0,0 +1,241 @@
...
+          ' - ' +
+          DateTime.fromISO(end).toLocaleString(locale),
+      })),
+    [stats],
+  )
+
+  const formatIntervals = (label: string): string => {
+    // check for default bounds
+    if (label.toString() !== '0' && label !== 'auto') {
+      const dt = DateTime.fromFormat(label, 'MMM d, t')
```
mastercactapus, line 115: We can't parse locale strings -- they will change for different systems (e.g., locally, the page crashes for me since the format doesn't match) and are meant to be human-readable vs. standardized.

Can we update the label to be ISO-strings and then use this formatter to handle rendering them correctly?

Forfold: @mastercactapus The label is coming from the `recharts` `XAxis` component- I copied the code from the Alert Counts Graph.

I'll look into it though since it's crashing on your end.

mastercactapus:
Jumping on a different issue, but possible TODO: fill in the gaps with zeros. The backend returns sparse data, so the graph only displays counts >0. We could fix this in either place, but it's probably the cleanest/easiest in Go from the service side.

web/src/app/admin/admin-message-logs/util.ts
```diff
@@ -0,0 +1,23 @@
...
+  start: string
+  end: string
+  graphInterval: string
+}
+
+export function useMessageLogsParams(): [
+  MessageLogsParams,
+  (params: Partial<MessageLogsParams>) => void,
+] {
+  const end = DateTime.now().startOf('hour').plus({ hour: 1 })
```
Forfold, line 15: This function in its own file kind of seems like overkill since they are both just grabbing some items from the URL

mastercactapus: For this to work correctly, all three components must have completely identical defaults, which is why I made the shared function. It would also be hard to debug if it were to get out of sync.

A shared `util.ts` is the convention we've used elsewhere for shared functions/hooks.

---

Elsewhere we've read the parameters in a single place and then passed the values down to other components as props -- that could be an alternative.

mastercactapus (dismissed):

m17ch (approved):
LGTM 👍

mastercactapus (approved):

m17ch (approved):

PR #2512, opened 2022-07-18
services: add maintenance mode option to snooze alert notifications

### LINKED ISSUE #2490: service maintenance: pause outgoing notifications
## Goals
- [x] Manually disable notifications for a single service
  - e.g. during a storm of alerts and the on-call engineer is aware of the situation but needs time to resolve
  - e.g. during planned maintenance to avoid heartbeat alerts and similar
- [x] can't "forget" and stay in maintenance mode (needs to expire)
- [x] integration keys still work (alerts can be created & closed- just no outgoing notifications)

### MVP Implementation Notes
- [x] manual escalation returns error ("in maintenance mode")
- [x] increment esc. mgr module
- [x] max of 8 hours (api validation)
- [x] optimistically update UI after create and cancel
- [x] Service{} prop `maintenanceExpiresAt`
- [x] only prevent any/all escalations
- [x] mode is always from current timestamp-on (not extending original/no stacking)
- [x] store expire time in DB
- [x] add `maint_expires_at` nullable column to `services` table
- [x] UpdateService{} prop `maintenanceExpiresAt`
- [x] cancel by setting expires at to time in the past

### Design
- [x] countdown time remaining in notice
- [x] warning notice at top of svc alerts page
- [x] warning notice at top of svc alert details page
- [x] use DetailsPage primary action
- [x] 1, 2, and 4 hour options
- [x] info/tooltip or text explaining what maintenance mode does
- [x] warning notice at top of svc details page
- [x] cancel maintenance mode button in notice

### smoketest
- [x] ensure existing notification rules continue for pre-existing alert
  - e.g. A user has two notification rules, text immediately and call after 5 minutes. An alert gets created, and then the service gets put in maintenance mode. As of today, after 5 minutes, we expect the alert to still get sent out even though the service is in maintenance mode.
- [ ] ep can't be escalated while in maintenance mode

### cypress test
- [x] start maintenance mode, see notice banner
- [x] create alert during maintenance mode, escalated to step 1 shouldnt exist
- [x] cancel maintenance mode

## Out of Scope
- update alert log when alert created, if in maintenance mode
- log an entry in all open alerts when maintenance mode start/stops
- integration key disable
- history/who set it and when
- "Specify..." option for up to 8 hrs
- scheduled maintenance
- add set maintenance mode reply option from bundled alert notifications

### Details page redesign (cards)
- similar to grafana panels how they fill the page dynamically
- summary/preview, open to route with full experience
- started discussion, add thoughts here: https://github.com/target/goalert/discussions/2489

Examples:
- maintenance mode with visual clock counting down, 1H/2H/4H buttons inline
- recently used integration keys
- unhealthy heartbeat monitors
- alert count graph

### slack notifications
- at start
- at stop
- an optional 5min warning?

Forfold:
Closes #2490

Adds a button to configure maintenance mode for 1, 2, or 4 hours from any service details page. All notifications during this time period will be silenced, and escalations will not be processed. Any incoming alerts will still be created (snoozing integration keys out of scope).

After maintenance mode ends, alerts will be sent out as normal to the on-call user.

**Screenshots:**
<details>
  <summary>
    Service details maintenance mode button
  </summary>
  <br />
  
</details>

<details>
  <summary>
    Maintenance mode dialog
  </summary>
  <br />
  
</details>

<details>
  <summary>
    Maintenance mode notice banners
  </summary>
  <br />
  

---

  

---

  
</details>

<details>
  <summary>
    Escalate alert GraphQL error while in maintenance mode
  </summary>
  <br />
  
</details>


alert/store.go
```diff
@@ -293,18 +300,37 @@ func (s *Store) EscalateMany(ctx context.Context, alertIDs []int) ([]int, error)
...
-	tx, err := s.db.BeginTx(ctx, nil)
+		err = tx.StmtContext(ctx, s.maintenanceMode).QueryRowContext(ctx, svcID).Scan(&maintExpiresAt)
+		if err != nil {
+			return nil, err
+		}
+
+		if maintExpiresAt.Valid {
+			return nil, validation.NewGenericError("escalate alert: in maintenance mode")
+		}
+	}
```
mastercactapus, line 326: Might be able to simplify this, but in any case, should happen _after_ the `lockAlertSvc` line; otherwise, we have a race condition from the moment we check the value and the time we actually grab the lock.

engine/escalationmanager/db.go
```diff
@@ -82,6 +83,12 @@ func NewDB(ctx context.Context, db *sql.DB, log *alertlog.Store) (*DB, error) {
 			returning ep_step_id, user_id
 		`),
 
+		cleanupMaintenanceMode: p.P(`
```
mastercactapus, line 86: nit: be a little more specific in what this one does in the name

```suggestion
		clearMaintExpiredSvc: p.P(`
```

engine/escalationmanager/db.go
```diff
@@ -82,6 +83,12 @@ func NewDB(ctx context.Context, db *sql.DB, log *alertlog.Store) (*DB, error) {
...
+				update services s
+				set maintenance_expires_at = null
+				where s.maintenance_expires_at notnull and now() > s.maintenance_expires_at
```
mastercactapus, line 89: This can be simplified as `>` already requires `notnull` and should be `>=` since it express *at*

Also, a nit, but `table_row` - `comparator` - `value` order would be more consistent with the rest of the module.
```suggestion
				where s.maintenance_expires_at <= now()
```

engine/escalationmanager/db.go
```diff
@@ -93,13 +100,15 @@ func NewDB(ctx context.Context, db *sql.DB, log *alertlog.Store) (*DB, error) {
...
+				select alert_id, step.id ep_step_id, step.delay, step.escalation_policy_id, a.service_id, s.maintenance_expires_at
 				from escalation_policy_state state
 				join escalation_policy_steps step on
 					step.escalation_policy_id = state.escalation_policy_id and
 					step.step_number = 0
 				join alerts a on a.id = state.alert_id and (a.status = 'triggered' or state.force_escalation)
-				where state.last_escalation isnull
+				join services s on a.service_id = s.id
+				where state.last_escalation isnull and
+				(s.maintenance_expires_at isnull or s.maintenance_expires_at < now())
```
mastercactapus, line 111: Just do the `isnull` comparison here, cleanup will handle updating it, and we can simplify the work Postgres needs to do for the join.

```suggestion
				(s.maintenance_expires_at isnull)
```

mastercactapus (changes requested):
For the escalation code, I had a couple of thoughts, what's the expected behavior when escalating multiple alerts when only one or two are in maintenance mode?

When we talked about it, we figured an error should be returned, but the method signatures suggest we should instead have it do nothing, as it's supposed to return only the alerts that were escalated.

`escalateAlerts(input: [1]) -> []`

As it stands, if one alert is for a service in maintenance mode, all will fail with a generic error. It's more a limitation of how the API is set up currently.

I'll reach out to chat more about it

Forfold:
@mastercactapus

> For the escalation code, I had a couple of thoughts, what's the expected behavior when escalating multiple alerts when only one or two are in maintenance mode?

totally did not think of this use case, we were thinking in terms of `escalateAlerts(input: [id])` the whole time since we were only testing with one alert. I think it makes sense to return maybe an array of true/error or something similar?

alert 1: active service
alert 2: maint. mode service

escalateAlerts(ids: [1, 2])

returns

```
{
  "errors": [{
    message: "escalate alert 2: in maintenance mode",
    path: ['escalateAlerts']
  }],
  "data": {
    "1": true
  }
}
```

web/src/app/services/ServiceMaintenanceDialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+  expiresAt?: string
+  onClose: () => void
+}
+
+function calcExp(hours: number): string {
+  return DateTime.now().plus({ hours }).toISO()
+}
+
+function label(hours: number): string {
+  return `Until ${DateTime.fromISO(calcExp(hours)).toFormat('t ZZZZ')}`
```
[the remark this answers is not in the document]
Forfold: good catch, ty 🙂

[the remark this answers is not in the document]
Forfold: Using `DateTime.DATETIME_MED` as it renders the same as `t ZZZZ` 👍

mastercactapus (approved):

mastercactapus (approved):

mastercactapus (approved):

KatieMSB (approved):

Issue #2488, opened 2022-07-01
ui/wizard: character limit on name


Forfold:
Names are currently limited to 64 characters. The setup wizard appends strings to the end of each target created, e.g. "Test Service" and "Test Primary Schedule". This can cause the 64-character limit to be reached prematurely.

mastercactapus:
- Remove suffix (e.g., Service, Schedule)
- Append Primary/Secondary only when two schedules are in-use
- Update validation on the UI when Primary/Secondary is used to show corrected limit (54 if 2 schedules are used instead of 64)

tony-tvu:
https://github.com/target/goalert/pull/2515

PR #2515, opened 2022-07-19
ui/wizard: fix wizard character limit

tony-tvu:

**Description:**
- This PR fixes the issue where a 64 character limit exceeded error occurred whenever users entered a name less than 64 characters but exceeded the limit whenever suffixes or timezones were appended. 
- Suffixes are no longer appended, i.e. "Service", "Escalation Policy", "Rotation"
- "Primary" and "Secondary" keywords are appended only if a service has two schedules
- Rotations that have Follow the Sun (FTS) support will have their timezones appended to the **rotation's name**. This solution was implemented to resolve `Name: already in use` errors whenever a schedule is created with FTS support that has a rotation with a timezone different than its original timezone, which resulted in two rotations with the same names sent to the backend. 
- Name length validation error messages are now customized based on user's needs. For example, if a user creates a service with one schedule without FTS, then they will see `cannot be more than 64 characters` error when they've exceeded the limit. If a user creates a service that has two schedules without FTS, then they will see `cannot be more than 54 characters` error since ' Secondary' is an additional 10 characters subtracted from 64. These scenarios are accounted for in this PR's cypress test.

https://github.com/target/goalert/issues/2488


KatieMSB (approved):
lgtm

KatieMSB (changes requested):
Unit tests ( `app/wizard/util.test.js`) for the wizard also need updated to reflect the new expected output

KatieMSB (approved):

Forfold (approved):

PR #2564, opened 2022-08-17
auth: link slack user

tony-tvu:

**Description:**
This PR facilitates linking slack users to their GoAlert accounts, which would allow them to acknowledge or close an alert from the Slack UI once linked.

**Out of Scope:**
Devtool's mock slack server has been updated to account for block elements

**Screenshots:**
1. Alert sent to unlinked slack user
Screen Shot 2022-08-17 at 10 49 34 AM

2. Clicking 'Link Account' button routes them to GoAlert's profile page to confirm linking
Screen Shot 2022-08-17 at 10 50 40 AM

3. Once linking is confirmed, it'll automatically acknowledge or close the alert based on the selection at stage 1
Screen Shot 2022-08-17 at 10 51 01 AM

Screen Shot 2022-08-17 at 10 51 48 AM

4. Linked user can now Close the alert from the slack UI
Screen Shot 2022-08-17 at 10 52 00 AM


mastercactapus (changes requested):
Here's what I think we have left:
- Update the messaging to be a little more informative, like: "Slack user <User.Name> (@<User.Username>) from <Team.Domain>.slack.com"
- We should show errors in the dialog before closing (e.g., network issue)
- Unknown/invalid/expired tokens should have consistent messaging
- We should probably add the details in the token, or at least a signature -- it's too easy to make a URL that will link a user in a way they didn't expect; we can have a query to validate the token before the user confirms

mastercactapus:
I moved the metadata to a table column, that way, to get _any_ info, you need a valid GoAlert session AND the token/url, and also handles ensuring the messaging the user gets is also validated and isn't subject to a malicious/misleading link.

mastercactapus (approved):

mastercactapus (approved):
Leaving integration test out-of-scope; we'll add that as a test when the mockslack server gains a UI

KatieMSB (approved):

PR #2575, opened 2022-08-18
feat: Adding more natural sounding voices to phone calls

### LINKED ISSUE #2574: Adding more natural sounding voices to phone calls
The default Twilio Basic Man voice is not good, and sounds very robotic.  Twilio supposed much better Text To Speech voice engines, such as Amazon Polly Neural.  When using TTS engines, it is also best to set the language so the TTS engine will cover the local vocal nuances.

Add options to default the language to `en-US`, like it currently is, and set the default TTS voice to Polly Neural.  It should also have the option to easily adjust it to something else.

NA

**Additional context:**
NA

### LINKED ISSUE #2576: Adjusting Phone Call Greetings to sound more natural
Currently the phone call structure of `This is {Appname} {details} {options}` sounds run together like once long sentence.

I propose to adjust the phone call structure to `Hello! This is {Appname} <700ms pause> {details} {options}`

**Additional context:**

AlaricWhitney:

**Description:**
This PR updates the application to have be able to set the Twilio Phone call language and Text To Speech engine. 
- Currently the default language is "" (which Twilio interprets as `us-EN`).  This PR allows you to adjust it to whatever you need.
- Currently the default TTS engine is Twilio Basic Man, which sounds like a robot.  This PR allows you to adjust it to whatever you need.
- The greeting is also adjusted to make it more natural sounding by adjusting the verbiage, and adding a pause between the greeting and the alert message.

Fixes #2574 
Fixes #2576

**Out of Scope:**
NA

**Screenshots:**
NA

By default, nothing changes for the user, however the user does have the new ability to change the TTS voice engine and the language for phone calls.

NA

**Additional Info:**
NA


config/config.go
```diff
@@ -360,6 +363,22 @@ func (cfg Config) ApplicationName() string {
...
+	return cfg.Twilio.VoiceName
+}
+
+// VoiceLanguage will return the Twilio.VoiceLanguage value
+func (cfg Config) VoiceLanguage() string {
+	if cfg.Twilio.VoiceLanguage == "" {
+		return "en-US"
+	}
+	return cfg.Twilio.VoiceLanguage
+}
```
mastercactapus, line 380: The default should be empty so that the voice config continues to be whatever is configured in the Twilio account unless a value is set

config/config.go
```diff
@@ -508,6 +527,14 @@ func (cfg Config) Validate() error {
 		)
 	}
 
+	if cfg.Twilio.VoiceName != "" {
+		err = validate.Many(
+			err,
+			validate.ASCII("Twilio.VoiceName", cfg.Twilio.VoiceName, 1, 50),
+			validate.ASCII("Twilio.VoiceLanguage", cfg.Twilio.VoiceLanguage, 0, 10),
+		)
+	}
```
mastercactapus, line 536: We can simplify this since neither is required

```suggestion
	err = validate.Many(
		err,
		validate.ASCII("Twilio.VoiceName", cfg.Twilio.VoiceName, 0, 50),
		validate.ASCII("Twilio.VoiceLanguage", cfg.Twilio.VoiceLanguage, 0, 10),
	)
```

notification/twilio/twiml.go
```diff
@@ -134,7 +142,7 @@ func (t *twiMLResponse) sendResponse() {
 		io.WriteString(t.w, `">`+"\n")
 	}
 	for _, s := range t.say {
-		io.WriteString(t.w, `<Say><prosody rate="slow">`)
+		io.WriteString(t.w, fmt.Sprintf(`<Say voice="%s" language="%s"><prosody rate="slow">`, t.voiceName, t.voiceLanguage))
```
mastercactapus, line 145: Only set the voice and/or language if the admin has explicitly set the new config option. An admin should expect no changes to the outgoing configuration from applying an update; so we need to introduce the feature in a way that is completely opt-in.

notification/twilio/voice.go
```diff
@@ -187,7 +187,7 @@ func (v *Voice) Send(ctx context.Context, msg notification.Message) (*notificati
 		Params:         make(url.Values),
 	}
 
-	prefix := fmt.Sprintf("This is %s", cfg.ApplicationName())
+	prefix := fmt.Sprintf("Hello! This is %s. <break strength=\"x-strong\" time=\"700ms\"/>", cfg.ApplicationName())
```
mastercactapus, line 190: We will need to update all the messages, too, with this change.

Currently, the prefix is spoken as part of the introduction, "This is {application} with alert notifications {details}" we'll want to address the text in the switch statement below.

Maybe drop a message in the chatroom about the proposal to ensure buy-in. My understanding is you're proposing we restructure voice messages from "{Appname & message type} {details} {options}" to "{appname} {details} {options}" or maybe the intent is "{appname} {type} {details} {options}"?

In any case, it's worth a few minutes to chat about the how and why so we can review and ensure all calls are consistent and all reviewers know what to look for.

notification/twilio/voice.go
```diff
@@ -572,10 +581,12 @@ func (v *Voice) ServeAlert(w http.ResponseWriter, req *http.Request) {
 	if call == nil {
 		return
 	}
+	cfg := config.FromContext(ctx)
 
 	// See Twilio Request Parameter documentation at
 	// https://www.twilio.com/docs/api/twiml/twilio_request#synchronous
 	resp := newTwiMLResponse(w)
```
mastercactapus, line 588: Let's add `ctx` as an arg to `newTwiMLResponse` and have it read in the config internally so we don't need to track calling `AddVoiceOptions` at every call site

mastercactapus (changes requested):

notification/twilio/voice.go
AlaricWhitney: Actually, this is just a minor wording an vocal pacing update.  The structure for phone calls will still remain `{Appname} {details} {options}`, and that is untouched.  The current Greeting is `This is GoAlert {details} {options}` which is all run together and sounds like it's all one long sentence.  This change simple adjusts the greeting to `Hello! This is GoAlert. <700ms pause> {details} {options}`, which adds a natural vocal separation between the greeting, and the message details.

config/config.go
AlaricWhitney: adjusted as per your request

config/config.go
AlaricWhitney: adjusted as per your request.  Technically you require a language if you specify a voice, which I adjusted the check to validate.

notification/twilio/twiml.go
AlaricWhitney: adjusted as per your request.  Now it will only be set if the config option is set, otherwise nothing changes.

notification/twilio/voice.go
AlaricWhitney: adjusted as per your request

notification/twilio/twiml.go
```diff
@@ -133,8 +142,18 @@ func (t *twiMLResponse) sendResponse() {
...
+
+	// add optional voice and/or language options
+	voiceAndLanuage := ""
+	if t.voiceName != "" && t.voiceLanguage != "" {
+		// language is required with a voice
+		voiceAndLanuage = fmt.Sprintf(` voice="%s" language="%s"`, t.voiceName, t.voiceLanguage)
+	} else if t.voiceLanguage != "" {
+		// language can be set without a voice
+		voiceAndLanuage = fmt.Sprintf(` language="%s"`, t.voiceLanguage)
+	}
```
mastercactapus, line 154: nit: we can simplify this and avoid the `else` by just doing an `if` for each, the name-then-lang requirement is already enforced by config validation.

```go
var opts string
if voice {
  opts += ` voice="%s"`
}
if lang {
  opts += ` lang="%s"`
}
```

**However,** the bigger issue here is that we need to escape the values otherwise this is vulnerable to TwiML injection through the UI.

We could lift and use the escaping code from the XML lib [here](https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/encoding/xml/xml.go;drc=8d57f4dcef5d69a0a3f807afaa9625018569010b;l=1961), but I'm thinking it might be a good sign that it's time to use the proper XML library to build this.

notification/twilio/twiml_test.go
```diff
@@ -1,19 +1,23 @@
...
 	"net/http/httptest"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/target/goalert/config"
 )
 
 func TestTwiMLResponse(t *testing.T) {
 	t.Run("hangup", func(t *testing.T) {
+		mockConfig := config.Config{}
```
mastercactapus, line 16: Use `var` for empty declarations, also since it's just a regular config object (not a fake/mock), I think we can call these `cfg`
```suggestion
		var cfg config.Config
```

notification/twilio/voice.go
mastercactapus: If you look at line 196, for example, we are appending the reason for the call to the prefix, which places it after the pause as a broken sentence.

The current greeting is:
`This is GoAlert with an alert notification. {details} {options}`

With this change it is:
`Hello! This is GoAlert. <700ms pause> with an alert notification. {details} {options}`

I think it should go before the pause, or be rephrased to be a complete sentence.

For example:
`Hello! This is GoAlert with an alert notification. <700ms pause> {details} {options}`

config/config.go
```diff
@@ -97,6 +97,9 @@ type Config struct {
 	Twilio struct {
 		Enable bool `public:"true" info:"Enables sending and processing of Voice and SMS messages through the Twilio notification provider."`
 
+		VoiceName     string `public:"true" info:"The Twilio voice to use for Text To Speech for phone calls."`
+		VoiceLanguage string `public:"true" info:"The Twilio voice language to use for Text To Speech for phone calls."`
```
mastercactapus, line 101: Are we using these in the UI outside of the admin config panel? If not, we can leave the `public` tag off. That is only intended to indicate fields that should be readable by non-admin users.

config/config.go
```diff
@@ -508,6 +511,14 @@ func (cfg Config) Validate() error {
 		)
 	}
 
+	if cfg.Twilio.VoiceName != "" {
+		err = validate.Many(
+			err,
+			validate.ASCII("Twilio.VoiceName", cfg.Twilio.VoiceName, 1, 50),
+			validate.ASCII("Twilio.VoiceLanguage", cfg.Twilio.VoiceLanguage, 1, 10), // language is required when a voice name is set
+		)
+	}
```
mastercactapus, line 520: Don't let language escape validation if name is unset

```suggestion
	if cfg.Twilio.VoiceName != "" {
		err = validate.Many(
			err,
			validate.ASCII("Twilio.VoiceName", cfg.Twilio.VoiceName, 1, 50),

			// language is required when a voice name is set
			validate.ASCII("Twilio.VoiceLanguage", cfg.Twilio.VoiceLanguage, 1, 10),
		)
	} else {
		err = validate.Many(
			err,
			validate.ASCII("Twilio.VoiceLanguage", cfg.Twilio.VoiceLanguage, 0, 10),
		)
	}
```

mastercactapus (changes requested):
I'm going to look into using the XML encoder to deal with the escaping issue (as a separate PR), I already started some of that with the mocktwilio refactor work, so I think it will be straightforward.

AlaricWhitney:
@mastercactapus Cool.  I'll wait until your XML PR #2582 goes in before making more updates to this PR.

mastercactapus:
@AlaricWhitney the XML PR has been merged

notification/twilio/twiml.go
AlaricWhitney: Updated to use the xml parser.

config/config.go
AlaricWhitney: not really.  removed the public tags.

config/config.go
AlaricWhitney: applied.

notification/twilio/voice.go
AlaricWhitney: Updated all of the pauses to do `Hello! This is GoAlert with an alert notification. <700ms pause> {details} {options}`

mastercactapus:
@AlaricWhitney, just a note on this, I haven't had a chance to dig in, but it doesn't seem to work with the Polly voices. The `<break>` blocks get read out literally. I think it all needs to be wrapped in `<speak>` or something like that to use the "advanced" syntax (I forget the acronym)

AlaricWhitney:
@mastercactapus This is ready for review again.  I've implemented the message to include an extra optional query parameter that indicates where in the string slice to insert pauses.  I've also added unit tests and benchmark tests to show how it works.  I also moved some of the message functionality into their own separate functions to make unit testing easier.

notification/twilio/voice.go
```diff
@@ -401,6 +357,17 @@ func (v *Voice) getCall(w http.ResponseWriter, req *http.Request) (context.Conte
 	msgID := q.Get(msgParamID)
 	subID, _ := strconv.Atoi(q.Get(msgParamSubID))
 	bodyData, _ := b64enc.DecodeString(q.Get(msgParamBody))
+
+	// add pause index values
+	var pauseIndexData []int
+	if q.Has(msgParamPause) {
+		for _, val := range q[msgParamPause] {
```
mastercactapus, line 364: Range works over `nil`, so we can skip the `Has` check

notification/twilio/voice.go
```diff
@@ -401,6 +357,17 @@ func (v *Voice) getCall(w http.ResponseWriter, req *http.Request) (context.Conte
...
+			if i, err := strconv.Atoi(val); err != nil {
```
mastercactapus, line 365: An error here should reject with a 400 bad request

notification/twilio/voice.go
```diff
@@ -638,3 +606,100 @@ func (v *Voice) FriendlyValue(ctx context.Context, value string) (string, error)
...
+		Params:         make(url.Values),
+	}
+
+	var message string
+	var pauseIndex []int
+	subID := -1
+	switch t := msg.(type) {
+	case notification.AlertBundle:
+		message = fmt.Sprintf("%s with alert notifications. Service '%s' has %d unacknowledged alerts.", prefix, t.ServiceName, t.Count)
+		pauseIndex = append(pauseIndex, len(fmt.Sprintf("%s with alert notifications.", prefix)))
```
mastercactapus, line 663: Instead of passing these indexes around everywhere, can we handle encoding/decoding at the request and use a string slice everywhere else?

notification/twilio/voice.go
```diff
@@ -471,13 +438,13 @@ func (v *Voice) ServeTest(w http.ResponseWriter, req *http.Request) {
...
 
-	resp := newTwiMLResponse(w)
+	resp := newTwiMLResponse(ctx, w)
 	switch call.Digits {
 	default:
 		resp.SayUnknownDigit()
 		fallthrough
 	case "", digitRepeat:
-		resp.Say(call.msgBody)
+		processSayBody(resp, call.msgBody, call.msgPauseIndex)
```
mastercactapus, line 447: This feels like it should just be a method

notification/twilio/voice_test.go
```diff
@@ -1,12 +1,393 @@
...
+	type mockInput struct {
+		resp          *twiMLResponse
+		msgBody       string
+		msgPauseIndex []int
+	}
+
+	testCases := map[string]struct {
+		input    mockInput
+		expected *twiMLResponse
+	}{
```
mastercactapus, line 241: resp is empty for all these, which adds noise, and specifying the indexes manually seems fragile.

mastercactapus (changes requested):
Having to manage/track indexes all over the place is going to be error-prone and noisy; we should be using a slice of strings throughout our application code to manage chunks of text.

The request itself can have the slice encoded in a `string` + `[]int` format, but we should parse that as soon as we get the request into a more usable format of `[]string`, and only switch to indexes at the last moment when we build the URL.

Doing so will let us simplify a lot of this.

notification/twilio/voice.go
AlaricWhitney: I can do it if you want, although I do think since it's voice specific, so I will make it part of the Voice object.

notification/twilio/voice.go
AlaricWhitney: I think it makes more sense to have it directly by the message for readability and comprehension.  If we move the index injection elsewhere, it won't really be obvious what's going on (ie, ensuring that the message text, and the injection are by each other so you can easily keep them in sync).

AlaricWhitney:
> Having to manage/track indexes all over the place is going to be error-prone and noisy; we should be using a slice of strings throughout our application code to manage chunks of text.
> 
> The request itself can have the slice encoded in a `string` + `[]int` format, but we should parse that as soon as we get the request into a more usable format of `[]string`, and only switch to indexes at the last moment when we build the URL.
> 
> Doing so will let us simplify a lot of this.

I chose to do it this way to ensure separation of functionality between what each object is supposed to do, and ensure backwards compatibility.  More specifically, I really think that the message injection processing should occur immediately next to where the message is defined.  Indexes are ordered, so it shouldn't be an issue to pass it around a little bit.

notification/twilio/voice.go
mastercactapus: Keeping it separate is basically a pseudo-type, like this:
```go
type MultiString struct {
	MsgBody string
	Indexes []int
}
```

While I agree to keep things surface-level and not add unnecessary abstraction, embedding it as two separate fields in an existing struct doesn't help with readability or comprehension over `[]string`, which will track the whole dataset, but without special knowledge or the necessity to keep indexes in sync.

Tracking indexes manually for the entire lifecycle also adds noise at every layer, like tests, where the reader needs to know why it's there, what it's doing, and how to keep it correct.

The separate-indexes-encoding stuff is ONLY relevant when building or parsing the query parameters, so we can keep that complexity deliberate and localized without maintaining it everywhere else.

https://google.github.io/styleguide/go/guide#simplicity
> - Does not assume that you already know what it is doing
> - Does not assume that you can memorize all of the preceding code

https://google.github.io/styleguide/go/guide#least-mechanism
> Where there are several ways to express the same idea, prefer the one that uses the most standard tools.

mastercactapus:
To help move this along, should we break #2574 into a PR on its own while the `<break>` stuff cycles? I think the voice/language stuff has been good for a while, doesn't require any pause/break changes, and is ready to be validated.

It would also help to limit the scope for review.

notification/twilio/voice.go
AlaricWhitney: @mastercactapus Updated to pass around a type rather than multiple inputs as suggested.

notification/twilio/voice.go
```diff
@@ -35,6 +35,11 @@ type Voice struct {
 	r notification.Receiver
 }
 
+type VoiceMultiString struct {
+	Body         string
+	PauseIndexes []int
+}
```
mastercactapus, line 41: multi-string already exists as `[]string` in Go

mastercactapus:
@AlaricWhitney Separating the two bodies of work should help reduce noise and get traction to move things along. Let's pull the config update changes into a separate PR specifically for #2574 and have this one be the greeting/API changes for #2576. That will also help with our release notes in the future.

Since the voice API callback parameters and rendering code need careful review, and the Twilio voice code is already a bit of a mess in master, it is more expensive to validate. Since we didn't have granular validation for this code, reviewing this PR is more complex. I like that your changes add some unit testing that has been missing, and I think we can break it up a little differently to make it more straightforward.

Let's chat/pair on this to reduce spin, at a high level, I'm thinking:
- To remove a bunch of nested conditionals and simplify, switch to using `[]string`
- To help the readability & reviewability of the tests, find a way to simplify the inputs/outputs for each test case (e.g., maybe changing `buildMessage` slightly)
- See if we can do something about `getCall` to make it testable, or at least a helper func that _is_

AlaricWhitney:
@mastercactapus Made the recommended updates to remove the pauses, and only include the voice settings.  ready for review again.

mastercactapus:
It looks like the `buildMessage`/refactor work was still left in. It might be cleaner/easier to open a PR from commit 612c1e2f3cf4c8fd0eb113c9ed4d162a4127946d, as that looks like the last one before the pause/break/refactor work started happening.

IIRC the only changes needed from that commit are to remove the XML break since we'll be handling that in the next PR(s), and it looked like some of the tests still had empty struct declarations instead of var.

It will be much easier to validate the functional voice changes against code that already has the rework & testing in place, so for that and to keep the PRs small, let's merge the voice refactor work as one PR on its own and then lastly the API changes for the `<break>` pause at the end.

AlaricWhitney:
@mastercactapus I decided to leave that element in based on our offline discussion on breaking up this large function into smaller functions to make unit testing easier.  I can leave it out of this PR if you want, but I do feel that it's an overall benefit to this application if left in, as you can now easily unit test the functional element to ensure that nothing is broken by my changes.

AlaricWhitney:
For the sake of getting this moving forward, I've removed the buildMessage enhancement for better testability from this.  I'll add it into the another PR.

mastercactapus (dismissed):

tony-tvu (commented):
Everything is working fine but I'm seeing these logs `fatal: No names found, cannot describe anything.` which I'm not seeing when running on master branch. Is anyone else seeing this on their on when running on this branch?

Screenshot 2023-01-25 at 10 45 15 AM

Forfold:
Testing this now and not running into that @tony-tvu 🤔

config/config.go
```diff
@@ -432,6 +435,8 @@ func (cfg Config) Validate() error {
 		validateKey("Twilio.AccountSID", cfg.Twilio.AccountSID),
 		validateKey("Twilio.AuthToken", cfg.Twilio.AuthToken),
 		validateKey("Twilio.AlternateAuthToken", cfg.Twilio.AlternateAuthToken),
+		validate.ASCII("Twilio.VoiceName", cfg.Twilio.VoiceName, 0, 50),
+		validate.ASCII("Twilio.VoiceLanguage", cfg.Twilio.VoiceLanguage, 0, 10),
```
Forfold, line 439: placeholder text on these could be useful. I wasn't totally sure if `english` would work, then I saw a test later on during the review using `en-US`

config/config.go
```diff
@@ -432,6 +435,8 @@ func (cfg Config) Validate() error {
 		validateKey("Twilio.AccountSID", cfg.Twilio.AccountSID),
 		validateKey("Twilio.AuthToken", cfg.Twilio.AuthToken),
 		validateKey("Twilio.AlternateAuthToken", cfg.Twilio.AlternateAuthToken),
+		validate.ASCII("Twilio.VoiceName", cfg.Twilio.VoiceName, 0, 50),
```
Forfold, line 438: Could we link to a list of names in the description? Or have the default name used filled in

Forfold (changes requested):

config/config.go
AlaricWhitney: @Forfold I didn't see any other examples of default values being used, so by doing so would deviate from the current config patterns.  If these values aren't set (like how it is currently today), it'll use the default setting within Twilio itself, so these don't need to be set at all.

config/config.go
AlaricWhitney: Having the default voice filled in would deviate from the config pattern of this app.  If it's not set (like how it is today) it will default to whatever the Twilio account has set to default.

I'll add the link to the voices in a comment here.

Forfold (approved):

Forfold (approved):

tony-tvu (approved):
I was able to functionally test this with a real phone call and change the voices. Hopefully the issue with `build | fatal: No names found, cannot describe anything` logs resolves itself when running on GoAlert's repo.

mastercactapus (approved):

PR #2582, opened 2022-08-22
twilio: use encoding/xml to generate TwiML response

mastercactapus:
**Description:**
This PR updates the `notification/twilio` package to use `encoding/xml` instead of raw strings to build the Response document.

**Additional Info:**
This is in preparation for user-definable values in #2575


Forfold (dismissed):
lgtm, functionally verified voice lines

Forfold (approved):

m17ch (approved):

PR #2588, opened 2022-08-24
twilio: add alternate auth token for use during key rotation

mastercactapus:
**Description:**
This PR adds the `AlternateAuthToken` field to the config allowing for key rotation without request failures.

To validate/perform a key rotation:
1. Verify calls and/or bi-directional SMS works
2. Generate secondary token in Twilio
3. Set the Auth Token to the new secondary, and the Alternate Auth Token to the current primary
4. Verify calls and/or bi-directional SMS works
5. Promote secondary to primary in Twilio
6. Verify calls and/or bi-directional SMS works
7. Clear Alternate Auth Token
8. Verify calls and/or bi-directional SMS works


frizzkitten (dismissed):

notification/twilio/validation.go
```diff
@@ -25,7 +25,14 @@ func validateRequest(req *http.Request) error {
```
Forfold, line 25: would it be worth adding additional validation returned if the user has `AlternateAuthToken` set and not `AuthToken`?

something like `"missing primary X-Twilio-Signature. only alternate token found."?

mastercactapus: It shouldn't be necessary; if Twilio is enabled, AuthToken must always be set; Config validation enforces it.

However, that did point out that I forgot to validate the AlternateAuthToken, just pushed a fix for that.

tony-tvu (dismissed):

tony-tvu (approved):

AlaricWhitney (approved):

Forfold (approved):

Issue #2554, opened 2022-08-10
Improve local development for external services


mastercactapus:
**What problem would you like to solve? Please describe:**
- Hard to know what local tools are available
- Onboarding currently requires creating accounts and credentials for external services
- Getting publicly-routable URLs to local developer machines isn't always feasible

**Describe the solution you'd like:**
- Re-use existing mock servers for dev, in addition to testing
- Link to existing tools (e.g., MailHog) where developers can find it

**Describe alternatives you've considered:**
Currently, developers need to have access to all services they are developing against, even for established features that have mock servers for testing.

**Additional context:**
![LocalDev](https://user-images.githubusercontent.com/595010/184021654-9491902c-86cf-4b68-8b09-aeaab99e51c0.png)

mastercactapus:
Wireframe for Twilio UI concept
![image](https://user-images.githubusercontent.com/595010/184933930-d1a67b26-bcb5-4586-8f9c-8b4a5a9a6451.png)

mastercactapus:
Will open new issues for:
- `mocktwilio` UI
- `mockslack` UI
- Admin/Config history
- Dev tool for sending things like Grafana alerts

PR #2602, opened 2022-08-29
dev: mocktwilio refactor

mastercactapus:
**Description:**
This PR makes several changes in the mocktwilio devtool to make it more usable as a standalone package.
- Assertions & interfaces are now part of the package itself (instead of in the harness)
- behavior is much more deterministic, and smoke tests run up to 4x faster in some cases
- interfaces were adjusted to represent better user options (to allow for UI in future PR)
- no longer depends on goalert packages (standalone)
- XML parsing/handling rather than relying on string scraping
- more explicit API for assertions, calls no longer call the application before being answered
- support for full TwiML flows (but only supported verbs)

Part of #2554


mastercactapus:
- validate start/stop tests (esp. with message SID)

PR #2606, opened 2022-08-31
update readme homepage screenshot

Forfold:
Adds the following screenshots to the main README.md, replacing the outdated one at the top.

homepage

alert_details

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,9 @@
 
 GoAlert provides on-call scheduling, automated escalations and notifications (like SMS or voice calls) to automatically engage the right person, the right way, and at the right time.
 
-![image](https://user-images.githubusercontent.com/11381794/78996595-889b7400-7b0a-11ea-90e9-dd10b0116b1a.png)
+![image](homepage.png)
+
+![image](alert_details.png)
 
 ## Installation
 


mastercactapus (changes requested):
Thoughts:
- I think just one image is enough for the README for the GitHub repo. The landing page/docs site is better suited to showcase more screens
- Light mode would be better suited for doc screenshots to show color. But maybe have the settings popover open with the dark mode option visible.
- We should keep the SMS example on the side (maybe update it if the template has changed). It's a good illustration of the primary use case.

Forfold:
I thought the alert details was more valuable than the SMS picture, but sounds good.

I also thought dark mode as the main screenshot would be more attractive to other devs out there looking at our product

mastercactapus:
I could see the value in using alert details instead of the alerts list for the screenshot, especially if we make sure the phone conversation matches the Event Log; that would be great!

As for dark mode, it wouldn't be bad to show that it's supported, but it's not the highlight of the application. As it is, I think it makes sense to show the default state, and also, the landing page/docs pages are all light-themed (white bg), so dark-mode screenshots would feel out of place. We also have stronger color theming in light mode, so it feels less greyscale to someone seeing it for the first time.

mastercactapus (approved):

KatieMSB (approved):

PR #2607, opened 2022-09-01
twilio: update SMS restrictions

mastercactapus:
**Description:**
This PR updates the SMS logic to account for additional SMS restrictions.
- URLs not allowed to some prefixes
- Added to list of prefixes that do not support 2-way SMS

### DIFF
--- a/notification/twilio/alertsms.go
+++ b/notification/twilio/alertsms.go
@@ -89,12 +89,16 @@ func mapGSM(r rune) rune {
 	return '?'
 }
 
-func oneOfPrefixes(s string, prefixes ...string) bool {
+// hasAnyPrefix returns true if any of the prefixes are present in the string.
+func hasAnyPrefix(s string, prefixes ...string) bool {
 	for _, p := range prefixes {
-		if strings.HasPrefix(s, p) {
-			return true
+		if !strings.HasPrefix(s, p) {
+			continue
 		}
+
+		return true
 	}
+
 	return false
 }
 
@@ -104,9 +108,10 @@ func canContainURL(ctx context.Context, number string) bool {
 		return false
 	}
 
-	// not extensive, but CN code numbers do not allow URLs in SMS
-	// https://www.twilio.com/guidelines/cn/sms
-	return !oneOfPrefixes(number, "+86")
+	return !hasAnyPrefix(number,
+		// Non-exhaustive list of dialing codes that forbid URLs.
+		"+86", // CN - https://www.twilio.com/guidelines/cn/sms
+	)
 }
 
 // hasTwoWaySMSSupport returns true if a number supports 2-way SMS messaging (replies).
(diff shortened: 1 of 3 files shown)


notification/twilio/alertsms.go
```diff
@@ -89,14 +89,34 @@ func mapGSM(r rune) rune {
 	return '?'
 }
 
+func oneOfPrefixes(s string, prefixes ...string) bool {
```
frizzkitten, line 92: Dunno how y'all feel about unit testing un-exported functions, but you could add a quick unit test for this

AlaricWhitney (approved):

notification/twilio/alertsms.go
```diff
@@ -89,14 +89,34 @@ func mapGSM(r rune) rune {
...
 // hasTwoWaySMSSupport returns true if a number supports 2-way SMS messaging (replies).
 func hasTwoWaySMSSupport(ctx context.Context, number string) bool {
 	if config.FromContext(ctx).Twilio.DisableTwoWaySMS {
 		return false
 	}
 
-	// India numbers do not support SMS replies.
-	return !strings.HasPrefix(number, "+91")
+	// not extensive
+	return !oneOfPrefixes(number, "+91", "+86", "+502", "+506", "+507", "+84")
```
frizzkitten, line 119: could add some constants for these so that it's clearer what countries don't allow two way support, so you could do 
```
return !oneOfPrefixes(number, codes.China, codes.India, ...)
```

frizzkitten (dismissed):

notification/twilio/alertsms.go
mastercactapus: https://github.com/target/goalert/blob/master/CONTRIBUTING.md#testing

As a general rule, we add a unit test if:
- The function is too complex to validate by visually looking at the code
- We need to benchmark for performance and/or memory concerns in a hot path
- We're writing an Example test for a constructor or exported API method
- The functionality is used from several call-sites (esp. if exported)
- There are edge cases that a provable table test would aid in review for
- Proving behavior for/during refactoring.


Something like this wouldn't have much value to be unit tested, other than increasing the number of unit tests, which would just add noise. This is a throwaway function to be used a few lines from its definition. It could just as easily live inside the two functions using it.

I did add a smoke test to functionally validate the system behavior, which would survive a refactor -- but in said refactor, these functions would all likely be deleted to use some other approach (like an API call).

That being said, a better name & comment would go a long way towards that visual validation, so I updated it.

notification/twilio/alertsms.go
mastercactapus: Similar to above, constants would add noise and indirection here that would take away from readability and add mental overhead; none of these dialing codes really have any value outside of the func here.

We don't have a _current_ use case for mapping countries to dialing codes, (or region names to dialing codes for that matter), though we could feasibly map them through `libphonenumber` but again, probably not worth the cost of constantly mapping to-from region codes vs. just listing the prefix which can be directly compared to the contact method value.

I do think you're right about adding clarity around which regions these represent, so I added some better comments for quick reference to the next/future readers.

notification/twilio/alertsms.go
frizzkitten: Sweet, thanks for the info!

frizzkitten (approved):

AlaricWhitney (approved):

m17ch (approved):

PR #2640, opened 2022-09-29
GraphQL: serviceID should be optional

theautoroboto:
**Description:**
When creating a new service with a heartbeat monitor, CreateHeartbeatMonitorInput is requiring a serviceID that has not yet been created.

This change will make serviceID optional in the schema for CreateHeartbeatMonitorInput, allowing createService to create the heartbeat monitor when the service is created.

### DIFF
--- a/graphql2/generated.go
+++ b/graphql2/generated.go
@@ -23801,7 +23801,7 @@ func (ec *executionContext) unmarshalInputCreateHeartbeatMonitorInput(ctx contex
 			var err error
 
 			ctx := graphql.WithPathContext(ctx, graphql.NewPathWithField("serviceID"))
-			it.ServiceID, err = ec.unmarshalNID2string(ctx, v)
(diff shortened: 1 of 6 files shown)


mastercactapus (changes requested):
A couple more changes will need to go with this; if you run `make check` locally, it should regenerate the necessary files and give you the error for what's missing.

Namely:
- in `graphql2/graphqlapp/service.go`, we just need to take the address of `result.ID`
- in `graphql2/graphqlapp/heartbeatmonitor.go` we need to account for it now being a pointer -- you can use the `CreateIntegrationKey` method in `graphql2/graphqlapp/integrationkey.go` for an example.

The underlying store methods will handle validation (e.g., checking for empty ServiceID), so it's enough to just handle the change from `string` to `*string` in the GraphQL code.

mastercactapus (approved):

tony-tvu (approved):

PR #2644, opened 2022-10-03
admin: add pagination to message logs query

### LINKED ISSUE #2600: message logs: access denied, auth limit
**Describe the Bug:**
Trying to view message logs on a busy system results in an error page.

**Steps to Reproduce:**
1. Go to Admin -> Message Logs on a system with ~1000 recent messages
2. See error page

**Expected Behavior:**
A list of recent messages.

**Observed Behavior:**
An error page.

**Screenshots/Stack Traces:**

image

**Application Version:**

```
$ goalert version
Version:   v0.28.0-3077-g823f65a69
GitCommit: 823f65a69fcbe536d4490980e11e3949ba3e3b33 (clean)
BuildDate: 2022-08-29T01:59:58-05:00
GoVersion: go1.19 (gc)
Platform:  linux/amd64
Migration: add-service-maintenance-expires-at (#225)

```

**Additional Context:**
The "auth limit" is a fail-safe check every time a permissions check is performed before issuing a DB query. Likely what is happening is the data being fetched all at once, isn't being batched into a single or low number of requests, and instead is resulting in many DB calls hitting the limit.

It may be worth changing the `debugMessages` query to be paginated rather than a fetch-all (it was only meant to be a temporary implementation detail anyway). If we build a query that matches behavior of things like the `alerts` query, we can also re-use frontend code easier when it comes to graphs.

Forfold:
Fixes #2600

- Adds a paginated query `MessageLogs` to the schema for use. Searching by direct value is supported.
- `DebugMessages` has been deprecated in favor of the `MessageLog` terminology.
- Pagination follows patterns of other pages (e.g. Rotations) for consistency.
- The list of logs as cards has been updated to use `PaginatedList` instead of using its own design pattern.

There's some work that could be cleaned up around fetching the slack destination. It's kind of sloppy right now while I think of a better solution- open to suggestions.

**Screenshots:**

Screen Shot 2022-10-04 at 2 43 44 PM

Screen Shot 2022-10-04 at 2 38 54 PM

**Out of Scope:** 
- Extending search options


graphql2/schema.graphql
```diff
@@ -2,7 +2,7 @@ type Query {
   phoneNumberInfo(number: String!): PhoneNumberInfo
 
   # Returns the list of recent messages.
-  debugMessages(input: DebugMessagesInput): [DebugMessage!]!
```
mastercactapus, line 5: This (and input/output) will need to stay identical for API compatibility reasons.

Add `messageLogs` as a net-new query.

For the GraphQL implementation, maybe just have the `debugMessages` query return the first page or something like that and mark it as deprecated, so we don't have the duplicate code.

Forfold: I thought about this once I was opening the draft- good call

notification/search.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/target/goalert/util/sqlutil"
+	"github.com/target/goalert/validation/validate"
+)
+
+// SearchOptions allow filtering and paginating the list of rotations.
+type SearchOptions struct {
+	Search string       `json:"s,omitempty"`
+	After  SearchCursor `json:"a,omitempty"`
+
+	// Omit specifies a list of rotation IDs to exclude from the results
```
Forfold, line 21:
```suggestion
// SearchOptions allow filtering and paginating the list of messages.
type SearchOptions struct {
	Search string       `json:"s,omitempty"`
	After  SearchCursor `json:"a,omitempty"`

	// Omit specifies a list of message IDs to exclude from the results
```

mastercactapus (approved):

mastercactapus (dismissed):

mastercactapus (approved):

notification/search.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	return &opts, err
+}
+
+func (opts renderData) QueryArgs() []sql.NamedArg {
+	return []sql.NamedArg{
+		sql.Named("search", opts.Search),
+		sql.Named("cursorCreatedAt", opts.After.CreatedAt),
+		sql.Named("createdAfter", opts.CreatedAfter),
```
KatieMSB, line 129:
```suggestion
		sql.Named("createdAfter", opts.CreatedAfter),
		sql.Named("afterID", opts.After.ID),
```
Missing `afterID` in sql above

KatieMSB (changes requested):

KatieMSB (approved):

mastercactapus (approved):

PR #2648, opened 2022-10-04
feat: Added help and cleaned up makefile

### LINKED ISSUE #2645: feat: Makefile doesn't include a help option
Makefile doesn't include a help option, which makes it difficult to know what you are supposed to do without looking through the source code.

Add `make help` that will output what the available options are.

Manually looking through the source code...

AlaricWhitney:

**Description:**
Added help to the makefile.
- Also cleaned up unused makefile options

Fixes #2645 

**Out of Scope:**
NA

**Screenshots:**
NA

Added `make help`
- Removed `make lint` due to deprecation of golint (https://pkg.go.dev/golang.org/x/lint#section-readme)
- Added `go fmt ./...` to `make check` for basic lint checking

NA

**Additional Info:**
NA


Makefile
```diff
@@ -125,23 +128,28 @@ start-integration: web/src/build/static/app.js bin/goalert bin/psql-lite bin/wai
...
 
-check-all: check test smoketest cy-wide-prod-run cy-mobile-prod-run
+check-go:
+	go vet ./...
+	go fmt ./...
+	# go run ./devtools/ordermigrations -check
+	go run github.com/gordonklaus/ineffassign ./...
+	CGO_ENABLED=0 go run honnef.co/go/tools/cmd/staticcheck ./...
+
+check-all: check lint test smoketest cy-wide-prod-run cy-mobile-prod-run ## Run all lint checks and integration tests
```
KatieMSB, line 152:
```suggestion
check-all: check test smoketest cy-wide-prod-run cy-mobile-prod-run ## Run all lint checks and integration tests
```
Screen Shot 2022-10-04 at 10 15 32 AM

Looks like a `lint` snuck in here

KatieMSB (changes requested):

Makefile
```diff
@@ -41,9 +41,9 @@ ifeq ($(PUSH), 1)
 PUSH_FLAG=--push
 endif
 
-all: test install
+all: test
 
-release: container-demo container-goalert bin/goalert-linux-amd64.tgz bin/goalert-linux-arm.tgz bin/goalert-linux-arm64.tgz bin/goalert-darwin-amd64.tgz bin/goalert-windows-amd64.zip
+release: container-demo container-goalert bin/goalert-linux-amd64.tgz bin/goalert-linux-arm.tgz bin/goalert-linux-arm64.tgz bin/goalert-darwin-amd64.tgz bin/goalert-windows-amd64.zip ## Launch the demo of this application
```
mastercactapus, line 46:
```suggestion
release: container-demo container-goalert bin/goalert-linux-amd64.tgz bin/goalert-linux-arm.tgz bin/goalert-linux-arm64.tgz bin/goalert-darwin-amd64.tgz bin/goalert-windows-amd64.zip ## Build all release artifacts
```

Makefile
```diff
@@ -125,23 +128,28 @@ start-integration: web/src/build/static/app.js bin/goalert bin/psql-lite bin/wai
...
+check-js: force-yarn generate node_modules
 	yarn run fmt
 	yarn run lint
 	yarn workspaces run check
 	./devtools/ci/tasks/scripts/codecheck.sh
 
-check-all: check test smoketest cy-wide-prod-run cy-mobile-prod-run
+check-go:
+	go vet ./...
+	go fmt ./...
```
mastercactapus, line 147: This won't cause a failure, only list unformatted files. It's already handled in `codecheck.sh` (which isn't go specific)
```suggestion
```

Makefile
```diff
@@ -125,23 +128,28 @@ start-integration: web/src/build/static/app.js bin/goalert bin/psql-lite bin/wai
...
-	go vet ./...
-	go run github.com/gordonklaus/ineffassign ./...
-	CGO_ENABLED=0 go run honnef.co/go/tools/cmd/staticcheck ./...
+check: check-go check-js  ## Run all lint checks
+
+check-js: force-yarn generate node_modules
 	yarn run fmt
 	yarn run lint
 	yarn workspaces run check
 	./devtools/ci/tasks/scripts/codecheck.sh
```
mastercactapus, line 143: This isn't a js check, should probably go under the `check` target as it needs to run after both the js and go fmt commands

Makefile
```diff
@@ -225,20 +232,18 @@ postgres: bin/waitfor
 		-p 5432:5432 \
 		docker.io/library/postgres:13-alpine && ./bin/waitfor "$(DB_URL)" && make regendb) || ($(CONTAINER_TOOL) start goalert-postgres && ./bin/waitfor "$(DB_URL)")
 
-regendb: bin/resetdb bin/goalert config.json.bak
+regendb: bin/resetdb bin/goalert config.json.bak ## Regenerate the database
```
mastercactapus, line 235:
```suggestion
regendb: bin/resetdb bin/goalert config.json.bak ## Reset the database and fill it with random data
```

Makefile
```diff
@@ -225,20 +232,18 @@ postgres: bin/waitfor
...
 	test -f config.json.bak && bin/goalert set-config --allow-empty-data-encryption-key "--db-url=$(DB_URL)" <config.json.bak || true
 	bin/goalert add-user --user-id=00000000-0000-0000-0000-000000000001 --user admin --pass admin123 "--db-url=$(DB_URL)"
 
-resetdb: config.json.bak
+resetdb: config.json.bak ## Reset the database
 	go run ./devtools/resetdb --no-migrate
 
-clean:
+clean: ## Clean up the application
+	@go mod tidy
```
mastercactapus, line 244: This shouldn't introduce any changes, only clean up build artifacts (e.g., similar to a fresh `git clone`). This should be done in `check` instead

```suggestion
```

Makefile
```diff
@@ -225,20 +232,18 @@ postgres: bin/waitfor
...
 	./bin/resetdb -with-rand-data -admin-id=00000000-0000-0000-0000-000000000001 -mult $(SIZE)
 	test -f config.json.bak && bin/goalert set-config --allow-empty-data-encryption-key "--db-url=$(DB_URL)" <config.json.bak || true
 	bin/goalert add-user --user-id=00000000-0000-0000-0000-000000000001 --user admin --pass admin123 "--db-url=$(DB_URL)"
 
-resetdb: config.json.bak
+resetdb: config.json.bak ## Reset the database
 	go run ./devtools/resetdb --no-migrate
 
-clean:
+clean: ## Clean up the application
```
mastercactapus, line 243:
```suggestion
clean: ## Clean up build artifacts
```

Makefile
```diff
@@ -125,23 +128,28 @@ start-integration: web/src/build/static/app.js bin/goalert bin/psql-lite bin/wai
...
 
-check-all: check test smoketest cy-wide-prod-run cy-mobile-prod-run
+check-go:
+	go vet ./...
+	go fmt ./...
+	# go run ./devtools/ordermigrations -check
+	go run github.com/gordonklaus/ineffassign ./...
+	CGO_ENABLED=0 go run honnef.co/go/tools/cmd/staticcheck ./...
+
+check-all: check test smoketest cy-wide-prod-run cy-mobile-prod-run ## Run all lint checks and integration tests
```
mastercactapus, line 152: Is it necessary to add a task that combines checks and tests? (if so we should come up with a better name) We already have all tests under a target `test-all` to put them into a single command:
```bash
make check test-all
```

Makefile
```diff
@@ -225,20 +232,18 @@ postgres: bin/waitfor
...
 		docker.io/library/postgres:13-alpine && ./bin/waitfor "$(DB_URL)" && make regendb) || ($(CONTAINER_TOOL) start goalert-postgres && ./bin/waitfor "$(DB_URL)")
 
-regendb: bin/resetdb bin/goalert config.json.bak
+regendb: bin/resetdb bin/goalert config.json.bak ## Regenerate the database
 	./bin/resetdb -with-rand-data -admin-id=00000000-0000-0000-0000-000000000001 -mult $(SIZE)
 	test -f config.json.bak && bin/goalert set-config --allow-empty-data-encryption-key "--db-url=$(DB_URL)" <config.json.bak || true
 	bin/goalert add-user --user-id=00000000-0000-0000-0000-000000000001 --user admin --pass admin123 "--db-url=$(DB_URL)"
 
-resetdb: config.json.bak
+resetdb: config.json.bak ## Reset the database
```
mastercactapus, line 240:
```suggestion
resetdb: config.json.bak ## Recreate the database leaving it empty (no migrations)
```

mastercactapus (changes requested):

Makefile
AlaricWhitney: Actually, it not only will list unformatted files, but it will also auto-correct them.  It'll fail if it can't auto correct it.

Makefile
AlaricWhitney: well, I actually didn't add this.  I only added a description as it seemed like something that you might want to publicly expose with make help.  I can remove this artifact if you want.

Makefile
mastercactapus: We already run `gofmt` as part of `codecheck.sh` and fail if there are non-formatted (_or non-simplified_) files, so it's redundant. Auto-correcting isn't good enough for CI where we want it to fail if the committed code isn't formatted.

That said, we can move that check here so it's not redundant, I just don't want there to be two different places doing the same thing in two slightly different ways, we should pick one.

Makefile
mastercactapus: Just put it in the `check` target; another target would just be noise for the sake of noise IMO. `codecheck.sh` is dependent on the other ones being run first, and isn't valid on it's own

Makefile
AlaricWhitney: If you run `go fmt ./...` on the master branch there's formatting errors, yet `make check` passes.  Based on the current behavior I would still say that `go fmt ./...` should be in there, as nothing else currently catches the formatting issues.

Makefile
AlaricWhitney: removed

Makefile
mastercactapus: To clarify: `codecheck.sh` MUST run last, after all linting/formatting/generating, and its task has to depend on both `check-js` and `check-go` running first.

Also, `check-go` also needs to depend on `generate` as that generates code for all languages

Makefile
AlaricWhitney: removed gofmt as it doesn't actually do anything at this point.  Using `go fmt ./...` is more useful based on current behavior.

Makefile
```diff
@@ -125,23 +128,27 @@ start-integration: web/src/build/static/app.js bin/goalert bin/psql-lite bin/wai
...
+	./devtools/ci/tasks/scripts/codecheck.sh
+
+check-js: force-yarn generate node_modules
 	yarn run fmt
 	yarn run lint
 	yarn workspaces run check
-	./devtools/ci/tasks/scripts/codecheck.sh
 
-check-all: check test smoketest cy-wide-prod-run cy-mobile-prod-run
+check-go:
```
mastercactapus, line 145: Need to ensure code generation is run _before_ validating
```suggestion
check-go: generate
```

Makefile
```diff
@@ -125,23 +128,27 @@ start-integration: web/src/build/static/app.js bin/goalert bin/psql-lite bin/wai
...
 
 force-yarn:
 	yarn install --no-progress --silent --frozen-lockfile --check-files
 
-check: force-yarn generate node_modules
-	# go run ./devtools/ordermigrations -check
-	go vet ./...
-	go run github.com/gordonklaus/ineffassign ./...
-	CGO_ENABLED=0 go run honnef.co/go/tools/cmd/staticcheck ./...
+check: check-go check-js check-code  ## Run all lint checks
```
mastercactapus, line 137:
```suggestion
check: check-go check-js  ## Run all lint checks
```

Makefile
```diff
@@ -125,23 +128,27 @@ start-integration: web/src/build/static/app.js bin/goalert bin/psql-lite bin/wai
...
 
-check: force-yarn generate node_modules
-	# go run ./devtools/ordermigrations -check
-	go vet ./...
-	go run github.com/gordonklaus/ineffassign ./...
-	CGO_ENABLED=0 go run honnef.co/go/tools/cmd/staticcheck ./...
+check: check-go check-js ## Run all lint checks
+	./devtools/ci/tasks/scripts/codecheck.sh
+
+check-js: force-yarn node_modules
```
mastercactapus, line 140: With `check-js` and `check-go` as separate tasks, `generate` must be a dependency on both since it generates both Go and js/ts. `make` will handle only running the task once but both tasks need to list it as a dependency.

Technically it should probably be a dependency of `check` _also_ but as long as it's on the other two, then `check`, `check-go`, and `check-js` will all behave correctly.

```suggestion
check-js: force-yarn node_modules generate
```

AlaricWhitney: Wouldn't that execute generate multiple times on `make check`?  what about adding a check if generate has already been run to prevent multiple runs?

AlaricWhitney: Actually, in my testing it didn't really add that much time.  Added `generate` to `check-js`

mastercactapus (approved):

Makefile
mastercactapus: `make` will only run the dependant task once per invocation, so it will just serve to ensure it always gets run, and in the correct order. Likewise, if you run targets in parallel with `-j` it would wait for generate complete before running the go, and js checks

KatieMSB (approved):

PR #2654, opened 2022-10-10
Added UI Configurable rate limits for each notification type

### LINKED ISSUE #2653: Configurable Throttle Limits for each Notification Type
Currently, we have rate limiting configuration set for each notification type per contact method. This is directly set in the code and not dynamic. If there's a need to change in the rate limits it needs redeployment of code every time.

Solution:
We can make the rate limits configurable in UI under admin panel. We have config limits set for unacknowledged alerts, contact methods per user etc..Similarly we can have it for throttling which can be modified by the admin according to their needs. We can have entries added in config_limits table with the maximum limit as value.

Example Screenshot:
Screenshot 2022-10-10 at 9 50 01 PM

Lakshmi2107:
**Description:**
Currently, we have rate limiting configuration set for each notification type per contact method. This is directly set in the code and not dynamic. If there's a need to change in the rate limits it needs redeployment of code everytime

Solution:
We can make the rate limits configurable in UI under admin panel. We have config limits set for unack alerts etc..Similarly we can have it for throttling which can be modified by the admin according to their needs. We can have entries added in config_limits table with the maximum limit as value.

Fixes #2653

**Screenshots:**
Screenshot 2022-10-10 at 9 50 01 PM

**Additional Info:**
Need to add few rows in config_limits table and add values to config limit enum
alter type enum_limit_type add value 'max_all_for_alert_status_per_20_minutes';
alter type enum_limit_type add value 'max_sms_per_15_minutes';
alter type enum_limit_type add value 'max_sms_per_hour';
alter type enum_limit_type add value 'max_sms_per_3_hours';
alter type enum_limit_type add value 'max_voice_per_15_minutes';
alter type enum_limit_type add value 'max_voice_per_hour';
alter type enum_limit_type add value 'max_voice_per_3_hours';
alter type enum_limit_type add value 'max_all_for_alert_status_per_3_minutes';
alter type enum_limit_type add value 'max_all_for_alert_status_per_2_hours';

Rows to be inserted in table config_limits

"max_sms_per_15_minutes"	5
"max_sms_per_hour"	11
"max_sms_per_3_hours"	20
"max_voice_per_15_minutes"	3
"max_voice_per_hour"	7
"max_voice_per_3_hours"	15
"max_all_for_alert_status_per_3_minutes"	1
"max_all_for_alert_status_per_2_hours"	8
"max_all_for_alert_status_per_20_minutes"	3

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -187,8 +187,12 @@ func (p *Engine) processMessages(ctx context.Context) {
 	defer recoverPanic(ctx, "MessageManager")
 	ctx, cancel := context.WithTimeout(ctx, time.Minute)
 	defer cancel()
+	err := p.msg.Init(ctx)
+	if err != nil {
+		fmt.Println(err)
+	}
 
-	err := p.msg.SendMessages(ctx, p.sendMessage, p.cfg.NotificationManager.MessageStatus)
+	err = p.msg.SendMessages(ctx, p.sendMessage, p.cfg.NotificationManager.MessageStatus)
 	if errors.Is(err, processinglock.ErrNoLock) {
 		return
 	}
--- a/engine/message/db.go
+++ b/engine/message/db.go
@@ -343,7 +343,7 @@ func NewDB(ctx context.Context, db *sql.DB, a *alertlog.Store, pausable lifecycl
 		`),
 
 		deleteAny:    p.P(`delete from outgoing_messages where id = any($1)`),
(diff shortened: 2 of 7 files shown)


mastercactapus (changes requested):
This will need a little more discussion in #2653, but on a high level, the configurable values will likely be put in the Config struct instead of in the limits table.

The system limits are only implemented the way they are because the checks are performed inside DB triggers.

PR #2702, opened 2022-11-28
fix items being filtered from options list

### LINKED ISSUE #2701: ui: userselect search display on no results
when there are no results when typing in a UserSelect formfield, the placeholder text *_sometimes_* displays "Start typing..." instead of "No options"

screenshots:

empty state

Screen Shot 2022-11-28 at 10 46 21 AM

with results state

Screen Shot 2022-11-28 at 10 46 31 AM

no results state

Screen Shot 2022-11-28 at 10 46 54 AM

expected behavior:
display "No options" after searching and array is still empty

Forfold:
Closes #2701

`AutoComplete` filters out options that don't match the value, so adding an override that passes all options through the menu resolves this:

before:

Screen Shot 2022-11-28 at 10 46 54 AM

after:

Screen Shot 2022-11-28 at 12 05 18 PM


mastercactapus (dismissed):

tony-tvu (dismissed):

web/src/app/selection/MaterialSelect.tsx
```diff
@@ -164,11 +164,12 @@ export default function MaterialSelect(
       data-cy-ready={!isLoading}
       classes={customCSS}
       multiple={multiple}
-      filterSelectedOptions={multiple}
+      filterSelectedOptions
```
mastercactapus, line 167: This was necessary so that searching for a value in single-select mode won't omit the current one -- since there is no indicator that it's currently selected otherwise (in multiple, the chip is displayed).

mastercactapus (changes requested):

web/src/app/selection/MaterialSelect.tsx
Forfold: @mastercactapus I noticed that even though this was on for single items, it wasn't working as intended- the selected item would show with the default load, but if I typed in a search term, my selected item wouldn't be visible anymore. I made this change after I noticed my Cypress tests fail- so hopefully with this change it resolves that:

default when clicking on select (with GoAlert selected)

Screen Shot 2022-12-05 at 9 19 20 AM


items after searching (with  GoAlert still selected)

Screen Shot 2022-12-05 at 9 19 35 AM

mastercactapus (dismissed):

web/src/app/selection/MaterialSelect.tsx
Forfold: Talked offline- going to keep the filter as it was before and update the other tests that failed to reflect the new functionality.

mastercactapus (approved):

KatieMSB (approved):

PR #2707, opened 2022-11-29
Fix problem removing user from rotation

### LINKED ISSUE #2543: Rotations: Problem removing user from rotation
**Describe the Bug:**
Rotations with at least 2 users, cannot remove last user if active.

**Steps to Reproduce:**
1. Create a rotation with at least 2 users
2. Set the 2nd person active
3. Remove the 2nd person from rotation

**Expected Behavior:**
First person in list remains and becomes active in rotation.

**Observed Behavior:**
The confirmation dialog displays "Confirm" and then "Retry" but never goes away.  The user is not removed from the rotation.

**Application Version:**
```
<meta http-equiv="x-goalert-version" content="v0.28.0-2861-g31cdeb056" />
<meta http-equiv="x-goalert-build-date" content="2022-07-28T05:21:17Z" />
<meta http-equiv="x-goalert-git-commit" content="31cdeb056175f2e67495e9c061551ec74bc82de3" />
```

ram191:
**Description**
This fixes the operation when mutating rotation participants. The bug initially occurs because the code is trying to update the active user index on an out of bound index since the user is removed from the rotation.

I updated the `updateRotationParticipants()` by adding the active user index to the parameter to determine if active user needs to be reset to index 0.

I also added a boolean value in the return of the function. The boolean value is used to decide whether or not an update in active user has been done in the function or not. This is used to prevent the origin of the bug where it is trying to update an out of bound index to occur.

**Which issue(s) this PR fixes**
Fixes #2543

### DIFF
--- a/graphql2/graphqlapp/rotation.go
+++ b/graphql2/graphqlapp/rotation.go
@@ -233,11 +233,11 @@ func (q *Query) Rotations(ctx context.Context, opts *graphql2.RotationSearchOpti
 	return conn, err
 }
 
-func (m *Mutation) updateRotationParticipants(ctx context.Context, tx *sql.Tx, rotationID string, userIDs []string, updateActive bool) (err error) {
+func (m *Mutation) updateRotationParticipants(ctx context.Context, tx *sql.Tx, rotationID string, userIDs []string, updateActive bool, activeUserIndex *int) (isActiveUpdated bool, err error) {
 	// Get current participants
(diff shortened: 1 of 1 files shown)


ram191:
Hi, thanks for the feedback on the merge request! I didn't pay attention to the invalid data sent from the UI there. I'll take a look at it and make an update to the PR

mastercactapus:
@ram191 is there still interest in fixing this?

ram191:
Hi! So sorry I had a hectic times this past weeks and completely forgot about this. I am still interested and will work on this asap.

Issue #2693, opened 2022-11-21
ui: make primary create buttons more visible


Forfold:
We often get user requests saying they don't know how to create or add something. Material's design focuses on a mobile-first design, but I think in this case the Fab button that we're currently using is too far out of the way for most desktop users.

mastercactapus:
I agree; maybe an ADD button in the Contact Methods and Notification Rules header as a first step? It seems to come up a lot

Forfold:
Agreed, I was thinking of maybe moving into the cards for details pages actions. Not sure on the list pages- we should get together and brainstorm

Forfold:
acceptance criteria:

- move profile fabs to contact method and notification method cards in top right
- include icon with relevant text (e.g. "+ ADD" or "+ CREATE")
- move list page fab buttons to top right of list
  - move search bar to the left

Forfold:
Sections remaining:

- [ ] rotation add user
- [ ] escalation policy add step
- [ ] Match contact/notification buttons styling to rest of app (incl. text, style button as contained w/ primary color)

PR #2709, opened 2022-11-29
ui: move profile add buttons

Forfold:
Part of #2693

Makes the create buttons for the profile page more visible when the browser width is large. On mobile, the two add buttons will remain as a FAB button in the bottom right.

Also updates the secondary action IconButton to be `color='secondary'` for `OtherActions`, and cleans up passing a prop through a couple components that can be pulled from an existing hook call.

Dark mode example

Screen Shot 2022-11-29 at 12 19 17 PM

Light mode example w/ notification rule add disabled when no contact methods are present

Screen Shot 2022-11-29 at 1 33 39 PM

### DIFF
--- a/web/src/cypress/integration/profile.ts
+++ b/web/src/cypress/integration/profile.ts
@@ -4,6 +4,7 @@ import profile from '../fixtures/profile.json'
 const c = new Chance()
 
 function countryCodeCheck(
+  screen: ScreenFormat,
   country: string,
   countryCode: string,
   value: string,
@@ -13,7 +14,12 @@ function countryCodeCheck(
     const name = 'CM SM ' + c.word({ length: 8 })
     const type = c.pickone(['SMS', 'VOICE'])
 
-    cy.pageFab('Contact')
+    if (screen === 'mobile') {
+      cy.pageFab('Add Contact Method')
+    } else {
+      cy.get('button[aria-label="Add contact method"]').click()
+    }
+
     cy.get('input[name=name]').type(name)
     cy.get('input[name=type]').selectByLabel(type)
     cy.get('input[name=value]').type(countryCode + value)
@@ -22,7 +28,7 @@ function countryCodeCheck(
   })
 }
 
-function testProfile(): void {
+function testProfile(screen: ScreenFormat): void {
   let cm: ContactMethod
 
   beforeEach(() =>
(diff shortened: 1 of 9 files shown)


web/src/cypress/integration/profile.ts
```diff
@@ -13,7 +14,12 @@ function countryCodeCheck(
     const name = 'CM SM ' + c.word({ length: 8 })
     const type = c.pickone(['SMS', 'VOICE'])
 
-    cy.pageFab('Contact')
+    if (screen === 'mobile') {
+      cy.pageFab('Add Contact Method')
```
Forfold, line 18: I considered having a second param to pageFab called "fallbackSelector" if a page fab isn't present. It just prioritizes mobile at that point. Thoughts?

mastercactapus: I think in this case doing the switch without the indirection is clearer.

The UI itself splits, and the test code reflects that. On wide, there's no FAB so calling `pageFab` would probably be confusing without context.

Forfold: Yeah that's pretty much where I landed. Sounds good 👍

web/src/cypress/integration/profile.ts
```diff
@@ -54,13 +60,15 @@ function testProfile(): void {
...
-          })
-          .engineTrigger()
-          .then(() => svc.id)
+        return (
+          cy
+            .createEPStep({
+              epID: svc.epID,
+              targets: [{ type: 'user', id: profile.id }],
+            })
+            // .engineTrigger()
```
tony-tvu, line 69: Should we remove this comment?

tony-tvu (dismissed):
LGTM!

web/src/cypress/integration/profile.ts
Forfold: Oh, this was causing tests to fail for me locally, but I don't think it was for Caza. I'll uncomment out and see what happens on CI. It also looks like there's a playwright test I need to update as well

mastercactapus: This is needed, so the test doesn't hang waiting for the engine to run in the background

tony-tvu (approved):

mastercactapus (approved):

PR #2713, opened 2022-11-30
db: skip CREATE EXTENSION if `gen_random_uuid` is available

### LINKED ISSUE #2710: db: drop pgcrypto requirement for postgres 13+
We currently have `CREATE EXTENSION pgcrypto` in a couple of migrations that require special permissions, which can be a pain point for deployment. This is only required for the `gen_random_uuid` function, which is standard in Postgres 13+

If Postgres 13.0 or higher is detected, filter out the `CREATE EXTENSION pgcrypto` lines in the initial migration(s).

None, this is just a pain point for new installs.

mastercactapus:
**Description:**
This PR removes the requirement for the `pgcrypto` extension on Postgres versions that have the `gen_random_uuid` function available.

Closes #2710 

**Additional Info:**
For validation:
- Ensure versions before 13 still work
- Ensure GoAlert starts on Postgres 13+, with `CREATE` permissions off, and the extension unavailable

Switching Postgres version in local dev can be done by setting the `PG` variable and running the `reset-postgres` target.

For example, to switch to Postgres 9.6:
```bash
make reset-postgres PG=9.6
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,8 @@
 .PHONY: smoketest generate check all test check-js check-go
 .PHONY: cy-wide cy-mobile cy-wide-prod cy-mobile-prod cypress postgres
 .PHONY: config.json.bak jest new-migration cy-wide-prod-run cy-mobile-prod-run
-.PHONY: goalert-container demo-container release force-yarn
+.PHONY: goalert-container demo-container release force-yarn delete-postgres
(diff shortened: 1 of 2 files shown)


mastercactapus:
Closing as this has gotten out of date, we also are using a digest function provided by the extension for alerts that will need to be considered as well.

PR #2715, opened 2022-11-30
ui: move create button on main list pages

Forfold:
part of #2693

- when viewing GoAlert with a widescreen browser width, display create buttons in top right instead of a FAB in the bottom right corner (mobile remains the same)
  - moves search bar to the left
  - swaps filter and ack/close all button placement on service alerts list
  - adds padding in-between grid items in pagination controls
- removes `SimpleListPage` ..for simplicity

screenshots:

Screen Shot 2022-11-30 at 2 15 18 PM

Screen Shot 2022-11-30 at 2 14 48 PM

Screen Shot 2022-11-30 at 12 17 55 PM


Forfold:
user list for admins is not up to date

Forfold:
alert checkbox dropdown carrot remove padding

mastercactapus:
@Forfold lots of conflicts, need a hand?

Forfold:
The test files got renamed to `.cy.ts` in the latest Cypress version, so it shouldn't be anything other than that! I'll take care of it

web/src/app/alerts/CreateAlertDialog/CreateAlertDialog.tsx
```diff
@@ -40,14 +41,15 @@ const useStyles = makeStyles((theme: Theme) => ({
 
 export default function CreateAlertDialog(props: {
   onClose: () => void
-  serviceID: string
 }): JSX.Element {
   const classes = useStyles()
   const [step, setStep] = useState(0)
+  const serviceID = useLocation()[0].split('/')[2] // if exists in URL
```
mastercactapus, line 47: This seems a little hacky and fragile, it really should inherit the `serviceID` from the AlertsList instead of making assumptions about the URL.

What about adding some sort of helper function to return a HOC for the one-off case?

```ts
CreateDialog={withServiceID(CreateAlertDialog, props.serviceID)}
```


Or have an optional `renderCreateDialog` prop for the one-off case:

```ts
renderCreateDialog={(props)=>(
  <CreateAlertDialog serviceID={props.serviceID} {props...} />
)}
```

web/src/app/lists/ControlledPaginatedList.tsx
```diff
@@ -57,6 +55,9 @@ export interface ControlledPaginatedListProps
   searchAdornment?: ReactElement
 
   items: CheckboxItemsProps[] | PaginatedListItemProps[]
+
+  CreateDialog?: ComponentType<{ onClose: () => void }>
```
mastercactapus, line 59: This should have the suffix `Component` since it's expected not to have been rendered yet.

web/src/app/lists/QueryList.tsx
```diff
@@ -84,6 +86,9 @@ export interface _QueryListProps extends ControlledPaginatedListProps {
 
   // mapVariables transforms query variables just before submission
   mapVariables?: (vars: OperationVariables) => OperationVariables
+
+  CreateDialog?: ComponentType<{ onClose: () => void }>
```
mastercactapus, line 90: Same as above, component props should have `Component` as a suffix.

mastercactapus (changes requested):

mastercactapus:
I'm happy with the `CreateDialogComponent` and `CreateDialogProps` pair, but I'm hesitant with the current implementation because it loses all type-checking info on the component's props.

Using `renderCreateDialogComponent(props)` would be a way to solve that. Still, I'm testing locally to see if we can annotate it differently so that `CreateDialogProps` is validated against the component you pass in.

<details>
<summary>what I'm trying so far</summary>

```ts
type CreateComponent = ComponentType<
  {
    onClose: () => void
  } & any
>

type Creatable<C extends CreateComponent> =
  | {}
  | {
      CreateDialogComponent: C
      createLabel: string
    }
  | {
      CreateDialogComponent: C
      CreateDialogProps: Omit<ComponentProps<C>, 'onClose'>
      createLabel: string
    }

type OtherProps = {
  title: string
}

type Props<C extends CreateComponent> = OtherProps & {
  details: string
} & Creatable<C>
```
</details>

web/src/app/lists/QueryList.tsx
```diff
@@ -199,6 +202,7 @@ export default function QueryList(props: QueryListProps): JSX.Element {
           noSearch={noSearch}
           CreateDialogComponent={CreateDialogComponent}
           CreateDialogProps={CreateDialogProps}
```
Forfold, line 204: whoops. i still need to remove these.

tony-tvu (dismissed):

mastercactapus (approved):

mastercactapus (approved):

AlaricWhitney (commented):
I see there's Cypress tests.  Is there a reason why there's no unit tests to quickly give confidence that this is working as expected?

Forfold:
@AlaricWhitney how would you recommend I write a unit test for a button that's moved? I generally reverse unit tests for functional updates that don't render anything, and UI testing with cypress or playwright otherwise

AlaricWhitney:
> @AlaricWhitney how would you recommend I write a unit test for a button that's moved? I generally reverse unit tests for functional updates that don't render anything, and UI testing with cypress or playwright otherwise

I think that's up to you.  All I care about is that the button will still work as expected, and I didn't see any unit tests to show that, but it could be that I don't know where to look for it.

Forfold:
@AlaricWhitney if the E2E tests pass we should be covered (cypress and playwright). They test the create buttons from every page. That being said, it looks like there's been changes to these tests specifically 😐 so I'll need to update that

AlaricWhitney (dismissed):

AlaricWhitney (approved):

AlaricWhitney (dismissed):

mastercactapus (approved):

AlaricWhitney (approved):

Issue #2714, opened 2022-11-30
Dev: Switch to GolangCI-Lint


ethan-haynes:
**What problem would you like to solve? Please describe:**
We want to switch from using multiple checks for linting to use a single Golang linting tool.

**Describe the solution you'd like:**
Be able to run `make check-go` which should execute the golangci-lint command. With error check disabled (out of scope).

Follow Up Issue:
#2579

mastercactapus:
Resolved by #2723

PR #2723, opened 2022-12-05
ci: switch to golangci-lint for Go linting

tony-tvu:

**Description:**
This PR updates the makefile command `go-check` to use [golangci-lint ](https://golangci-lint.run/)for Go linting.

https://github.com/target/goalert/issues/2714

**Out of Scope:**
Add config file to exclude errcheck and other specific errors


mastercactapus (changes requested):

Makefile
```diff
@@ -59,6 +59,9 @@ $(BIN_DIR)/tools/protoc-gen-go: go.mod
 $(BIN_DIR)/tools/protoc-gen-go-grpc: go.mod
 	GOBIN=$(abspath $(BIN_DIR))/tools go install google.golang.org/grpc/cmd/protoc-gen-go-grpc
 
+$(BIN_DIR)/tools/golangci-lint: golangci-lint.version
+	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(BIN_DIR)/tools $(shell cat golangci-lint.version)
```
mastercactapus, line 63: This will execute whatever is hosted in that install script -- instead, use `gettool` so we only download what we need directly from the releases page.

Makefile
```diff
@@ -142,13 +145,10 @@ check-js: force-yarn generate node_modules
 	yarn run lint
 	yarn workspaces run check
 
-check-go: generate
+check-go: generate $(BIN_DIR)/tools/golangci-lint bin/tools/golangci-lint
```
mastercactapus, line 148:
```suggestion
check-go: generate $(BIN_DIR)/tools/golangci-lint
```

Makefile
```diff
@@ -142,13 +145,10 @@ check-js: force-yarn generate node_modules
...
 	@go mod tidy
-	go vet ./...
-	go fmt ./...
 	# go run ./devtools/ordermigrations -check
-	go run github.com/gordonklaus/ineffassign ./...
-	CGO_ENABLED=0 go run honnef.co/go/tools/cmd/staticcheck ./...
+	./bin/tools/golangci-lint run
```
mastercactapus, line 151:
```suggestion
	$(BIN_DIR)/tools/golangci-lint run
```

mastercactapus (changes requested):

Makefile
tony-tvu: I switched it to use gettool. Their release for Windows only has **.zip** whereas linux/mac is **.gz**, so my current solution won't work for Windows. Should we account for Windows users as well? That would probably require a combination of the logic in `getPrometheus` and `getProtoC`.

mastercactapus: We should; I played with it this AM and pushed a fix for that. On a separate note, I think the `gettool` stuff should be refactored, especially with the increase in external tools.

Thanks for making the update!

mastercactapus (approved):

mastercactapus (approved):

KatieMSB (approved):

PR #2728, opened 2022-12-07
ui: use calendar override dialog for schedule and override lists

Forfold:
Part of #2693

Removes use of the override speed dial and replaces it with the dialog that the calendar uses, where the user is prompted with what each type of override (or temp schedule) is, and can then make a choice.

I plan to make a follow-up PR for the assignment and on-call notification buttons, but figured since this one is using context it should be on its own.

- Changes calendar button text from `Override` to `Add Override`
- Removes speed dial from shifts and overrides lists, using CreateFab instead if mobile
- Render an `Add Override` button in the top right of the shifts and overrides lists if on `> md` browser width
- Changes `ScheduleCalendarContext` to `OverrideDialogContext` for better reuse

Screenshots:

Screen Shot 2022-12-07 at 11 18 37 AM

Screen Shot 2022-12-07 at 11 19 33 AM

Screen Shot 2022-12-07 at 11 19 03 AM

Which all render this dialog (including mobile):

Screen Shot 2022-12-07 at 11 21 29 AM


mastercactapus:
@Forfold, I'm a little concerned with using "ADD" specifically because today's state has a different meaning for that word in the scope of overrides. Also, for developers, under the covers, we still have "add", "remove", and "replace" as override types.

If keeping it as a verb is important, maybe stick with "OVERRIDE" on the calendar page and  "CREATE OVERRIDE" on the override list page. Or "MAKE OVERRIDE" or "NEW OVERRIDE" IDK -- I think avoiding the word ADD might prevent some confusion for current users.

Forfold:
@mastercactapus Solid point, I didn't even consider that. Will make the changes to just say "+ OVERRIDE"

web/src/app/schedules/calendar/ScheduleCalendar.tsx
```diff
@@ -364,7 +364,7 @@ function ScheduleCalendar(props: ScheduleCalendarProps): JSX.Element {
               startIcon={<GroupAdd />}
               title='Make temporary change to schedule'
             >
-              Override
+              Add Override
```
mastercactapus, line 367: To completely avoid the overlap with ADD, how do you feel about having this just be `Override`? as in "override what you see on the calendar"

web/src/app/schedules/ScheduleOverrideDialog.js
```diff
@@ -35,7 +32,7 @@ const useStyles = makeStyles({
     marginTop: '.3rem',
   },
 })
-export default function ScheduleCalendarOverrideDialog(props) {
+export default function ScheduleOverrideDialog(props) {
```
tony-tvu, line 35: The proptypes on lines 149 and 153 need to be renamed to `ScheduleOverrideDialog.propTypes` otherwise they'll fail on `make check`.

tony-tvu (changes requested):

web/src/app/schedules/ScheduleOverrideDialog.js
Forfold: Thank you!

tony-tvu (approved):

tony-tvu (approved):

mastercactapus (approved):

PR #2729, opened 2022-12-07
ui/services: sort on call users by escalation step

### LINKED ISSUE #2126: sort on call users in order of escalation
_A clear and concise description of what the problem is._

Currently a service shows **On Call Users** sorted by name.  This means the first person to get alerted could be in the middle of the list.

_A clear and concise description of what you want to happen._

I think it'd make more sense to sort the **On Call Users** in order of _Step #_.  This would make the first person in the list the first person who would be paged if an alert was created against the service.

**Additional context:**

KatieMSB:

**Description:**
Updates the Service Details On Call Users list to have collapsable sections for each step for on call users.

Fixes #2126

**Screenshots:**
Screen Shot 2022-12-19 at 1 58 38 PM


mastercactapus (approved):
Looks good!

One thing of note we didn't consider: if there are no users on a step, it's omitted entirely -- probably okay, but I could also see users raising questions, "hey, why isn't step 2 showing up?"

We *could* pull in the escalation policy to display all steps, though we should probably add # of on-call users in each section as subtext or something for that as a hint.

Or maybe it's all the better as a follow-up _if_ someone requests it.

@Forfold Thoughts?

Forfold:
Yeah that sounds like a good idea. Display the step if even if there's nobody on it. The accordion details could just have some typography like "Nobody on-call for this step" and there could be a warning icon rendered to the left of the Step # text in the accordion summary

johnmarksilly (approved):
Changes look good!

mastercactapus (approved):

PR #2735, opened 2022-12-09
ui: add desktop service create buttons

Forfold:
Part of #2715

Adds create buttons when on wide browser width for:
- Heartbeat monitors
- Labels
- Integration Keys

Also adds a header note for the service labels list

### DIFF
--- a/web/src/app/services/ServiceLabelList.tsx
+++ b/web/src/app/services/ServiceLabelList.tsx
@@ -85,7 +85,7 @@ export default function ServiceLabelList(props: {
         data-cy='label-list'
         emptyMessage='No labels exist for this service.'
         items={items}
-        headerNote='Labels are a way associate services with each other throughout GoAlert. Search using the format key1/key2=value'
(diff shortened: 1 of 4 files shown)


web/src/app/services/ServiceLabelList.tsx
```diff
@@ -81,6 +85,18 @@ export default function ServiceLabelList(props: {
         data-cy='label-list'
         emptyMessage='No labels exist for this service.'
         items={items}
+        headerNote='Labels are a way associate services with each other throughout GoAlert. Search using the format key1/key2=value'
```
johnmarksilly, line 88: Typo on this line.

johnmarksilly (changes requested):
Typo but good otherwise

web/src/app/services/ServiceLabelList.tsx
Forfold, line 88:
```suggestion
        headerNote='Labels are a way to associate services with each other throughout GoAlert. Search using the format key1/key2=value'
```

mastercactapus (approved):

johnmarksilly (approved):

PR #2737, opened 2022-12-09
ui: some small fixes

Forfold:
From the [design discussion here](https://github.com/target/goalert/discussions/2489#discussioncomment-4210469):

- Reduces spacing on alerts checkbox controls between the box and the dropdown arrow
- Disables the checkbox controls if no items/no items in the list are being rendered
- Hides the alert tab filters from the dropdown menu when viewing on a wide screen browser (tabs are already present)
- Clicking the GoAlert logo in the top left will route you to the homepage ('/alerts')

### DIFF
--- a/web/src/app/main/AppRoutes.tsx
+++ b/web/src/app/main/AppRoutes.tsx
@@ -71,6 +71,7 @@ const alertQuery = gql`
 // Allow any component to be used as a route.
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const routes: Record<string, JSXElementConstructor<any>> = {
+  '/': AlertsList,
   '/alerts': AlertsList,
   '/alerts/:alertID': AlertDetailPage,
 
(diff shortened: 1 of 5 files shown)


johnmarksilly (approved):
Looks good. 🦖

web/src/app/main/AppRoutes.tsx
```diff
@@ -71,6 +71,7 @@ const alertQuery = gql`
 // Allow any component to be used as a route.
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const routes: Record<string, JSXElementConstructor<any>> = {
+  '/': AlertsList,
```
mastercactapus, line 74: I don't think this will ever render; we redirect `/` to `/alerts` before that happens

Forfold: Without this change, you'll see a flicker of "Page does not exist" before the alerts page renders

mastercactapus (approved):

PR #2738, opened 2022-12-12
ui: Fix default time not being set correctly for new schedule assignments

### LINKED ISSUE #2736: schedule assignments: default time should be in schedule TZ
**Describe the Bug:**
If a user is editing a schedule in a different timezone from their local, new rules default to midnight in their local timezone instead of the schedules.

**Steps to Reproduce:**
1. Create a schedule in a different timezone than your own
2. Go to assignments and add a rotation or user
3. Observe the input field default

**Expected Behavior:**
It should default to `12:00 AM` (or `00:00` depending on your locale)

**Observed Behavior:**
It displays midnight in _your_ timezone instead of the schedules.

**Screenshots/Stack Traces:**
image

**Application Version:**
v0.30.1 & master

cmarquis:

**Description:**
Fixed an issue with the ScheduleRuleCreateDialog where it was using the users timezone for the default start and end times rather than the schedules timezone.

Fixes #2736

### DIFF
--- a/web/src/app/schedules/ScheduleRuleCreateDialog.js
+++ b/web/src/app/schedules/ScheduleRuleCreateDialog.js
@@ -34,10 +34,10 @@ export default function ScheduleRuleCreateDialog(props) {
     targetID: '',
     rules: [
       {
-        start: DateTime.local({ zone: data?.schedule.timeZone ?? 'UTC' })
+        start: DateTime.local({ zone: data.schedule.timeZone })
           .startOf('day')
           .toISO(),
-        end: DateTime.local({ zone: data?.schedule.timeZone ?? 'UTC' })
+        end: DateTime.local({ zone: data.schedule.timeZone })
           .plus({ day: 1 })
           .startOf('day')
           .toISO(),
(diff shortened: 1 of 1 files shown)


web/src/app/schedules/ScheduleRuleCreateDialog.js
```diff
@@ -25,20 +25,27 @@ const query = gql`
...
       {
-        start: DateTime.local().startOf('day').toUTC().toISO(),
-        end: DateTime.local().plus({ day: 1 }).startOf('day').toUTC().toISO(),
+        start: DateTime.local({ zone: data?.schedule.timeZone ?? 'UTC' })
+          .startOf('day')
+          .toISO(),
+        end: DateTime.local({ zone: data?.schedule.timeZone ?? 'UTC' })
+          .plus({ day: 1 })
+          .startOf('day')
+          .toISO(),
```
mastercactapus, line 43: UTC fallback won't work this way because once the real data loads, it won't update and would lead to subtle bugs.

But, since we are guaranteed to have the schedule info cached, we can probably just omit the check.

```suggestion
        start: DateTime.local({ zone: data.schedule.timeZone })
          .startOf('day')
          .toISO(),
        end: DateTime.local({ zone: data.schedule.timeZone })
          .plus({ day: 1 })
          .startOf('day')
          .toISO(),
```

mastercactapus (changes requested):

KatieMSB (approved):

mastercactapus (approved):

PR #2742, opened 2022-12-13
admin/message-log: reflect sent time in sort order

### LINKED ISSUE #2650: admin: message log does not reflect sent time
**Describe the Bug:**
Currently, the message logs admin tool displays and sorts based on `created_at` time only, ignoring `sent_at`.

This is important in cases where messages are being throttled or retried since throttling is based on `sent_at`.

**Expected Behavior:**
Sort by `created_at` IF `sent_at` is null; otherwise, sort by `sent_at`, keeping unsent messages at the top.

**Observed Behavior:**
Sort is purely based on `created_at` time, sent time, and retries, all are not available.

cmarquis:

**Description:**
The admin message log will now display messages that have not yet been sent at the top of the list and then sort by the created time thereafter. 

Fixes #2650 

- The sorting of the admin message logs was affected as described in the description above.
- Added the sent at time to the message details
- Added the retry count to the message details, if greater than 0

Added `sentAt` and `retryCount` to the `DebugMessage` type.

**Additional Info:**
- Updated the sql query and model to retrieve the `sent_at` and `retry_count` values from the database
- Updated the sql query to sort messages first by any messages where `sent_at` is null and then `created_at`
- Added a new smoke test to validate the sorting is being applied properly

### DIFF
--- a/graphql2/graphqlapp/messagelog.go
+++ b/graphql2/graphqlapp/messagelog.go
@@ -173,12 +173,13 @@ func (q *Query) MessageLogs(ctx context.Context, opts *graphql2.MessageLogSearch
 		}
 
 		dm := graphql2.DebugMessage{
-			ID:        log.ID,
-			CreatedAt: log.CreatedAt,
-			UpdatedAt: log.LastStatusAt,
-			Type:      strings.TrimPrefix(log.MessageType.String(), "MessageType"),
-			Status:    msgStatus(notification.Status{State: log.LastStatus, Details: log.StatusDetails}),
-			AlertID:   &log.AlertID,
+			ID:         log.ID,
+			CreatedAt:  log.CreatedAt,
+			UpdatedAt:  log.LastStatusAt,
+			Type:       strings.TrimPrefix(log.MessageType.String(), "MessageType"),
+			Status:     msgStatus(notification.Status{State: log.LastStatus, Details: log.StatusDetails}),
+			AlertID:    &log.AlertID,
+			RetryCount: log.RetryCount,
 		}
 		if dest.ID != "" {
 			dm.Destination, err = q.formatDest(ctx, dest)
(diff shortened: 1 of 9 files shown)


graphql2/graphqlapp/messagelog.go
```diff
@@ -211,6 +212,9 @@ func (q *Query) MessageLogs(ctx context.Context, opts *graphql2.MessageLogSearch
 		if log.ProviderMsgID != nil {
 			dm.ProviderID = &log.ProviderMsgID.ExternalID
 		}
+		if log.SentAt != nil {
+			dm.SentAt = log.SentAt
+		}
```
mastercactapus, line 217: These can both be nil, so the `if` check isn't necessary.
```suggestion
		dm.SentAt = log.SentAt
```

notification/search.go
```diff
@@ -98,7 +102,7 @@ var searchTemplate = template.Must(template.New("search").Funcs(search.Helpers()
 		OR (om.created_at = :cursorCreatedAt AND om.id > :afterID)
 	{{end}}
 		AND om.last_status != 'bundled'
-	ORDER BY om.created_at desc, om.id asc
+	ORDER BY (CASE WHEN om.sent_at IS NULL THEN 1 ELSE 0 END) desc, om.created_at desc, om.id asc
```
mastercactapus, line 105: Hmm, this won't entirely give us what we want -- there are failure scenarios where `sent_at` will always be NULL; we also want to sort by the value of `sent_at` and not _only_ if it's null or not.

What about doing `sent_at` but falling back to `last_status_at`? That way, as messages are retried/updated, they will bounce to the top but, once sent, will keep their place.

Also, any messages with the `PENDING` state should probably stay at the top, since those can be important to consider.


```suggestion
	ORDER BY om.last_status = 'pending' desc, coalesce(om.sent_at, om.last_status_at) desc, om.created_at desc, om.id asc
```

mastercactapus (changes requested):

graphql2/graphqlapp/messagelog.go
cmarquis: Moved the setting of dm.SentAt into the struct instantiation.

notification/search.go
cmarquis: That makes sense. Change made and test updated as well.

Forfold (approved):

mastercactapus (approved):

Issue #2579, opened 2022-08-19
Add golang lint checking to Github actions for PR's


AlaricWhitney:
**What problem would you like to solve? Please describe:**
Currently for PR's, there is only a Github action that does a nodeJS lint check.  There is no lint checking on PR's for golang currently.

**Describe the solution you'd like:**
Add a github action for a robust golang lint check, such as https://github.com/golangci/golangci-lint by utilizing the following docs: https://golangci-lint.run/usage/install/#ci-installation

**Describe alternatives you've considered:**
You could add a check that runs `make check` to use whatever the Makefile specifies.

**Additional context:**
Had to submit https://github.com/target/goalert/issues/2577 to address pre-existing lint issues within the codebase.

mastercactapus:
We should start by adding `make check` as a GitHub action as an easy win that would help speed things up.

Separately let's look into adding the linter to the `check` target, I expect we'll have a fair amount of lining issues to work through.

PR #2752, opened 2022-12-21
feat: Add error check on tx.Rollback

AlaricWhitney:

**Description:**
In preparation for golangci-lint checks, this PR adds error checks to all tx.Rollback commands, using the golang logger whenever possible.

Part of #2579 

**Out of Scope:**
Not making any functional changes to the app.  Logging only!


mastercactapus (changes requested):
As a general rule, sweeping style changes (like how we handle defers) should have some form of discussion/buy-in before being implemented, not a hard requirement, but it will prevent substantial rework before things go in the wrong direction.

I'll include my review of what should change, but we should have a quick talk with the team for a second opinion since this is a coding style change.

---

As-is, this adds a little too much noise and overhead to the code IMO. It's also a problem because new engineers are going to need to know to open a closure and know to use our log framework just to manage transactions. Also, setting this particular pattern as the standard is problematic because we lose the capture feature of `defer`. If `tx` is reassigned later in the function, the code will break in subtle ways.

`defer` is designed to solve this by evaluating the call arguments immediately, but deferring the actual execution until later. We can use that to solve the reference issue by making a helper. That should also help reduce the noise/complexity at all of these call sites.

I'm thinking something like:
```go
defer errutil.Rollback(tx, "update alert metrics")
```

Or maybe a more generic version we can use elsewhere:

```go
defer errutil.Do(tx.Rollback, "update alert metrics: rollback")
```

`tx` or `tx.Rollback` will be captured immediately thus avoiding the reference issue.

---

For wrapping errors, use `fmt.Errorf` instead of `errors.Wrap` now that it's part of the standard library, we're moving away from `pkg/errors`

Finally, on error messages: in Go, errors should not start with a capital letter and should be terse. Instead of logging "error=Issue with EscalateMany rollback: some-error" we want it to say "error=escalate many: rollback: some-error". Each wrap should add context and not attempt to make it a sentence.

More info on best practices regarding errors:
https://github.com/golang/go/wiki/CodeReviewComments#error-strings
https://go.dev/doc/effective_go#errors

AlaricWhitney:
sounds good.  I'm not used to using `pkg/errors`, but I decided to copy the error patterns found within this project.  I'll setup a discussion next week on how to handle the defers, and if we really care about logging rollback errors at all (or just ignore the error).

AlaricWhitney:
switched to using shim utility functions to perform rollbacks, depending on the usage context.

alert/store.go
```diff
@@ -481,7 +481,7 @@ func (s *Store) Create(ctx context.Context, a *Alert) (*Alert, error) {
 	if err != nil {
 		return nil, err
 	}
-	defer tx.Rollback()
+	defer sqlutil.Rollback(ctx, "Create", tx)
```
mastercactapus, line 484: Seeing "Create" in the logs is going to be challenging to troubleshoot by itself, add the package and describe the action "alert: create" or "create alert"

alert/store.go
```diff
@@ -308,7 +308,7 @@ func (s *Store) EscalateMany(ctx context.Context, alertIDs []int) ([]int, error)
 	if err != nil {
 		return nil, err
 	}
-	defer tx.Rollback()
+	defer sqlutil.Rollback(ctx, "EscalateMany", tx)
```
mastercactapus, line 311:
```suggestion
	defer sqlutil.Rollback(ctx, "escalate alert", tx)
```

app/cmd.go
```diff
@@ -517,7 +518,7 @@ Migration: %s (#%d)
 			if err != nil {
 				return errors.Wrap(err, "begin tx")
 			}
-			defer tx.Rollback()
+			defer sqlutil.Rollback(ctx, "testCmd", tx)
```
mastercactapus, line 521: not sure `testCmd` would be clear what's happening
```suggestion
			defer sqlutil.Rollback(ctx, "add-user", tx)
```

With something like that ^ we'd see a message like "add-user: rollback tx: some-error"

mastercactapus (changes requested):
Error messaging needs a little tweaking, try to match existing context from call sites where it makes sense. A lot of times, having the package name might be helpful too; as a general rule, use lower-case words to describe the operation. (e.g., instead of "LinkAccount" do "auth: link account")

Function names can be ambiguous and don't always inform about the operation (e.g., from an admin, not dev perspective).

Admittedly error messaging across the app hasn't been consistent, but since we're adding new context info for these, I think it makes sense to do so for the new code so it can be seen as an example.

I flagged some but not all -- happy to discuss and iron out some better guidelines if that would be helpful.

Examples & more info:
- https://github.com/golang/go/wiki/CodeReviewComments#error-strings
- https://go.dev/doc/effective_go#errors
- https://go.dev/blog/error-handling-and-go

test/smoke/overrideconflict_test.go
```diff
@@ -10,12 +10,13 @@ import (
...
 	"github.com/target/goalert/permission"
 	"github.com/target/goalert/test/smoke/harness"
+	"github.com/target/goalert/util/sqlutil"
 )
 
 func TestOverrideConflict(t *testing.T) {
 	t.Parallel()
 
-	sql := `
+	sqlstmt := `
```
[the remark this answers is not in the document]
AlaricWhitney: i originally had to rename it due to a naming conflict with importing the sql library when I originally was doing the anonymous functions method. I'll rename it back now that it's no longer needed.

util/sqlutil/rollback.go
```diff
@@ -0,0 +1,38 @@
...
+func RollbackTest(t *testing.T, errMsg string, tx *sql.Tx) {
+	if err := tx.Rollback(); err != nil {
+		if err != sql.ErrTxDone && err != sql.ErrConnDone {
+			t.Errorf("tx rollback issue at %s: %v", errMsg, err)
+		}
+	}
+}
+
+// ContextRollback perform a DB rollback function with the context
+func ContextRollback(ctx context.Context, errMsg string, tx pgx.Tx) {
```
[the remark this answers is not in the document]
AlaricWhitney: Done (had to update manually due to the function name change)

util/sqlutil/rollback.go
```diff
@@ -0,0 +1,38 @@
...
+func Rollback(ctx context.Context, errMsg string, tx *sql.Tx) {
+	if err := tx.Rollback(); err != nil {
+		if err != sql.ErrTxDone && err != sql.ErrConnDone {
+			log.Log(ctx, fmt.Errorf("tx rollback issue at %s: %v", errMsg, err))
+		}
+	}
+}
+
+// RollbackTest will perform a DB rollback for use only within tests
+func RollbackTest(t *testing.T, errMsg string, tx *sql.Tx) {
```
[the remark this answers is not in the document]
AlaricWhitney: Sure thing.  It all depends on how you want to handle it.  I personally would use `t.Fatal` in this case, because it's still tied to the testing framework in case you want to do anything with it.

alert/store.go
AlaricWhitney: sure thing.  went with alert: create

AlaricWhitney:
I took a swing at updating the error messages to include the package where they came from.  I'm confident at this point a developer could troubleshoot where in the code an issue be generated from.

Forfold (approved):
lgtm

PR #2753, opened 2022-12-21
feat: Handle errors from w.Write

AlaricWhitney:

**Description:**
In preparation for golangci-lint checks, this PR adds error checks to all w.Write commands, using the logging patterns of each function

**Out of Scope:**

### DIFF
--- a/auth/handler.go
+++ b/auth/handler.go
@@ -276,7 +276,10 @@ func (h *Handler) ServeProviders(w http.ResponseWriter, req *http.Request) {
 	if errutil.HTTPError(req.Context(), w, err) {
 		return
 	}
-	w.Write(data)
+	_, err = w.Write(data)
+	if errutil.HTTPError(req.Context(), w, err) {
+		return
+	}
 }
 
 // IdentityProviderHandler will return a handler for the given provider ID.
--- a/auth/handler.go
+++ b/auth/handler.go
@@ -276,10 +276,7 @@ func (h *Handler) ServeProviders(w http.ResponseWriter, req *http.Request) {
 	if errutil.HTTPError(req.Context(), w, err) {
 		return
 	}
-	_, err = w.Write(data)
-	if errutil.HTTPError(req.Context(), w, err) {
-		return
-	}
+	_, _ = w.Write(data)
 }
 
 // IdentityProviderHandler will return a handler for the given provider ID.
--- a/calsub/http.go
+++ b/calsub/http.go
@@ -50,5 +50,8 @@ func (s *Store) ServeICalData(w http.ResponseWriter, req *http.Request) {
 	}
 
 	w.Header().Set("Content-Type", "text/calendar")
-	w.Write(calData)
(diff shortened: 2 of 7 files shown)


auth/handler.go
```diff
@@ -276,7 +276,10 @@ func (h *Handler) ServeProviders(w http.ResponseWriter, req *http.Request) {
 	if errutil.HTTPError(req.Context(), w, err) {
 		return
 	}
-	w.Write(data)
+	_, err = w.Write(data)
+	if errutil.HTTPError(req.Context(), w, err) {
+		return
+	}
```
mastercactapus, line 282: This is incorrect; this code will attempt to send and log a 500 status code on a write failure, which occurs when the client has already disconnected, _and_ calling `w.Write` already sends the status code and headers; attempts to do so a second time is also invalid.

Since it's normal/expected to have a write error (e.g., when someone closes a tab), this error should probably be ignored. If there's a way to filter out connection errors, logging would be okay, but it would not be helpful to have errors in the logs every time someone closes a tab.

AlaricWhitney: done.  I errored on the side of caution and tried to copy error patterns within the function.  I'm totally fine with ignoring the error in this case if you are.

calsub/http.go
```diff
@@ -50,5 +50,8 @@ func (s *Store) ServeICalData(w http.ResponseWriter, req *http.Request) {
 	}
 
 	w.Header().Set("Content-Type", "text/calendar")
-	w.Write(calData)
+	_, err = w.Write(calData)
+	if errutil.HTTPError(ctx, w, err) {
```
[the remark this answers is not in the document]
AlaricWhitney: done.  I errored on the side of caution and tried to copy error patterns within the function.  I'm totally fine with ignoring the error in this case if you are.

devtools/mocktwilio/lookup.go
```diff
@@ -49,5 +49,8 @@ func (s *Server) serveLookup(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 	w.Header().Set("Content-Type", "application/json")
-	w.Write(data)
+	if _, err := w.Write(data); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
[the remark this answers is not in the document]
AlaricWhitney: done.  I errored on the side of caution and tried to copy error patterns within the function.  I'm totally fine with ignoring the error in this case if you are.

calsub/http.go
```diff
@@ -50,5 +50,8 @@ func (s *Store) ServeICalData(w http.ResponseWriter, req *http.Request) {
...
+		return
+	}
```
mastercactapus, line 56: missed one:
```suggestion
	_, _ = w.Write(calData)
```

mastercactapus (changes requested):

mastercactapus (approved):

cmarquis (approved):

PR #2755, opened 2022-12-21
feat: Error check Process.Kill()

AlaricWhitney:

***Description:***
In preparation for golangci-lint checks, this PR adds error checks to all Process.Kill() commands, using the logging patterns of each function

***Which issue(s) this PR fixes:***

***Out of Scope:***

### DIFF
--- a/devtools/runproc/process.go
+++ b/devtools/runproc/process.go
@@ -141,7 +141,9 @@ func (p *Process) Kill() {
 	}
 
 	p.logAction("Killing...")
-	p.cmd.Process.Kill()
+	if err := p.cmd.Process.Kill(); err != nil {
+		p.logError(err)
+	}
 	p.state <- ProcessStateKilling
 
 	<-p.exited
--- a/devtools/sendit/readme_test.go
+++ b/devtools/sendit/readme_test.go
@@ -51,7 +51,11 @@ func TestReadme(t *testing.T) {
 	r, w := io.Pipe()
 	cmd.Stderr = io.MultiWriter(w, os.Stdout)
 	require.NoError(t, cmd.Start())
-	defer cmd.Process.Kill()
+	defer func() {
+		if err := cmd.Process.Kill(); err != nil {
+			t.Error(err)
+		}
+	}()
 
 	rd := bufio.NewReader(r)
 	s, err := rd.ReadString('\n')
@@ -78,7 +82,11 @@ func TestReadme(t *testing.T) {
 	cmd.Stderr = w
 	rd = bufio.NewReader(r)
 	require.NoError(t, cmd.Start())
-	defer cmd.Process.Kill()
+	defer func() {
+		if err := cmd.Process.Kill(); err != nil {
+			t.Error(err)
+		}
+	}()
 
 	for {
 		s, err = rd.ReadString('\n')


devtools/procwrap/main.go
```diff
@@ -168,7 +167,9 @@ func stopWith(lock bool, sig os.Signal) int {
 	log.Println("stopping", flag.Arg(0))
 	close(stopping)
 
-	cmd.Process.Signal(sig)
+	if err := cmd.Process.Signal(sig); err != nil {
+		log.Println("ERROR: ", err)
```
mastercactapus, line 171: `panic` would be better here, the code below has no condition to handle if the signal failed to send, and it's assumed it will never fail.

devtools/runproc/process.go
```diff
@@ -129,7 +129,9 @@ func (p *Process) gracefulTerm() {
 		return
 	}
 
-	p.cmd.Process.Signal(os.Interrupt)
+	if err := p.cmd.Process.Signal(os.Interrupt); err != nil {
+		p.logError(err)
```
mastercactapus, line 133: same, this was always assumed to never fail, so just panic.

devtools/runproc/process.go
```diff
@@ -141,7 +143,9 @@ func (p *Process) Kill() {
 	}
 
 	p.logAction("Killing...")
-	p.cmd.Process.Kill()
+	if err := p.cmd.Process.Kill(); err != nil {
+		p.logError(err)
```
mastercactapus, line 147: same

devtools/sendit/readme_test.go
```diff
@@ -51,7 +51,11 @@ func TestReadme(t *testing.T) {
 	r, w := io.Pipe()
 	cmd.Stderr = io.MultiWriter(w, os.Stdout)
 	require.NoError(t, cmd.Start())
-	defer cmd.Process.Kill()
+	defer func() {
+		if err := cmd.Process.Kill(); err != nil {
+			t.Error(err)
+		}
+	}()
```
mastercactapus, line 58: This loses reference to `cmd` and defeats the purpose of using defer.

If we can't clean up the process we start, the test itself is invalid, rather than the testable code failing; this, too, should probably be a `panic`. We should also ignore the case that the returned error is `ErrProcessDone`

```suggestion
	defer mustExit(cmd.Process.Kill)
```

devtools/sendit/readme_test.go
```diff
@@ -78,7 +82,11 @@ func TestReadme(t *testing.T) {
 	cmd.Stderr = w
 	rd = bufio.NewReader(r)
 	require.NoError(t, cmd.Start())
-	defer cmd.Process.Kill()
+	defer func() {
+		if err := cmd.Process.Kill(); err != nil {
```
mastercactapus, line 86: same

mastercactapus (changes requested):

devtools/sendit/readme_test.go
AlaricWhitney: I tested out your suggestion, and `mustExit` doesn't exist.  Would you rather I make a shim function of `mustExit` and check if we need to panic inside of that, or just pass `cmd` into this anonymous function and check for panic in there?

mastercactapus: Yes, sorry for the confusion, intended it as pseudo code.

I was thinking something like this:
```go
func mustExit(fn func() error) {
	err := fn()
	if err != nil && !errors.Is(err, os.ErrProcessDone) {
		panic(err)
	}
}
```

Though your suggestion of passing `cmd` instead of the `cmd.Process.Kill` method may read cleaner, I like that.

AlaricWhitney: I certainly don't mind doing a shim function like you've shown above.  When looking at the bigger picture, my only thought on it would be we probably want to be consistent across the app with handling defers that require some kind of error checking (even if `_ = something()`).  I'll wait until after our planed offline discussion before taking any action on this.

AlaricWhitney: added shim function as suggested

devtools/sendit/readme_test.go
AlaricWhitney: added shim function as suggested

AlaricWhitney:
Changed the defers to use a shim function as suggested

devtools/sendit/readme_test.go
```diff
@@ -51,7 +52,7 @@ func TestReadme(t *testing.T) {
 	r, w := io.Pipe()
 	cmd.Stderr = io.MultiWriter(w, os.Stdout)
 	require.NoError(t, cmd.Start())
-	defer cmd.Process.Kill()
+	defer mustExit(cmd.Process.Kill())
```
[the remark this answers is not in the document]
AlaricWhitney: I see the missing element I had was to not include the parentheses.  I removed them and modified the mustExit function to use the generic like you suggested, and it works.

mastercactapus (approved):

PR #2756, opened 2022-12-21
feat: Error check on io.WriteString

AlaricWhitney:

**Description:**
In preparation for golangci-lint checks, this PR adds error checks to all io.WriteString() commands, using the logging patterns of each function

**Out of Scope:**

### DIFF
--- a/app/healthcheck.go
+++ b/app/healthcheck.go
@@ -59,8 +59,6 @@ func (app *App) engineCycle(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	if _, err := io.WriteString(w, app.Engine.NextCycleID().String()); err != nil {
-		http.Error(w, "engine unable to move to next cycle", http.StatusBadRequest)
-		return
-	}
+	_, _ = io.WriteString(w, app.Engine.NextCycleID().String())
+
 }
--- a/app/healthcheck.go
+++ b/app/healthcheck.go
@@ -59,5 +59,8 @@ func (app *App) engineCycle(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	io.WriteString(w, app.Engine.NextCycleID().String())
+	if _, err := io.WriteString(w, app.Engine.NextCycleID().String()); err != nil {
+		http.Error(w, "engine unable to move to next cycle", http.StatusBadRequest)
+		return
+	}
 }
--- a/auth/handler.go
+++ b/auth/handler.go
@@ -421,7 +421,9 @@ func (h *Handler) handleProvider(id string, p IdentityProvider, refU *url.URL, w
 			} else {
 				w.WriteHeader(400)
 			}
-			io.WriteString(w, err.Error())
+			if _, err := io.WriteString(w, err.Error()); err != nil {
+				log.Log(ctx, errors.Wrap(err, "Issue with handleProvider when using io.WriteString"))
+			}
 			return
 		}
 		http.Redirect(w, req, refU.String(), http.StatusFound)
@@ -502,7 +504,9 @@ func (h *Handler) handleProvider(id string, p IdentityProvider, refU *url.URL, w
 	}
 
 	if noRedirect {
-		io.WriteString(w, tokStr)
+		if _, err := io.WriteString(w, tokStr); err != nil {
+			log.Log(ctx, errors.Wrap(err, "Issue with handleProvider when using io.WriteString"))
+		}
 		return
 	}
 
--- a/auth/handler.go
+++ b/auth/handler.go
@@ -421,9 +421,7 @@ func (h *Handler) handleProvider(id string, p IdentityProvider, refU *url.URL, w
 			} else {
 				w.WriteHeader(400)
 			}
-			if _, err := io.WriteString(w, err.Error()); err != nil {
-				log.Log(ctx, errors.Wrap(err, "Issue with handleProvider when using io.WriteString"))
-			}
+			_, _ = io.WriteString(w, err.Error())
 			return
 		}
 		http.Redirect(w, req, refU.String(), http.StatusFound)
@@ -504,9 +502,7 @@ func (h *Handler) handleProvider(id string, p IdentityProvider, refU *url.URL, w
 	}
 
 	if noRedirect {
-		if _, err := io.WriteString(w, tokStr); err != nil {
-			log.Log(ctx, errors.Wrap(err, "Issue with handleProvider when using io.WriteString"))
-		}
+		_, _ = io.WriteString(w, tokStr)
 		return
 	}
 
--- a/devtools/pgdump-lite/dumpdata.go
+++ b/devtools/pgdump-lite/dumpdata.go
@@ -152,11 +152,17 @@ func DumpData(ctx context.Context, conn *pgx.Conn, out io.Writer, skip []string)
 			}
 			for i, v := range vals {
 				if i > 0 {
-					_, _ = io.WriteString(out, "\t")
+					if _, err := io.WriteString(out, "\t"); err != nil {
+						return err
+					}
 				}
-				_, _ = io.WriteString(out, string(*v.(*scannable)))
+				if _, err := io.WriteString(out, string(*v.(*scannable))); err != nil {
+					return err
+				}
 			}
-			_, _ = io.WriteString(out, "\n")
+			if _, err := io.WriteString(out, "\n"); err != nil {
+				return err
+			}
 		}
 		rows.Close()
 
--- a/devtools/pgdump-lite/dumpdata.go
+++ b/devtools/pgdump-lite/dumpdata.go
@@ -152,17 +152,11 @@ func DumpData(ctx context.Context, conn *pgx.Conn, out io.Writer, skip []string)
 			}
 			for i, v := range vals {
 				if i > 0 {
-					if _, err := io.WriteString(out, "\t"); err != nil {
-						return err
-					}
+					_, _ = io.WriteString(out, "\t")
 				}
-				if _, err := io.WriteString(out, string(*v.(*scannable))); err != nil {
-					return err
-				}
+				_, _ = io.WriteString(out, string(*v.(*scannable)))
 			}
-			if _, err := io.WriteString(out, "\n"); err != nil {
-				return err
-			}
+			_, _ = io.WriteString(out, "\n")
 		}
 		rows.Close()
 
--- a/devtools/pgdump-lite/dumpdata.go
+++ b/devtools/pgdump-lite/dumpdata.go
@@ -152,11 +152,17 @@ func DumpData(ctx context.Context, conn *pgx.Conn, out io.Writer, skip []string)
 			}
 			for i, v := range vals {
 				if i > 0 {
-					io.WriteString(out, "\t")
+					if _, err := io.WriteString(out, "\t"); err != nil {
+						return err
+					}
 				}
-				io.WriteString(out, string(*v.(*scannable)))
+				if _, err := io.WriteString(out, string(*v.(*scannable))); err != nil {
+					return err
+				}
 			}
-			io.WriteString(out, "\n")
+			if _, err := io.WriteString(out, "\n"); err != nil {
+				return err
+			}
 		}
 		rows.Close()
 
--- a/devtools/runproc/process.go
+++ b/devtools/runproc/process.go
@@ -124,11 +124,11 @@ func (p *Process) Stop() {
 func (p *Process) gracefulTerm() {
 	if p.pty != nil {
 		if _, err := io.WriteString(p.pty, "\x03"); err != nil {
-			p.logError(err)
+			panic(err)
 		}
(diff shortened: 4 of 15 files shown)


app/healthcheck.go
```diff
@@ -59,5 +59,8 @@ func (app *App) engineCycle(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	io.WriteString(w, app.Engine.NextCycleID().String())
+	if _, err := io.WriteString(w, app.Engine.NextCycleID().String()); err != nil {
+		http.Error(w, "engine unable to move to next cycle", http.StatusBadRequest)
+		return
+	}
```
mastercactapus, line 65: This is incorrect; It's also not a bad request, and the error comes from the `Write` method being unable to write to the `ResponseWriter` (presumably because the other end canceled the request) and not from the engine.

An error in writing a response to a request that's already been canceled should be ignored. As noted in a previous PR.

graphql2/cmtype.go
```diff
@@ -1,17 +1,21 @@
...
+
 	graphql "github.com/99designs/gqlgen/graphql"
 	"github.com/pkg/errors"
 )
 
 func MarshalContactMethodType(t contactmethod.Type) graphql.Marshaler {
 	return graphql.WriterFunc(func(w io.Writer) {
-		io.WriteString(w, `"`+string(t)+`"`)
+		if _, err := io.WriteString(w, `"`+string(t)+`"`); err != nil {
+			log.Println("ERROR with MarshalContactMethodType when using io.WriteString:", err)
```
mastercactapus, line 17: Does this writer write to the ResponseWriter or a buffer?

If it's the former, we likely should ignore it because it means the other end is disconnected; if it's later, it will never return error (panic internal). In both cases, it should be ignored.

Lastly:
- don't use the stdlib logger; use the application logging package
- use the proper error message formatting

mailgun/mailgun.go
```diff
@@ -67,8 +67,12 @@ func httpError(ctx context.Context, w http.ResponseWriter, err error) bool {
...
 func validSignature(ctx context.Context, req *http.Request, apikey string) bool {
 	h := hmac.New(sha256.New, []byte(apikey))
-	io.WriteString(h, req.FormValue("timestamp"))
-	io.WriteString(h, req.FormValue("token"))
+	if _, err := io.WriteString(h, req.FormValue("timestamp")); err != nil {
+		log.Log(ctx, errors.Wrap(err, "Issue with validSignature when using io.WriteString"))
+	}
+	if _, err := io.WriteString(h, req.FormValue("token")); err != nil {
+		log.Log(ctx, errors.Wrap(err, "Issue with validSignature when using io.WriteString"))
+	}
```
mastercactapus, line 75: These should be ignored to eliminate noise; the package documentation guarantees it will never return an error:
https://pkg.go.dev/hash#Hash

notification/twilio/twiml.go
```diff
@@ -195,7 +195,9 @@ func (t *twiMLResponse) sendResponse() {
 	}
 
 	var buf bytes.Buffer
-	io.WriteString(&buf, xml.Header)
+	if _, err := io.WriteString(&buf, xml.Header); err != nil {
```
mastercactapus, line 198: `bytes.Buffer` will never return an error; if out of memory, it will panic internally. The error value should be ignored

notification/twilio/twiml.go
```diff
@@ -205,5 +207,8 @@ func (t *twiMLResponse) sendResponse() {
 	}
 
 	t.w.Header().Set("Content-Type", "application/xml; charset=utf-8")
-	io.WriteString(t.w, buf.String())
+	if _, err := io.WriteString(t.w, buf.String()); err != nil {
+		http.Error(t.w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+		panic(err)
```
mastercactapus, line 212: We should not try to write a second set of response headers, nor panic the application if twilio disconnects early (the error state this represents)

remotemonitor/monitor.go
```diff
@@ -76,7 +76,12 @@ func NewMonitor(cfg Config) (*Monitor, error) {
 		m.ServeHTTP(w, req)
 	}), m.tw)
 	mux := http.NewServeMux()
-	mux.HandleFunc("/health", func(w http.ResponseWriter, req *http.Request) { io.WriteString(w, "ok") })
+	mux.HandleFunc("/health", func(w http.ResponseWriter, req *http.Request) {
+		if _, err := io.WriteString(w, "ok"); err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
```
mastercactapus, line 83: Same as elsewhere

test/smoke/mailgunalerts_test.go
```diff
@@ -67,8 +67,12 @@ func TestMailgunAlerts(t *testing.T) {
...
 
 	hm := hmac.New(sha256.New, []byte(cfg.Mailgun.APIKey))
-	io.WriteString(hm, timestamp)
-	io.WriteString(hm, token)
+	if _, err := io.WriteString(hm, timestamp); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := io.WriteString(hm, token); err != nil {
+		t.Fatal(err)
+	}
```
mastercactapus, line 75: same as above

util/timeutil/isoduration.go
```diff
@@ -164,11 +164,15 @@ func ParseISODuration(s string) (d ISODuration, err error) {
...
+		if _, err := io.WriteString(w, "null"); err != nil {
+			panic(err)
+		}
 		return
 	}
 
-	io.WriteString(w, `"`+dur.String()+`"`)
+	if _, err := io.WriteString(w, `"`+dur.String()+`"`); err != nil {
+		panic(err)
+	}
```
mastercactapus, line 175: ^^

util/timeutil/isorinterval.go
```diff
@@ -135,11 +135,15 @@ func ParseISORIntervalFrom(t time.Time, s string) (ISORInterval, error) {
...
+		if _, err := io.WriteString(w, "null"); err != nil {
+			panic(err)
+		}
 		return
 	}
 
-	io.WriteString(w, `"`+r.String()+`"`)
+	if _, err := io.WriteString(w, `"`+r.String()+`"`); err != nil {
+		panic(err)
+	}
```
mastercactapus, line 146: ^^

mastercactapus (changes requested):

app/healthcheck.go
AlaricWhitney: sure thing.  I'll set it to ignore the error if you're ok with it (I decided to err on the side of caution).

graphql2/cmtype.go
AlaricWhitney: I'll just have it ignore the error since we probably don't care about it.  Let me know if you'd rather a panic instead.

(as an fyi, i didn't use the app logger here because ctx didn't exist, and it would've involved application changes to get ctx passed in here, which I figured was outside the scope of work).  Either way it's moot once I set it to ignore the error.

auth/handler.go
```diff
@@ -421,7 +421,9 @@ func (h *Handler) handleProvider(id string, p IdentityProvider, refU *url.URL, w
 			} else {
 				w.WriteHeader(400)
 			}
-			io.WriteString(w, err.Error())
+			if _, err := io.WriteString(w, err.Error()); err != nil {
+				log.Log(ctx, errors.Wrap(err, "Issue with handleProvider when using io.WriteString"))
+			}
```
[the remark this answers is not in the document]
AlaricWhitney: done.  set to ignore (repeated the pattern across the PR where I think it makes sense).

auth/handler.go
```diff
@@ -502,7 +504,9 @@ func (h *Handler) handleProvider(id string, p IdentityProvider, refU *url.URL, w
 	}
 
 	if noRedirect {
-		io.WriteString(w, tokStr)
+		if _, err := io.WriteString(w, tokStr); err != nil {
+			log.Log(ctx, errors.Wrap(err, "Issue with handleProvider when using io.WriteString"))
```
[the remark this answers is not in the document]
AlaricWhitney: done. set to ignore (repeated the pattern across the PR where I think it makes sense).

devtools/runproc/process.go
```diff
@@ -123,9 +123,13 @@ func (p *Process) Stop() {
 
 func (p *Process) gracefulTerm() {
 	if p.pty != nil {
-		io.WriteString(p.pty, "\x03")
+		if _, err := io.WriteString(p.pty, "\x03"); err != nil {
+			p.logError(err)
```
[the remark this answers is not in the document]
AlaricWhitney: agreed.  I wasn't sure if you wanted a panic here, or just error with a log, so I err'd on the side of caution.  I've changed this to panic.

graphql2/isotimestamp.go
```diff
@@ -13,10 +14,14 @@ import (
...
+			if _, err := io.WriteString(w, "null"); err != nil {
+				log.Println("ERROR: Issue with MarshalISOTimestamp when using io.WriteString:", err)
+			}
 			return
 		}
-		io.WriteString(w, `"`+t.UTC().Format(time.RFC3339Nano)+`"`)
+		if _, err := io.WriteString(w, `"`+t.UTC().Format(time.RFC3339Nano)+`"`); err != nil {
+			log.Println("ERROR: Issue with MarshalISOTimestamp when using io.WriteString:", err)
+		}
 	})
```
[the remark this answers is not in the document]
AlaricWhitney: changed to ignore the error since I personally don't think we need the error message, and if you're ok with ignoring it in these cases where we don't really care if it errors anyways.

mailgun/mailgun.go
AlaricWhitney: changed to ignore the error since I personally don't think we need the error message, and if you're ok with ignoring it in these cases where we don't really care if it errors anyways.

notification/twilio/twiml.go
AlaricWhitney: done.  set to ignore

notification/twilio/twiml.go
AlaricWhitney: sure thing.  I simply copied the err pattern from earlier in the same function.  I've now updated it to ignore the error.

remotemonitor/monitor.go
AlaricWhitney: sure thing.  I simply copied the err pattern from earlier in the same function.  I've now updated it to ignore the error.

test/smoke/mailgunalerts_test.go
AlaricWhitney: done.  set to ignore the error

util/timeutil/isoduration.go
AlaricWhitney: done.  set to ignore the error.  I only set to panic based on looking at the surrounding functions, but i've updated it to ignore the error.

util/timeutil/isorinterval.go
AlaricWhitney: done. set to ignore the error. I only set to panic based on looking at the surrounding functions

graphql2/cmtype.go
mastercactapus: TBH, probably why we did that to begin with. The complexity of getting context here (since `gqlgen` doesn't support it here) would not be worth it.

graphql2/isotimestamp.go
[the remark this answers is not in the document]
mastercactapus: Yeah, I'm reasonably sure gqlgen builds this in a buffer before sending, so no error should be possible. If it is, it would be because someone closed a tab, which wouldn't be helpful.

devtools/pgdump-lite/dumpdata.go
```diff
@@ -152,11 +152,11 @@ func DumpData(ctx context.Context, conn *pgx.Conn, out io.Writer, skip []string)
...
 			for i, v := range vals {
 				if i > 0 {
-					io.WriteString(out, "\t")
+					_, _ = io.WriteString(out, "\t")
 				}
-				io.WriteString(out, string(*v.(*scannable)))
+				_, _ = io.WriteString(out, string(*v.(*scannable)))
 			}
-			io.WriteString(out, "\n")
+			_, _ = io.WriteString(out, "\n")
```
mastercactapus, line 159: If we fail to write to the output file, we want the operation to fail immediately with the proper error message. Otherwise, we'll continue dumping data and writing blindly to a broken pipe with no indication of a problem.

This is a legitimate case where the linter would have found a bug (errors that shouldn't be ignored); your original code returning the error for these was the proper fix.

mastercactapus (changes requested):

devtools/pgdump-lite/dumpdata.go
AlaricWhitney: sounds good.  I'll put the panic back in for this section of code.

mastercactapus (approved):

cmarquis (approved):

PR #2761, opened 2022-12-23
Fix UI bugs on removing user from rotation

ram191:
**Description:**
Following the previously [closed PR](https://github.com/target/goalert/pull/2707) on fixing the same issue, this PR fixes the operation when mutating rotation participants. The bug initially occurs because the code is trying to update the active user index on an out of bound index since the user is removed from the rotation. 

As suggested, I added the error props in the Dialog component so that the error is properly shown if there's any when submitting the mutation. I also omitted the `activeUserIndex` from the mutation and it immediately fixes this issue.

Fixes https://github.com/target/goalert/issues/2543

**Screenshots:**
Screen Shot 2022-12-23 at 13 43 59

### DIFF
--- a/web/src/app/rotations/RotationUserDeleteDialog.tsx
+++ b/web/src/app/rotations/RotationUserDeleteDialog.tsx
@@ -13,7 +13,6 @@ const query = gql`
         id
         name
       }
-      activeUserIndex
     }
   }
 `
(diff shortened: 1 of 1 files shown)


mastercactapus (approved):
Functionally validated error display and bugfix locally, LGTM.

I also appreciate the explanation and screenshot; thanks for the fix!

KatieMSB (approved):

ram191:
Glad to help with the issue!

PR #2764, opened 2022-12-29
user/cm: cleanup unverified contact methods

### LINKED ISSUE #2727: user/cm: cleanup contact methods that do not get verified
Contact methods can be created, without verification, that sit forever. This can be an issue if a user mistypes a contact method value (like phone number) and then never deletes it because other users cannot add the same value.

Cleanup contact methods when their verification code expires if they've been created recently.

N/A this would need to be handled manually by an admin or individual users.

**Additional context:**
Should we also clean up contact methods that have been disabled for long periods?

cmarquis:

**Description:**
Unverified contact methods will be automatically removed when the verification code expires. This prevents contact methods that may have been created by mistake or were otherwise orphaned from hanging around in the system.

Fixes #2727

**Screenshots:**
Contact Methods

A UI message is displayed on unverified contact methods.

Added a `pending` value to the `UserContactMethod`.

### DIFF
--- a/engine/verifymanager/db.go
+++ b/engine/verifymanager/db.go
@@ -3,6 +3,7 @@ package verifymanager
 import (
 	"context"
 	"database/sql"
+
 	"github.com/target/goalert/engine/processinglock"
 	"github.com/target/goalert/util"
 )
@@ -50,12 +51,16 @@ func NewDB(ctx context.Context, db *sql.DB) (*DB, error) {
 
 		cleanupExpired: p.P(`
 			with rows as (
-				select id
+				select id, contact_method_id
 				from user_verification_codes
 				where now() >= expires_at
 				limit 100
 				for update skip locked
-			)
+			), _cms as (
+        delete from user_contact_methods cm
+        using rows
+        where cm.id = rows.contact_method_id and cm.pending = true
+      )
 			delete from user_verification_codes code
 			using rows
 			where code.id = rows.id
--- a/migrate/migrations/20221228092646-add-pending-to-contact-methods.sql
+++ b/migrate/migrations/20221228092646-add-pending-to-contact-methods.sql
@@ -1,0 +1,36 @@
+-- +migrate Up
+
+BEGIN;
+
+-- Add pending column with default to false for existing records and update default to true for new records later on
+ALTER TABLE user_contact_methods ADD COLUMN pending BOOLEAN NOT NULL DEFAULT FALSE;
+
+-- Update default to true for new records
+ALTER TABLE user_contact_methods ALTER COLUMN pending SET DEFAULT TRUE;
+
+-- Add a function to set pending to false when disabled is set to false
+-- +migrate StatementBegin
+CREATE OR REPLACE FUNCTION set_pending_to_false() RETURNS TRIGGER AS $$
+BEGIN
+  IF NEW.disabled = FALSE THEN
+    NEW.pending = FALSE;
+  END IF;
+  RETURN NEW;
+END;
+$$ language plpgsql;
+-- +migrate StatementEnd
+
+-- Bind the trigger function to user_contact_methods
+CREATE TRIGGER set_pending_to_false
+BEFORE UPDATE OF disabled ON user_contact_methods
+FOR EACH ROW EXECUTE PROCEDURE set_pending_to_false();
+
+COMMIT;
+
+-- +migrate Down
+
+-- Drop pending column
+ALTER TABLE user_contact_methods DROP COLUMN pending;
+
+-- Drop trigger
+DROP FUNCTION set_pending_to_false();
--- a/test/smoke/verifymanager_test.go
+++ b/test/smoke/verifymanager_test.go
@@ -1,0 +1,177 @@
+package smoke
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/target/goalert/permission"
+	"github.com/target/goalert/test/smoke/harness"
+)
+
+// TestUserVerificationCleanup tests that verification codes are cleaned up
+// when they are expired. Also tests that contact methods that are not verified
+// are cleaned up with the verification code.
+func TestUserVerificationCleanup(t *testing.T) {
+	t.Parallel()
+
+	type cm struct {
+		ID       string `json:"id"`
+		Disabled bool   `json:"disabled"`
+		Pending  bool   `json:"pending"`
+	}
+
+	type data struct {
+		User struct {
+			ID             string `json:"id"`
+			ContactMethods []cm   `json:"contactMethods"`
+		} `json:"user"`
+	}
+
+	doQL := func(t *testing.T, h *harness.Harness, query string, res interface{}) {
+		t.Helper()
+		g := h.GraphQLQuery2(query)
+		for _, err := range g.Errors {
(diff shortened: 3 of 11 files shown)


test/smoke/verifymanager_test.go
```diff
@@ -0,0 +1,177 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/target/goalert/permission"
+	"github.com/target/goalert/test/smoke/harness"
+)
+
+// TestUserVerificationCleanup tests that verification codes are cleaned up
+// when they are expired. Also tests that contact methods that are not verified
+// are cleaned up with the verification code.
+func TestUserVerificationCleanup(t *testing.T) {
```
mastercactapus, line 19: This test is a little hard to follow. Also, it's several separate tests. `t.Run` is generally intended for table tests (i.e., from a `for`-loop), which is different from how test suites work in the js world.

I'd suggest just one test to validate this feature:
- two existing contact methods in the initial SQL, A, and B, where B is disabled
- create two with GraphQL, C and D; verify C
- assert that only D is deleted after fast-forwarding

That will validate that existing contact methods aren't deleted and that new ones will expire & get cleaned up if not verified.

test/smoke/verifymanager_test.go
```diff
@@ -0,0 +1,177 @@
...
+			insert into user_contact_methods (id, user_id, name, type, value, disabled) 
+			values
+					({{uuid "cm1"}}, {{uuid "user"}}, 'personal1', 'SMS', {{phone "1"}}, true);
+			insert into user_verification_codes (id, contact_method_id, code, expires_at)
+			values
+				({{uuid "code"}}, {{uuid "cm1"}}, '1234', now() + '15 minutes'::interval);
+	`
+		h := harness.NewHarness(t, sql, "switchover-mk2")
+		defer h.Close()
+		h.Migrate("add-pending-to-contact-methods")
```
mastercactapus, line 92: `NewHarness` will already migrate to the newest version
```suggestion
```

migrate/migrations/20221228092646-add-pending-to-contact-methods.sql
```diff
@@ -0,0 +1,40 @@
+-- +migrate Up
+
+BEGIN;
+
+-- Add pending column with default to false for existing records and update default to true for new records later on
+ALTER TABLE user_contact_methods ADD COLUMN pending BOOLEAN NOT NULL DEFAULT FALSE;
+
+-- Update default to true for new records
```
mastercactapus, line 8: nit: we can describe the why above, the what is redundant and adds noise
```suggestion
```

mastercactapus (changes requested):

engine/verifymanager/db.go
```diff
@@ -50,11 +51,15 @@ func NewDB(ctx context.Context, db *sql.DB) (*DB, error) {
...
-				select id
+				select id, contact_method_id
 				from user_verification_codes
 				where now() >= expires_at
 				limit 100
 				for update skip locked
+			), _cms as (
+				delete from user_contact_methods cm
+				using rows
+				where cm.id = rows.contact_method_id and cm.pending = true
```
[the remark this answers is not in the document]
cmarquis: indeed it is

migrate/migrations/20221228092646-add-pending-to-contact-methods.sql
```diff
@@ -0,0 +1,40 @@
+-- +migrate Up
+
+BEGIN;
```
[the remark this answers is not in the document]
cmarquis: Got it, good to know.

test/smoke/verifymanager_test.go
cmarquis: I've refactored the test.

web/src/app/users/UserContactMethodList.tsx
```diff
@@ -148,6 +149,9 @@ export default function UserContactMethodList(
         </React.Fragment>
       )
     }
+    if (cm.pending) {
+      return `${cm.formattedValue} - this contact method will be automatically deleted if not verified`
+    }
```
KatieMSB, line 154: This warning also needs to be added for unverified webhook contact methods

KatieMSB (changes requested):

web/src/app/users/UserContactMethodList.tsx
cmarquis: Makes sense, added.

KatieMSB (dismissed):

test/smoke/verifymanager_test.go
```diff
@@ -0,0 +1,153 @@
...
+		insert into users (id, name, email) 
+		values 
+			({{uuid "user"}}, 'bob', 'joe');
+		insert into user_contact_methods (id, user_id, name, type, value, disabled) 
+		values
+				({{uuid "cm1"}}, {{uuid "user"}}, 'personal1', 'SMS', {{phone "1"}}, false),
+				({{uuid "cm2"}}, {{uuid "user"}}, 'personal2', 'SMS', {{phone "2"}}, true);
+	`
+	h := harness.NewHarness(t, sql, "")
+	defer h.Close()
```
mastercactapus, line 68: nit: move this up to the top of the test, with helpers below

test/smoke/verifymanager_test.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	verifyCM(t, cm3.CreateUserContactMethod.ID)
+	verifyCM(t, cm4.CreateUserContactMethod.ID)
+
+	tw := h.Twilio(t)
+	d1 := tw.Device(h.Phone("3"))
+	d2 := tw.Device(h.Phone("4"))
+
+	d1Msg := d1.ExpectSMS("verification")
+	_ = d2.ExpectSMS("verification")
```
mastercactapus, line 107: nit: this doesn't return an error, the `_ =` isn't necessary

```suggestion
	d2.ExpectSMS("verification")
```

test/smoke/verifymanager_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+	codeStr := strings.Map(func(r rune) rune {
+		if r >= '0' && r <= '9' {
+			return r
+		}
+		return -1
+	}, d1Msg.Body())
+
+	code, _ := strconv.Atoi(codeStr)
```
mastercactapus, line 116: Do you know if we need to convert this to an int? We can pass it as a string below (we're already asserting 0-9)

test/smoke/verifymanager_test.go
```diff
@@ -0,0 +1,153 @@
...
+						pending
+					}
+				}
+			}
+		`, h.UUID("user")), &d)
+	assert.Len(t, d.User.ContactMethods, 3)
+
+	expectedCMs := []string{h.UUID("cm1"), h.UUID("cm2"), cm3.CreateUserContactMethod.ID}
+	sort.Strings(expectedCMs)
+	actualCMs := []string{}
```
mastercactapus, line 147: nit: use var for empty declarations in Go
```suggestion
	var actualCMs []string
```

mastercactapus (changes requested):

KatieMSB (dismissed):

KatieMSB (dismissed):

mastercactapus (approved):

cmarquis:
@mastercactapus @KatieMSB had to remake the db-schema to fix build issues.

mastercactapus (approved):

PR #2768, opened 2023-01-03
feat: breaking out voice calls into a testable function

AlaricWhitney:

**Description:**
Broke out the building of a phone call from a large monolithic function into its own, smaller, and testable function.

Part of addressing https://github.com/target/goalert/issues/2576

**Out of Scope:**
changes to the voice xml building functionality.

**Screenshots:**
NA

NA

NA

**Additional Info:**
NA

### DIFF
--- a/notification/twilio/voice.go
+++ b/notification/twilio/voice.go
@@ -181,52 +181,11 @@ func (v *Voice) Send(ctx context.Context, msg notification.Message) (*notificati
 		"Type":   "TwilioVoice",
 	})
 
-	opts := &VoiceOptions{
-		ValidityPeriod: time.Second * 10,
-		CallbackParams: make(url.Values),
-		Params:         make(url.Values),
+	// prefix is placed here due to the cfg dependancy, making it less complex to test buildMessage
+	opts, err := buildMessage(fmt.Sprintf("Hello! This is %s", cfg.ApplicationName()), msg)
+	if err != nil {
+		return nil, err
 	}
-
-	prefix := fmt.Sprintf("This is %s", cfg.ApplicationName())
-
-	var message string
-	subID := -1
-	switch t := msg.(type) {
-	case notification.AlertBundle:
-		message = fmt.Sprintf("%s with alert notifications. Service '%s' has %d unacknowledged alerts.", prefix, t.ServiceName, t.Count)
-		opts.Params.Set(msgParamBundle, "1")
-		opts.CallType = CallTypeAlert
-	case notification.Alert:
-		if t.Summary == "" {
-			t.Summary = "No summary provided"
-		}
-		message = fmt.Sprintf("%s with an alert notification. %s.", prefix, t.Summary)
-		opts.CallType = CallTypeAlert
-		subID = t.AlertID
-	case notification.AlertStatus:
-		message = rmParen.ReplaceAllString(t.LogEntry, "")
-		message = fmt.Sprintf("%s with a status update for alert '%s'. %s", prefix, t.Summary, message)
-		opts.CallType = CallTypeAlertStatus
-		subID = t.AlertID
-	case notification.Test:
-		message = fmt.Sprintf("%s with a test message.", prefix)
-		opts.CallType = CallTypeTest
-	case notification.Verification:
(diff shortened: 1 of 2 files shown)


notification/twilio/voice.go
```diff
@@ -638,3 +597,56 @@ func (v *Voice) FriendlyValue(ctx context.Context, value string) (string, error)
 	}
 	return libphonenumber.Format(num, libphonenumber.INTERNATIONAL), nil
 }
+
+// buildMessage is a function that will build the VoiceOptions object with the proper message contents
+func buildMessage(prefix string, msg notification.Message) (opts *VoiceOptions, err error) {
+	if prefix == "" {
+		return nil, errors.New("No prefix provided")
```
cmarquis, line 604: Small change to lowercase the "no"

Also not sure where this falls in the call chain above as to what the error would be logged/outputted as but should it have a bit more context like `build message error: no prefix provided`?

cmarquis (changes requested):

notification/twilio/voice.go
AlaricWhitney: done.  changed to `buildMessage error: no prefix provided`

notification/twilio/voice.go
```diff
@@ -638,3 +597,56 @@ func (v *Voice) FriendlyValue(ctx context.Context, value string) (string, error)
 	}
 	return libphonenumber.Format(num, libphonenumber.INTERNATIONAL), nil
 }
+
+// buildMessage is a function that will build the VoiceOptions object with the proper message contents
+func buildMessage(prefix string, msg notification.Message) (opts *VoiceOptions, err error) {
```
mastercactapus, line 602: While having this logic moved out of `Send` makes it a bit easier to test, we still have a nested structure in and out along with the base64 encoding of the message AND rendering said message all in one place. That makes the test cases bulky and because they need to validate multiple concerns, hard to read/maintain.

If we break it into some smaller pieces around each concern, it will be easier to write (and, more importantly, read) test cases for them.

Let's try:
- one func, `buildMessage`, that takes the app name & `notification.Message` and outputs a string
- one method, `VoiceOptions.setMsgBody`, to handle encoding the message returned from `buildMessage` (this will be expanded in a future PR for the pause work)
- one method, `VoiceOptions.setMsgParams`, to set the remaining parameters given a `notification.Message`


```go
// buildMessage will generate a voice call message for the provided notification.Message.
func buildMessage(appName string, msg notification.Message) (string, error) {
	// ...
}

// setMsgBody will encode and set/update the message body parameter.
func (vo *VoiceOptions) setMsgBody(body string) {
	if vo.Params == nil {
		vo.Params = make(url.Values)
	}
	// Encode the body, so we don't need to worry about
	// buggy apps not escaping URL params properly.
	vo.Params.Set(msgParamBody, b64enc.EncodeToString([]byte(body)))
}

// setMsgParams will set parameters for the provided message.
func (vo *VoiceOptions) setMsgParams(n notification.Message) error {
	// ...
}
```

The `Send` function will then have something like:
```go
func (v *Voice) Send(// ...	
	opts := &VoiceOptions{
		ValidityPeriod: time.Second * 10,
	}

	err = opts.setMsgParams(msg)
	if err != nil {
		return err
	}

	msgBody, err := buildMessage(cfg.ApplicationName(), msg)
	if err != nil {
		return err
	}
	opts.setMsgBody(msgBody)
```
---
We'll be sacrificing the single switch statement for two. Still, I think it's necessary to have one testable function that takes in a `notification.Message` and outputs spoken text that's easy to validate.

The voice-equivalent of `renderAlertMessage` for SMS.

mastercactapus (dismissed):

cmarquis (approved):

notification/twilio/voice_test.go
```diff
@@ -1,16 +1,179 @@
...
 }
 
+func TestBuildMessage(t *testing.T) {
+	prefix := "This is GoAlert"
+	type mockInput struct {
+		prefix string
+		msg    notification.Message
+	}
+
+	testCases := map[string]struct {
```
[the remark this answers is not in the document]
AlaricWhitney: updated.

allending313 (approved):

PR #2774, opened 2023-01-10
ui/EscalationPolicyStep: Add support for webhook in escalation policy steps

nataliekorzh:
- [x] Identified the issue which this PR solves.
- [x] Read the [**CONTRIBUTING**](https://github.com/target/goalert/blob/master/CONTRIBUTING.md) document.
- [x] Code builds clean without any errors or warnings.
- [x] Added appropriate tests for any new functionality.
- [x] All new and existing tests passed.
- [x] Added comments in the code, where necessary.
- [x] Ran `make check` to catch common errors. Fixed any that came up.

**Description:**
Adds support for webhooks as a notification channel, available as an option for escalation policy steps. Webhooks are only available to the escalation policies they were created in.

**Which issue(s) this PR fixes:**
Closes #2691 

**Out of Scope:**
Future: look into better ways of differentiating webhook names as chips displayed in the UI

**Screenshots:**

Screen Shot 2023-01-10 at 3 58 19 PM

Screen Shot 2023-01-10 at 3 58 40 PM

**Describe any introduced API changes:**
Adds new queries for filtering and searching for webhooks.
```graphql
type Webhook {
  id: String!
  name: String!
}

input WebhookSearchOptions {
  first: Int = 15
  after: String = ""
  search: String = ""
  omit: [ID!]
  escalationPolicyID: ID
}

type WebhookConnection {
  nodes: [Webhook!]!
  pageInfo: PageInfo!
}

# Allows searching for webhooks.
webhooks(input: WebhookSearchOptions): WebhookConnection!

# Returns a single webhook with the given URL.
webhook(id: String!): Webhook
```

Issue #2691, opened 2022-11-18
webhook support for escalation policy steps


mastercactapus:
**What problem would you like to solve? Please describe:**
Non-user-specific webhook support.

**Describe the solution you'd like:**
Calling a webhook from an escalation policy step

**Describe alternatives you've considered:**
A workaround is to create a user with a webhook configured and assign them to the policy step.

mastercactapus:
this has been supported for a while now

PR #2809, opened 2023-02-02
feat: adding webhooks as a notification channel type

nataliekorzh:

**Description:**
Adds support for webhooks as a notification channel. 

Part of https://github.com/target/goalert/issues/2691
Splitting up #2774


app/app.go
```diff
@@ -245,7 +245,7 @@ func NewApp(c Config, db *sql.DB) (*App, error) {
 
 // WaitForStartup will wait until the startup sequence is completed or the context is expired.
 func (a *App) WaitForStartup(ctx context.Context) error {
-	return a.mgr.WaitForStartup(log.WithLogger(ctx, a.cfg.Logger))
+	return a.mgr.WaitForStartup(a.Context(ctx))
```
Forfold, line 248: I'm getting `app/app.go:248:32: a.Context undefined (type *App has no field or method Context)` when attempting to start the app, so I think this line should be reverted?

Forfold:
```suggestion
	return a.mgr.WaitForStartup(log.WithLogger(ctx, a.cfg.Logger))
```

nataliekorzh: Oops yeah I noticed that and meant to add it, nice catch 👍 thanks

Forfold (dismissed):

mastercactapus (changes requested):
For breaking this into pieces, can we put a feature flag around this? That way if one of the other PRs gets delayed, it doesn't block a release.

Just around the register sender bit and creating the notification channel.

app/app.go
```diff
@@ -245,7 +245,7 @@ func NewApp(c Config, db *sql.DB) (*App, error) {
 
 // WaitForStartup will wait until the startup sequence is completed or the context is expired.
 func (a *App) WaitForStartup(ctx context.Context) error {
-	return a.mgr.WaitForStartup(a.Context(ctx))
+	return a.mgr.WaitForStartup(log.WithLogger(ctx, a.cfg.Logger))
```
mastercactapus, line 248: Merge issue? `a.Context` should be used here

app/startup.go
```diff
@@ -70,7 +70,12 @@ func (app *App) startup(ctx context.Context) error {
 
 	app.initStartup(ctx, "Startup.Slack", app.initSlack)
 	app.notificationManager.RegisterSender(notification.DestTypeUserEmail, "smtp", email.NewSender(ctx))
-	app.notificationManager.RegisterSender(notification.DestTypeUserWebhook, "webhook", webhook.NewSender(ctx))
+	if expflag.ContextHas(ctx, expflag.WebhookNC) {
+		app.notificationManager.RegisterSender(notification.DestTypeUserWebhook, "userWebhook", webhook.NewSender(ctx))
+		app.notificationManager.RegisterSender(notification.DestTypeWebhook, "webhook", webhook.NewSender(ctx))
+	} else {
+		app.notificationManager.RegisterSender(notification.DestTypeUserWebhook, "webhook", webhook.NewSender(ctx))
```
mastercactapus, line 77: We shouldn't change the ID based on the flag since these are used for logging. Update the user webhook to `user-webhook` always, and then conditionally load the new one.

alert/alertlog/store.go
```diff
@@ -339,6 +341,7 @@ func (s *Store) logAny(ctx context.Context, tx *sql.Tx, insertStmt *sql.Stmt, id
 			case notification.DestTypeUserEmail:
 				r.subject.classifier = "Email"
 			case notification.DestTypeUserWebhook:
+			case notification.DestTypeWebhook:
```
mastercactapus, line 344: Go switch statements have an implicit `break`, so we need to use `fallthrough` here explicitly.
```suggestion
			case notification.DestTypeWebhook:
				fallthrough
			case notification.DestTypeUserWebhook:
```

mastercactapus (changes requested):
As I'm going through this, I wonder if we should qualify `Webhook` everywhere to reduce ambiguity; `nc-webhook` and `user-webhook` vs. `webhook` on its own for either one.

Thoughts?

AlaricWhitney:
Since there will be more than one type of Webhook, we should ensure that there's a prefix to help everyone know what kind of webhook it is.

nataliekorzh:
> As I'm going through this, I wonder if we should qualify `Webhook` everywhere to reduce ambiguity; `nc-webhook` and `user-webhook` vs. `webhook` on its own for either one.
> 
> Thoughts?

Just added the new `chanWebhook` type

assignment/targettype.go
```diff
@@ -61,6 +62,8 @@ func (tt *TargetType) UnmarshalText(data []byte) error {
 		*tt = TargetTypeNotificationChannel
 	case "slackChannel":
 		*tt = TargetTypeSlackChannel
+	case "webhook":
```
mastercactapus, line 65:
```suggestion
	case "chanWebhook":
```

assignment/targettype.go
```diff
@@ -111,6 +114,8 @@ func (tt TargetType) MarshalText() ([]byte, error) {
 		return []byte("notificationChannel"), nil
 	case TargetTypeSlackChannel:
 		return []byte("slackChannel"), nil
+	case TargetTypeChanWebhook:
+		return []byte("webhook"), nil
```
mastercactapus, line 118:
```suggestion
		return []byte("chanWebhook"), nil
```

app/startup.go
```diff
@@ -70,7 +70,10 @@ func (app *App) startup(ctx context.Context) error {
 
 	app.initStartup(ctx, "Startup.Slack", app.initSlack)
 	app.notificationManager.RegisterSender(notification.DestTypeUserEmail, "smtp", email.NewSender(ctx))
-	app.notificationManager.RegisterSender(notification.DestTypeUserWebhook, "webhook", webhook.NewSender(ctx))
+	app.notificationManager.RegisterSender(notification.DestTypeUserWebhook, "userWebhook", webhook.NewSender(ctx))
+	if expflag.ContextHas(ctx, expflag.ChanWebhook) {
+		app.notificationManager.RegisterSender(notification.DestTypeChanWebhook, "chanWebhook", webhook.NewSender(ctx))
```
mastercactapus, line 75: Since these strings are primarily used for logging, let's use a hyphen for scan-ability and to match the style for Twilio SMS vs voice of `type-subtype`

```suggestion
	app.notificationManager.RegisterSender(notification.DestTypeUserWebhook, "webhook-user", webhook.NewSender(ctx))
	if expflag.ContextHas(ctx, expflag.ChanWebhook) {
		app.notificationManager.RegisterSender(notification.DestTypeChanWebhook, "webhook-channel", webhook.NewSender(ctx))
```

mastercactapus (changes requested):

mastercactapus:
Trying to validate this functionally, I'm not able to create a webhook channel via the API -- I'm attempting to do so via creating an EP target:

image

nataliekorzh:
If I add the new `escalation/store.go` file originally included in #2774 webhooks can be created successfully via the API. I was planning to include that file in a future PR where I set up the backend for adding a webhook to an escalation policy step. This PR was just for setting webhooks up as a new notification channel type. Do you think it would be better to include that file in this PR and allow for the functionality now?

mastercactapus:
I see; having it here would certainly help validate the API and DB changes -- is it possible to include just the minimal API <-> DB bits? If not, we can leave it out of scope.

nataliekorzh:
Sure, just added it.

escalation/store.go
```diff
@@ -286,6 +287,22 @@ func (s *Store) _updateStepTarget(ctx context.Context, stepID string, tgt assign
 	return err
 }
 
+func (s *Store) newWebhook(ctx context.Context, tx *sql.Tx, webhookTarget assignment.Target) (assignment.Target, error) {
```
mastercactapus, line 290: nit: call this `chanWebhook` since it doesn't always create a new one

second: add a check here to reject with a validation error if the experimental flag isn't set. See `contactmethod.go:89` for an example

mastercactapus (approved):

AlaricWhitney (approved):

Issue #2712, opened 2022-11-30
slack: add DM contact method and use for linking


mastercactapus:
**What problem would you like to solve? Please describe:**
Linking and managing the association with a Slack account is complex and "magical" currently.
- The only ways to "link" a Slack user to GoAlert is by creating a test alert or having an admin use a special-case tool (`goalert-slack-email-sync`)
- The status of being linked is invisible
- Makes it difficult for `@` mentions to work for schedules
- Blocker for #1228 

**Describe the solution you'd like:**
Allow users to add a contact method for Slack and use this to identify actions coming from Slack.

**Describe alternatives you've considered:**
A suggested alternative was to add a `/goalert link` command, but this keeps the linking status invisible, and users/devs/admins are unable to remove the link (e.g., for testing).

**Additional context:**
Notes from the design session:
- need to keep `auth_subjects` in sync with Slack contact methods
- existing links must result in new, verified, contact methods
- existing flows should remain untouched for MVP

To be added:
  - New contact method type `SLACK_DM` to be added
  - Mimic phone number entry but use `@` instead of `+`
  - same format as channel notifications
  - Same validation & test flow
  - Use metadata column to store and display user name
  - store value as `team_id:user_id`
  - add `cm_id` as a nullable column to `auth_subjects` defaulting to null (on delete cascade)
  - create auth_subjects entry when creating new contact method (with cm_id set)
  - new engine `compat` module
  - create verified contact methods for providers starting with `slack:` and null `cm_id`, update `cm_id`

MVP+1:
- update to use CM as the source of truth for slack user lookup
- update slack username over time (maybe when a message is sent up to once per day?)

mastercactapus:
Last items for the finish line:
- [ ] #2824 needs to get merged
- [ ] Validate and fix any remaining issues with slack dms (e.g., feature flag-on)
- [ ] Remove the feature flag (always on)

PR #2815, opened 2023-02-10
slack: add UI for Slack DM contact method type (experimental)

mastercactapus:
**Description:**
This PR adds UI support for Slack DMs as a contact method type.

Part of #2712
This is a follow-up to #2805 

**Out of Scope:**
It only adds support in the UI for creating & deleting the new contact method type.

**Additional Info:**
Start with `make start EXPERIMENTAL=slack-dm` to create a Slack DM contact method.

image

### DIFF
--- a/web/src/app/users/UserContactMethodList.tsx
+++ b/web/src/app/users/UserContactMethodList.tsx
@@ -19,6 +19,7 @@ import AppLink from '../util/AppLink'
 import { styles as globalStyles } from '../styles/materialStyles'
 import { UserContactMethod } from '../../schema'
 import UserContactMethodCreateDialog from './UserContactMethodCreateDialog'
+import { useExpFlag } from '../util/useExpFlag'
 
 const query = gql`
   query cmList($id: ID!) {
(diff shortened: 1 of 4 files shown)


web/src/app/users/UserContactMethodList.tsx
```diff
@@ -102,6 +104,9 @@ export default function UserContactMethodList(
       },
     ]
 
+    // don't show send test for slack DMs if disabled
+    if (cm.type === 'SLACK_DM' && !hasSlackDM) return actions
```
Forfold, line 108: is the experimental flag needed here? could probably leave it out here

mastercactapus: If the flag is turned off after a user has added a Slack contact method, we don't want to show the Test/Verify actions because they will be broken; otherwise, it will generate failures/errors in the backend every time anyone tries to use them.

cmarquis (approved):

Forfold (dismissed):

Forfold (approved):

cmarquis (approved):

PR #2818, opened 2023-02-20
Feature/escalate via ivr

### LINKED ISSUE #2775: Alert Escalation via IVR and SMS
Currently, GoAlert supports alert escalation via web app only and IVR/SMS doesn't have the same options for the users to escalate the issues immediately. 

Users need to have `escalate` alerts like ack and close alert actions in the IVR and SMS(text). 4 is used to ACK and 6 is used to Close the alert. Probably, 5 can be used to escalate the alert to the next level. For SMS(Text), 1a - ack, 1c-close and 1e-escalate alert.

Lakshmi2107:
**Description:**
Currently, GoAlert supports alert escalation via web app only and IVR/SMS doesn't have the same options for the users to escalate the issues immediately. We can have escalated options for alerts like ack and close alert actions in the IVR and SMS(text). 4 is used to ack and 6 is used to close the alert. Probably, 5 can be used to escalate the alert to the next level. For SMS(Text), 1a - ack, 1c-close and 1e-escalate alert.

Fixes #2775 

In Voice calls and SMS, we will have an additional message saying press 5 to escalate and {code}e to escalate, respectively.

We need to think about introducing an already escalated option.


engine/engine.go
```diff
@@ -317,6 +317,10 @@ func (p *Engine) ReceiveSubject(ctx context.Context, providerID, subjectID, call
 		newStatus = alert.StatusActive
 	case notification.ResultResolve:
 		newStatus = alert.StatusClosed
+	case notification.ResultEscalate:
+		newStatus = alert.StatusTriggered
+		alertIDS := []int{cb.AlertID}
+		p.a.EscalateMany(ctx, alertIDS)
```
mastercactapus, line 323: Performing the escalation and returning here may make more sense since the code below is for status updates.

The escalation action will handle updating the status automatically elsewhere.

Lakshmi2107: @mastercactapus thanks for the comments. I have updated the code. Since we escalate only one alert at a time, I have assigned cb.AlertID the updated alert id (alertIds[0]).

engine/engine.go
```diff
@@ -317,6 +317,13 @@ func (p *Engine) ReceiveSubject(ctx context.Context, providerID, subjectID, call
 		newStatus = alert.StatusActive
 	case notification.ResultResolve:
 		newStatus = alert.StatusClosed
+	case notification.ResultEscalate:
+		newStatus = alert.StatusTriggered
+		alertIds, err := p.a.EscalateMany(ctx, []int{cb.AlertID})
+		if err != nil {
+			log.Logf(ctx, "Error while escalating "+err.Error())
+		}
+		cb.AlertID = alertIds[0]
```
mastercactapus, line 326: We actually don't want to do any status updates here, and just leave it at escalation

```suggestion
		_, err := p.a.EscalateMany(ctx, []int{cb.AlertID})
		if err != nil {
			return fmt.Errorf("escalate alert: %w", err)
		}
		return nil
```

notification/twilio/twiml.go
```diff
@@ -48,6 +48,7 @@ const (
 	optionCancel
 	optionConfirmStop
 	optionAck
+	optionEsc
```
mastercactapus, line 51: nit: let's spell this out since `esc` is a little ambiguous without context
```suggestion
	optionEscalate
```

notification/twilio/voice.go
```diff
@@ -54,6 +54,7 @@ const (
 	digitConfirm  = "3"
 	digitOldAck   = "8"
 	digitOldClose = "9"
+	digitEsc      = "5"
```
mastercactapus, line 57: Same, let's spell out escalate so it's not confused as an "escape digit" or something like that
```suggestion
	digitEscalate      = "5"
```

mastercactapus (changes requested):

engine/engine.go
Lakshmi2107: Got it, updated the code. Thanks.

mastercactapus:
This is in a fairly good state. However, it may be good to make some adjustments for escalations to detect when an alert is already closed since, without it, this will reply "alert escalated" even after the alert is closed.

I'll open a separate PR for fixing that, but for this PR:
- The `Receive` handler: it's missing the escalation case, so this doesn't work yet when I tested locally. It should be the same as `ReceiveSubject`
- The reply message should probably be "escalation requested" since escalations are always handled async

Lakshmi2107:
@mastercactapus thanks for ur comments. I have pushed the changes. Please review.

mastercactapus:
@Lakshmi2107 hey, I had a little time today to update this with master, and made some changes based on the escalation checks PR I recently merged.

However, a couple of the unit tests still need to be updated (alertsms_test.go, for the changes to the message) before we can get this in.

The expected text will need to be updated with the new options, and part of the summary will need to be truncated a bit more (since it leaves less space for the rest of the text).

Let me know if you need help with this, you can run all the preflight checks with `make check` and just the unit tests with `make test`.

Lakshmi2107:
@mastercactapus thanks a lot. I have modified the unit tests and pushed the changes. Please review.

mastercactapus (approved):

KatieMSB (approved):

PR #2825, opened 2023-02-27
feat: creating chanWebhook store and endpoints for searching

nataliekorzh:

**Description:**
This PR sets up the store for the new chanWebhook notification channel type and adds endpoints for searching for chanWebhooks. 

Part of https://github.com/target/goalert/issues/2691
Splitting up #2774


graphql2/schema.graphql
```diff
@@ -90,6 +90,12 @@ type Query {
   # Allows searching for user overrides.
   userOverrides(input: UserOverrideSearchOptions): UserOverrideConnection!
 
+  # Allows searching for webhook notification channels.
+  chanWebhooks(input: ChanWebhookSearchOptions): ChanWebhookConnection!
```
mastercactapus, line 94: Can you give some context regarding the use case for a global webhook search?

We _do_ have something similar for Slack channels, but those are a special case, and the search is fed by the Slack API and not the `notification_channels` table.

nataliekorzh: The global webhook search is used to gather options to populate the webhook select component that opens up when a user tries to add a webhook to an escalation policy step. In practice/the UI the search is limited to webhooks previously added to that escalation policy step only.

mastercactapus: Since we wouldn't want (nor allow) duplicates on a single step and, in practice, it's limited to that step, could it be left out in favor of populating via the step query?

I may be misunderstanding

nataliekorzh: Just removed `chanWebhook()` and `chanWebhooks()` search from this PR

PR #2827, opened 2023-02-28
ui/time: use single Time component for displaying timestamps and relative times

mastercactapus:
**Description:**
This PR adds a new `<Time>` component to handle formatting various timestamps throughout the application.

**Out of Scope:**
A couple of places have been left out to aid in review, due to how much code change is required to clean them up.
- Schedule calendar
- Temp schedules
- A `TimeRange` component

**Screenshots**
image

Experience should be largely the same, though with minor changes due to the added consistency of using a single formatting function, and using the browsers configured locale instead of our custom logic.


mastercactapus:
This has been merged in pieces.

PR #2830, opened 2023-03-01
ui: add Time component

mastercactapus:
**Description:**
This PR adds a new `<Time>` component to render various timestamps and durations throughout the application.

**Out of Scope:**
Only a few call sites have been updated to keep the PR focused on the new component.

**Screenshots**
An example of the rendered times (this change to rotations is **_not_** in this PR)
image

Experience should be the same, with minor changes due to the added consistency of using a single formatting function and the browsers configured locale instead of our custom logic.

**Additional Context**
Places updated to use the new component for validating this PR:
- Alert details event log timestamps
- Alert list timestamps
- Heartbeat monitor timeout display
- Heartbeat monitor status display
- User session list

### DIFF
--- a/web/src/app/util/Time.tsx
+++ b/web/src/app/util/Time.tsx
@@ -67,6 +67,7 @@ const TimeTimestamp: React.FC<TimeTimestampProps> = (props) => {
 type TimeDurationProps = TimeBaseProps & {
   precise?: boolean
   duration: DurationLikeObject | string | Duration
+  units?: readonly (keyof DurationLikeObject)[]
 }
 
 const TimeDuration: React.FC<TimeDurationProps> = (props) => {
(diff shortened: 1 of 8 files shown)


web/src/app/util/Time.tsx
```diff
@@ -0,0 +1,109 @@
...
+type TimeTimestampProps = TimeBaseProps &
+  Omit<FormatTimestampArg, 'time'> & {
+    time: string | null | undefined
+    zero?: string
+  }
+
+const TimeTimestamp: React.FC<TimeTimestampProps> = (props) => {
+  const [, setTS] = useState('') // force re-render
+  useEffect(() => {
+    if (!['relative', 'relative-date'].includes(props.format || '')) return
```
Forfold, line 25: Can we simplify this, or are strings with xyz_relative expected?

```suggestion
    if (props.format !== 'relative' || props.format !== 'relative-date' || !props.format) {
      return
    }
```

Forfold (approved):

web/src/app/util/Time.tsx
mastercactapus: It's testing if format === relative or relative-date (not substrings)

KatieMSB (approved):

PR #2853, opened 2023-03-10
ui/ServiceLink: convert to typescript

NickHackman:

**Description:**

### DIFF
--- a/web/src/app/alerts/components/AlertDetails.tsx
+++ b/web/src/app/alerts/components/AlertDetails.tsx
@@ -355,7 +355,7 @@ export default function AlertDetails(props: AlertDetailsProps): JSX.Element {
   return (
     <Grid container spacing={2}>
       <ServiceMaintenanceNotice
-        serviceID={props.data?.service?.id ?? ''}
+        serviceID={alert?.service?.id ?? ''}
         extraNotices={alert.pendingNotifications.map((n) => ({
           type: 'WARNING',
           message: `Notification Pending for ${n.destination}`,
--- a/web/src/app/alerts/components/AlertDetails.tsx
+++ b/web/src/app/alerts/components/AlertDetails.tsx
@@ -369,11 +369,13 @@ export default function AlertDetails(props: AlertDetailsProps): JSX.Element {
         <Card sx={{ width: '100%' }}>
           <CardContent data-cy='alert-summary'>
             <Grid container spacing={1}>
-              <Grid item xs={12}>
-                <Typography variant='body1'>
-                  {ServiceLink(alert.service)}
-                </Typography>
-              </Grid>
+              {alert.service && (
+                <Grid item xs={12}>
+                  <Typography variant='body1'>
+                    {ServiceLink(alert.service)}
+                  </Typography>
+                </Grid>
+              )}
               <Grid item xs={12}>
                 <Typography component='h2' variant='h5'>
                   {alert.alertID}: {alert.summary}
--- a/web/src/app/links/ServiceLink.tsx
+++ b/web/src/app/links/ServiceLink.tsx
@@ -1,0 +1,9 @@
+import React from 'react'
+import { Service } from '../../schema'
+import AppLink from '../util/AppLink'
+
+export const ServiceLink = (
+  service: Service | null | undefined,
+): JSX.Element => {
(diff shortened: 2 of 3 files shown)


web/src/app/links/ServiceLink.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+import { Service } from '../../schema'
+import AppLink from '../util/AppLink'
+
+export const ServiceLink = (
+  service: Service | null | undefined,
+): JSX.Element => {
+  return <AppLink to={`/services/${service?.id}`}>{service?.name}</AppLink>
```
NickHackman, line 8: Service is nullable and potentially undefined in its use in AlertDevice

https://github.com/target/goalert/blob/a9d8684de6f59b7349f3f8cf98aed7ee6cbb5ef7/web/src/app/alerts/components/AlertDetails.tsx#L374

Handling the nullability/undefined here results in potentially invalid or dead links. Is there a preferred method here? Could always conditionally render the AppLink 🤔

mastercactapus: Yeah, I think the `ServiceLink` component should require it to be non-null as it's impossible to render a link to a service without its info. The JS version would have just thrown an error and broken the page, so the type info should reflect its requirement.

The call site rendering has a potential bug; conditionally rendering it is fine.

That said, I *think* having it undefined and throwing an error would have broken the page, so it's likely the data is guaranteed by some previous loading check anyhow. If that is the case, there might be a way to have the type information reflect that it's not nullable in practice.

NickHackman: Went up to the Page level and found the assertion being made on `alert` being present

https://github.com/target/goalert/blob/38d5a2c0e27263ed695790d76359b294b3e8332e/web/src/app/alerts/pages/AlertDetailPage.js#L52

I couldn't find an assertion on `service` being present, the nullability/undefined is being handled

https://github.com/target/goalert/blob/077d5091f4a01613c61bf7fbf1177dcdd063e1a7/web/src/app/alerts/components/AlertDetails.tsx#L160

https://github.com/target/goalert/blob/077d5091f4a01613c61bf7fbf1177dcdd063e1a7/web/src/app/alerts/components/AlertDetails.tsx#L331

https://github.com/target/goalert/blob/077d5091f4a01613c61bf7fbf1177dcdd063e1a7/web/src/app/alerts/components/AlertDetails.tsx#L358

From here, conditionally rendering makes the most sense to me 👍

NickHackman: Conditional rendering done in https://github.com/target/goalert/pull/2853/commits/42dd1360c5404718f48c2a4798864853a0b12586

web/src/app/alerts/components/AlertDetails.tsx
```diff
@@ -355,7 +355,7 @@ export default function AlertDetails(props: AlertDetailsProps): JSX.Element {
   return (
     <Grid container spacing={2}>
       <ServiceMaintenanceNotice
-        serviceID={props.data?.service?.id ?? ''}
+        serviceID={alert?.service?.id ?? ''}
```
NickHackman, line 358: Super minor refactor: above the props are destructured and renamed to `alert`

web/src/app/alerts/components/AlertDetails.tsx
```diff
@@ -369,11 +369,13 @@ export default function AlertDetails(props: AlertDetailsProps): JSX.Element {
         <Card sx={{ width: '100%' }}>
           <CardContent data-cy='alert-summary'>
             <Grid container spacing={1}>
-              <Grid item xs={12}>
-                <Typography variant='body1'>
-                  {ServiceLink(alert.service)}
-                </Typography>
-              </Grid>
+              {alert.service && (
```
NickHackman, line 372: From my understanding of JS, the `&&` should be fine here since Objects are never falsy

docs: https://developer.mozilla.org/en-US/docs/Glossary/Falsy

this should avoid the common gotcha with `&&` and conditional rendering. Since undefined and null won't render and `alert.service` is always truthy

https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator

mastercactapus (approved):

KatieMSB (approved):

PR #2855, opened 2023-03-10
ui/schedules: Convert ScheduleShiftList to ts

andrewbenington:

**Description:**
- Converted the `ScheduleShiftList` component to TypeScript
  - Added the `truncated` field to the `Shift` object

Part of [2318](https://github.com/target/goalert/issues/2318)


web/src/app/schedules/ScheduleShiftList.tsx
```diff
@@ -171,7 +179,7 @@ function ScheduleShiftList({ scheduleID }) {
           shiftDetails = `Active after ${startTime}`
         }
         result.push({
-          title: s.userName,
+          title: s.user?.name || '',
```
andrewbenington, line 182: userName is not an attribute of a Shift in the Go app, so I do this instead for consistency (see the change on line 115 of the old version/124 of the new version)

mastercactapus: `s.user!.name` could be used instead since we know it will be set.

We can do the same for `user.id` I think too

mastercactapus (approved):

KatieMSB (approved):

PR #2871, opened 2023-03-15
feat: add UI for chanWebhooks escalation policy step option

nataliekorzh:

**Description:**
This PR adds UI support for adding chanWebhooks (webhook notification channels) to escalation policy steps.

Part of #2691 
This is a follow-up to #2809 

**Additional Info:**
When the escalation policy step create dialog is opened for the first time, a bad `setState()` call warning is rendered. This is from the feature flag updating, so it won't be relevant once this PR is approved and the feature flag is removed.

**Screenshots:**
Screenshot 2023-03-15 at 4 01 06 PM
Screenshot 2023-03-15 at 3 51 20 PM


graphql2/graphqlapp/escalationpolicy.go
```diff
@@ -55,6 +57,14 @@ func (m *Mutation) CreateEscalationPolicyStep(ctx context.Context, input graphql
 		)
 	}
 
+	if len(input.Targets) != 0 {
```
mastercactapus, line 60: This isn't necessary as `for ... range` already handles nil/empty slice correctly

nataliekorzh: Got it, just removed it

cmarquis (dismissed):

web/src/app/links/ChanWebhookLink.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+import AppLink from '../util/AppLink'
+
+export const ChanWebhookLink = (webhook: {
+  id: string
+  name: string
+}): JSX.Element => {
+  return <AppLink to={webhook.id}>{webhook.name}</AppLink>
```
mastercactapus, line 8: We can't/shouldn't link to the webhook -- it's an API target, rather than a UI landing page. We'd only put a link here if we had something like a details page within GoAlert for it.

web/src/app/selection/ChanWebhookSelect.tsx
```diff
@@ -0,0 +1,91 @@
...
+  } catch {
+    return false
+  }
+}
+
+export const ChanWebhookSelect = (
+  props: ChanWebhookSelectProps,
+): JSX.Element => {
+  const [newURL, setNewURL] = useState<string>('')
+  const { value, onChange = () => {} } = props
```
mastercactapus, line 25: nit: I think we can omit the default since it's listed as a required prop above

web/src/app/selection/ChanWebhookSelect.tsx
```diff
@@ -0,0 +1,91 @@
...
+          variant='outlined'
+          fullWidth
+          name='webhooks'
+          value={newURL}
+          onChange={(e) => {
+            setNewURL(e.target.value.trim())
+          }}
+          error={
+            (!isValidURL(newURL) && newURL.length > 0) ||
+            value.indexOf(newURL) > -1
```
mastercactapus, line 54: nit: use `.includes` instead
```suggestion
            value.includes(newURL)
```

mastercactapus (changes requested):

cmarquis (dismissed):

mastercactapus (approved):

PR #2880, opened 2023-03-20
ui: fix snackbar not showing latest alert notification

Forfold:
On the alerts list, if multiple updates are made, only the first snackbar notification appears then it glitches out.

This change uses context at the top level in `App.js` to always shows the latest notification, pushing new notification data to the context on mutation `onComplete`s/`onError`s.

the bug:
https://youtu.be/CEMMgKroCQg


KatieMSB (changes requested):
Looks like some cypress tests need updated, seeing it failing at `alerts.ts - should acknowledge, escalate, and close multiple alerts`

Forfold:
@KatieMSB that test should be good now. I had to add the onClose clickaway handler to the snackbar component as well again

KatieMSB (approved):

PR #2891, opened 2023-03-23
add notices to services when nearing unackd alerts system limit

### LINKED ISSUE #2869: make active alert cap more obvious
There is a system-wide limit (admin configurable) on the number of active/unacknowledged alerts per-service. When a service hits this limit, it isn't obvious outside of the API call that there could be an issue creating new alerts.

A notice on a service, and any alert details pages, when a service is at or near the limit.
- @ >= 90% display as warning
- @ >= 100% display as error (since no new alerts will be accepted)

The system works as intended currently, but with the cost of admin and end-user time to determine when this happens.

Forfold:
Closes #2869

This PR introduces the notices API to a `service` (see escalation policies for a notice example). Adds it to the graphql api, fetches the data using sqlc, and shows the data using urql and a `ServiceNotices` component.

Screenshots (using a system limit of 10 to functionally test):

Screenshot 2023-03-23 at 10 03 58 AM

Screenshot 2023-03-23 at 10 04 52 AM

Screenshot 2023-03-23 at 10 05 05 AM


notice/store.go
```diff
@@ -78,3 +83,58 @@ func (s *Store) FindAllPolicyNotices(ctx context.Context, policyID string) ([]No
...
+	if err != nil {
+		return nil, err
+	}
+
+	count, err := gadb.New(s.db).CountUnackedAlertsByService(ctx, uuid.MustParse(serviceID))
+	if err != nil {
+		return nil, err
+	}
+
+	ls, err := limit.NewStore(ctx, s.db)
```
mastercactapus, line 105: Don't create a new store -- this will be inefficient and prepare all statements for every invocation. Pass the existing store, or add a query to `queries.sql`.

Maybe update the query for the use case:
```sql
-- name: NoticeUnackedAlertsByService :one
SELECT
    count(*),
    (
        SELECT
            max
        FROM
            config_limits
        WHERE
            id = 'unacked_alerts_per_service')
FROM
    alerts
WHERE
    service_id = $1::uuid
    AND status = 'triggered';

```

notice/store.go
```diff
@@ -78,3 +83,58 @@ func (s *Store) FindAllPolicyNotices(ctx context.Context, policyID string) ([]No
...
+	// hit system limit notice
+	if count >= unackedLim {
+		return []Notice{
+			{
+				Type:    TypeError,
+				Message: "Unacknowledged alert limit reached",
+				Details: details,
+			},
+		}, nil
+	}
```
mastercactapus, line 137: nit: do the error case first, and then drop the `count < unackedLim` from the warning check.

mastercactapus (changes requested):

notice/store.go
mastercactapus: ^ that way we only make a single call to the DB and it would simplify the code in the notice package.

mastercactapus (approved):

KatieMSB (approved):

PR #2898, opened 2023-03-24
feat: adding error handlers to devtools

AlaricWhitney:

**Description:**
Adding error handling to the methods within the devtools folder

**Out of Scope:**
modifying functionality, or adding complexity.

**Screenshots:**
NA

NA

NA

**Additional Info:**
NA


devtools/runproc/process.go
```diff
@@ -169,7 +169,11 @@ func (p *Process) run() error {
 		p.cmd.Stdin = tty
 		err = p.cmd.Start()
 		if err == nil {
-			go io.Copy(p.p, p.pty)
+			go func(p *Process) {
+				if _, err := io.Copy(p.p, p.pty); err != nil {
+					p.logError(err)
+				}
+			}(p)
```
mastercactapus, line 176: Passing the *Process is redundant since we're in a method handler, and `p` doesn't get re-assigned. However, `p.p` and `p.pty` can be, so this could be problematic in a background thread.

How about a `copyIO` method that doesn't return an error to capture the Reader and Writer?

```go
func (p *Process) copyIO(dst io.Writer, src io.Reader) {
	err := io.Copy(dst, src)
	if err != nil {
		p.logError(err)
	}
}
```

And then use like
```go
go p.copyIO(p.p, p.pty)
```

The receiver allows using `logError` but the value of `p.p` and `p.pty` gets captured synchronously, preventing race conditions with the background goroutine.

mastercactapus (approved):

PR #2899, opened 2023-03-24
feat: adding error checks to tests folder

AlaricWhitney:

**Description:**
Adding error checks to the methods inside the `tests` folder.

**Out of Scope:**
adding any complexity to the methods

**Screenshots:**
NA

NA

NA

**Additional Info:**
NA

### DIFF
--- a/test/smoke/harness/harness.go
+++ b/test/smoke/harness/harness.go
@@ -314,7 +314,9 @@ func (h *Harness) Start() {
 	h.TwilioNumber("") // register default number
 	h.slack.SetActionURL(h.slackApp.ClientID, h.backend.URL()+"/api/v2/slack/message-action")
 
-	go h.backend.Run(context.Background())
+	go func(h *Harness) {
+		_ = h.backend.Run(context.Background()) // can't pass to test fatal on error
+	}(h)
 	err = h.backend.WaitForStartup(ctx)
 	if err != nil {
 		h.t.Fatalf("failed to start backend: %v", err)
@@ -507,7 +509,7 @@ func (h *Harness) AddNotificationRule(userID, cmID string, delayMinutes int) {
 func (h *Harness) Trigger() {
 	id := h.backend.Engine.NextCycleID()
 	go h.backend.Engine.Trigger()
-	h.backend.Engine.WaitCycleID(context.Background(), id)
+	_ = h.backend.Engine.WaitCycleID(context.Background(), id)
 }
 
 // Escalate will escalate an alert in the database, when 'level' matches.
@@ -549,7 +551,9 @@ func (h *Harness) dumpDB() {
 	if err != nil {
 		h.t.Fatalf("failed to get abs dump path: %v", err)
 	}
-	os.MkdirAll(filepath.Dir(file), 0o755)
+	if err := os.MkdirAll(filepath.Dir(file), 0o755); err != nil {
+		h.t.Fatalf("failed to create abs dump path: %v", err)
+	}
 
 	conn, err := pgx.Connect(context.Background(), h.dbURL)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


test/smoke/harness/harness.go
```diff
@@ -507,7 +509,7 @@ func (h *Harness) AddNotificationRule(userID, cmID string, delayMinutes int) {
 func (h *Harness) Trigger() {
 	id := h.backend.Engine.NextCycleID()
 	go h.backend.Engine.Trigger()
-	h.backend.Engine.WaitCycleID(context.Background(), id)
+	_ = h.backend.Engine.WaitCycleID(context.Background(), id)
```
mastercactapus, line 512: Since this is for tests, maybe we should `assert.NoError` here

test/smoke/harness/harness.go
```diff
@@ -314,7 +314,9 @@ func (h *Harness) Start() {
 	h.TwilioNumber("") // register default number
 	h.slack.SetActionURL(h.slackApp.ClientID, h.backend.URL()+"/api/v2/slack/message-action")
 
-	go h.backend.Run(context.Background())
+	go func(h *Harness) {
+		_ = h.backend.Run(context.Background()) // can't pass to test fatal on error
```
mastercactapus, line 318: Should we `assert.NoError` here too?

test/smoke/harness/harness.go
```diff
@@ -314,7 +314,9 @@ func (h *Harness) Start() {
 	h.TwilioNumber("") // register default number
 	h.slack.SetActionURL(h.slackApp.ClientID, h.backend.URL()+"/api/v2/slack/message-action")
 
-	go h.backend.Run(context.Background())
+	go func(h *Harness) {
+		require.NoError(h.t, h.backend.Run(context.Background())) // can't pass to test fatal on error
```
mastercactapus, line 318: We can't use `require` here because it calls `FailNow` which can't be called from a separate goroutine, use `assert`.

https://pkg.go.dev/testing#T.FailNow

https://medium.com/freethreads/golang-testify-require-vs-assert-b3bbfb4e0b8f

test/smoke/migrations_test.go
```diff
@@ -550,7 +550,7 @@ func TestMigrations(t *testing.T) {
 	if err != nil {
 		t.Fatal("failed to create db:", err)
 	}
-	defer db.Exec("drop database " + sqlutil.QuoteID(dbName))
+	defer func(db *sql.DB) { _, _ = db.Exec("drop database " + sqlutil.QuoteID(dbName)) }(db)
```
mastercactapus, line 553: No need to pass it in, we can reference it from the outer scope.
- we're not in a loop
- it doesn't get reassigned

```suggestion
	defer func() { _, _ = db.Exec("drop database " + sqlutil.QuoteID(dbName)) }()
```

mastercactapus (changes requested):

test/smoke/harness/harness.go
AlaricWhitney: done.  I just used require to be consistent with the rest of the method.

test/smoke/migrations_test.go
```diff
@@ -550,7 +550,7 @@ func TestMigrations(t *testing.T) {
 	if err != nil {
 		t.Fatal("failed to create db:", err)
 	}
-	defer db.Exec("drop database " + sqlutil.QuoteID(dbName))
+	defer func() { _, _ = db.Exec("drop database " + sqlutil.QuoteID(dbName)) }()
```
andrewbenington, line 553: Should this use fmt.Sprintf?

mastercactapus (changes requested):

andrewbenington (approved):
LGTM

mastercactapus (approved):

PR #2900, opened 2023-03-24
feat: adding error checks to notification methods

AlaricWhitney:

**Description:**
Adding error checks to methods inside of the `notification` folder

**Out of Scope:**
adding complexity to existing methods

**Screenshots:**
NA

NA

NA

**Additional Info:**
NA

### DIFF
--- a/notification/slack/servemessageaction.go
+++ b/notification/slack/servemessageaction.go
@@ -118,19 +118,20 @@ func (s *ChannelSender) ServeMessageAction(w http.ResponseWriter, req *http.Requ
 	case alertCloseActionID:
 		res = notification.ResultResolve
 	case linkActActionID:
-		if err = s.withClient(ctx, func(c *slack.Client) error {
+		err = s.withClient(ctx, func(c *slack.Client) error {
 			// remove ephemeral 'Link Account' button
 			_, err = c.PostEphemeralContext(ctx, payload.Channel.ID, payload.User.ID,
 				slack.MsgOptionText("", false), slack.MsgOptionReplaceOriginal(payload.ResponseURL),
 				slack.MsgOptionDeleteOriginal(payload.ResponseURL))
 			if err != nil {
-				return err
+				return fmt.Errorf("delete ephemeral message: %w", err)
 			}
 			return nil
-		}); err != nil {
-			errutil.HTTPError(ctx, w, validation.NewFieldErrorf("action_id", "issue with link action on action ID '%s'", act.ActionID))
+		})
+		if errutil.HTTPError(ctx, w, err) {
 			return
 		}
+
 		return
 	default:
 		errutil.HTTPError(ctx, w, validation.NewFieldErrorf("action_id", "unknown action ID '%s'", act.ActionID))
--- a/notification/slack/servemessageaction.go
+++ b/notification/slack/servemessageaction.go
@@ -118,7 +118,7 @@ func (s *ChannelSender) ServeMessageAction(w http.ResponseWriter, req *http.Requ
 	case alertCloseActionID:
 		res = notification.ResultResolve
 	case linkActActionID:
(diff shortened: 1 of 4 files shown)


notification/slack/servemessageaction.go
```diff
@@ -127,7 +127,9 @@ func (s *ChannelSender) ServeMessageAction(w http.ResponseWriter, req *http.Requ
 				return err
 			}
 			return nil
-		})
+		}); err != nil {
+			errutil.HTTPError(ctx, w, validation.NewFieldErrorf("action_id", "issue with link action on action ID '%s'", act.ActionID))
```
mastercactapus, line 131: Failing to post to Slack on a separate request shouldn't result in a Bad Request error to the original callback request. In this case, the request was valid, and we should log the error. Deleting the ephemeral message isn't super important, as at this point account linking already succeeded.

notification/twilio/validation.go
```diff
@@ -13,7 +13,9 @@ import (
 
 func validateRequest(req *http.Request) error {
 	if req.Method == "POST" {
-		req.ParseForm()
+		if err := req.ParseForm(); err != nil {
+			return errors.New("unable to parse form input")
+		}
```
mastercactapus, line 18: If Twilio sends an invalid payload, we should probably return a validation error (400). This will result in a 500 error

mastercactapus (changes requested):

notification/slack/servemessageaction.go
AlaricWhitney: also given the rest of this function returns an http response code on failure, I'm just trying to follow the established pattern of what this method does.

mastercactapus: If we specifically can't delete the ephemeral message, that's not a fatal error here, as the account linking is 100% on the GoAlert side (save for the response).

The rest of the function initiates the account-linking process, and if that fails, then the intended click action has failed. The only purpose for this piece is a best-effort cleanup since the link can only be used once.

That said, if we abort here, it would be safe to re-try clicking the button anyway.

---
For some added context:

When this code was first written, it was decided to intentionally allow things to continue if this call fails (i.e., worst case, the button remains after clicking but linking still works) since it doesn't impact the ability to act in any way.

The fact that the error was never logged, though, was an oversight. As well as not including a *why* comment. I'll be okay with either logging it or returning it.

In any case, I agree with your point of making the problem visible: returning the error can always guarantee consistent behavior (message deleted) or no linking possible.

notification/twilio/validation.go
mastercactapus: That was my mistake; I'm used to needing to wrap with our validation package for the catch-all everywhere else.

notification/slack/servemessageaction.go
mastercactapus: ^ for the return route, put it in the format:
```go
err = s.withClient(...)
if errutil.HTTPError(ctx, err) {
	return
}
```

It looks like what you currently have here was copy-pasted and forgot to fix it

AlaricWhitney: added the extra return.  I didn't add it as the very next line is `return` anyways.

mastercactapus: That wasn't quite the issue. Some pointers:
- Where possible use `if errutil.HTTPError(ctx, w, err) {` with as little noise as possible (e.g., similar to `if err != nil {`)
- Put the context in the returning func where possible (instead of in the call to HTTPError)
- Only return validation errors for validation problems, not server errors.
 
In this case, `validation.NewFieldErrorf` is the wrong thing to use, and we'd lose all the error information from our logs, returning a 400 to slack instead of a 500. We'd only want to use that if there was an issue with the `action_id`.

---

I'll push up a commit with the issue addressed to speed this along.

allending313 (approved):
LGTM

PR #2907, opened 2023-03-27
slack: add user group queries to graphql schema

mastercactapus:
**Description:**
This PR adds the necessary GraphQL queries for selecting and searching Slack user groups.

**Additional Info**
- Start with `make start EXPERIMENTAL=slack-ug`
- `slackUserGroups { nodes { id, name, handle } }`
- `slackUserGroup(id: "id-here") { id, name handle }`

### DIFF
--- a/graphql2/models_gen.go
+++ b/graphql2/models_gen.go
@@ -485,6 +485,18 @@ type SlackChannelSearchOptions struct {
 	Omit   []string `json:"omit"`
 }
 
+type SlackUserGroupConnection struct {
+	Nodes    []slack.UserGroup `json:"nodes"`
+	PageInfo *PageInfo         `json:"pageInfo"`
+}
+
+type SlackUserGroupSearchOptions struct {
+	First  *int     `json:"first"`
+	After  *string  `json:"after"`
+	Search *string  `json:"search"`
+	Omit   []string `json:"omit"`
+}
+
 type StringConnection struct {
 	Nodes    []string  `json:"nodes"`
 	PageInfo *PageInfo `json:"pageInfo"`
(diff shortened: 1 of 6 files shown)


graphql2/models_gen.go
```diff
@@ -485,6 +485,18 @@ type SlackChannelSearchOptions struct {
 	Omit   []string `json:"omit"`
 }
 
+type SlackUserGroupConnection struct {
+	Nodes    []slack.UserGroup `json:"nodes"`
+	PageInfo *PageInfo         `json:"pageInfo"`
```
AlaricWhitney, line 490: Target standards for json is to use snake case, not camelCase.  Is this json for our app, or how we interact with Slack?

https://apiplatform-docsite.prod.target.com/docs/customer-engagement/api-design-standards/#use-snake_case-not-camelcase

AlaricWhitney (changes requested):

graphql2/models_gen.go
mastercactapus: This is generated code and based on the schema

graphql2/schema.graphql
```diff
@@ -220,6 +226,24 @@ type MessageLogConnection {
...
+
+type SlackUserGroup {
+  id: ID!
+  name: String!
+  handle: String!
+}
+
+type SlackUserGroupConnection {
+  nodes: [SlackUserGroup!]!
+  pageInfo: PageInfo!
```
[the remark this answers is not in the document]
mastercactapus: While GraphQL technically uses JSON as a transport, it is an independent standard to which camelCase is the established convention.

This schema is only used for GraphQL and has no REST-style endpoints.

Forfold (approved):

PR #2931, opened 2023-04-10
yarn: use version 3.5.0 with corepack and pnp

mastercactapus:
**Description:**
Updates yarn usage to be managed by `corepack` (node >= 16.10) and using the pnp install method.
- no more `node_modules`
- much faster install times
- fixes issues where incorrect local yarn version would cause issues


cmarquis (dismissed):

cmarquis (dismissed):

tsconfig.json
```diff
@@ -11,7 +11,8 @@
     "noImplicitThis": true,
     "noUnusedLocals": true,
     "skipLibCheck": true,
-    "strict": true
+    "strict": true,
+    "lib": ["ES2017"]
```
AlaricWhitney, line 15: do you want to use `ES2017`, or use a newer version (or `ESNext`)?

AlaricWhitney (commented):
just one question regarding js standard to adhere to.

tsconfig.json
mastercactapus: Setting it to something like ESNext would be problematic as it would mean that features not-yet-supported by some browsers could make it into the app (which are not polyfilled or otherwise transpiled).

ES2017 was offered by an error message, which happens to be the lowest version that supports all the features we currently use and is well-supported by all browsers we'd expect to see:
https://caniuse.com/async-functions,object-values,object-entries,mdn-javascript_builtins_object_getownpropertydescriptors,pad-start-end,mdn-javascript_grammar_trailing_commas_trailing_commas_in_functions

This config in the type checker essentially works as our "gate," so only the code and features able to run in production will be allowed.

We can always add individual features too, but those discussions will be as needed:
https://www.typescriptlang.org/tsconfig#individual-library-components

---

Note: Adding it here was necessary to _increase_ the list of allowed features rather than to restrict them. The effective default was lower than ES2017, and it refused to build.

cmarquis (dismissed):

AlaricWhitney (approved):

AlaricWhitney:
@mastercactapus it looks like TS is failing in CICD: https://github.com/target/goalert/actions/runs/4798634642/jobs/8537179878?pr=2931

mastercactapus:
@AlaricWhitney Turns out the reviewdog stuff depends on `node_modules` to work, so I updated the script to tell yarn to use that linker for the checks.

I'm guessing the failure was due to a version mismatch.

Now it should use the correct version of yarn and all locked package versions.

cmarquis (approved):

AlaricWhitney (approved):

PR #2962, opened 2023-04-20
chore: eslint updates part2

cmarquis:

**Description:**
Updates the eslint cypress plugin. Due to changes several cypress tests were updated to remove unsafe chaining (https://docs.cypress.io/guides/core-concepts/retry-ability#Actions-should-be-at-the-end-of-chains-not-the-middle)


mastercactapus (changes requested):

web/src/cypress/support/select-by-label.ts
```diff
@@ -59,7 +59,7 @@ function findByLabel(
         .find('button[aria-label="Open"]')
         .should('be.visible')
         .click()
-        .should('not.have.focus')
+      cy.should('not.have.focus')
```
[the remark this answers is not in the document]
cmarquis: Removing it doesn't seem to break anything so I think I'm fine with that option.

AlaricWhitney (approved):

AlaricWhitney:
@cmarquis it looks like something is failing in the cypress tests

AlaricWhitney (approved):

AlaricWhitney:
@cmarquis it looks like one of the cypress tests are failing in ci

web/src/cypress/e2e/escalationPolicySteps.cy.ts
```diff
@@ -202,7 +202,6 @@ function testSteps(): void {
             .should('contain', s1.delayMinutes)
             .parent('[tabindex]')
             .focus()
-            .type(' ')
```
mastercactapus, line 205: I think this is significant because the spacebar is used to lift/drop items, maybe add a `cy.type` after

mastercactapus (changes requested):

web/src/cypress/e2e/escalationPolicySteps.cy.ts
cmarquis: Yeah, that was causing the test to fail. Should be resolved now.

PR #2977, opened 2023-04-28
ui: fix slack link typography in dialog

Forfold:
Adds proper typography to the dialog when linking from Slack to GoAlert.

From:

Screenshot 2023-04-28 at 10 31 04 AM

To:

Screenshot 2023-04-28 at 10 30 44 AM


allending313 (commented):
There seems to be some node errors but it looks good otherwise

Forfold:
@allending313 yeah it looks like it's affecting multiple PRs. I think its stemming from the luxon update here: #2677

allending313:
> @allending313 yeah it looks like it's affecting multiple PRs. I think its stemming from the luxon update here: #2849
oh, that's fun

allending313 (dismissed):

mastercactapus (changes requested):
Can we revert the lockfile change and address that separately?

mastercactapus (approved):

allending313 (approved):

PR #2978, opened 2023-04-28
ui: update lockfile

Forfold:
make clean && yarn && make start


Forfold:
I ran `yarn —version` and it gave me `3.5.0`. Is this correct?

I don't understand— did a recent PR introduce breaking changes?

mastercactapus:
Yes that is correct, but looking at your first commit, it looks like it reverted the lockfile to v1 which is where I got that idea. I'm also able to run `yarn install` and `make check` on master without any changes

I also noticed this which seems to indicate you may have something odd happening on your machine:
image

Forfold:
I don't think anything odd happened. There was an expectation that yarn should've been upgraded globally for this project that I wasn't aware of.

I'll close this PR.

PR #2984, opened 2023-05-01
dev: add yarnrc and stable version

mastercactapus:
**Description:**
Fixes issues where running `yarn` before `make start` could result in unexpected lockfile changes and other package install issues due to version mismatch.


Forfold (approved):
Functionally tested, lgtm

KatieMSB (approved):

Issue #2985, opened 2023-05-01
auth: create user keeps transaction open while calculating bcrypt hash


mastercactapus:
**Describe the Bug:**
When an admin creates a user using the GraphQL mutation, a transaction is started to insert the new user, and to insert the auth info (user/pass-hash) for the new user. The bcrypt hash is calculated during this transaction which keeps the connection open longer than necessary.

This also becomes a more significant issue if the available compute resources cannot compute the hash faster than the transaction timeout, resulting in an "Unexpected error" to the UI and "idle-in-transaction timeout (SQLSTATE 25P03)" in the logs.

**Steps to Reproduce:**
1. Create a user via GraphQL
2. Optionally, set the bcrypt value high, or add a `time.Sleep` to observe timeout

**Expected Behavior:**
The DB transaction should only be open to store the data (calculations should happen before).

**Observed Behavior:**
The DB transaction is held open during the calculation.

tong-bluehill:
For workaround, beside using master branch, can I set a longer timeout value?

PR #2991, opened 2023-05-04
ui/user: Allow setting/changing basic auth passwords in UI

### LINKED ISSUE #2667: Self reset password function for user
Dear Team,

Is there reset password function available? As I check, there is no way to reset password on GUI.
If this function still not implemented, kindly consider and enhance.
Thank you.

allending313:

**Description:**
Provides a way for users and admins to reset a basic auth password through the UI.
- Updated the edit user detail modal to support resetting a password. 
- The edit detail button will be available to non-admin users when viewing their own user profile page (edit role checkbox will be unavailable if they are not an admin). 
- Admins will be able to update users' passwords without needing to enter the user's old password. 

Fixes #2667 

**Out of Scope:**
Forgot password feature

**Screenshots:**
Non-Admin view
Screenshot 2023-05-04 at 6 08 17 PM

Admin view
Screenshot 2023-05-04 at 6 08 41 PM

- Updated user edit details modal to support resetting password. 
- Slightly refactored user profile page option buttons (delete, edit, favorite)

- Created gql endpoint to update user password


auth/basic/db.go
```diff
@@ -60,6 +63,42 @@ func (b *Store) CreateTx(ctx context.Context, tx *sql.Tx, userID, username, pass
...
+		if err != nil {
+			return validation.NewFieldError("oldPassword", "Invalid Password")
+		}
+
+		if validatedUserId != userID {
+			return err
+		}
+	}
+
+	hashedPassword, err := bcrypt.GenerateFromPassword([]byte(newPassword), 12)
```
mastercactapus, line 94: This would create a new occurrence of bug #2985 

We should merge #2990 first which fixes it, and then update this PR to use the `HashedPassword` type.

auth/basic/db.go
```diff
@@ -60,6 +63,42 @@ func (b *Store) CreateTx(ctx context.Context, tx *sql.Tx, userID, username, pass
...
+	err := permission.LimitCheckAny(ctx, permission.System, permission.Admin, permission.User, permission.MatchUser(userID))
+	if err != nil {
+		return err
+	}
+
+	err = validate.Many(
+		validate.UUID("UserID", userID),
+		validate.Username("Username", username),
+		validate.Text("oldPassword", oldPassword, 0, 200),
+		validate.Text("newPassword", newPassword, 8, 200),
```
mastercactapus, line 77:
```suggestion
		validate.Text("OldPassword", oldPassword, 0, 200),
		validate.Text("NewPassword", newPassword, 8, 200),
```

auth/basic/db.go
```diff
@@ -60,6 +63,42 @@ func (b *Store) CreateTx(ctx context.Context, tx *sql.Tx, userID, username, pass
...
+	)
+	if err != nil {
+		return err
+	}
+
+	if !permission.Admin(ctx) {
+		validatedUserId, err := b.Validate(ctx, username, oldPassword)
```
mastercactapus, line 84: Let's talk/think about how we can do this OUTSIDE of the transaction.

Can we implement a verification token that can be passed in maybe?

go.mod
```diff
@@ -31,7 +31,7 @@ require (
 	github.com/matcornic/hermes/v2 v2.1.0
 	github.com/pelletier/go-toml v1.9.5
 	github.com/pkg/errors v0.9.1
-	github.com/prometheus/client_golang v1.15.0
+	github.com/prometheus/client_golang v1.15.1
```
mastercactapus, line 34: noise: Can we drop the updates from this PR, or were they required?

graphql2/graphqlapp/user.go
```diff
@@ -146,6 +147,24 @@ func (a *Mutation) UpdateUser(ctx context.Context, input graphql2.UpdateUserInpu
 	return err == nil, err
 }
 
+func (a *Mutation) UpdateUserPassword(ctx context.Context, input graphql2.UpdateUserPassword) (bool, error) {
+	err := withContextTx(ctx, a.DB, func(ctx context.Context, tx *sql.Tx) error {
+		username, err := a.UserStore.FindAuthBasicUsername(ctx, tx, input.ID)
+		if err != nil {
+			fmt.Println("Happened here", err)
```
mastercactapus, line 154: Debug?

graphql2/schema.graphql
```diff
@@ -515,6 +515,8 @@ type Mutation {
 
   createUser(input: CreateUserInput!): User
 
+  updateUserPassword(input: UpdateUserPassword!): Boolean!
```
mastercactapus, line 518: Can we merge this with the existing `updateUser` mutation?

user/store.go
```diff
@@ -153,6 +156,13 @@ func NewStore(ctx context.Context, db *sql.DB) (*Store, error) {
 			FOR UPDATE
 		`),
 
+		findAuthBasicUsername: p.P(`
```
mastercactapus, line 159: `findAuthSubjects` already does this (the `subject_id` column is the username)

web/src/app/users/UserDetails.tsx
```diff
@@ -148,6 +148,36 @@ export default function UserDetails(props: {
...
+      key='secondary-action-favorite'
+      id={userID}
+      type='user'
+    />,
+  ]
+
+  if (isAdmin || userID === currentUserID) {
+    options.unshift({
+      label: 'Edit',
+      icon: <EditIcon />,
```
mastercactapus, line 169: We can change this to the Key icon since it is about access.

yarn.lock
```diff
@@ -2396,12 +2396,12 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@types/react-dom@npm:18.0.11":
-  version: 18.0.11
-  resolution: "@types/react-dom@npm:18.0.11"
+"@types/react-dom@npm:18.2.1":
```
mastercactapus, line 2399: noise: were dependency updates required?

mastercactapus (changes requested):

graphql2/graphqlapp/user.go
allending313: ah oops

graphql2/schema.graphql
allending313: The updateUser mutation applies to the users db entry whereas the basic auth info we want to update is stored in auth_basic_users so I created a new mutation for it

go.mod
allending313: Most likely not required, I'll drop them

yarn.lock
mastercactapus: There was a fix for that merged recently; if you didn't need to update any deps doing `git checkout origin/master -- yarn.lock` should clear everything up

graphql2/schema.graphql
mastercactapus: I see where you're coming from, but I think that would be leaking implementation details. In other words, the API design should be based on what is most predictable for the API layer, and not influenced by how things work under the covers.

There are always things to consider, but in this case, I feel like updating a user can include updating the user's password. From the UI side, it makes sense because it's just a single mutation backing the edit form, and from the UX perspective they are submitting the new password along with the admin true/false info.

allending313: Ah that makes sense, will try it this way

AlaricWhitney:
@allending313 it looks like there's some conflicts that need to be resolved.

web/src/app/users/UserDetails.tsx
Forfold: A cogwheel would be the most familiar icon imo

mastercactapus: Wouldn't that be general settings rather than passwords/authentication?

auth/basic/db.go
```diff
@@ -41,6 +46,12 @@ type HashedPassword interface {
 	_private() // prevent external implementations
 }
 
+type ValidatedPassword interface {
+	Hash() string
```
mastercactapus, line 50: Have this return the user ID it was validated against rather than the hash. We can use that to ensure the ValidatedPassword is for the correct user in `UpdateTx`

auth/basic/db.go
```diff
@@ -84,6 +95,21 @@ func (b *Store) CreateTx(ctx context.Context, tx *sql.Tx, userID, username strin
...
+	err := permission.LimitCheckAny(ctx, permission.Admin, permission.MatchUser(userID))
+	if err != nil {
+		return err
+	}
+
+	if oldPassword == nil && !permission.Admin(ctx) {
+		return validation.NewFieldError("oldPassword", "Previous password required")
+	}
+
+	_, err = tx.StmtContext(ctx, b.update).ExecContext(ctx, userID, newPassword.Hash())
```
mastercactapus, line 109: Validate `userID` is a `UUID` before passing it to the DB

auth/basic/db.go
```diff
@@ -115,3 +141,39 @@ func (b *Store) Validate(ctx context.Context, username, password string) (string
 
 	return userID, nil
 }
+
+// ValidatePassword should check if the password matches the user's stored password
+func (b *Store) ValidatePassword(ctx context.Context, userID string, password string) (ValidatedPassword, error) {
```
mastercactapus, line 146: nit: simplify
```suggestion
func (b *Store) ValidatePassword(ctx context.Context, userID, password string) (ValidatedPassword, error) {
```

auth/basic/db.go
```diff
@@ -115,3 +141,39 @@ func (b *Store) Validate(ctx context.Context, username, password string) (string
...
+	err := validate.Many(
```
mastercactapus, line 147: Do permissions check first before validating. Use `permission.MatchUser` -- this technically validates the password without requiring a username, so instead we're requiring the current request is already authorized for the intended userID.

auth/basic/db.go
```diff
@@ -115,3 +141,39 @@ func (b *Store) Validate(ctx context.Context, username, password string) (string
...
+		return nil, err
+	}
+
+	row := b.getByID.QueryRowContext(ctx, userID)
+	var hash string
+	err = row.Scan(&hash)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, errors.New("invalid userID")
+		}
```
mastercactapus, line 161: nit: flatten also use `unknown` instead of `invalid` (we already ensured it's valid)
```suggestion
	if errors.Is(err, sql.ErrNoRows) {
		return nil, errors.New("unknown userID")
	}
	if err != nil {
```

auth/basic/db.go
```diff
@@ -115,3 +141,39 @@ func (b *Store) Validate(ctx context.Context, username, password string) (string
...
+
+	err = bcrypt.CompareHashAndPassword([]byte(hash), []byte(password))
+	if err != nil {
+		return nil, validation.NewFieldError("oldPassword", "invalid password")
+	}
+
+	hashedPassword, err := bcrypt.GenerateFromPassword([]byte(password), 12)
+	if err != nil {
+		return nil, err
+	}
```
mastercactapus, line 176: We don't need this for validation, and it results in slowing things down by double.

We only do that in the new password because we need to store the generated hash. For verification, we just want to ensure we validated for the correct user.

web/src/app/users/UserEditDialog.tsx
```diff
@@ -19,40 +19,116 @@ interface UserEditDialogProps {
...
 
-  function handleChange(e: React.ChangeEvent<HTMLInputElement>): void {
-    setAdminChecked(e.target.checked)
+  // Checks if any of the password fields are used. Used to skip any unnecessary updateUserPassword mutation
+  function passwordChanged(): boolean {
+    return !(
+      value.oldPassword === '' &&
+      value.newPassword === '' &&
+      value.confirmNewPassword === ''
+    )
```
mastercactapus, line 60: nit: Can we invert this? from `NOT (... AND ...)` to `... OR ...`
```suggestion
    return Boolean(
      value.oldPassword||value.newPassword||value.confirmNewPassword
    )
```

web/src/app/users/UserEditDialog.tsx
```diff
@@ -19,40 +19,116 @@ interface UserEditDialogProps {
...
+  }
+
+  // Validates inputs to the newPassword and confirmNewPassword fields
+  function handleValidation(): FieldError[] {
+    let err: FieldError[] = []
+    if (!passwordChanged()) return err
+    if (value.newPassword.length < 8 || value.newPassword.length > 20) {
```
mastercactapus, line 67: We should probably leave most of this validation to the backend, since that should be the source of truth.

mastercactapus (changes requested):

web/src/app/users/UserEditForm.tsx
```diff
@@ -0,0 +1,91 @@
...
+        onChange(val)
+      }}
+    >
+      <Grid container spacing={2}>
+        {!admin && (
+          <Grid item xs={12}>
+            <FormField
+              InputProps={{
+                'data-cy': 'oldPassword',
+              }}
```
Forfold, line 38: you can probably test off of `label` or something so we don't have this artifact (and the ones below) in the future if/when we are migrating to playwright

(no worries if this was the best solution you found though, we have a lot of `'data-cy'` tags throughout the app)

web/src/app/users/UserDetails.tsx
allending313: I'm okay with the key icon in this case as it aligns more closely to the function of the edit modal. The only two functions of this modal as of now are to reset passwords and change admin status, with the former of which being what I assume will have the more prominent usage (esp as only admins can access changing the status). If we add any more options to the modal then I think it would be worth considering a more general icon.

Forfold: Yeah that makes sense since we aren't editing a ton that you didn't use that icon, but we should probably keep the key icon reserved for services. I'd be open to other options.

allending313: Thoughts on the open lock icon?
Screenshot 2023-05-24 at 9 50 32 AM

Forfold (dismissed):
LGTM! Functionally tested and reviewed.

Forfold (dismissed):

Forfold (dismissed):

mastercactapus (changes requested):
pretty much just some code styling stuff

auth/basic/db.go
```diff
@@ -84,6 +101,32 @@ func (b *Store) CreateTx(ctx context.Context, tx *sql.Tx, userID, username strin
...
+
+	if oldPassword != nil {
+		if userID != oldPassword.UserId() {
+			return validation.NewFieldError("oldPassword", "Password does not match User")
+		}
+	} else {
+		if !permission.Admin(ctx) {
+			return validation.NewFieldError("oldPassword", "Previous password required")
+		}
+	}
```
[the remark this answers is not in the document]
allending313: looking back at this I have no idea why I nested the conditionals this way :/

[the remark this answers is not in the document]
mastercactapus: It's the struggle of re-using a single `oldPassword != nil` check in two conditionals but not having an elegant way to do both without nesting at least one.

I find myself doing the same thing occasionally in this situation. In the end, keeping it flatter and separate short-circuit returns with a bit of repetition vs. `else` tends to read better.

It may be better to swap the second one to `if !permission.Admin(ctx) && old == nil {` because `old != nil` and `old == nil` look very similar.
https://google.github.io/styleguide/go/guide#concision

Forfold (dismissed):

allending313:
> 

for accounts created via OIDC/Github, should we hide the pw reset fields for those user accounts?

mastercactapus:
@allending313 Based on recent discussions, disable rather than hide. It tends to be confusing when pieces of the UI are missing entirely. A little text about why it's disabled would also probably help.

I'm thinking maybe we can add a `hasPassword` to the `User` type, which is only accessible to the current user, or an admin, and in the case of `hasPassword === false`, then we'd hide the "Old password" field and not require it.

We just need to be sure that while an admin can "upsert" a new password, a user can only update an existing one if the old password is specified.

mastercactapus (dismissed):

mastercactapus (approved):

andrewbenington:
Screenshot 2023-06-08 at 10 51 27 AM
What is the Username field used for? it doesn't seem to correspond to the display name or email

m17ch (approved):

allending313:
> Screenshot 2023-06-08 at 10 51 27 AM What is the Username field used for? it doesn't seem to correspond to the display name or email

The Username field is there for admins to create a basic auth AuthSubject for a user. Normally when a user already has basic auth as an AuthSubject, the pw reset feature would just update that basic auth password. However, if they don't already have an existing basic auth AuthSubject we want admins to have the option of creating one for the user which requires a username for login.

andrewbenington (approved):

PR #3001, opened 2023-05-08
meta: add dependabot limits

Forfold:
Adds some reasonable limits to Dependabot now that we are allowing more types of dependencies to be automated.

Puts `npm` back to a limit of 3 (same as before), and puts `gomod` at 3 as well. The rest I set to a limit of `1` to be handled in order, so as to not clog up the repo.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,6 +13,7 @@ updates:
       day: 'sunday'
       time: '00:00'
       timezone: 'Asia/Kolkata'
+    open-pull-requests-limit: 1
     allow:
       - dependency-type: 'direct'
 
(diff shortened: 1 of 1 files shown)


AlaricWhitney (changes requested):
seeing as depenabot was really behind with the limit of 3 previously, i don't think it's realistic with limiting it to 3.  IMO opinion either the better option is to leave it alone, and address the CICD issues with building each image to increase the speed and confidence of the merges (so we can start talking about auto merges for these dependabot PR's, and only have to manually approve ones with CICD failures.

Forfold:
@AlaricWhitney I personally think this will make it worse/more overwhelming for me to review PRs when there is an over-abundance of information to look at.

PR #3026, opened 2023-05-16
ui: preflight changes for message logs graph

Forfold:
Part of #2460
Pre-flight PR to #2481

- Renames `AdminDebugMessagesLayout` to `AdminMessageLogsLayout`
- Renames `DebugMessagesControls` to `AdminMessageLogsControls`
- Renames `DebugMessageDetails` to `AdminMessageLogDrawer`
- Removes unused `useFuse` hook
- Extends `ListItem` for the `FlatListItem` interface (the props are spread onto a `ListItem` element)
- Updates PaginatedList to rely on `items.length` for no results, instead of `pageCount`
- Updates PaginatedList to use a `div` for the secondary typography element, if the children is anything other than a string (helps with console warnings, DOM shouldn't be affected visually)
- Adds title casing to both the type and ID for the `ToolbarPageTitle` component
- Updates `AlertAveragesGraph` to use the `sx` prop for css instead of classes via `makeStyles`
- Updates `ISOPickers` to update the input value if the incoming value changes
- Updates `ISOPickers` coloring UX


mastercactapus (dismissed):
Looks good, functionally validated

mastercactapus (approved):

tony-tvu (approved):

Issue #2914, opened 2023-03-29
api or webhook request for on-call shifts


jjjeroenski:
Is there a way to get the information of who is on-call from-DT --> untill-DT with contact details for another automated system?

Possible reason could be to link this data to Twilio so they could also transfer the incoming human support phonecalls to the correct person who is on-call in that shift/period? Otherwise we need to maintain two different shedules.

Or a sheduled script/cron with an json export who could be used in Twilio or PBX
Or an api/webhook request

mastercactapus:
On-call notifications send the list of currently on-call users from a schedule to a Slack channel. This can happen at a set time or when on-call user(s) change.

With the recent introduction of webhook support on EP steps, adding webhook support to that feature should be relatively easy.

---

Alternatively, we could expand the calendar subscription to, optionally, include all user shifts instead of only the current user.

jjjeroenski:
Both solutions sounds good, unfortunately slack isn't very popular here in our region.
i've seen also the calendare subscription button with the export, if you're using that, you will also need to add the user or an unique userid in the output.

i guess the most flexible solution for the future is an request for shedule or shifts where you get the same output like you choose "shifts" in the webinterface.

andrewbenington:
Webhook support added with #3036

PR #3036, opened 2023-05-22
feat: webhook on call notification

andrewbenington:

**Description:**
Adds webooks as a possible destination for Schedule On-Call Notifications.
- ScheduleOnCallNotificationsForm's `value` uses a `channelFields` sub-object instead of including `slackChannel` and `slackUserGroup` in the base object
  - Dropdown added to select type of channel (slack/webhook)
  - Slack Channel/User Group form fields moved to their own component, which is able to be swapped for the webhook field
- On-call notifications sent to a webhook include a JSON body with the following
  - User (ID, URL, Name)
  - Schedule ID
  - Schedule Name
  - Schedule URL

https://github.com/target/goalert/issues/2914

Users can specify a webhook destination for Schedule On-Call Notifications if webhooks are enabled by an admin. Under a Schedule, the Quick Link for On-Call Notifications will now show up if webhooks are enabled and the chan-webhook experimental flag is set. When creating/editing an On-Call Notification rule, there is now a dropdown to select between Slack and a webhook, which will also display the appropriate form fields.

`setScheduleOnCallNotificationRules` can now accept a rule with a target of type "chanWebhook", if it includes an ID that conforms with an approved webhook URL.

**UI Screenshots:**
Screenshot 2023-06-07 at 1 05 39 PM
Screenshot 2023-06-07 at 1 06 04 PM


Makefile
```diff
@@ -181,7 +181,7 @@ ensure-yarn: # Yarn ensures the correct version of yarn is installed
 
 yarn:
 	corepack enable
-	corepack prepare yarn@stable --activate
```
andrewbenington, line 184: changed this because corepack complains about using `stable`:
`Invalid package manager specification in CLI arguments; expected a semver version`

Makefile
```diff
@@ -181,7 +181,7 @@ ensure-yarn: # Yarn ensures the correct version of yarn is installed
...
+	corepack prepare yarn@3.5.0 --activate
```
andrewbenington, line 184: changed this because corepack complains about using `stable`:
`Invalid package manager specification in CLI arguments; expected a semver version`

notification/webhook/sender.go
```diff
@@ -57,6 +57,16 @@ type POSTDataVerification struct {
 	Code    string
 }
 
+// POSTDataOnCallNotification represents fields in outgoing on call notification.
+type POSTDataOnCallNotification struct {
+	AppName      string
+	Type         string
+	Users        []notification.User
```
mastercactapus, line 64: Because this struct defines an API, let's add a struct def here instead of re-using one. Otherwise, changing a different package (`notification.User`) could result in a difference in our API without a compiler error.

Adding an explicit def here also has the benefit of having all webhook API code and fields in one place.

```go
// We use types defined in this package to insulate against unintended API
// changes.
users := make([]POSTDataOnCallUser, len(m.Users))
for i, u := range m.Users {
	users[i] = POSTDataOnCallUser(u)
}
```

Since the structs are identical, we can do a direct assignment, but this will insulate from tag changes in the notification package and create a compiler error if there are any incompatible changes.

mastercactapus (changes requested):
Small change for some backend code.

Since this PR contains UI changes, it would be helpful to add a screenshot or two to make sure everyone is on the same page for the design/layout

notification/webhook/sender.go
andrewbenington: That makes sense, thanks for the explanation. I just made the change

mastercactapus:
Do you think we should mark this as a draft in the meantime? I think you mentioned this is blocked until the dropdown changes are made to the rule form.

No one is actively working on that for the moment, to my knowledge, though. I created a separate issue to track that specific change:
- https://github.com/target/goalert/issues/3045

mastercactapus (changes requested):
It looks like we're displaying the full URL, including sensitive parts:
image

In contact methods, we mask the password, strip query params, and truncate the path:
image
image

And escalation policies only show the hostname:
image

We should be consistent everywhere as a follow-up, but let's update this to follow one of the two existing methods.

andrewbenington:
> It looks like we're displaying the full URL, including sensitive parts:
> 
> In contact methods, we mask the password, strip query params, and truncate the path:
> 
> And escalation policies only show the hostname: 
> 
> We should be consistent everywhere as a follow-up, but let's update this to follow one of the two existing methods.

@mastercactapus I think it's good to have the full URL to differentiate between webhooks with the same hostname. The name is now set to the "FriendlyValue", but the full URL will still be visible on edit. Should this be hidden too?
Screenshot 2023-06-06 at 4 14 14 PM
The Webhook contact method disables editing the field, and masks the password (but not query params):
Screenshot 2023-06-06 at 4 19 46 PM

mastercactapus (approved):

mastercactapus:
@andrewbenington It's mostly consistent with contact method behavior, and I'd consider changing that to be out-of-scope for this PR.

Mainly because the notification channel can be edited, and contact methods can not (since they also convey ownership). With it in experimental state, I'm okay with leaving it open for iteration in the future (e.g., maybe only show the FriendlyValue _or_ allow replacing it entirely) but then consider how we can get EP and contact methods to all match

KatieMSB (approved):
lgtm! 👍

PR #3040, opened 2023-05-23
fix: webhook check on edit escalation policy step

### LINKED ISSUE #3039: escalation policies: restrict webhooks to admin allowed URLs
**Describe the Bug:**
When an admin enables webhooks as a notification destination, they can opt to restrict them to a limited set of URLs. When updating an existing escalation policy step, there is no check to make sure a webhook conforms with the restricted URLs.

**Steps to Reproduce:**
1. Allow webhooks as an admin, and restrict URLs to a domain (e.g. www.example.com)
2. Create an escalation policy
3. Create a new step
4. Edit the step to add a webhook notification with a URL that should be disallowed (e.g. www.notexample.com)

**Expected Behavior:**
The user should receive an error upon trying to add a disallowed webhook notification to an escalation policy

**Observed Behavior:**
The user can add a webhook with any URL if editing an existing step

**Screenshots/Stack Traces:**
If applicable, add screenshots and/or stack traces to help explain your problem.

Admin settings:
admin_restrictions
Error should have occurred when editing step, but did not:
esc_policy

**Application Version:**
Output of `goalert version` and/or version information from view-source on the UI.

`
Version:   v0.30.1-1543-g8e07d53c9-dirty
GitCommit: 8e07d53c9c310241f94ac60d1ace8bb93b8f1bb6 (dirty)
BuildDate: 2023-05-23T09:58:16-05:00
GoVersion: go1.20.4 (gc)
Platform:  darwin/arm64
Migration: slack-ug (#235)
`

**Desktop:**
 - OS: MacOS
 - Browser: Firefox
 - Version 113.0.1

andrewbenington:

**Description:**
Adding backend check for webhooks added to existing escalation steps.
Uses the same validation logic as creating a new step.

Fixes #3039 

**Attachments**
This error will appear when editing a step to include a forbidden webhook:
Screenshot 2023-05-30 at 3 00 52 PM

### DIFF
--- a/graphql2/graphqlapp/escalationpolicy.go
+++ b/graphql2/graphqlapp/escalationpolicy.go
@@ -59,6 +59,7 @@ func (m *Mutation) CreateEscalationPolicyStep(ctx context.Context, input graphql
 
 	for _, tgt := range input.Targets {
 		if tgt.Type == assignment.TargetTypeChanWebhook && !cfg.ValidWebhookURL(tgt.ID) {
+			// UI code expects targets to be un-indexed
 			return nil, validation.NewFieldError("targets", "URL not allowed by administrator")
 		}
 	}
@@ -248,7 +249,8 @@ func (m *Mutation) UpdateEscalationPolicyStep(ctx context.Context, input graphql
 			step.Targets = make([]assignment.Target, len(input.Targets))
 			for i, tgt := range input.Targets {
 				if tgt.Type == assignment.TargetTypeChanWebhook && !cfg.ValidWebhookURL(tgt.ID) {
-					return validation.NewFieldError("webhook-targets", "URL not allowed by administrator")
+					// UI code expects targets to be un-indexed
+					return validation.NewFieldError("targets", "URL not allowed by administrator")
 				}
 				step.Targets[i] = tgt
 			}
--- a/graphql2/graphqlapp/escalationpolicy.go
+++ b/graphql2/graphqlapp/escalationpolicy.go
@@ -226,8 +226,8 @@ func (m *Mutation) UpdateEscalationPolicy(ctx context.Context, input graphql2.Up
 }
 
 func (m *Mutation) UpdateEscalationPolicyStep(ctx context.Context, input graphql2.UpdateEscalationPolicyStepInput) (bool, error) {
-	cfg := config.FromContext(ctx)
 	err := withContextTx(ctx, m.DB, func(ctx context.Context, tx *sql.Tx) error {
+		cfg := config.FromContext(ctx)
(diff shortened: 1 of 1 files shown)


graphql2/graphqlapp/escalationpolicy.go
```diff
@@ -226,6 +226,7 @@ func (m *Mutation) UpdateEscalationPolicy(ctx context.Context, input graphql2.Up
 }
 
 func (m *Mutation) UpdateEscalationPolicyStep(ctx context.Context, input graphql2.UpdateEscalationPolicyStepInput) (bool, error) {
+	cfg := config.FromContext(ctx)
 	err := withContextTx(ctx, m.DB, func(ctx context.Context, tx *sql.Tx) error {
 		step, err := m.PolicyStore.FindOneStepForUpdateTx(ctx, tx, input.ID) // get delay
```
mastercactapus, line 231: nit: use the closest-scoped context for the config

```suggestion
	err := withContextTx(ctx, m.DB, func(ctx context.Context, tx *sql.Tx) error {
		cfg := config.FromContext(ctx)
		step, err := m.PolicyStore.FindOneStepForUpdateTx(ctx, tx, input.ID) // get delay
```

graphql2/graphqlapp/escalationpolicy.go
```diff
@@ -246,6 +247,9 @@ func (m *Mutation) UpdateEscalationPolicyStep(ctx context.Context, input graphql
 		if input.Targets != nil {
 			step.Targets = make([]assignment.Target, len(input.Targets))
 			for i, tgt := range input.Targets {
+				if tgt.Type == assignment.TargetTypeChanWebhook && !cfg.ValidWebhookURL(tgt.ID) {
+					return validation.NewFieldError("targets", "URL not allowed by administrator")
```
mastercactapus, line 251: Use `[%d]` to denote the error within a list
```suggestion
					return validation.NewFieldError(fmt.Sprintf("Targets[%d]", i), "URL not allowed by administrator")
```

andrewbenington: I tried this and it resulted in the error not displaying properly on the form. Tried lowercase `targets` as well.

mastercactapus: ah, I see, if that already works properly, and is being used elsewhere, would you mind just adding a comment explaining the _why_? Something like `the UI code expects targets to be un-indexed` should suffice.

Thank you for the context and for digging into it.

andrewbenington: no problem! just added the comment to both instances of the code

mastercactapus (approved):

allending313 (approved):
lgtm!

Issue #3045, opened 2023-05-24
Improve Slack User Group Notifications and Channel Updates


mastercactapus:
**What problem would you like to solve? Please describe:**
Users find the current layout confusing in which they select a channel and then, optionally, update a user group as well. We received feedback from early adopters that this process could be improved.

**Describe the solution you'd like:**
We would like to clearly separate the intent of updating a user group and creating individual channel notifications by implementing the following changes:

1. Add a dropdown to Schedule on-call notifications, making it similar to contact methods.
2. Decouple Slack user group notification rules from channel notifications. User group notification rules will still include a channel for errors, but will update silently.
3. Keep Slack channel notifications unchanged.

**Additional context:**
Primary user group issue: https://github.com/target/goalert/issues/1228

PR #3046, opened 2023-05-24
feat: on call notification channel separation

andrewbenington:

**Description:**

- The dialog for creating/editing schedule on-call notifications now features a dropdown to select channel type
- Slack channel/user groups have been separated into two channel types, selectable from the dropdown
  - The checkbox was removed, as its functionality is replaced by the dropdown
- User Group notifications are no longer sent to the Slack channel unless there is an error with the User Group
- The `Value` type for `ScheduleOnCallNotificationsForm` has been altered to be more accommodating of different notification channel types
- Minor dev fix: use `YARN_VERSION` instead of assumed stable version
  - Fixes `Usage Error: Invalid package manager specification in CLI arguments; expected a semver version` from recent build changes

#3045 

**Screenshots:**
Screenshot 2023-05-24 at 4 21 43 PM
Screenshot 2023-05-24 at 4 21 47 PM

image

To use Slack User Groups for schedule on-call notifications, the user must select "SLACK USER GROUP" from the new dropdown instead of checking a checkbox.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -181,7 +181,7 @@ ensure-yarn: # Yarn ensures the correct version of yarn is installed
 
 yarn:
 	corepack enable
-	corepack prepare yarn@stable --activate
+	corepack prepare yarn@$(YARN_VERSION) --activate
 
 check-js: generate $(NODE_DEPS)
 	$(MAKE) ensure-yarn
--- a/notification/slack/usergrouperror.go
+++ b/notification/slack/usergrouperror.go
@@ -19,8 +19,6 @@ type userGroupError struct {
 	ScheduleName string
 	Missing      []notification.User
 
-	ErrorMsg string
-
 	callbackFunc func(string, ...url.Values) string
 }
 
@@ -56,23 +54,17 @@ func slackLink(url, label string) string {
 }
 
 // userGroupErrorMissing is a template for when a user-group update fails because one or more users are missing from Slack.
-var userGroupErrorMissing = template.Must(template.New("userGroupErrorMissing").Parse(`{{.ErrorMsg}}
-
-Hey everyone! I couldn't update {{.GroupRef}} because I couldn't find the following user(s) in Slack: {{.MissingUserRefs}}
+var userGroupErrorMissing = template.Must(template.New("userGroupErrorMissing").Parse(`Hey everyone! I couldn't update {{.GroupRef}} because I couldn't find the following user(s) in Slack: {{.MissingUserRefs}}
 
 If you could have them add a SLACK_DM contact method from their respective GoAlert profile page(s), that would be great! Hopefully I'll be able to update the user-group next time.`))
 
 // userGroupErrorEmpty is a template for when a user-group update fails because there are no users on-call.
-var userGroupErrorEmpty = template.Must(template.New("userGroupErrorEmpty").Parse(`{{.ErrorMsg}}
-
-Hey everyone! I couldn't update {{.GroupRef}} because there is nobody on-call for {{.ScheduleRef}}.
+var userGroupErrorEmpty = template.Must(template.New("userGroupErrorEmpty").Parse(`Hey everyone! I couldn't update {{.GroupRef}} because there is nobody on-call for {{.ScheduleRef}}.
 
 Since a Slack user-group cannot be empty, I'm going to leave it as-is for now.`))
 
 // userGroupErrorUpdate is a template for when a user-group update fails for any reason other than missing users.
-var userGroupErrorUpdate = template.Must(template.New("userGroupErrorUpdate").Parse(`{{.ErrorMsg}}
-
-Hey everyone! I couldn't update {{.GroupRef}} because I ran into a problem. Maybe touch base with the GoAlert admin(s) to see if they can help? I'm sorry for the inconvenience!
+var userGroupErrorUpdate = template.Must(template.New("userGroupErrorUpdate").Parse(`Hey everyone! I couldn't update {{.GroupRef}} because I ran into a problem. Maybe touch base with the GoAlert admin(s) to see if they can help? I'm sorry for the inconvenience!
 
 Here's the ID I left with the error in my logs so they can find it:
 {{.ErrorRef}}`))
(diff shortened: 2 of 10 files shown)


mastercactapus (changes requested):
For the layout:
- can we move the notification type selection below the notification options?
- The Slack channel for user groups is for errors/issues, not as a fallback (from a user perspective)

Also, this needs the behavioral change in the backend -- ensuring user group updates don't send the regular channel notification.

andrewbenington:
> For the layout:
> 
>     * can we move the notification type selection below the notification options?
> 
>     * The Slack channel for user groups is for errors/issues, not as a fallback (from a user perspective)
> 
> 
> Also, this needs the behavioral change in the backend -- ensuring user group updates don't send the regular channel notification.

Just to clarify - no messages should be sent at all if there isn't an error updating the UG?

mastercactapus:
Correct, if the notification channel is `SLACK_USER_GROUP` no messages should be sent to the channel unless there is an error.

web/src/app/schedules/on-call-notifications/util.ts
```diff
@@ -4,20 +4,115 @@ import { DateTime } from 'luxon'
...
+  return target
+}
+
 export type Value = {
-  slackChannelID?: string | null
-  slackUserGroup?: string | null
   time: string | null
   weekdayFilter: WeekdayFilter
+  type: NotificationChannelType
+  channelFields: ChannelFields
```
andrewbenington, line 105: Future-proofing for non-slack channels

notification/slack/usergrouperror.go
```diff
@@ -19,7 +19,7 @@ type userGroupError struct {
 	ScheduleName string
 	Missing      []notification.User
 
-	OnCall string
+	ErrorMsg string
```
mastercactapus, line 22: This field and the `{{.ErrorMsg}}` line can be removed entirely, that was only used as the placeholder for the on call message. The entire template _is_ the error message

notification/slack/usergroupsender.go
```diff
@@ -132,27 +131,31 @@ func (s *UserGroupSender) Send(ctx context.Context, msg notification.Message) (*
...
+	// Only send to the channel if an error occurred
+	if stateDetails != "" {
+		err = s.withClient(ctx, func(c *slack.Client) error {
+			_, ts, err = c.PostMessageContext(ctx, chanID, slack.MsgOptionText(errorMsg, false))
+			if err != nil {
+				return fmt.Errorf("post message to channel '%s': %w", chanID, err)
+			}
+			return nil
+		})
 		if err != nil {
```
mastercactapus, line 156: Let's flatten this.
https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

We can invert the check for the quick/easy return (since there's no state details or external ID). Secondly, we can just return the error to `withClient` and add the context in the external error check.
```go
if stateDetails == "" {
	return &notification.SentMessage{State: notification.StateDelivered}, nil
}

err = s.withClient(ctx, //...
	_, ts, err = c.PostMessageContext( //...
	return err
})
if err != nil {
	return nil, fmt.Errorf("post message to channel '%s': %w", chanID, err)
}
```

Forfold:
We can probably disable the type field if there is only 1 type present/user groups are disabled:

Screenshot 2023-05-31 at 12 16 29 PM

andrewbenington:
> We can probably disable the type field if there is only 1 type present/user groups are disabled:
> 
> Also maybe make the types Title Case?
> 
> Screenshot 2023-05-31 at 12 16 29 PM

Good idea about the type field.

And I made the types in all caps to stay in line with the personal contact method dropdown, which uses all caps. I'm fine with changing that to title case as well, but I think we should be consistent

Forfold:
Good point, could probably be a small change in its own PR since it's on the profile too.

When I go to create a duplicate user group, it only validates that the channel is the same (even though it's for error reporting). I think we should error on a duplicate user-group and set that as the first error to return if using that type, even if the channel for error messaging is different.

Screenshot 2023-05-31 at 12 27 12 PM
Screenshot 2023-05-31 at 12 27 28 PM

andrewbenington:
> Good point, could probably be a small change in its own PR since it's on the profile too.
> 
> When I go to create a duplicate user group, it only validates that the channel is the same (even though it's for error reporting). I think we should error on a duplicate user-group and set that as the first error to return if using that type, even if the channel for error messaging is different. This could also be a separate issue I think but we could maybe fix here.
> 
> Screenshot 2023-05-31 at 12 27 12 PM Screenshot 2023-05-31 at 12 27 28 PM

It will only show the error if both the user group and channel are the same as another existing rule. @mastercactapus says he thinks that's fine for now, and that we can refactor this in the future if we want to make the user group unique between rules.

web/src/app/schedules/on-call-notifications/ScheduleOnCallNotificationsEditDialog.tsx
```diff
@@ -32,7 +26,7 @@ export default function ScheduleOnCallNotificationsEditDialog(
       ? DateTime.fromFormat(rule.time, 'HH:mm', { zone }).toISO()
       : null,
     weekdayFilter: rule?.time ? rule.weekdayFilter || EVERY_DAY : NO_DAY,
-    type: channelTypeFromTarget(rule?.target),
+    type: rule?.target?.type ?? 'slackChannel',
```
Forfold, line 29: Does the type of slack channel no longer come in? If there's no type it seems odd to assume slackChannel

I feel like this may produce an issue in the future if we add more types- the code before was explicit in what types come in using a switch statement

Makefile
```diff
@@ -181,7 +181,7 @@ ensure-yarn: # Yarn ensures the correct version of yarn is installed
 
 yarn:
 	corepack enable
-	corepack prepare yarn@stable --activate
+	corepack prepare yarn@$(YARN_VERSION) --activate
```
Forfold, line 184: Is this change intended? It doesn't seem applicable to this PR

andrewbenington: `make yarn` fails with `Usage Error: Invalid package manager specification in CLI arguments; expected a semver version` otherwise. I figured I would include it as a minor dev fix, but if you think it would be better as its own PR then I can revert it.

Forfold: Sounds good. I ran into this too at some point. I updated the description.

Forfold (approved):

PR #2824, opened 2023-02-27
status updates: switch to per-contact-method behavior

mastercactapus:
**Description:**
This PR changes the behavior of status updates to be a per-contact-method setting, instead of a global user-level setting.

**Which issue(s) this PR fixes:**
Related to #2712 

**Screenshots:**
image

**Describe any introduced user-facing changes:**
- Users can now control if status updates are enabled for a specific contact method.
- Status updates will only activate on contact methods that receive an initial notification.

**Describe any introduced API changes:**
- New `statusUpdates` field on `UserContactMethod` and `enableStatusUpdates` on `UpdateUserContactMethodInput`

**Additional Context**
- Creating a Slack DM contact method should create an entry in `auth_subjects`
- Deleting the contact method should remove it
- This should be reflected in alert notifications (e.g., ack should work with a CM)
- "linking" via an alert ack button should result in a contact method being created

Slack DM functionality is behind a feature flag, test with `make start EXPERIMENTAL=slack-dm`

PR #3060, opened 2023-06-01
fix slack dm status updates

### LINKED ISSUE #2889: slack DM: status updates not working
When an alert is initially created, a Slack DM is sent. However, if the alert is modified, such as an update in the status, there is no update to the Slack DM for users to see.

Implement the ability for GoAlert to send status updates to Slack DM notifications.

N/A

**Additional context:**
N/A

mastercactapus:
**Description:**
This PR removes leftover code around `alert_status_log_contact_method_id`, causing problems with status updates.

A smoke test was added that reproduces the issue, fails before the fix, and passes after.

The status update cancel test was also updated to use the new API.

Fixes #2889 

The deprecated status method CM ID fields no longer do anything and always return an empty string.

**Additional Info:**
Root issue: Previous to the recent status update change in #2824, changing the `alert_status_log_contact_method_id` in the user table would retroactively update any pending status update messages to the new contact method.

With the new behavior, subscriptions are now per contact method. When a user previously had a contact method set up for status updates (before the update), all individual updates would still mistakenly be redirected to the original.

This meant that status updates for Slack would be redirected to the user's SMS, for instance. If status updates were disabled for the SMS contact method (in the new format), then the message would be canceled/deleted, which left no trace of why status updates weren't being sent to Slack.

This issue doesn't only impact Slack, but it was the most common/visible manifestation.

### DIFF
--- a/engine/message/db.go
+++ b/engine/message/db.go
@@ -48,7 +48,6 @@ type DB struct {
 
 	sentByCMType *sql.Stmt
 
-	updateCMStatusUpdate      *sql.Stmt
 	cleanupStatusUpdateOptOut *sql.Stmt
 
 	tempFail     *sql.Stmt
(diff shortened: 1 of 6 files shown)


tony-tvu (approved):
LGTM!

m17ch (approved):

PR #3072, opened 2023-06-06
ui/profile: fix test message dialog

mastercactapus:
**Description:**
Fixes sending test notifications. A recent change from the Luxon upgrade enforces strict parsing of DateTime strings. The test dialog had a silent error where an empty string was parsed.

This PR fixes the issue by only parsing the timestamp when available.

To reproduce/validate:
- On `master` (before this change), `Send Test` from a contact method will fail with broken page
- On this branch, the test notifications should be sent, and status printed in the dialog

### DIFF
--- a/web/src/app/users/SendTestDialog.tsx
+++ b/web/src/app/users/SendTestDialog.tsx
@@ -14,7 +14,7 @@ import {
 import toTitleCase from '../util/toTitleCase'
 import DialogContentError from '../dialogs/components/DialogContentError'
 import { DateTime } from 'luxon'
-import { ContactMethodType } from '../../schema'
+import { ContactMethodType, UserContactMethod } from '../../schema'
 
 const query = gql`
   query ($id: ID!) {
@@ -45,7 +45,9 @@ export default function SendTestDialog(
 
   const [sendTestStatus, sendTest] = useMutation(mutation)
 
-  const [{ data, fetching, error }] = useQuery({
+  const [{ data, fetching, error }] = useQuery<{
+    userContactMethod: UserContactMethod
+  }>({
     query,
     variables: {
       id: messageID,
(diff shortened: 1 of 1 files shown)


web/src/app/users/SendTestDialog.tsx
```diff
@@ -57,34 +59,33 @@ export default function SendTestDialog(
   const [now] = useState(DateTime.utc())
   const status = data?.userContactMethod?.lastTestMessageState?.status ?? ''
   const cmDestValue = data?.userContactMethod?.formattedValue ?? ''
-  const cmType: ContactMethodType = data?.userContactMethod?.type ?? ''
-  const lastTestVerifyAt = data?.userContactMethod?.lastTestVerifyAt ?? ''
-  const timeSinceLastVerified = now.diff(DateTime.fromISO(lastTestVerifyAt))
+  const cmType: ContactMethodType | '' = data?.userContactMethod.type ?? ''
```
andrewbenington, line 62:
```suggestion
  const cmType: ContactMethodType | '' = data?.userContactMethod?.type ?? ''
```
In typescript, optional chaining needs to continue to the end

KatieMSB (approved):

web/src/app/users/SendTestDialog.tsx
```diff
@@ -57,34 +59,33 @@ export default function SendTestDialog(
...
+  const lastSent = data?.userContactMethod?.lastTestVerifyAt
+    ? DateTime.fromISO(data.userContactMethod.lastTestVerifyAt)
+    : now.plus({ day: -1 })
   const fromValue =
     data?.userContactMethod?.lastTestMessageState?.formattedSrcValue ?? ''
   const errorMessage = (error?.message || sendTestStatus.error?.message) ?? ''
 
+  const hasSent =
+    Boolean(data?.userContactMethod.lastTestMessageState) &&
```
andrewbenington, line 71:
```suggestion
    Boolean(data?.userContactMethod?.lastTestMessageState) &&
```
Same as above

andrewbenington (changes requested):

web/src/app/users/SendTestDialog.tsx
andrewbenington: If `userContactMethod` isn't present on `data`, wouldn't this throw an error? because it's trying to access `lastTestMessageState` of `undefined`? Or are you saying that `userContactMethod` will always be present in this scenario

Forfold: I believe it tries them in order and once it hits undefined on a ? it returns the wanted result, or undefined if none given after the ??.

My IDE isn't throwing any errors or warning regarding this code

Forfold: This code reads to me that `userContactMethod` may not always be present on `data`, but if it is, `lastTestMessageState` will be available

Forfold (approved):

web/src/app/users/SendTestDialog.tsx
andrewbenington: Not getting the error on my IDE either. Not sure why it's different from the TypeScript playground, but I dismissed my change request

andrewbenington (approved):

PR #3073, opened 2023-06-06
sms: Allow multi-segment messages

### LINKED ISSUE #3058: Support Multi-Segment SMS and Improve Debugging for Alert Length Issues
When using GoAlert to send SMS alerts, some users are experiencing issues where the SMS fails due to size, especially when using bundles with long public URLs. The alert delivery fails instead of sending multi-segment messages for these cases. Additionally, debugging such scenarios is tricky as the logs show little helpful information, making it difficult to investigate these issues.

1. Add support for sending multi-segment messages in GoAlert when an SMS alert's content exceeds the max length for a single SMS.
2. Improve the logger to include the attempted message string that was too long for debugging purposes. Additionally, add the option to log the request body for POSTs to assist in debugging issues related to alert length.

Possible workarounds include:
- Disabling SMS links, which is not ideal as it limits functionality.
- Use a shorter domain just for the links by setting it to General.ShortURL. GoAlert will issue a redirect to the Public URL when clicking a link. However, considering possible domain constraints, this workaround may only suit some users.

**Additional context:**
This issue was brought up by community members discussing difficulties sending SMS alerts via Twilio when using alert bundle messages. Implementing these solutions could help improve the user experience when faced with alerts that exceed the maximum allowed length.

mastercactapus:
**Description:**
Updates the SMS code to fallback to sending multi-segment messages instead of refusing delivery.

The SMS code will truncate the details of an alert to try to fit it into a single segment (160 characters) message. When this isn't possible, it will attempt to truncate into two segments, and then three, etc...

The factors that control the minimum message size are configured at the admin level, application name, public URL, and related options. Fields such as alert summary and service name (user controlled) are subject to truncation.

This ensures misconfiguration by a user can not cause multiple message segments to go out but allows alerts to be sent in cases where the public URL, for instance, is too long for certain message types to fit.

Fixes #3058 

**Out of Scope:**
- No rate limit/throttling code changes are made to account for extra segments.
- No changes were made to logging as message rejection based on size is no longer possible

**Additional context**
- The entire SMS message is now rendered in the template. Previously, the application name was added as a prefix after calculation.


notification/twilio/alertsms.go
```diff
@@ -138,26 +139,53 @@ func normalizeGSM(str string) (s string) {
...
+		}
+
+		return result, nil
+	}
+}
+
 // renderAlertMessage will render a single-segment SMS for an Alert.
 //
 // Non-GSM characters will be replaced with '?' and fields will be
 // truncated (if needed) until the output is <= maxLen characters.
```
tony-tvu, line 172: Can these comments be updated since `maxLen` has been changed to `appName`? It's a bit unclear now.

tony-tvu (changes requested):

tony-tvu (approved):
LGTM

notification/twilio/alertsms.go
```diff
@@ -197,26 +228,27 @@ func renderAlertStatusMessage(maxLen int, a notification.AlertStatus) (string, e
 	return result, nil
 }
 
-// renderAlertBundleMessage will render a single-segment SMS for an Alert Bundle.
+// renderAlertBundleMessage will render a SMS message for an Alert Bundle.
```
Forfold, line 231: nit
```suggestion
// renderAlertBundleMessage will render an SMS message for an Alert Bundle.
```

tony-tvu (approved):

PR #3114, opened 2023-06-21
auth: Validate redir on return from auth

mastercactapus:
**Description:**
This PR adds additional validation on the `login_redir` cookie before issuing a redirect on successful login. The added validation is identical to the validation performed before setting the cookie value initially.

There are no changes in behavior: in user flows, the validation happens before the cookie is set. However, invalid cookie values will now be rejected.

**Additional Context**
- A recent integration test (oidc) validates that authentication works as it did before.


tony-tvu (approved):
LGTM

KatieMSB (approved):

PR #3132, opened 2023-06-27
go: update dependencies with major-version updates

mastercactapus:
**Description:**
Updates remaining Go packages to their latest versions, including major-version updates.

- the jwt library now takes validation options at the call to Parse, instead of manually calling validation methods

**Out of Scope:**
- The update to `github.com/jackc/pgx/v5` will be handled separately


app/cmd.go
```diff
@@ -354,13 +354,13 @@ Migration: %s (#%d)
...
 
-			t, err := toml.LoadFile(file)
+			data, err := os.ReadFile(file)
 			if err != nil {
 				return err
 			}
 
 			var cfg remotemonitor.Config
-			err = t.Unmarshal(&cfg)
+			err = toml.Unmarshal(data, &cfg)
```
Forfold, line 363: How is it able to see `toml` now that the import has been changed?

devtools/sendit/token.go
```diff
@@ -38,20 +38,12 @@ func TokenSubject(secret []byte, aud, token string) (string, error) {
 			return nil, jwt.ErrInvalidKeyType
 		}
 		return secret, nil
-	})
+	}, jwt.WithAudience(aud), jwt.WithIssuer(TokenIssuer), jwt.WithValidMethods([]string{"HS256"}))
```
Forfold, line 41:
```suggestion
	}, jwt.WithAudience(aud), jwt.WithIssuer(TokenIssuer), jwt.WithValidMethods([]string{"ES256"}))
```

Forfold: Not too familiar with signatures, but it looks like this should be "ES"?

app/cmd.go
mastercactapus: `toml` is the actual package name (i.e., in the `.go` files), so the prefix on the import line was redundant.

The change from `t` to `toml` here is due to the API change with v2 of the package.

Forfold (approved):

KatieMSB (approved):

PR #3133, opened 2023-06-28
add option to set noise reasons for an alert

### LINKED ISSUE #3003: Ability to Record Alert Usefulness (noise)
Currently, there is no way to differentiate between actionable alerts and false positives within GoAlert. It would be beneficial to have a mechanism to capture and report on the usefulness of alerts to improve response efficiency.

- Allow users to mark alerts as one of three states: actionable, false positive, or none (default).

- Teams can manually review alerts regularly, but this may not provide the same level of context and efficiency as marking alerts in the moment.

**Out of Scope:**
- Implement a reporting feature that provides an overview of alerts in each of the three states to help teams identify trends and improve alert management.

**Additional context:**
- This feature would help teams better understand the effectiveness of their alerting system and identify areas for improvement.

Forfold:
**Description:**
- Adds the ability to set notes on an alert to document signal to noise.
- When set, a notice will appear at the top of the alert with the option to undo.
- Moves the alert update buttons to the left (Ack, Escalate, Close), making them primary contained text buttons instead of secondary icon buttons.

- Closes #3003

**Out of Scope:**
- Updating alert metrics table & graph to include section for alerts with notes set

**Screenshots:**

**New alert, no action taken**

Screenshot 2023-06-28 at 11 03 57 AM

**New alert, some notes checked**

Screenshot 2023-06-28 at 11 04 34 AM

**Notes set on alert, notice shown**

Screenshot 2023-06-28 at 11 04 42 AM

- Added type `feedback` to `alert` with the field `note`


tony-tvu (dismissed):
Functionally checks outs :rocket:

alert/store.go
```diff
@@ -202,6 +205,21 @@ func NewStore(ctx context.Context, db *sql.DB, logDB *alertlog.Store) (*Store, e
 			FROM services
 			WHERE id = $1
 		`),
+
+		feedback: p(`
```
mastercactapus, line 209: Can we use `sqlc` for this since it's new code?

mastercactapus (changes requested):

alert/store.go
Forfold: Yeah sorry, this is what's most familiar to me. Can do.

andrewbenington (approved):
looks good to me, plus the sql files are more readable now

alert/store.go
```diff
@@ -791,3 +791,41 @@ func (s *Store) State(ctx context.Context, alertIDs []int) ([]State, error) {
...
+		return err
+	}
+
+	f, err := (s).Feedback(ctx, feedback.AlertID)
+	if err != nil {
+		return err
+	}
+	f.NoiseReason = feedback.NoiseReason
+
+	err = gadb.New(s.db).SetAlertFeedback(ctx, gadb.SetAlertFeedbackParams{
```
mastercactapus, line 822: We need to validate the content and length of `NoiseReason` before passing it to the DB (e.g., `validate.Text`).

graphql2/graphqlapp/alert.go
```diff
@@ -361,6 +360,28 @@ func (m *Mutation) CreateAlert(ctx context.Context, input graphql2.CreateAlertIn
 	return m.AlertStore.Create(ctx, a)
 }
 
+func (a *Alert) NoiseReason(ctx context.Context, raw *alert.Alert) (*string, error) {
+	am, err := a.AlertStore.Feedback(ctx, raw.ID)
+	if err != nil {
+		return nil, err
+	}
+	return &am.NoiseReason, nil
```
mastercactapus, line 368: According to the schema, this is a null-able field, so we should probably return `nil` here if `NoiseReason == ""`

alert/store.go
```diff
@@ -791,3 +791,41 @@ func (s *Store) State(ctx context.Context, alertIDs []int) ([]State, error) {
 
 	return list, nil
 }
+
+func (s *Store) Feedback(ctx context.Context, alertID int) (f Feedback, err error) {
```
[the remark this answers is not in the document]
Forfold: Why check if we allow anything?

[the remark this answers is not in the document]
mastercactapus: It validates that it has some sort of authentication (e.g., they must be logged in, vs. open to the public). Unauthenticated calls will be rejected by `Any` since they don't have any type of access.

alert/store.go
```diff
@@ -791,3 +791,39 @@ func (s *Store) State(ctx context.Context, alertIDs []int) ([]State, error) {
...
+	}, err
+}
+
+func (s Store) UpdateFeedback(ctx context.Context, feedback *Feedback) error {
+	err := permission.LimitCheckAny(ctx, permission.System, permission.User)
+	if err != nil {
+		return err
+	}
+
+	err = validate.Text("Noise Reason", feedback.NoiseReason, 1, 255)
```
mastercactapus, line 815:
```suggestion
	err = validate.Text("NoiseReason", feedback.NoiseReason, 1, 255)
```

andrewbenington (dismissed):
👍

alert/store.go
```diff
@@ -791,3 +791,44 @@ func (s *Store) State(ctx context.Context, alertIDs []int) ([]State, error) {
...
 }
+
+func (s *Store) Feedback(ctx context.Context, alertID int) (*Feedback, error) {
+	err := permission.LimitCheckAny(ctx, permission.System, permission.User)
+	if err != nil {
+		return nil, err
+	}
+
+	row, err := gadb.New(s.db).AlertFeedback(ctx, int64(alertID))
+	if errors.Is(err, sql.ErrNoRows) {
```
mastercactapus, line 802: Missing `err != nil` check after this (would be a silent failure)

migrate/migrations/20230616110941-add-alerts-feedback.sql
```diff
@@ -0,0 +1,9 @@
+-- +migrate Up
+CREATE TABLE alert_feedback
+(
+    alert_id BIGINT PRIMARY KEY REFERENCES alerts (id) ON DELETE CASCADE,
+    noise_reason TEXT
```
mastercactapus, line 5: Should we make this not null?

It's not immediately obvious, but we're actually handling three different empty states in code:
- no rows
- noise_reason is `null`
- noise_reason is `""`

As written in this PR, only no rows and `""` are possible. If we need to migrate this in the future, it may be good to enforce consistency, plus getting rid of the `null` lets us use a Go `string` instead of `sql.NullString` that we aren't using.

mastercactapus (changes requested):

mastercactapus (approved):

PR #3138, opened 2023-06-30
feat: show disabled notification options

### LINKED ISSUE #3008: Show Disabled Email, SMS, and Voice Options with Explanation when Unavailable
Currently, the options for Email, SMS, and Voice integration keys and contact methods are hidden when the required configurations (Mailgun and Twilio) are not enabled. This can lead to confusion for users and make it difficult to understand why the options are not available.

Instead of hiding the options for Email, SMS, and Voice integration keys and contact methods when they are not available, display them as disabled/greyed out, along with a message explaining why they are unavailable. This would help users understand the need to configure Mailgun and Twilio in the admin panel and prevent confusion when the UI does not match the documentation.

An alternative could be to provide more explicit documentation and instructions, but this still leaves room for confusion and does not directly address the issue in the user interface.

**Additional context:**
This feature request was prompted by the bug report discussed in #2631, where a user was unsure how to create an Email integration key because the option was not visible in their environment. The proposed solution aims to make the requirements for enabling these options more discoverable.

andrewbenington:

**Description:**
User contact method and on-call notification channel selections now show unavailable options in the dropdown, disabled with tooltip explanations.
Disabled menu items are always listed last.

Resolves #3008 

**Screenshots:**
Webhooks enabled, Slack disabled:
Screenshot 2023-07-05 at 2 11 47 PM
Slack enabled, webhooks disabled:
Screenshot 2023-07-05 at 2 09 58 PM
User contact methods:
Screenshot 2023-06-30 at 12 38 58 PM
Screenshot 2023-06-30 at 12 38 53 PM

Users now will see the disabled options in dropdowns, but will still not be able to select them without administrator action.


mastercactapus (approved):

web/src/app/users/UserContactMethodForm.tsx
```diff
@@ -176,17 +210,7 @@ export default function UserContactMethodForm(
...
-              <MenuItem value='VOICE'>VOICE</MenuItem>
-            )}
-            {(edit || emailEnabled) && <MenuItem value='EMAIL'>EMAIL</MenuItem>}
-            {(edit || webhookEnabled) && (
-              <MenuItem value='WEBHOOK'>WEBHOOK</MenuItem>
-            )}
-            {(edit || slackEnabled) && (
-              <MenuItem value='SLACK_DM'>SLACK DM</MenuItem>
-            )}
+            {contactMethods.map(renderMenuItem)}
```
Forfold, line 213: This call confused me for a second. It feels like `renderMenuItem` should be called like a component, such as `<MenuItemWithDisable />`, since the function being called is essentially a functional JSX component if it was named differently

`contactMethods.map((cm) => <MenuItemWithDisable contactMethod={cm} />)` or something.

Just my 2¢

andrewbenington: I absolutely agree, and I tried to get that to work a lot of different ways. The tricky part is that the Form is looking for a component with the right `value` prop, and it seems like the only elegant way for it to work is for this to be a function. It didn't work as a component.

https://stackoverflow.com/questions/68378474/mui-select-custom-menuitem-not-working-properly
^^ this stackoverflow thread covers the exact issue

Forfold (approved):

mastercactapus (approved):

Forfold (approved):

PR #3156, opened 2023-07-10
dev: wait for goalert to start before grpc tool

mastercactapus:
**Description:**
Fixes a race condition with `make start` where the grpcui tool will exit (causing `make start` to fail) if GoAlert doesn't bind the port in time.

The fix is to use the `waitfor` tool to simply wait for GoAlert first.


allending313 (approved):
lgtm

cmarquis (approved):

PR #3158, opened 2023-07-10
auth: use samesite=lax for oauth cookies

mastercactapus:
**Description:**
Fixes a regression that prevents login when redirecting from an OIDC server by setting `SameSite=Lax` for OAuth cookies, and retaining `SameSite=Strict` for the session cookie.

**Additional Info:**
1. `SameSite=None`: Cookies will be sent in all contexts. This was how cookies worked by default prior to the SameSite attribute implementation.

2. `SameSite=Lax`: Cookies are not sent on normal cross-site subrequests (for example, to load images or frames into a third-party site), but __are sent when a user is navigating to the origin site__. This is the default value in modern browsers if the attribute is not specified.

3. `SameSite=Strict`: Cookies will only be sent in a first-party context and not be sent along with requests initiated by third-party websites. This means that the cookies won't be sent if the user is not already on your site.

### DIFF
--- a/auth/cookies.go
+++ b/auth/cookies.go
@@ -10,12 +10,12 @@ import (
 )
 
 // SetCookie will set a cookie value for all API prefixes, respecting the current config parameters.
-func SetCookie(w http.ResponseWriter, req *http.Request, name, value string) {
-	SetCookieAge(w, req, name, value, 0)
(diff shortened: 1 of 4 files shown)


allending313 (approved):
lgtm!

Forfold (approved):
lgtm & functionally tested

PR #3162, opened 2023-07-11
feat: add smoke test for on call notification webhooks

allending313:

**Description:**
Added additional smoke test ensuring that webhooks can receive notifications from a schedule's on-call notification rule.

### DIFF
--- a/test/smoke/webhookoncall_test.go
+++ b/test/smoke/webhookoncall_test.go
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 	"github.com/target/goalert/expflag"
 	"github.com/target/goalert/test/smoke/harness"
 )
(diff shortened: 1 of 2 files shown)


test/smoke/webhookoncall_test.go
```diff
@@ -0,0 +1,84 @@
...
+	ScheduleID   string
+	ScheduleName string
+	ScheduleURL  string
+}
+
+// TestWebhookOnCallNotification tests that the configured rule sends the intended notification to the webhook.
+func TestWebhookOnCallNotification(t *testing.T) {
+	t.Parallel()
+
+	ch := make(chan POSTDataOnCallNotification)
```
mastercactapus, line 36: When setting up a receiver like this, add a buffer if possible. In this case, it will alow the handler goroutine to exit when it's done, even if the test were to exit early or something. (otherwise it will block forever on `ch <- alert`)
```suggestion
	ch := make(chan POSTDataOnCallNotification, 1)
```

test/smoke/webhookoncall_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var alert POSTDataOnCallNotification
+
+		data, err := io.ReadAll(r.Body)
+		require.Nil(t, err)
```
mastercactapus, line 42: Use `assert.*` in here -- `require.*` isn't allowed in background goroutines (like that of an HTTP request handler) since it uses `panic` internally.

mastercactapus (changes requested):

test/smoke/webhookoncall_test.go
allending313: this would apply to the other webhook test as well right? https://github.com/target/goalert/blob/05efe4a36d0e7ce87e0b1c825cd170d87211ef68/test/smoke/webhook_test.go#L25C25-L25C25

mastercactapus: yeah good catch!

Forfold (approved):
lgtm!

mastercactapus (approved):

PR #3167, opened 2023-07-12
add API for listing supported integration key types

### LINKED ISSUE #3166: Drive Available Integration Key Types by API
The existing implementation limits the addition of multiple providers behind a single integration key type. This problem adversely impacts forward progress, like with the SMTP integration (issue #3136) for instance. Also, adding new key types and working with experimental flags is more challenging due to current hard-coding logic.

Enhance the backend to provide a GraphQL query endpoint of available integration key types, replacing the existing approach of passing this as a config to the UI. This will simplify the UI, make the code more maintainable, and pave the way to introduce plugins. 

**Additional context:**
The enhancement should be straightforward. We can provide an array of name, type id, description, and a flag indicating if the type is enabled as a baseline.

mastercactapus:
**Description:**
Adds a GraphQL query for listing available integration key types.

Closes #3166 

When Mailgun is disabled, Email is now listed as an option, but disabled when creating an integration key.
image

Added a new GraphQL query:
```graphql
type Query {
  integrationKeyTypes: [IntegrationKeyTypeInfo!]!
}

type IntegrationKeyTypeInfo {
  id: ID!

  # User-displayable name of the integration key type.
  name: String!

  # User-displayable description of the integration key value (i.e., copy/paste instructions).
  label: String!

  # Indicates if the type is currently enabled.
  enabled: Boolean!
}
```


web/src/app/util/useIntKeyTypes.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export function useIntKeyTypes(): Array<IntegrationKeyTypeInfo> {
+  const [{ data, fetching, error }] = useQuery({
+    query,
+    requestPolicy: 'cache-first',
+  })
+  if (fetching) return []
+  if (error) return []
+
+  return data.integrationKeyTypes || []
```
Forfold, line 23: Might as well return the useQuery so we can do something with the loading and error states? Or not use this hook at all since it's only used in 2 spots and just making a query call. They could both handle loading states differently I imagine

mastercactapus: In this case, there are a few important things not immediately obvious:
- This needs to handle caching differently than the rest of the application (aside from config). Whether it uses `useQuery` or `useContext` under the covers shouldn't matter to the caller (separation of concerns). 10 integration keys using this hook must not result in 10 requests every 3 seconds.
- We will be doing the same thing for contact methods, schedule notifications, and escalation policy steps in the future (i.e., plugins), and expecting `useQuery` calls with identical settings and use throughout the application won't be maintainable; so consider this an experimental precedent. Imagine if instead of `useConfig` we had `useQuery` everywhere without type safety and having to remember and maintain cache settings in every spot we use config.
- As this is configuration, not user data, it's following how we're handling configuration in the application -- by separating the implementation from its use. The caller cares about the config, not how we get it.

Today it's hard-coded, tomorrow it could be embedded in the page, this PR moves it to an API call. Regardless of the implementation, the usage of config is the same, so having some abstraction there keeps us from having to refactor unrelated chunks of the application when we need to make a change we didn't predict.

That all said -- maybe a better implementation would be to mix this into the config helper (i.e., add it to that query) and `useContext`.

As for the separate hook -- that's separation of concerns and setting things up for the future:
- `useIntegrationKeyTypes`
- `useEPStepTypes`
- `useSchedNotificationTypes`
- `useCMTypes`

While each one may only be used in 2-4 places, the purpose of breaking them apart is to separate the internals of the building block (i.e., config fetching, logic, section, and structure) from the thing being built (e.g., a form, a list, etc...). Someone could make a hook for contact method types that returns a constant and start refactoring the UI while someone starts working on backend logic to feed that through.

I hope that helps explain the reasoning behind the design. I think I will switch this to use context, though, as that will help ensure we don't end up with a bunch of extra calls for config data.

mastercactapus: Your points about the necessity of a more complex structure are well taken, but I would like to clarify the approach. When we talk about separating concerns and using abstraction where it makes sense, the goal isn't just to complicate things for the sake of it. Rather, we aim to streamline the process of working on different parts of the code.

Consider the case of updating something like a form, especially when we're dealing with configuration data and items such as labels. If we don't adopt a modular approach, anyone who wants to make a change would need to understand everything from the backend logic, the API, to the query, to the rendering method. That would be a steep learning curve for anyone, especially if they aren't working on GoAlert full-time or are new to the project.

Our goal is to strike a balance. We want to keep things modular enough that individual pieces can be worked on without needing to untangle the entire application's workings. The adage from the Go programming language you mentioned is often interpreted as "a little copying is better than a little dependency." We're not talking about introducing new dependencies here. Instead, we're aiming for a design that helps us manage complexity and keep the code maintainable and approachable for everyone.

Therefore, while I appreciate your perspective and understand your preference for a more direct approach, I believe a structure like what I proposed is justified for this application. I hope this gives you a better understanding of the reasoning behind the decision to use typed hooks for loading configuration data rather than raw queries.

KatieMSB (approved):

tony-tvu (approved):
LGTM

Issue #3180, opened 2023-07-19
GoAlert fails on Postgres 14 due to SQL scan error


mastercactapus:
**Describe the Bug:**
Using GoAlert on Postgres 14 (or possibly newer versions) causes some queries to fail, with an error indicating a failure in SQL scan. Specifically, the error `append alert log: lookup heartbeat monitor interval by ID: sql: Scan error on column index 0, name "?column?": converting driver.Value type string ("60.0000000000000000") to a int: invalid syntax` is observed.

**Steps to Reproduce:**
1. Setup a dev environment with Postgres 14 or newer database.
2. Run `make smoketest`.

**Expected Behavior:**
The `make smoketest` command should execute successfully with no errors.

**Observed Behavior:**
The `make smoketest` command throws a SQL scan error on certain queries.

**Application Version:**
This issue appears to affect all GoAlert versions.

**Additional Context:**
The problem seems to be only occurring with Postgres 14 and newer versions. The application works fine with Postgres 13 which is the recommended version.

mastercactapus:
A good way to start is to look for errors/failures with the smoke & integration tests and convert the problematic queries to use `sqlc` (what we're using for new work anyhow). The generated code/scanning methods seem to handle whatever discrepancy is causing this.

mastercactapus:
This was fixed

PR #3181, opened 2023-07-19
alert/alertlog: migrate insert queries to use sqlc

mastercactapus:
**Description:**
This PR moves some problematic queries (see: #3180) to use `sqlc`. This transition will be done in stages.

The queries were moved as-is with minor changes (e.g., removing use of `ANY` and using `bigint` instead of `int`) and formatting.

Lastly, `sqlc` has been updated to the latest version.

Part of #3180


KatieMSB (approved):

allending313 (approved):
lgtm!

PR #3229, opened 2023-08-15
ui: add user shifts calendar to profile

### LINKED ISSUE #3164: Single user calendar view across schedules
For users who are in multiple rotations, such as `primary` and `secondary`, it can difficult to visualize availability or on-call coverage for a specific user.

It would be nice to have a calendar view which consolidates all active schedules for just a single user; accessible perhaps via another `quick link` or calendar tab on a specific user's page.

The current method to get around this is to have multiple side-by-side tabs, one for each schedule, but this also shows all other users/overrides and can get messy quickly.

Forfold:
**Description:**
- This PR introduces a calendar (found within the Profile Quick Links) that will by default display an aggregate of all schedule shifts the user currently 

- Closes #3164

- Shows personalized calendar under contact methods and notifications rules on the profile (when not mobile)
- ~~Schedule Calendar Subscriptions` quick link renamed to `External Calendar Subscriptions`~~
- ~~Updates quick link URL `/profile/sessions` to `/profile/active-sessions` to match quick link text~~

**Screenshots:**

Screenshot 2023-10-27 at 2 44 47 PM

Screenshot 2023-10-27 at 2 46 08 PM


graphql2/schema.graphql
```diff
@@ -93,6 +93,10 @@ type Query {
   # Returns a single user override with the given ID.
   userOverride(id: ID!): UserOverride
 
+  # Returns shifts for a given user ID.
+  # If no ID is provided, the current user's context is used
+  userShifts(input: UserShiftsOptions): [OnCallShift!]!
```
Forfold, line 98: instead of `userShifts`, this can be `shifts` inside of `User`

Forfold:
Found two related issues when working on this with @KatieMSB, though they can probably be fixed as separate bodies of work:
- #3462
- #3463

KatieMSB (approved):

Forfold:
Cypress passing locally but failing in CI again 🤔

KatieMSB (dismissed):

web/src/app/users/UserContactMethodList.tsx
```diff
@@ -78,7 +78,7 @@ export default function UserContactMethodList(
   if (!data?.user) return <ObjectNotFound type='user' />
   if (error) return <GenericError error={error.message} />
 
-  const contactMethods = data.user.contactMethods
+  const contactMethods = data?.user?.contactMethods ?? []
```
mastercactapus, line 81: Is this necessary with suspense? It _should_ be all-or-nothing (under the covers suspense throws if the data isn't loaded)

web/src/app/util/calendar/CalendarEventWrapper.tsx
```diff
@@ -53,6 +55,10 @@ export default function ScheduleCalendarEventWrapper({
   const [showEditDialog, setShowEditDialog] = useState('')
   const [showDeleteDialog, setShowDeleteDialog] = useState('')
 
+  // for showing "Visit Schedule" buttons when on user profile
+  const [path] = useLocation()
+  const onProfile = path.includes('users') || path.includes('profile')
```
mastercactapus, line 60: Can we do this through a prop like `showScheduleLink` rather than relying on magic?

mastercactapus (dismissed):
I appreciate the walkthrough; In the future, let's try to do renames/cleanup/refactors/prereq-changes in separate PRs so they aren't tied to the new code. Great work!

web/src/app/users/UserContactMethodList.tsx
Forfold: I was getting console errors. I'll remove and try again.

web/src/app/users/UserShiftsCalendar.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  const [queryStart, queryEnd] = weekly
+    ? [
+        getStartOfWeek(DateTime.fromISO(start)).toISO(),
+        getEndOfWeek(DateTime.fromISO(start)).toISO(),
+      ]
+    : [
+        getStartOfWeek(DateTime.fromISO(start).startOf('month')).toISO(),
+        getEndOfWeek(DateTime.fromISO(start).endOf('month')).toISO(),
+      ]
```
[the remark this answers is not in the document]
Forfold: ok

web/src/app/util/calendar/CalendarEventWrapper.tsx
Forfold: what is magical about this

web/src/app/users/UserNotificationRuleList.tsx
```diff
@@ -66,7 +65,7 @@ export default function UserNotificationRuleList(props: {
     return <ObjectNotFound type='notifcation rules list' />
   if (error) return <GenericError error={error.message} />
 
-  const { user }: { user: User } = data
+  const user = data?.user
```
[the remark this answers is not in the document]
Forfold: See above.

web/src/app/util/calendar/CalendarEventWrapper.tsx
Forfold: This component is called like this:

```
components={{
            // @ts-expect-error Property 'children' does not exist on type - yes it does
            eventWrapper: CalendarEventWrapper,
            toolbar: () => null,
          }}
```

Where recently you removed it being called as `eventWrapper: ({ children }) => (<CalendarEventWrapper />)` or something similar so it wouldn't render multiple times.

ethan-haynes (dismissed):

KatieMSB (dismissed):

ethan-haynes (approved):

web/src/app/util/calendar/CalendarEventWrapper.tsx
mastercactapus: When I refer to the approach as "magical," I'm emphasizing the indirect nature of rendering based on the URL path, rather than through explicit props. The core of my concern is about maintainability and clarity of the component's behavior.

Using explicit props like showScheduleLink for rendering decisions aligns better with best practices in component design. It ensures that the component's behavior is directly controlled by its parent, making it more predictable and easier to manage. This approach is especially beneficial in scenarios like routing changes, URL modifications, and testing. If we rely on the URL path, it introduces an additional layer of complexity and fragility. Changes in routing or URL structure would directly impact the component, potentially leading to unforeseen issues.

In essence, by using props, we're maintaining a clear contract between the parent and the child component. It makes the data flow and rendering logic more transparent and cohesive. This not only aids in current understanding and maintenance but also simplifies any future modifications, whether it's changing routes, refactoring, or writing tests.

Forfold: https://github.com/target/goalert/commit/57a61e89af988a578d09b58d0e962c1bb09cf945#diff-aa1aba9c2ce6c7a6f0e6b6771d34bdec24c2c56b1be30a0e264712cebbd22d4aL387

mastercactapus: Swap the component, rather than render inline:
```
            eventWrapper: props.showScheduleLink
              ? CalendarEventWrapperWithLink
              : CalendarEventWrapper,
```

90% of the code in `CalendarEventWrapper` is about temp schedules and overrides, so rather than adding more complexity to it, we could just make a simple one for the new event type.

Alternatively, you could do something like this:
```

function CalendarEventWrapperWithLink(
  props: ScheduleCalendarEventWrapperProps,
): JSX.Element {
  return <CalendarEventWrapper {...props} showScheduleLink />
}
```

However, `CalendarEventWrapper` will still need all the conditionals throughout since it's being used in two different ways.

Either option will solve the problem without re-introducing the render bug.

Forfold: So... Swap one complexity for the other. I'll make a wrapper component, sure.

Forfold: Done. Resolving.

KatieMSB (dismissed):

tony-tvu (approved):
Functionally working as expected :thumbsup:

tony-tvu (approved):

PR #3231, opened 2023-08-16
gql: add field "assignedSchedules" to type "User"

Forfold:
**Description:**
Adds the field `assignedSchedules` to type `User` so that each schedule's shifts can be fetched for a given user. This will be especially important for adding a calendar to a user's profile, so they know the context of each shift they are on-call for.

This PR is precedent to #3229 and is a part of #3164

**Screenshots:**

Screenshot 2023-10-11 at 11 47 02 AM

**Other changes:**
Also adds the argument `user` on query `shifts` to use for filtering:
```
type Schedule {
  // ...
  shifts(start: ISOTimestamp!, end: ISOTimestamp!, userIDs: [ID!]): [OnCallShift!]!
  // ...
}
```


andrewbenington (changes requested):

oncall/state.go
```diff
@@ -124,33 +125,44 @@ func (s *state) CalculateShifts(start, end time.Time) []Shift {
...
 		}
 		now := time.Unix(t.Unix(), 0)
 		for _, id := range userIDs {
 			stillOnCall[id] = true
-			s := isOnCall[id]
-			if s != nil {
+			o := isOnCall[id]
+			if o != nil {
 				continue
 			}
```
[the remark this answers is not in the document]
Forfold: I only did a rename because it was using `s`, which is the same as the function's `(s State)` variable, so I had to choose a different letter to access the state within that scope

andrewbenington (dismissed):

allending313 (approved):
lgtm!

mastercactapus:
I propose we expand the existing `shifts` field on `Schedule` by adding a `userIDs` argument that will filter the returned shifts to the provided list.

Secondly, an `assignedSchedules` field on `User` that will return the list of schedules the user is associated with (i.e., assigned directly, on a future add_override /w end_time in the future, or participant of a rotation assigned to the schedule)

```graphql
type Schedule {
  // ...
  shifts(start: ISOTimestamp!, end: ISOTimestamp!, userIDs: [ID!]): [OnCallShift!]!
  // ...
}

type User {
  // ...
  assignedSchedules: [Schedule!]
  // ...
} 
```

Forfold:
When attempting to find any temporary schedules, the application gets stuck in an infinite timeout. No idea why. It happens on this line:

`err := s.grp.Get(ctx, timeKey("userIDs", time.Minute), groupcache.AllocatingByteSliceSink(&data))`

Forfold:
Blocked by #3286

oncall/store.go
```diff
@@ -196,7 +197,7 @@ func (s *Store) OnCallUsersBySchedule(ctx context.Context, scheduleID string) ([
 }
 
 // HistoryBySchedule will return the list of shifts that overlap the start and end time for the given schedule.
-func (s *Store) HistoryBySchedule(ctx context.Context, scheduleID string, start, end time.Time) ([]Shift, error) {
+func (s *Store) HistoryBySchedule(ctx context.Context, scheduleID string, start, end time.Time, userIDs []string) ([]Shift, error) {
```
mastercactapus, line 200: I'm thinking since this is already a complicated method, we move the user ID filtering to the caller (graphql code).

It's a relatively simple thing to do and then we can avoid having to update methods, tests, etc...

Filtering is done at the very end, and is how we're already doing it elsewhere like calendar subscriptions.

schedule/rotation/store.go
```diff
@@ -259,6 +266,42 @@ func (s *Store) FindMany(ctx context.Context, ids []string) ([]Rotation, error)
 	return result, nil
 }
 
+func (s *Store) FindManyByUserID(ctx context.Context, tx *sql.Tx, userID string) (rotationIDs []string, err error) {
```
mastercactapus, line 269: This should return `[]Rotation` to match things like `FindMany` also to reduce round-trips.

Could be done in a number of ways:
```sql
select * from rotations
where id = any(select rotation_id from rotation_participants where user_id = $1)
```

graphql2/graphqlapp/schedule.go
```diff
@@ -81,14 +81,22 @@ func (q *Query) Schedule(ctx context.Context, id string) (*schedule.Schedule, er
...
 	if end.Before(start) {
 		return nil, validation.NewFieldError("EndTime", "must be after StartTime")
 	}
 	if end.After(start.AddDate(0, 0, 50)) {
 		return nil, validation.NewFieldError("EndTime", "cannot be more than 50 days past StartTime")
 	}
-	return s.OnCallStore.HistoryBySchedule(ctx, raw.ID, start, end)
+
+	for _, userID := range userIDs {
+		err := validate.UUID("UserID", userID)
```
mastercactapus, line 93: Use `validate.ManyUUID` instead with a limit on the length

graphql2/schema.graphql
```diff
@@ -1404,6 +1404,8 @@ type User {
   onCallSteps: [EscalationPolicyStep!]!
 
   isFavorite: Boolean!
+
+  assignedSchedules: [Schedule!]
```
mastercactapus, line 1408:
```suggestion
  assignedSchedules: [Schedule!]!
```

schedule/rotation/store.go
```diff
@@ -94,6 +95,12 @@ func NewStore(ctx context.Context, db *sql.DB) (*Store, error) {
 			WHERE r.id = ANY($1)
 		`),
 
+		findManyByUserID: p.P(`
+			SELECT rp.rotation_id
+			FROM rotation_participants rp
+			WHERE rp.user_id = $1
+		`),
```
mastercactapus, line 102: Can we put this in a sqlc query since it's new?

schedule/store.go
```diff
@@ -115,6 +127,48 @@ func (store *Store) FindMany(ctx context.Context, ids []string) ([]Schedule, err
 
 	return result, nil
 }
+
+func (store *Store) FindMany(ctx context.Context, ids []string) ([]Schedule, error) {
+	return store.FindManyTx(ctx, nil, ids)
+}
+
+func (store *Store) FindManyByAssignments(ctx context.Context, tx *sql.Tx, userID string, rotationIDs []string) (scheduleIDs []string, err error) {
```
mastercactapus, line 135: Same as rotations, the signature should be `[]Schedule, error` to reduce round-trips and to match the other methods

mastercactapus (changes requested):

graphql2/graphqlapp/schedule.go
Forfold: What is the benefit of that?

schedule/rotation/store.go
Forfold: would it be okay to return a `[]Rotation` slice, where only the `ID` is set on each el?

graphql2/graphqlapp/schedule.go
mastercactapus: It exists specifically for this type of validation:
- It validates the length (i.e., a max number of UserIDs instead of unlimited)
- Consistency with all other code in the repo taking a slice of UUIDs already using it
- Returns the proper field name (i.e., `UserID[23]`) when there is an error on an individual value

schedule/rotation/store.go
```diff
@@ -259,6 +260,47 @@ func (s *Store) FindMany(ctx context.Context, ids []string) ([]Rotation, error)
 	return result, nil
 }
 
+func (s *Store) FindManyByUserID(ctx context.Context, tx *sql.Tx, userID string) ([]Rotation, error) {
```
mastercactapus, line 263: see above: may not be needed

schedule/store.go
```diff
@@ -115,6 +129,63 @@ func (store *Store) FindMany(ctx context.Context, ids []string) ([]Schedule, err
...
+	}
+
+	err = validate.UUID("UserID", userID)
+	if err != nil {
+		return nil, err
+	}
+	uid, err := uuid.Parse(userID)
+	if err != nil {
+		return nil, err
+	}
```
mastercactapus, line 150:
```suggestion
	uid, err := validate.ParseUUID("UserID", userID)
	if err != nil {
		return nil, err
	}
```

Alternatively, we could just require the `userID` to be `uuid.UUID` instead of string and parse it in the GraphQL layer

schedule/store.go
```diff
@@ -115,6 +129,63 @@ func (store *Store) FindMany(ctx context.Context, ids []string) ([]Schedule, err
...
+	var rids = []uuid.UUID{}
+	for _, r := range rotationIDs {
+		rid, err := uuid.Parse(r)
+		if err != nil {
+			return nil, err
+		}
+		rids = append(rids, rid)
+	}
+
+	rows, err := gadb.New(store.db).FindManyByAssignments(ctx, gadb.FindManyByAssignmentsParams{
```
mastercactapus, line 161: Use/require the DB to be passed into this method -- as it is it accepts a transaction but uses the connection pool directly (could trigger a deadlock scenario if the pool is exhausted)

schedule/store.go
```diff
@@ -115,6 +129,63 @@ func (store *Store) FindMany(ctx context.Context, ids []string) ([]Schedule, err
 
 	return result, nil
 }
+
+func (store *Store) FindMany(ctx context.Context, ids []string) ([]Schedule, error) {
+	return store.FindManyTx(ctx, nil, ids)
+}
+
+func (store *Store) FindManyByAssignments(ctx context.Context, tx *sql.Tx, userID string, rotationIDs []string) ([]Schedule, error) {
```
mastercactapus, line 137:
```suggestion
func (store *Store) FindManyByAssignments(ctx context.Context, db gadb.DBTX, userID string, rotationIDs []string) ([]Schedule, error) {
```

schedule/store.go
```diff
@@ -115,6 +129,63 @@ func (store *Store) FindMany(ctx context.Context, ids []string) ([]Schedule, err
...
+
+	if errors.Is(err, sql.ErrNoRows) {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	var result []Schedule
+
```
mastercactapus, line 177: nit: blank line between decl. and loop
```suggestion
```

mastercactapus (changes requested):

schedule/queries.sql
```diff
@@ -0,0 +1,3 @@
+-- name: FindManyByAssignments :many
+SELECT * FROM schedules
+WHERE id = any(SELECT schedule_id FROM schedule_rules WHERE tgt_user_id = $1 OR tgt_rotation_id = any(@TgtRotationIDs::UUID[]));
```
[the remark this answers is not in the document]
Forfold: Yeah this works! The intention is the same, but looks like from your example we can achieve it easily with an inner select instead.

KatieMSB (approved):

graphql2/graphqlapp/schedule.go
```diff
@@ -81,14 +82,37 @@ func (q *Query) Schedule(ctx context.Context, id string) (*schedule.Schedule, er
...
 		return nil, validation.NewFieldError("EndTime", "cannot be more than 50 days past StartTime")
 	}
-	return s.OnCallStore.HistoryBySchedule(ctx, raw.ID, start, end)
+
+	err := validate.ManyUUID("userID", userIDs, 50)
+	if err != nil {
+		return nil, err
+	}
+
+	shifts, err := s.OnCallStore.HistoryBySchedule(ctx, raw.ID, start, end)
```
ethan-haynes, line 98: Small comment, but should we avoid "magic numbers" like the 50 used here and above in `end.After(start.AddDate(0, 0, 50`?

ethan-haynes (approved):
LGTM 👍

graphql2/graphqlapp/schedule.go
Forfold: Looks like it's an issue with most of our calls to `validate.ManyUUID`, so I'm going to document that for its own issue, if possible

PR #3234, opened 2023-08-17
fix: prefix search

### LINKED ISSUE #3217: bug: user search select not displaying exact search results
when searching for a user in a large set of usernames that are similar, there's a chance you won't see the proper user in the dropdown.

In this instance, I'll use my own name as an example, we want to search for `Nathaniel.N`

but what shows are the following users:

- Random.Nathaniel
- Random.NathanielRandom
- NathanielRand.Koch
- NathanielRand.Smith
- NathanielRand.Weeks

andrewbenington:

**Description:**

- Searching list pages (users, escalation policies, rotations, schedules, notifications, and services) only shows results that match the order of the search terms.
- Words can be skipped (e.g. the search "Serv Two" will return both "Service Two" and "Service One Two")

Before:
Screenshot 2023-08-17 at 12 57 45 PM
Screenshot 2023-08-17 at 12 57 35 PM

After:
Screenshot 2023-08-17 at 12 20 49 PM
Screenshot 2023-08-17 at 12 20 55 PM

Fixes #3217 

Anything using `prefixSearch` now uses `orderedPrefixSearch`, which queries with a single regex.

Examples (`\m` matches beginning of word):
`\m{term 1}`
`\m{term 1}.*\m{term 2}`
`\m{term 1}.*\m{term 2}.*\m{term 3}`

### DIFF
--- a/search/render.go
+++ b/search/render.go
@@ -23,8 +23,8 @@ type RenderData interface {
 // Helpers returns a map of all the helper functions that can be used in a template.
 func Helpers() template.FuncMap {
 	return template.FuncMap{
-		"prefixSearch": func(argName string, columnName string) string {
-			return fmt.Sprintf("lower(%s) ~ all(:~%s)", columnName, argName)
+		"orderedSearch": func(argName string, columnName string) string {
+			return fmt.Sprintf("lower(%s) ~ :~%s", columnName, argName)
 		},
 		"textSearch": func(argName string, columnNames ...string) string {
 			var buf strings.Builder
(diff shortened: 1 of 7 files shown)


search/render.go
```diff
@@ -55,16 +55,26 @@ func splitRxTerms(rx string) pgtype.TextArray {
...
 		}
 
-		terms = append(terms, "\\m"+cur)
+		// prefix match terms with the \m "word start" symbol
+		if rx == "" {
+			rx = "\\m" + cur
+		} else {
+			// extra words in between are allowed with .*
+			rx = rx + ".*\\m" + cur
+		}
```
mastercactapus, line 64: Maybe a simpler way would be to change the original `splitRxTerms` func return `[]string` and then have a new `orderedPrefixRx` func call it and join with `.*`

mixing the join with the term separation is a little hard to follow

mastercactapus (approved):

tony-tvu (approved):
lgtm

PR #3241, opened 2023-08-21
dev: use pgdump-lite for schema generation

mastercactapus:
**Description:**
This PR updates the `pgdump-lite` tool to support outputting the DB schema.
- `make db-schema` no longer depends on the `pg_dump` tool & version
- All OSes should output the same schema
- All outputs/types/columns/etc... are sorted, ensuring uniform, stable, deterministic output

By removing an implicit dependency on the `pg_dump` tool, this change resolves issues where different users (with varying OS and/or postgres versions) would get inconsistent results when running `make db-schema`.

**Out of Scope**
The migrations test still uses `pg_dump` and updating this will be addressed separately.


allending313 (approved):
LGTM!

tony-tvu (approved):

Issue #3196, opened 2023-07-24
Oncall rotation for a specific date


Lakshmi2107:
**What problem would you like to solve? Please describe:**
Currently, we have daily, weekly, and hourly rotations. It would be nice if we can have a monthly on-call rotation. For Ex: 1st of every month.

**Describe the solution you'd like:**
Any other alternative solution with the current setup for handling this scenario( like shift length=28 => daily rotation and extra coverage for remaining 2/3 days)

PR #3243, opened 2023-08-22
feat: add monthly rotation option

allending313:

**Description:**
Currently, we have daily, weekly, and hourly rotations. It would be nice if we can have a monthly on-call rotation. For Ex: 1st of every month.
Added support for monthly rotations where rotations start at the beginning of every month and end at the end of the month.
This PR is for the backend component of this feature.

https://github.com/target/goalert/issues/3196

Introduce new monthly rotation type

**Other changes:**
The smoketest for a monthly rotation has to fast-forward time more than the 30-day session expiration; The harness has been updated to handle refreshing its session.


graphql2/schema.graphql
```diff
@@ -972,7 +973,8 @@ input CalcRotationHandoffTimesInput {
   handoff: ISOTimestamp!
   from: ISOTimestamp
   timeZone: String!
-  shiftLengthHours: Int!
+  shiftLength: Int!
+  shiftType: RotationType!
```
mastercactapus, line 977: We can't make breaking API changes; there is already code in the wild using `shiftLengthHours` -- we *can* make `shiftLengthHours` optional and add a new field, however.

I think using `ISODuration` (`shiftLength: ISODuration`) would be better than adding two fields, since it already supports encoding hour vs day vs month vs etc.

TBH, it might make sense to look at deprecating (at least from UI/API) the concept of a "rotation type" and instead just treat it as an ISODuration

migrate/migrations/20170530135027-schedule-rotation.sql
```diff
@@ -11,6 +11,7 @@ INSERT INTO schedules (id, name, description, time_zone)
     SELECT id::UUID, name, description, 'America/Chicago' FROM schedule;
 
 CREATE TYPE enum_rotation_type AS ENUM (
+    'monthly',
```
mastercactapus, line 14: We cannot edit existing migrations, only add new ones.

Migrations will not be applied again (without wiping the DB) and new installations will then not match existing. Add a new value with `ALTER TYPE ... ADD VALUE` with `notransaction` set.

mastercactapus (changes requested):
Since rotations have a "start time" which can include a date, we should discuss how that will work and how to document it

Forfold (approved):

Forfold (dismissed):

test/smoke/harness/harness.go
```diff
@@ -366,6 +366,21 @@ func (h *Harness) FastForward(d time.Duration) {
 	}
 }
 
+// FastForwardExtended handles time advances of > 30 days refreshing DefaultGraphQLAdminUserID auth which would otherwise be expired.
+func (h *Harness) FastForwardExtended(d time.Duration) {
```
mastercactapus, line 370: Instead of adding a new method, let's just use the existing, and fix the GraphQL code's bug by auto refreshing the session if/when needed (e.g., attempt a single retry upon 401).

devtools/pgmocktime/mocker.go
```diff
@@ -156,6 +156,12 @@ func (m *Mocker) AdvanceTime(ctx context.Context, d time.Duration) error {
 	return m.readErr("advance time")
 }
 
+// AdvanceTime advances the time by the given duration in seconds (workaround for pgmocktime milliseconds interval field value out of range error (~2147520000 milliseconds)).
```
mastercactapus, line 159: I'm thinking we just fix the bug in `AdvanceTime` rather than creating a new method that isn't broken -- we can do so without changing behavior.

Maybe break it down into hours + milliseconds

`base_time = %s.now() + '%d hours'::interval + '%d milliseconds'::interval`

graphql2/schema.graphql
```diff
@@ -972,7 +973,8 @@ input CalcRotationHandoffTimesInput {
   handoff: ISOTimestamp!
   from: ISOTimestamp
   timeZone: String!
-  shiftLengthHours: Int!
+  shiftLengthHours: Int
```
mastercactapus, line 976: Can we add a deprecation note here? and then we can call the new one `shiftLength` since the type is already clear

mastercactapus (changes requested):

graphql2/schema.graphql
```diff
@@ -972,7 +973,11 @@ input CalcRotationHandoffTimesInput {
   handoff: ISOTimestamp!
   from: ISOTimestamp
   timeZone: String!
-  shiftLengthHours: Int!
+
+  # Deprecated shiftLength as an integer of length in hours
+  shiftLength: Int
```
mastercactapus, line 978: We can't change the existing field name as that would be a breaking change, this needs to remain `shiftLengthHours`

Make the new field `shiftLength` since units are included with it's type

test/smoke/harness/graphql.go
```diff
@@ -24,18 +24,23 @@ const DefaultGraphQLAdminUserID = "00000000-0000-0000-0000-000000000002"
...
-	var err error
 	if userID == DefaultGraphQLAdminUserID {
 		permission.SudoContext(context.Background(), func(ctx context.Context) {
-			_, err = h.backend.UserStore.Insert(ctx, &user.User{
-				Name: "GraphQL User",
-				ID:   userID,
-				Role: permission.RoleAdmin,
-			})
+			foundUser, err := h.backend.UserStore.FindOne(ctx, userID)
+			if foundUser == nil {
```
mastercactapus, line 30: This creates a bit of a race condition -- maybe separate the insert from the session creation

graphql2/graphqlapp/rotation.go
```diff
@@ -413,3 +425,26 @@ func (a *Query) CalcRotationHandoffTimes(ctx context.Context, input *graphql2.Ca
 
 	return result, nil
 }
+
+func getRotationFromISO(start time.Time, dur *timeutil.ISODuration) (rotation.Rotation, error) {
```
mastercactapus, line 429: we should validate that the iso duration only matches one (e.g., if months are specified, days & timepart must be zero and the same for the rest)

test/smoke/harness/graphql.go
```diff
@@ -91,43 +96,55 @@ func (h *Harness) GraphQLQueryT(t *testing.T, query string) *QLResponse {
...
+			Name:  auth.CookieName,
+			Value: tok,
+		})
+		resp, err = http.DefaultClient.Do(req)
+		if err != nil {
+			t.Fatal("failed to make http request:", err)
+		}
+		if resp.StatusCode == 401 && retry > 0 {
+			h.mx.Lock()
+			tok = h.insertGraphQLUser(userID)
```
mastercactapus, line 137: As mentioned above, in this case we already know the user was created, so we can just call a method that creates a new session

test/smoke/harness/graphql.go
```diff
@@ -91,43 +96,55 @@ func (h *Harness) GraphQLQueryT(t *testing.T, query string) *QLResponse {
...
+			h.mx.Unlock()
+			retry--
+			continue
```
mastercactapus, line 140: call `resp.Body.Close()` before continuing

mastercactapus (changes requested):

mastercactapus (dismissed):

schedule/rotation/rotation.go
```diff
@@ -32,10 +32,57 @@ func (r Rotation) shiftClock() timeutil.Clock {
...
+	}
+
+	// t is before start of rotation
+	prev := r.Start.AddDate(0, -r.ShiftLength*n, 0)
+	if prev.Before(t) {
+		return prev
+	}
+
+	return r.monthStartTime(t, n+1)
+}
```
Forfold, line 61: can we add some more context here please?

Is the month start time of what? The shift? The month itself?

Why are we multiplying by a random int always sent as `1` and returning different things based off that? A comment above these and/or StartTime would be helpful.

There is the comment "For daily and weekly rotations, start time will be the previous handoff time (from start)."

It would be useful if that comment also says why monthly is being calculated differently

test/smoke/harness/graphql.go
```diff
@@ -22,22 +22,23 @@ import (
 // DefaultGraphQLAdminUserID is the UserID created & used for GraphQL calls by default.
 const DefaultGraphQLAdminUserID = "00000000-0000-0000-0000-000000000002"
 
-func (h *Harness) insertGraphQLUser(userID string) string {
+func (h *Harness) createGraphQLUser(userID string) {
```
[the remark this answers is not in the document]
mastercactapus: The smoketest for a monthly rotation has to fast-forward time more than the 30-day session expiration; so it was necessary to update the harness to handle refreshing its session.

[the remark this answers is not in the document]
Forfold: I will update the PR description with this information.

schedule/rotation/rotation.go
allending313: comments added,
monthStartTime/monthEndTime follow the same behavior as daily/weekly rotation in the startTime/endTime functions. The previous handoff time of the rotation starting from or active at time t is calculated for monthStartTime.
The handoff time of the rotation starting from or active at time t is calculated for monthEndTime.

Recursion was used here as the length of months vary meaning that we could not simply add/subtract the r.shiftClock() value. It made more sense to recursively calculate with the AddDate() function to accurately determine the start/end times.

Forfold (approved):

schedule/rotation/rotation.go
Forfold: Thank you for the clarification! It should help with maintenance in the future.

Forfold (dismissed):

mastercactapus (approved):

Forfold (approved):

PR #3254, opened 2023-08-30
feat: add google analytics support

KatieMSB:

**Description:**
This pr adds support for integration with Google Analytics 4 to track user interaction across the app. 

- Re-enables the previously deprecated Google Analytics ID config value.
- Adds custom event tracking for alert actions (i.e. Acknowledge, Escalate, Close) from an alert details page or the alert lists page.


web/src/app/main/App.tsx
```diff
@@ -54,6 +56,9 @@ const useStyles = makeStyles((theme: Theme) => ({
 }))
 
 export default function App(): JSX.Element {
+  const [analyticsID] = useConfigValue('General.GoogleAnalyticsID') as [string]
+  if (analyticsID) ReactGA.initialize(analyticsID)
```
Forfold, line 60: should this be in a useEffect so it doesn't get called every render, or is it safe here?

KatieMSB: yea that's a good point, I'll throw that in a useEffect

mastercactapus (approved):

Forfold (approved):

PR #3258, opened 2023-09-01
db: switch driver to pgx5

mastercactapus:
**Description:**
Updates the rest of the application to use pgx v5 from v4.

**Out of Scope:**
Migrating from `github.com/jackc/pgtype` -> `github.com/jackc/pgx/v5/pgtype` will be done in a separate PR.

**Technical Details:**
During the migration, we encountered the question of whether to use `pgxpool.Pool` or continue using `sql.DB` in a few places. Here are some insights into why certain choices were made:

- `pgxpool.Pool` is a native PGX type, while `sql.DB` is from the Go standard library.
- The `swomsg` package was previously using `sql.DB` and converting it into a `pgx` connection and back again, adding unnecessary boilerplate. This has been streamlined by directly using `pgxpool.Pool`.
- The `pgxpool` was used directly where possible elsewhere in the `swo` code to eliminate boilerplate code.
- The old `AcquireConn` method, which allowed fetching a PGX-native connection object from a `sql.DB`, has been removed in `pgx v5`, further justifying the use of `pgxpool.Pool` where possible.

The GoAlert application code continues to use Go's standard library interfaces (`sql.DB`), while the `swo` components have been upgraded to use the native PGX interfaces (`pgxpool.Pool`).


KatieMSB (approved):

swo/swoinfo/db.go
```diff
@@ -22,11 +22,11 @@ func DBInfo(ctx context.Context, conn *pgx.Conn) (*DB, error) {
 	if err != nil {
 		return nil, err
 	}
-	if info.ID == uuid.Nil {
+	if !info.ID.Valid {
 		return nil, fmt.Errorf("no database ID")
```
Forfold, line 26:
```suggestion
		return nil, fmt.Errorf("invalid database ID")
```

Forfold (approved):

swo/swoinfo/db.go
mastercactapus: talked about this offline -- `Valid` in this context refers to the state of the `ID` (`NullUUID` type), `!Valid` means a `NULL` value was scanned and thus there is no database ID.

PR #3265, opened 2023-09-06
mark many alerts as noise from alerts list

### LINKED ISSUE #3188: Bulk Application of Feedback to Alerts
Currently, there is no efficient way to assign feedback to multiple noisy alerts at once. Users are required to individually go through each alert to mark them as noise. In cases where there is a large number of noisy alerts due to misconfiguration or known non-actionable issue, this becomes very time-consuming.

A bulk feedback application feature can solve this problem. One possible solution could be to incorporate a "Mark Noisy" action in the alerts list, which can be applied to multiple selected alerts at once. This can be implemented through checkboxes and a subsequent dialog for confirmation.

The alternative to this would be to continue with the current method of providing feedback to each alert individually. However, this process is laborious and consumes a considerable amount of time, especially when confronted with a large number of alerts.

**Additional context:**
This feature will greatly enhance the user experience, especially in scenarios when there is a surge in the number of alerts due to known issues or misconfigurations.

Forfold:
**Description:**
- Adds the option to mark alerts from within the alerts list as noise. This allows users to select many alerts at once, rather than needing to open each alert in a new tab to set the alert's noise reason.

**Todo**
- GraphQL smoketest? If needed

- Closes #3188

**Screenshots:**

Screenshot 2023-09-12 at 9 47 45 AM

Screenshot 2023-09-06 at 11 00 39 AM

Screenshot 2023-09-06 at 11 00 45 AM

Screenshot 2023-09-06 at 11 01 00 AM

- New icon for marking alerts as noise to click on when checking alerts from alert list

- Adds the `noiseReason` field to the `updateAlerts` mutation


mastercactapus:
I like how on the details page we ask `Is this alert noise?` with a list of reasons -- can we do something similar here?

Maybe:
- Instead of the hover text "Set Noise Reason" maybe "Mark as Noise" 
- Instead of the title 'Set Selected Alerts Noise Reason' maybe "Mark Alerts as Noise" or "Why are these noise?" since the next line is `Alerts:` with a list.

Forfold:
I originally had "Mark as Noise" but wasn't sure if it was fitting. I'll play around with it!

mastercactapus (dismissed):

andrewbenington (dismissed):

andrewbenington (approved):

mastercactapus (approved):

PR #3272, opened 2023-09-11
timeutil: preserve field/type information when parsing ISODuration

mastercactapus:
**Description:**
This PR updates the `timeutil.ISODuration` to preserve information about weeks, hours, minutes, and seconds.

Convenience methods were added for code that previously relied on the combined field values:
- The struct now has explicit fields for each part (e.g., `.YearPart` and `.WeekPart`)
- A `.TimePart()` method was added to convert hours/minutes/seconds into a `time.Duration`
- A `.SetTimePart(time.Duration)` method was added to do the reverse
- A `.Days()` method was added to give `.WeekPart` and `.DayPart` together

This allows the backend to use this type information and is necessary work for #3243


allending313 (approved):
lgtm!

KatieMSB (approved):

Issue #3007, opened 2023-05-11
Admin Ability to Create API Keys for Integration with External Systems


mastercactapus:
**What problem would you like to solve? Please describe:**
External systems and teams often want to build tools that integrate with GoAlert directly but currently lack the ability to authenticate and securely access GoAlert's API.

**Describe the solution you'd like:**
Allow administrators to create and manage API keys for external system integration. Offer direct tokens with optional expiration time and support for tokens generated by an OAuth server. Limit the ability to generate keys to users with admin access, but the keys themselves may not have admin-level access. For now, keep the ability to limit keys to specific resources out of scope, as the current auth model doesn't support it.

**Describe alternatives you've considered:**
User-generated API keys can be considered, but starting with user access limited to admins can help with adoption and maintain security.

**Additional context:**
Some use cases for API keys include:
- External systems subscribing to alert events for improved communication and reporting
- Building a Terraform plugin for GoAlert
- Integrating with external scheduling systems
- Synchronizing multiple GoAlert instances

mastercactapus:
<img width="2301" alt="untitled-Map 1" src="https://github.com/target/goalert/assets/595010/7d62087b-16c0-4a5e-a2c1-e0b3b7bccc95">

mastercactapus:
UI Design Notes:

New page `Admin > API Keys` as a list page displaying keys:
- no search
- no pagination
- Name, Expiration, # of fields, is read-only (no `Mutation.*`), time last used
- CREATE API KEY button (top on wide, FAB on mobile)
- Can edit keys (name and description only)
- Clicking on a key opens a details pane (like the `Admin > Message Logs` page) to allow edit & delete
  - Show allowed fields
  - Show creation time
  - Show created by user
  - Show last user agent string

New API Key form/dialog:
- input: Name
- input: Description
- input: Expiration time
- input: Allowed fields (multi-select)
  - Optional query input (paste/enter gql query, update selected fields)
- Upon creation show non-click-away dialog state with copy button (only way to get the token)

API Reqs:
- `apiKeys` query to list all API keys
- `createAPIKey` mutation to generate a new key
- `queryFields` query to list the fields of a provided query
- `allowedFields` query to list all available schema fields

mastercactapus:
- duplicate should open the create dialog with all the values pre-populated
- expired keys should probably have a warning icon
- integration test can wait until the query field is done (so it can be a full test)

Forfold:
@mastercactapus How do we use this interface with input variables?

[...]

PR #3274, opened 2023-09-12
api: add base admin API key support

mastercactapus:
**Description:**
This PR introduces the ability to create, delete, and use admin-only API keys for the GraphQL API.

This functionality is behind the experimental flag `gql-api-keys`.

`make start EXPERIMENTAL=gql-api-keys`

Part of #3007 

**Out of Scope:**
- Policy caching (perf. optimization) will be done in a separate PR
- Update and listing will be done in a separate PR, this is just base functionality
- UI for managing and viewing usage will be done in a separate PR

- `createGQLAPIKey` and `deleteGQLAPIKey` mutations were added

**Additional Context**
Since this is a large PR, here's a walkthrough of the changes:
- There are context helpers for storing/retrieving the API key policy from a request
- The tokens themselves are JWT, with expiration and an embedded hash of the policy (prevents policy changes after token generation)
- last-used data is kept for each key and includes the user agent and IP address
- the policy for a key includes all allowed GraphQL fields, as well as the role (Admin or User)
- policy info is stored in the DB, the hash must match the embedded hash of the token
- sqlc was used for all DB queries
- this PR supports Delete, Authorize, and Create
- Keys are only valid for `/api/graphql` and the `GQLAPIKey` experimental flag must be enabled
- Field enforcement is done as `AroundFields` moddleware
- schema has the 2 mutations added (create/delete), creating a key is the only way to retrieve it's token
- The key information is stored in a `gql_api_keys` table, with soft-deletes
- The usage information is a separate `gql_api_key_usage` table, since it will receive frequent updates


andrewbenington (approved):
looks good and seems to work well locally

andrewbenington (changes requested):
looks like `expflag.d.ts` and `schema.d.ts` are out of date, running `go generate ./web/src` updated them for me

andrewbenington (approved):

KatieMSB (approved):

Issue #3235, opened 2023-08-18
go: use sqlc for all db calls


Forfold:
**Finish migrating db calls to use `sqlc`:**

- [ ] `alert/store.go`
- [ ] `alert/alerlog/store.go`
- [ ] `alert/alertmetrics/store.go`
- [ ] `auth/handler.go`
- [ ] `auth/basic/db.go`
- [ ] `auth/nonce/store.go`
- [ ] `config/store.go`
- [ ] `engine/backend.go`
- [ ] `engine/cleanupmanager/db.go`
- [ ] `engine/compatmanager/db.go`
- [ ] `engine/escalationmanager/db.go`
- [ ] `engine/escalationmanager/update.go`
- [ ] `engine/heartbeatmanager/db.go`
- [ ] `engine/message/db.go`
- [ ] `engine/metricsmanager/db.go`
- [ ] `engine/npcyclemanager/db.go`
- [ ] `engine/processinglock/conn.go`
- [ ] `engine/processinglock/lock.go`
- [ ] `engine/rotationmanager/db.go`
- [ ] `engine/schedulemanager/db.go`
- [ ] `engine/verifymanager/db.go`
- [ ] `engine/verifymanager/update.go`
- [ ] `escalation/store.go`
- [ ] `gadb/db.go`
- [ ] `heartbeat/store.go`
- [ ] `integrationkey/store.go`
- [ ] `keyring/store.go`
- [ ] `label/store.go`
- [ ] `limit/store.go`
- [ ] `notice/store.go`
- [ ] `notification/store.go`
- [ ] `notification/twilio/dbsms.go`
- [ ] `notificationchannel/store.go`
- [ ] `oncall/store.go`
- [ ] `override/store.go`
- [ ] `schedule/store.go`
- [ ] `schedule/rotation/store.go`
- [ ] `schedule/rule/store.go`
- [ ] `service/store.go`
- [ ] `user/store.go`
- [ ] `user/contactmethod/store.go`
- [ ] `user/favorite/store.go`
- [ ] `user/notificationrule/store.go`
- [ ] `util/sqlprepare.go`

**Files calling `gadb` to migrate:**
- [ ] `alert/store.go`
- [ ] `alert/alertlog/store.go`
- [ ] `auth/authlink/store.go`
- [ ] `calsub/http.go`
- [ ] `calsub/store.go`
- [ ] `engine/statusmgr/update.go`
- [ ] `gadb/db.go`
- [ ] `gadb/models.go`
[...]

Issue #3108, opened 2023-06-20
Migrate from util.Prepare to sqlc queries


mastercactapus:
**What problem would you like to solve? Please describe:**
Improve the development experience by adopting `sqlc` as a replacement for the majority of our DB queries, which are currently using "stores" with strings directly in Go as prepared statements.

**Describe the solution you'd like:**
Migrate these stores from using the `util.Prepare` helper to using `sqlc` queries instead. This should make the queries more readable and maintainable.

**Additional context:**
This is an enhancement focused on improving the developer experience and overall project maintainability.

mastercactapus:
Guidelines:
- Create a `queries.sql` file in the package (if not there already)
- Add the new `.sql` file to `sqlc.yaml` in the root
- Update the store code to use the generated method instead of the `sql.Stmt` & `.Scan` method (e.g., `gadb.New(s.db).SomeMethod()` see `calsub/store.go` for examples)
- Only migrate part of the store per PR for large files/packages.

Example of the current state:
![image](https://github.com/target/goalert/assets/595010/c668711d-e690-4a0c-96cb-1a412dfcf059)

Where we want things to go:
![image](https://github.com/target/goalert/assets/595010/7997ec48-2f16-4754-9e04-f6281ddb89be)

mastercactapus:
Closing in favor of #3235, which tracks the progress of individual files across the codebase (ty @Forfold )

PR #3288, opened 2023-09-19
user/favorite: use sqlc for user favorites

mastercactapus:
**Description:**  
This PR focuses on converting the `user/favorite` package to use sqlc queries. I have updated the store methods to accept a database interface (`sql.DB` or `sql.Tx`) as an argument rather than using an embedded database. This is a standard PR that is part of a larger effort to migrate to sqlc queries, as outlined in issues #3108 and #3235.

Fixes part of #3235

**Out of Scope:**  
No additional features or fixes are included that are not directly related to the issue #3235.

**Screenshots:**  
N/A (Not a user-facing change)

None. This change is behind-the-scenes and does not affect the user experience.

No API changes. The store methods now accept a database interface as an argument, but this is not user-facing.

**Additional Info:**  
This PR is one of many that are working towards migrating GoAlert to use sqlc queries.


user/favorite/store.go
```diff
@@ -180,55 +117,38 @@ func (s *Store) FindAll(ctx context.Context, userID string, filter []assignment.
...
-				allowUsers = true
-			}
-		}
+	args := gadb.UserFavFindAllParams{
+		UserID:                  uuid.MustParse(userID),
+		AllowServices:           len(filter) == 0 || slices.Contains(filter, assignment.TargetTypeService),
+		AllowSchedules:          len(filter) == 0 || slices.Contains(filter, assignment.TargetTypeSchedule),
+		AllowRotations:          len(filter) == 0 || slices.Contains(filter, assignment.TargetTypeRotation),
+		AllowEscalationPolicies: len(filter) == 0 || slices.Contains(filter, assignment.TargetTypeEscalationPolicy),
+		AllowUsers:              len(filter) == 0 || slices.Contains(filter, assignment.TargetTypeUser),
```
Forfold, line 126: Do we want to set all of these to true is `len(filter) == 0`? It looks like before we were only doing so for `allowServices`, and the rest were dependent on check within the loop

mastercactapus: I believe that was a bug. If we don't specify a filter, all favorites should be returned. 

Originally only services supported favorites, and the code was just never updated when we added additional types.

KatieMSB (approved):

Forfold (approved):

PR #3291, opened 2023-09-20
apikey: cache policy info lookup & last usage updates

mastercactapus:
**Description:**
This PR adds a cache for API key policy info and last-usage updates.

This ensures that last-used info isn't updated more than once per minute (per instance). The policy cache ensures that repeated use of the same key doesn't require fetching and parsing the policy JSON for each request.

Additionally, invalid keys (that otherwise have a valid signature) are put in a negative cache. This means when a key is deleted, only the first request will require a DB lookup and subsequent requests from the same key will be rejected immediately.

Part of #3007

### DIFF
--- a/apikey/lastusedcache.go
+++ b/apikey/lastusedcache.go
@@ -1,0 +1,34 @@
+package apikey
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/golang/groupcache/lru"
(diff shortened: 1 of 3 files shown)


cmarquis (approved):

allending313 (approved):
LGTM!

PR #3293, opened 2023-09-20
chore: integration keys sqlc

andrewbenington:

**Description:**
Migrated all integration key SQL statements to use sqlc. Also modified UUID validation to optionally return the parsed UUIDs.

Part of #3235 

**Out of Scope:**

**Screenshots:**

**Additional Info:**


integrationkey/queries.sql
```diff
@@ -0,0 +1,20 @@
+-- name: GetServiceID :one
```
mastercactapus, line 1: Since all queries get put into the same package, let's prefix these with something like `IntKey`, like `IntKeyServiceID` `IntKeyCreate` `IntKeyFindOne` `IntKeyFindByService` etc...

Also, run this file through `pgFormatter`

integrationkey/store.go
```diff
@@ -16,26 +15,17 @@ import (
...
+	return &Store{db: db}
+}
 
-		getServiceID:     p.P("SELECT service_id FROM integration_keys WHERE id = $1 AND type = $2"),
-		create:           p.P("INSERT INTO integration_keys (id, name, type, service_id) VALUES ($1, $2, $3, $4)"),
-		findOne:          p.P("SELECT id, name, type, service_id FROM integration_keys WHERE id = $1"),
-		findAllByService: p.P("SELECT id, name, type, service_id FROM integration_keys WHERE service_id = $1"),
-		delete:           p.P("DELETE FROM integration_keys WHERE id = any($1)"),
-	}, p.Err
+func (s *Store) queries(tx *sql.Tx) *gadb.Queries {
```
mastercactapus, line 24: I wonder if this might be a good opportunity to do away with the "maybe Tx" logic in the store, and just require the caller pass a `gadb.DBTX` instead?

mastercactapus (changes requested):

integrationkey/store.go
andrewbenington: So a caller would always create the transaction first and pass it into the Store call?

andrewbenington: Or (for example) would we just change functions like `Store.Delete()` to do this:
```go
func (s *Store) Delete(ctx context.Context, id string) error {
	tx, err := s.db.BeginTx(ctx, nil)
	if err == nil {
		return err
	}
	return s.DeleteTx(ctx, tx, id)
}
```
instead of passing in `nil`

mastercactapus: `gadb.DBTX` is an interface type, so they can pass in `*sql.DB` OR `*sql.Tx` but we'd require they pass in something, rather than implementing fallback logic.

My recommendation is to delete `Create` and rename `CreateTx` to just `Create` -- that way you can follow the compiler errors to ensure you don't miss any code updates

mastercactapus (approved):

allending313 (approved):
lgtm!

PR #3296, opened 2023-09-21
gql/ui: segment message log graph data

### LINKED ISSUE #3271: Introduce Detailed Breakdowns in Message Logs Graph
Currently, the message logs graph only presents the number of messages with optional filter criteria. However, it lacks specificity, failing to provide adequate insights into traffic spikes and where they're coming from.

I suggest introducing breakdowns in the message logs graph. Let's allow the data to be viewed per-type, per-user, and/or per-service. This way, users can have more detailed insight into any traffic spikes, identifying their sources more accurately.

An alternative solution could be to add additional filter criteria to the existing graph, but this might not provide as clear or versatile insights as the suggested breakdowns.

Forfold:
**Description:**
Adds the option to segment the message log graph stats data by service, user, or message type.

Closes #3271

**Out of Scope:**
Update interval when start/end date changes drastically
- e.g. start-end is a short interval, breaking down by minute. If the start changes to a time far in the past, the interval should change to daily/weekly/monthly as necessary

**Screenshots:**

Screenshot 2023-10-24 at 11 44 40 AM

Screenshot 2023-10-24 at 11 44 56 AM


Forfold:
TODO notes from demo:
- Move "Interval Duration" to above dropdown
- Option to reset/toggle off segmentation
- Add max-height to tooltip if many users/services present for a given time on the x-axis

graphql2/schema.graphql
```diff
@@ -245,19 +245,28 @@ type MessageLogConnection {
...
+}
+
 type MessageLogConnectionStats {
   timeSeries(input: TimeSeriesOptions!): [TimeSeriesBucket!]!
 }
 
 input TimeSeriesOptions {
   bucketDuration: ISODuration!
   bucketOrigin: ISOTimestamp
+  segmentBy: MessageLogSegmentBy
```
mastercactapus, line 262: Since `MessageLogSegmentBy` is message-log specific, let's add it as an optional parameter to `timeSeries` instead

```graphql
type MessageLogConnectionStats {
  timeSeries(input: TimeSeriesOptions!, segmentBy: MessageLogSegmentBy): [TimeSeriesBucket!]!
}
```

notification/search.go
```diff
@@ -67,6 +67,15 @@ var searchTemplate = template.Must(template.New("search").Funcs(search.Helpers()
...
 		count(*)
+		{{if .SegmentByService}}
+		, s.name
+		{{end}}
+		{{if .SegmentByUser}}
+		, u.name
+		{{end}}
+		{{if .SegmentByMessageType}}
+		, om.message_type
+		{{end}}
```
mastercactapus, line 78: I expect them to be functionally equiv. but is there a reason this is using different logic than the GROUP BY below? (e.g., `else if and .TimeSeries .SegmentByUser`) if they were identical it would be more clear they work the same

graphql2/graphqlapp/messagelog.go
```diff
@@ -113,18 +117,44 @@ func (q *MessageLogConnectionStats) TimeSeries(ctx context.Context, opts *notifi
...
+	// if input.SegmentBy != nil {
+	// 	var segments []string
+	// 	for _, segment := range graphql2.AllMessageLogSegmentBy {
+	// 		segments = append(segments, "\""+segment.String()+"\"")
+	// 	}
+
+	// 	if !slices.Contains(segments, input.SegmentBy.String()) {
+	// 		return nil, fmt.Errorf("invalid segment type. please use one of: %v", strings.Join(segments, ", "))
+	// 	}
+	// }
```
mastercactapus, line 130: still needed?

notification/search.go
```diff
@@ -193,10 +222,13 @@ func (s *Store) TimeSeries(ctx context.Context, opts TimeSeriesOpts) ([]TimeSeri
...
-		TimeSeries:         true,
-		TimeSeriesOrigin:   opts.TimeSeriesOrigin,
-		TimeSeriesInterval: opts.TimeSeriesInterval,
+		SearchOptions:        opts.SearchOptions,
+		TimeSeries:           true,
+		TimeSeriesOrigin:     opts.TimeSeriesOrigin,
+		TimeSeriesInterval:   opts.TimeSeriesInterval,
+		SegmentByService:     opts.SegmentBy == "service",
+		SegmentByUser:        opts.SegmentBy == "user",
+		SegmentByMessageType: opts.SegmentBy == "messageType",
```
mastercactapus, line 231: Should we make these a const/special type and link gql to them to ensure they match?

web/src/app/admin/admin-message-logs/AdminMessageLogsGraph.tsx
```diff
@@ -31,17 +38,13 @@ import { gql, useQuery } from 'urql'
 import { Time } from '../../util/Time'
 import { getValidIntervals, useMessageLogsParams } from './util'
 
-type Stats = Array<{
+type Stat = {
```
mastercactapus, line 41: Can this be imported from `schema.d.ts`?

web/src/app/admin/admin-message-logs/AdminMessageLogsGraph.tsx
```diff
@@ -77,24 +81,23 @@ export default function AdminMessageLogsGraph(): JSX.Element {
...
+  // get list of segment labels from data to map out Lines
+  type LabelDict = { [key: string]: Stats }
+  const [segmentLabels, setSegmentLabels] = useState<LabelDict>({})
+  useEffect(() => {
+    const sl: LabelDict = {}
+    stats.forEach((stat) => {
+      if (!sl[stat.segmentLabel]) sl[stat.segmentLabel] = [stat]
+      else sl[stat.segmentLabel].push(stat)
+    })
+    setSegmentLabels(sl)
```
mastercactapus, line 99: Is there a reason `useMemo` doesn't work here? It should be equal to `useEffect` mixed with `useState` but simpler (since that's it's whole purpose)
https://react.dev/reference/react/useMemo

The main difference is `useMemo` won't cause a double-first-render since it can get it's value inline.

mastercactapus (changes requested):

web/src/app/admin/admin-message-logs/AdminMessageLogsGraph.tsx
Forfold: You're right

graphql2/schema.graphql
Forfold: It should be optional from within the TimeSeriesOptions input type as is, with only `bucketDuration` being required

notification/search.go
Forfold: I did the else if and was getting errors on my syntax, I'll mess with it more to try that approach 👍

graphql2/graphqlapp/messagelog.go
Forfold: This should still be needed, I couldn't decide if I wanted the validation in this function or the store function. I'll un-comment and it should be good to go with its import statement

notification/search.go
Forfold: Though I have it nested inside of the first `{{.if .TimeSeries}}` so we don't have to rewrite out
```
SELECT (trunc((extract('epoch' from om.created_at)-:timeSeriesOrigin)/:timeSeriesInterval))::bigint AS bucket,
```
several times, since in the group by doing `GROUP BY bucket` a few times isn't as noticeable

notification/search.go
Forfold: Do you think the validation I uncommented will be sufficient? If not I can do something like that

graphql2/schema.graphql
mastercactapus: The issue isn't that it's optional, but that `TimeSeriesOptions` is generic and meant to be used for more than message logs going forward (otherwise it would be `MessageLogTimeSeriesOptions`), so tying it to `MessageLogSegmentBy` doesn't make sense here.

notification/search.go
mastercactapus: So maybe for group by we can just do:
```sql
	{{if .SegmentByService}}
	GROUP BY bucket, s.name
	{{else if .SegmentByUser}}
	GROUP BY bucket, u.name
	{{else if .SegmentByMessageType}}
	GROUP BY bucket, om.message_type
```

As for the `else if` -- ping me the error and I can try to help; but it would make it clear that only one (at most) of the three will end up in the query

notification/search.go
mastercactapus: From an API standpoint, it should validate fine. Using constants just makes it cleaner to follow, adds type-checking, and potential refactoring made easier.

notification/search.go
Forfold: That is how the group by is currently, so we should be good 👍

graphql2/schema.graphql
Forfold: Ah, didn't realize that this is trying to set a precedent- having the option there was confusing to me due to everywhere else mainly using an input object **or** having the input args be comma separated, but I can make this change

Forfold: Pushed these changes, though I don't necessarily like the feel of it.

allending313 (changes requested):

allending313:
In the case of a broad search parameter with lots of data points, should we consider hiding the zero value labels?

Screenshot 2023-10-11 at 12 45 19 PM

Forfold:
@allending313 yeah that's a good point- we probably shouldn't show data that doesn't exist for the timeframe. I'm also noticing no lines get drawn after the primary and secondary colors are used (I thought it would use the library default, instead of nothing).

I'll work on this now, thanks!

Forfold:
Addressed the requested changes, but I'm now noticing the data isn't being sorted so the names move around on the graph even though the data is the same

allending313:
Screenshot 2023-10-25 at 12 23 01 PM

The graph's user labels doesn't seem to be limited causing the section to be overloaded squishing the graph if the time range isn't set to a narrow span. This also causes quite a bit of lag loading all these labels, perhaps we should just limit the number of labels for now or give an option to expand if needed.

devtools/resetdb/config.go
```diff
@@ -4,11 +4,11 @@ package main
 const (
 	UserCount            = 1619  // select count(id) from users
 	CMMax                = 7     // select count(id) from user_contact_methods group by user_id order by count desc limit 1
-	NRMax                = 15    // select count(id) from user_notification_rules group by user_id order by count desc limit 1
```
Forfold, line 7: I'll revert this file

allending313 (dismissed):

allending313 (dismissed):

graphql2/gqlgen.yml
```diff
@@ -107,3 +107,5 @@ models:
       - github.com/target/goalert/schedule.RuleID
   MessageLogConnectionStats:
     model: github.com/target/goalert/notification.SearchOptions
+  MessageLogSegmentBy:
+    model: github.com/target/goalert/graphql2.MessageLogSegmenyBy
```
mastercactapus, line 111: Can we just let gqlgen generate this? It already handles enums natively.

graphql2/graphqlapp/messagelog.go
```diff
@@ -113,18 +118,35 @@ func (q *MessageLogConnectionStats) TimeSeries(ctx context.Context, opts *notifi
...
-		out[i] = graphql2.TimeSeriesBucket(b)
+
+		var gb graphql2.TimeSeriesBucket
+
+		gb.Count = b.Count
+		gb.Start = b.Start
+		gb.End = b.End
+		gb.SegmentLabel = b.SegmentLabel
+
+		out[i] = graphql2.TimeSeriesBucket(gb)
```
mastercactapus, line 149: Can this be simplified to `out[i] = graphql2.TimeSeriesBucket(b)`? They should have the same fields

notification/search.go
```diff
@@ -193,10 +226,13 @@ func (s *Store) TimeSeries(ctx context.Context, opts TimeSeriesOpts) ([]TimeSeri
...
 	data := &renderData{
-		SearchOptions:      opts.SearchOptions,
-		TimeSeries:         true,
-		TimeSeriesOrigin:   opts.TimeSeriesOrigin,
-		TimeSeriesInterval: opts.TimeSeriesInterval,
+		SearchOptions:        opts.SearchOptions,
+		TimeSeries:           true,
+		TimeSeriesOrigin:     opts.TimeSeriesOrigin,
+		TimeSeriesInterval:   opts.TimeSeriesInterval,
+		SegmentByService:     opts.SegmentBy == "service",
```
mastercactapus, line 233: I feel like these should be constant values rather than arbitrary strings

notification/search.go
```diff
@@ -123,6 +143,10 @@ type renderData struct {
 	TimeSeries         bool
 	TimeSeriesOrigin   time.Time
 	TimeSeriesInterval time.Duration
+
+	SegmentByService     bool
+	SegmentByUser        bool
+	SegmentByMessageType bool
```
mastercactapus, line 149: Can we define these as methods instead that just check `SearchOptions`? that way we're not duplicating data (i.e., which is source of truth, these books or the value in SearchOptions)

notification/search.go
```diff
@@ -217,35 +253,69 @@ func (s *Store) TimeSeries(ctx context.Context, opts TimeSeriesOpts) ([]TimeSeri
 	}
 	defer rows.Close()
 
-	counts := make(map[int]int)
+	buckets := make(map[bucketID]int) // bucketID{}: count
```
mastercactapus, line 256:
```suggestion
	buckets := make(map[bucketID]int)
```

notification/search.go
```diff
@@ -217,35 +253,69 @@ func (s *Store) TimeSeries(ctx context.Context, opts TimeSeriesOpts) ([]TimeSeri
...
+			}
+		} else {
+			err := rows.Scan(&bucket, &count)
+			if err != nil {
+				return nil, err
+			}
+		}
+		label := segmentLabel.String
+		if label == "" {
+			label = "Message Counts"
```
mastercactapus, line 274: Should we handle this in the UI? If `label` is empty, there are no segments, which feels a little misleading.

web/src/app/admin/admin-message-logs/AdminMessageLogsControls.tsx
```diff
@@ -28,6 +28,12 @@ export default function AdminMessageLogsControls(): JSX.Element {
           />
         </Grid>
         <Grid item>
+          {/* TODO: allow temporary invalid state if typing.
+        
+          e.g. start: 9/14. with an end set at 9/30, the user wants to
+          set end to 9/15 by typing "1" then "5" in the day field.
+          This currently results in a no-op as "9/1" is less than "9/14"
+         */}
```
mastercactapus, line 36: Maybe open this as a bug for the ISODateTimePicker?

notification/search.go
Forfold: the comment is to let anyone know what the `int` represents

notification/search.go
Forfold: Since the query handles both segmentation and all data, if there are no segments then the total message count is returned for each time series.

web/src/app/admin/admin-message-logs/AdminMessageLogsControls.tsx
Forfold: https://github.com/target/goalert/issues/3426

graphql2/gqlgen.yml
Forfold: I needed to convert it before without it erroring out, but I'll remove this and regenerate and see what happens.

notification/search.go
mastercactapus: Right, with no segmentation (i.e., there is a single group for all datapoints)

Forfold: I thought a default value for the label would be a nice ux if it's returning all counts, API-wise as well. Would there be a better way to represent that here?

allending313 (approved):

allending313 (approved):

allending313 (dismissed):

mastercactapus (changes requested):

web/src/app/admin/admin-message-logs/AdminMessageLogsGraph.tsx
```diff
@@ -157,75 +180,160 @@ export default function AdminMessageLogsGraph(): JSX.Element {
...
+                                <Time time={payload[0].payload.end} />
+                              </Typography>
+                              {payload.map((p) => {
+                                if (p.payload.count === 0) return null
+
+                                return (
+                                  <React.Fragment key={p.name}>
+                                    <Typography variant='body2'>
+                                      {p.payload.segmentLabel === ''
+                                        ? 'Count'
```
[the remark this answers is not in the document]
Forfold: Just count should suffice as "Message Counts" is the x-axis label

notification/search.go
Forfold: Well, it is a leftover note, but I thought it would be useful for others to see in the future, as `int` is ambiguous.

web/src/app/admin/admin-message-logs/AdminMessageLogsGraph.tsx
```diff
@@ -78,23 +76,24 @@ export default function AdminMessageLogsGraph(): JSX.Element {
...
+  const graphLines = useMemo(() => {
+    const gl: { [key: string]: TimeSeriesBucket[] } = {}
+    if (!segmentBy) gl['Message Counts'] = stats
+    else {
+      stats.forEach((stat) => {
+        if (!gl[stat.segmentLabel]) gl[stat.segmentLabel] = [stat]
+        else gl[stat.segmentLabel].push(stat)
+      })
+    }
+    return gl
```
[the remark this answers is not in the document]
Forfold: That's what it's doing now (if.. else..), but I can change it to be more "go like" with an if and then a return.

allending313 (approved):

Forfold:
Not sure what's happening here. Hitting timeouts?

Screenshot 2023-11-21 at 12 48 37 PM

allending313 (approved):
lgtm

Forfold:
Closing as this does not perform well. Can rethink this in the future if we want.

PR #3298, opened 2023-09-21
chore: batch js updates

mastercactapus:
**Description:**
This PR updates most javascript dependencies to their latest versions.

**Out of Scope:**
- graphiql's major version update is left out because it has invalid depdendencies
- cypress is left out as it requires a large number of test updates
- remark-gfm is left out because v4 doesn't work with the current version of react-markdown https://github.com/remarkjs/remark-gfm/issues/57

**Additional Info:**
- yarn was updated to `3.6.3` to fix some issues
- CreateAlertConfirm was refactored to work better with the typescript updates
- MaterialSelect added missing argument types added
- HeartbeatMonitorForm had a missing argument type added
- app/urlql.ts was updated due to a nodejs type name change
- graphql explorer app was udpated to use `createRoot`
- yarn.lock was regenerated due to a large number of duplicate packages (which broke typescript in many places)


KatieMSB (dismissed):

KatieMSB (dismissed):

andrewbenington (commented):
for consistency, should `stable` be replaced with `$(YARN_VERSION)` here?
https://github.com/target/goalert/blob/b25e705e00902d4f0485ecedfd9c5c2d77c74f1e/Makefile#L192-L193

KatieMSB (approved):

andrewbenington (approved):

PR #3310, opened 2023-09-25
chore: user contactmethod sqlc

allending313:

**Description:**
Converted user/contactmethod store SQL statements to use sqlc.
Removed db field from store in favor of passing in dbtx.

https://github.com/target/goalert/issues/3235

**Additional Changes**
- added dataloader newStoreLoader wrapper to support passing in gadb.dbtx
- added db field to twilio config to support passing non-nil dbtx


user/contactmethod/store.go
```diff
@@ -17,94 +19,13 @@ import (
...
-	enable       *sql.Stmt
-	disable      *sql.Stmt
-	metaTV       *sql.Stmt
-	setMetaTV    *sql.Stmt
-	now          *sql.Stmt
 }
 
 // NewStore will create a DB backend from a sql.DB. An error will be returned if statements fail to prepare.
 func NewStore(ctx context.Context, db *sql.DB) (*Store, error) {
 	p := &util.Prepare{DB: db, Ctx: ctx}
```
andrewbenington, line 26:
```suggestion
func NewStore(ctx context.Context, db *sql.DB) *Store {
```
You can get rid of `p` and the error it might return

user/contactmethod/store.go
```diff
@@ -267,31 +186,35 @@ func (s *Store) DeleteTx(ctx context.Context, tx *sql.Tx, ids ...string) error {
...
+			uids[i] = uuid.MustParse(id)
+		}
+		err = gadb.New(tx).DeleteContactMethod(ctx, uids)
 		return err
 	}
 
-	rows, err := wrapTx(ctx, tx, s.lookupUserID).QueryContext(ctx, sqlutil.UUIDArray(ids))
+	uids := make([]uuid.UUID, len(ids))
+	for i, id := range ids {
+		uids[i] = uuid.MustParse(id)
```
andrewbenington, line 199: you could use `validate.ParseManyUUID()` here instead of a loop, that would also give an error instead of panicking

user/contactmethod/store.go
```diff
@@ -306,12 +229,23 @@ func (s *Store) FindOneTx(ctx context.Context, tx *sql.Tx, id string) (*ContactM
 		return nil, err
 	}
 
-	var c ContactMethod
-	row := wrapTx(ctx, tx, s.findOneUpd).QueryRowContext(ctx, id)
-	err = row.Scan(&c.ID, &c.Name, &c.Type, &c.Value, &c.Disabled, &c.UserID, &c.lastTestVerifyAt, &c.StatusUpdates, &c.Pending)
+	row, err := gadb.New(s.db).FindOneUpdateContactMethod(ctx, uuid.MustParse(id))
```
andrewbenington, line 232: replace line 227 with:
`err, methodUUID := validate.ParseUUID("ContactMethodID", id)`
and then you can do this:
```suggestion
	row, err := gadb.New(s.db).FindOneUpdateContactMethod(ctx, methodUUID)
```

user/contactmethod/store.go
```diff
@@ -327,12 +261,23 @@ func (s *Store) FindOne(ctx context.Context, id string) (*ContactMethod, error)
 		return nil, err
 	}
 
-	var c ContactMethod
-	row := s.findOne.QueryRowContext(ctx, id)
-	err = row.Scan(&c.ID, &c.Name, &c.Type, &c.Value, &c.Disabled, &c.UserID, &c.lastTestVerifyAt, &c.StatusUpdates, &c.Pending)
+	row, err := gadb.New(s.db).FindOneContactMethod(ctx, uuid.MustParse(id))
```
andrewbenington, line 264: Same as above

user/contactmethod/store.go
```diff
@@ -388,27 +333,32 @@ func (s *Store) FindMany(ctx context.Context, ids []string) ([]ContactMethod, er
 		return nil, err
 	}
 
-	rows, err := s.findMany.QueryContext(ctx, sqlutil.UUIDArray(ids))
+	uids := make([]uuid.UUID, len(ids))
+	for i, id := range ids {
+		uids[i] = uuid.MustParse(id)
+	}
```
andrewbenington, line 339: replace line 326 with:
`err, uids := validate.ParseManyUUID("ContactMethodID", ids, 50)`
and you can remove this loop

user/contactmethod/store.go
```diff
@@ -423,11 +373,25 @@ func (s *Store) FindAll(ctx context.Context, userID string) ([]ContactMethod, er
 		return nil, err
 	}
 
-	rows, err := s.findAll.QueryContext(ctx, userID)
+	rows, err := gadb.New(s.db).FindAllContactMethod(ctx, uuid.MustParse(userID))
```
andrewbenington, line 376: same as above, you can use `validate.ParseUUID` on line 366 instead

andrewbenington (changes requested):

user/contactmethod/store.go
```diff
@@ -194,9 +192,9 @@ func (s *Store) DeleteTx(ctx context.Context, tx *sql.Tx, ids ...string) error {
 		return err
 	}
 
-	uids := make([]uuid.UUID, len(ids))
-	for i, id := range ids {
-		uids[i] = uuid.MustParse(id)
+	uids, err := validate.ParseManyUUID("UserIDs", ids, len(ids))
```
andrewbenington, line 195: Sorry, I meant you should replace line 181 (to avoid calling validate twice)

e.g., change line 181 from this:
```
err = validate.ManyUUID("ContactMethodID", ids, 50)
```
to this:
```
uids, err = validate.ParseManyUUID("ContactMethodID", ids, 50)
```
and then you can get rid of line 195 here

andrewbenington: looks like you can also get rid of lines 187-190

allending313: oops missed that one, thanks

andrewbenington (approved):

andrewbenington (approved):

user/contactmethod/queries.sql
```diff
@@ -0,0 +1,132 @@
...
+    user_contact_methods
+WHERE
+    type = $1
+    AND value = $2;
+
+-- name: UpdateMetaTVContactMethod :exec
+UPDATE
+    user_contact_methods
+SET
+    metadata = jsonb_set(jsonb_set(metadata, '{CarrierV1}', @carrier_v1::jsonb), '{CarrierV1,UpdatedAt}',('"' || NOW()::timestamptz AT TIME ZONE 'UTC' || '"')::jsonb)
```
Forfold, line 117: I see some of the queries aren't a one-to-one migration- is there any changes to the query we should know about?

It looks like this is the new replacement for:
```
setMetaTV: p.P(`
	UPDATE user_contact_methods
	SET metadata = $3
	WHERE type = $1 AND value = $2
`),
```

allending313: Yeah this one created some issues when migrating it one-to-one. This resolves the errors by pulling only the carrierV1 field from the input instead of the whole metadata object of type `pqtype.NullRawMessage` which was causing issues. The functionality should remain the same.

Forfold: Cool just wanted to make sure since all of them are being converted here. Thank you for the clarification 🙂

notification/twilio/config.go
```diff
@@ -24,4 +25,7 @@ type Config struct {
 
 	// CMStore is used for storing and fetching metadata (like carrier information).
 	CMStore *contactmethod.Store
+
+	// todo
```
mastercactapus, line 29: TODO

user/contactmethod/queries.sql
```diff
@@ -0,0 +1,132 @@
+-- name: AddContactMethod :exec
```
mastercactapus, line 1: Can we prefix these (e.g., ContactMethodAdd)? It will make it easier to maintain/use since all queries end up in a single package.

user/contactmethod/store.go
```diff
@@ -261,83 +154,69 @@ func (s *Store) DeleteTx(ctx context.Context, tx *sql.Tx, ids ...string) error {
...
 	}
 
 	if permission.Admin(ctx) {
-		_, err = wrapTx(ctx, tx, s.delete).ExecContext(ctx, sqlutil.UUIDArray(ids))
+		err = gadb.New(dbtx).DeleteContactMethod(ctx, uids)
 		return err
 	}
 
-	rows, err := wrapTx(ctx, tx, s.lookupUserID).QueryContext(ctx, sqlutil.UUIDArray(ids))
+	rows, err := gadb.New(s.db).LookupUserIDContactMethod(ctx, uids)
```
mastercactapus, line 167: This should be running in the same transaction (`dbtx` not `s.db`)

user/contactmethod/store.go
```diff
@@ -4,166 +4,61 @@ import (
...
-	"github.com/target/goalert/util"
 	"github.com/target/goalert/util/log"
-	"github.com/target/goalert/util/sqlutil"
 	"github.com/target/goalert/validation"
 	"github.com/target/goalert/validation/validate"
 )
 
 // Store implements the lookup and management of ContactMethods against a *sql.Store backend.
 type Store struct {
 	db *sql.DB
```
mastercactapus, line 18: Let's delete this to ensure it's not being used (since we should be passing the dbtx in)

mastercactapus (changes requested):

user/contactmethod/store.go
allending313: removing this field conflicts with the `FindMany` func as adding a new parameter `dbtx` messes with the implementation of 
```ctx = context.WithValue(ctx, dataLoaderKeyCM, dataloader.NewStoreLoader(ctx, a.CMStore.FindMany))```

**Review by andrewbenington — CHANGES_REQUESTED** [review]

File: user/contactmethod/store.go [hunk 13]
```diff
@@ -4,171 +4,65 @@ import (
...
-				UPDATE user_contact_methods
-				SET name = $2, disabled = $3, enable_status_updates = $4
-				WHERE id = $1
-			`),
-		delete: p.P(`
-				DELETE FROM user_contact_methods
-				WHERE id = any($1)
-			`),
-	}, p.Err
+func NewStore(ctx context.Context, db *sql.DB) *Store {
```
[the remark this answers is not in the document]
andrewbenington: Actually you could probably remove the function altogether

andrewbenington (dismissed):

mastercactapus (approved):

Forfold (approved):

PR #3316, opened 2023-09-28
signals: graphql prototype

### LINKED ISSUE #3319: GraphQL Schema for Signals and Service Rules
Update the GraphQL schema to include
- A Signal data type
  - A Signal should be able to include the Service and Service Rule it is tied to
- `signal(id)` query for a single Signal
- `signals(searchParams)` query for a paginated search of Signals
- A Service Rule data type
- `createServiceRule()`
- `updateServiceRule()`
- `deleteServiceRule()`

andrewbenington:

**Description:**
- New GraphQL queries:
  - Get one signal
  - Search signals
  - Get one service rule
  - Create service rule
  - Update service rule
  - Delete service rule
- Services now have the `rules` graphql field to return all service rules associated with the service

Closes #3319

**Out of Scope:**
`filterString` field in a `ServiceRule` will eventually be replaced with a list of Filter objects, and the frontend will never handle the raw filterString

**Screenshots:**
Screenshot 2023-09-28 at 10 21 52 AM
Screenshot 2023-09-28 at 10 21 45 AM
Screenshot 2023-09-28 at 10 21 13 AM
Screenshot 2023-09-28 at 10 23 07 AM

- new signal and service rule graphql fields and queries
- services have `rules` graphql field


Forfold:
I didn't see any issues related to this when I did a search on "signal". Could you create one please so we can have a description of what we should expect a signal be and do?

andrewbenington:
> I didn't see any issues related to this when I did a search on "signal". Could you create one please so we can have a description of what we should expect a signal to be and do?

Done. The larger Signal/Service Rule issue is #3318, and the specific issue for this PR is #3319

signal/store.go
```diff
@@ -18,6 +23,63 @@ func NewStore(ctx context.Context, db *sql.DB) *Store {
 	return &Store{db: db}
```
allending313, line 23: This is more of a nitpick but we could pass in the db transactions as a `gadb.DBTX` parameter instead of storing the db as a field in the store obj. It'll be more explicit and readable this way but just less convenient i suppose.

allending313: Looks good aside from that though!

andrewbenington: The HTTP endpoint currently doesn't have direct access to the database. Do you think we should pass it into the handler config? That would work, but it goes against what the other HTTP handlers currently do.

andrewbenington: for reference, here are the Store calls in the http handler:
https://github.com/target/goalert/blob/521869d33c3ac92ed38fae4bad283a33e5088c94/signalapi/notify.go#L91
https://github.com/target/goalert/blob/521869d33c3ac92ed38fae4bad283a33e5088c94/signalapi/rule.go#L16

allending313: I think it's fine to add db to the handler config and setting it to `app.db` in inithttp.go

allending313 (approved):
lgtm!

Issue #3240, opened 2023-08-21
ui: migrate from @apollo to urql


Forfold:
We are in the process of migrating from the npm package `@apollo` for our GraphQL queries to instead use `urql`. There are _several_ files that still need to be migrated:

- [ ] web/src/app/apollo.js
- [ ] web/src/app/index.tsx
- [ ] web/src/app/admin/AdminConfig.tsx
- [ ] web/src/app/admin/AdminLimits.tsx
- [ ] web/src/app/admin/AdminNumberLookup.tsx
- [ ] web/src/app/admin/AdminSMSSend.tsx
- [ ] web/src/app/admin/SlackActions.tsx
- [ ] web/src/app/alerts/AlertDetailLogs.tsx
- [ ] web/src/app/alerts/AlertsList.tsx
- [ ] web/src/app/alerts/components/AlertDetails.tsx
- [ ] web/src/app/alerts/CreateAlertDialog/useCreateAlerts.ts
- [ ] web/src/app/alerts/CreateAlertDialog/StepContent/CreateAlertServiceListItem.tsx
- [ ] web/src/app/alerts/CreateAlertDialog/StepContent/CreateAlertServiceSelect.tsx
- [ ] web/src/app/alerts/pages/AlertDetailPage.tsx
- [ ] web/src/app/escalation-policies/PolicyCreateDialog.tsx
- [ ] web/src/app/escalation-policies/PolicyStepsCard.js
- [ ] web/src/app/escalation-policies/PolicyStepsQuery.tsx
- [ ] web/src/app/links/SlackChannelLink.tsx
- [ ] web/src/app/lists/QueryList.tsx
- [ ] web/src/app/rotations/RotationAddUserDialog.tsx
- [ ] web/src/app/rotations/RotationCreateDialog.tsx
- [ ] web/src/app/rotations/RotationForm.tsx
- [ ] web/src/app/rotations/RotationUserList.tsx
- [ ] web/src/app/schedules/ScheduleCreateDialog.tsx
- [ ] web/src/app/schedules/ScheduleDeleteDialog.tsx
- [ ] web/src/app/schedules/ScheduleDetails.tsx
- [ ] web/src/app/schedules/ScheduleOverrideCreateDialog.tsx
- [ ] web/src/app/schedules/ScheduleOverrideDialog.js
- [ ] web/src/app/schedules/ScheduleOverrideList.js
- [ ] web/src/app/schedules/ScheduleRuleCreateDialog.js
[...]

PR #3324, opened 2023-09-29
3007 api key UI

1ddo:

**Description:**
Adds a page for API key creation and deletion. When a key is created, it can be viewed and copied only until the dialog is closed. After that, a new one must be created if lost.

Use the command `make start EXPERIMENTAL=gql-api-keys` to functionally test.

**Out of Scope:**

**Screenshots:**

**Additional Info:**


web/src/app/admin/AdminAPIKeys.tsx
```diff
@@ -0,0 +1,229 @@
...
+  Typography,
+  Card,
+  ButtonBase,
+  CardHeader,
+} from '@mui/material'
+import { Add } from '@mui/icons-material'
+import AdminAPIKeysDrawer from './admin-api-keys/AdminAPIKeysDrawer'
+import { GQLAPIKey, CreatedGQLAPIKey } from '../../schema'
+import { Time } from '../util/Time'
+import { gql, useQuery } from '@apollo/client'
```
mastercactapus, line 15: We're moving away from apollo in favor of urql #3240 -- the APIs are almost identical, so it shouln't be to bad to update these

web/src/app/admin/AdminAPIKeys.tsx
```diff
@@ -0,0 +1,229 @@
...
+    gqlAPIKeys {
+      id
+      name
+      description
+      createdAt
+      createdBy {
+        id
+        role
+        name
+        email
```
mastercactapus, line 34: We can leave out `role` and `email` that are unused to keep this query smaller

web/src/app/admin/AdminAPIKeys.tsx
```diff
@@ -0,0 +1,229 @@
...
+    '& .MuiListItem-root': {
+      padding: '0px',
+    },
+  },
+  containerSelected: {
+    [theme.breakpoints.up('md')]: {
+      maxWidth: '70%',
+      transition: `max-width ${theme.transitions.duration.enteringScreen}ms ease`,
+    },
+    '& .MuiListItem-root': {
```
mastercactapus, line 77: I'm not sure if these classnames are stable/should be referenced directly, @Forfold can you weigh in?

web/src/app/admin/admin-api-keys/AdminAPIKeysDeleteDialog.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+// query for deleting API Key which accepts API Key ID
+const deleteGQLAPIKeyQuery = gql`
+  mutation DeleteGQLAPIKey($id: ID!) {
+    deleteGQLAPIKey(id: $id)
+  }
+`
+
+export default function AdminAPIKeysDeleteDialog(props: {
+  apiKey: GQLAPIKey | null
```
mastercactapus, line 21: Rather than passing in the entire API key, let's just ask for the ID and then have the dialog itself grab the key info. Add a query here to fetch the keys and grab the one we need to display it's name in the dialog.

Use `PolicyStepDeleteDialog` as an example of how to do this.

Additionally:
- Use `<FormDialog>` and ensure to pass along the possible mutation error (so the retry button works)
- Make sure to set `additionalTypenames: ['GQLAPIKey']` so the app reloads data properly

web/src/app/admin/admin-api-keys/AdminAPIKeysActionDialog.tsx
```diff
@@ -0,0 +1,114 @@
...
+      allowedFields: apiKey.allowedFields,
+      expiresAt: apiKey.expiresAt,
+      role: apiKey.role,
+    }
+
+    apiKeyAction({
+      input: create ? createKey : updateKey,
+    }).then((result) => {
+      if (!result.error) {
+        props.setReloadFlag(Math.random())
```
mastercactapus, line 68: Switch to `urql` and set `additionalTypenames` in the mutation (see delete dialog comment) -- that will cause `urql` to automatically reload any relevant data as needed.

web/src/app/admin/admin-api-keys/AdminAPIKeysActionDialog.tsx
```diff
@@ -0,0 +1,114 @@
...
+  apiKey: GQLAPIKey
+  setAPIKey: (param: GQLAPIKey) => void
+  setSelectedAPIKey: (param: GQLAPIKey) => void
+}): JSX.Element {
+  let query = updateGQLAPIKeyQuery
+  const { create, apiKey, setAPIKey, setSelectedAPIKey } = props
+
+  if (props.create) {
+    query = newGQLAPIKeyQuery
+  }
```
mastercactapus, line 41:
```suggestion
  const { create, apiKey, setAPIKey, setSelectedAPIKey } = props
  const query = create ? newGQLAPIKeyQuery : updateGQLAPIKeyQuery
```

web/src/app/admin/admin-api-keys/AdminAPIKeyForm.tsx
```diff
@@ -0,0 +1,174 @@
...
+            />
+          ) : (
+            <FormField
+              fullWidth
+              component={ISODateTimePicker}
+              name='expiresAt'
+              required
+              disabled
+            />
+          )}
```
mastercactapus, line 135: Can we make `AdminAPIKeyExpirationField` support the `disabled` prop so we don't need a ternary here?

Then this can just become `<AdminAPIKeyExpirationField ... disabled={!create} />`

web/src/app/admin/admin-api-keys/AdminAPIKeyExpirationField.tsx
```diff
@@ -0,0 +1,122 @@
...
+import { DateTime } from 'luxon'
+
+const useStyles = makeStyles(() => ({
+  expiresCon: {
+    'padding-top': '15px',
+  },
+}))
+// props object for this compoenent
+interface FieldProps {
+  setValue: (val: string) => void
```
mastercactapus, line 20: nit: use `onChange` to match other components

web/src/app/admin/admin-api-keys/AdminAPIKeyExpirationField.tsx
```diff
@@ -0,0 +1,122 @@
...
+          minute: '2-digit',
+          second: '2-digit',
+        }),
+      )
+    }
+  }
+  // set expirate at date value to state
+  useEffect(() => {
+    setValue(dateVal)
+  })
```
mastercactapus, line 75: Not sure why this is necessary -- it will cause an immediate re-render on mount, but `dateVal` should already be `props.value` so I don't think it's needed.

web/src/app/admin/admin-api-keys/AdminAPIKeyExpirationField.tsx
```diff
@@ -0,0 +1,122 @@
...
+          }),
+      )
+    }
+  }
+  // handles custon expiration date field option changes: sets and computes expires at value based on the selected date
+  const handleDatePickerChange = (val: string): void => {
+    // eslint-disable-next-line prettier/prettier
+    if (val != null) {
+      setDateVal(
+        new Date(val).toLocaleString([], {
```
mastercactapus, line 60: Always use ISO strings to store values in state/props and then convert to locale for rendering.

Local timestamps and locale-formatted strings aren't reliable to parse and can be difficult to debug. Always using ISO timestamps insulates the data from any nuances with the users machine/browser and makes it easier if we need to change or refactor our time formatting tools.

web/src/app/admin/admin-api-keys/AdminAPIKeyExpirationField.tsx
```diff
@@ -0,0 +1,122 @@
...
+            second: '2-digit',
+          }),
+      )
+    }
+  }
+  // handles custon expiration date field option changes: sets and computes expires at value based on the selected date
+  const handleDatePickerChange = (val: string): void => {
+    // eslint-disable-next-line prettier/prettier
+    if (val != null) {
+      setDateVal(
```
mastercactapus, line 59: Let's invert this so it is easier to follow: Short-circuit rather than nest to keep things flatter. Also, if disabling the linter is necessary, leave a comment explaining why.

Instead of:
```js
function example() {
  if (val !== null) {
    // do something
  }
}
```

Do:
```js
function example() {
  if (val === null) return
  
  // do something
}
```

mastercactapus (changes requested):

web/src/app/admin/admin-api-keys/AdminAPIKeysDeleteDialog.tsx
1ddo: I think the gqlAPIKeys query needs to be updates to accepts GQLAPIKey input before I could integrate fetching by ID, is that right?

web/src/app/admin/AdminAPIKeys.tsx
Forfold: I believe the recommended solution is to use the `sx` prop for styling. I removed this locally and didn't see a change, so I think it can be removed

web/src/app/admin/AdminAPIKeys.tsx
1ddo: Noted and this has been addressed.

web/src/app/admin/AdminAPIKeys.tsx
1ddo: Noted and this has been addressed.

web/src/app/admin/admin-api-keys/AdminAPIKeysActionDialog.tsx
1ddo: Noted and this has been addressed.

web/src/app/admin/admin-api-keys/AdminAPIKeysActionDialog.tsx
1ddo: Noted and this has been addressed.

web/src/app/admin/admin-api-keys/AdminAPIKeyForm.tsx
1ddo: Noted and this has been addressed

web/src/app/admin/AdminAPIKeys.tsx
```diff
@@ -0,0 +1,217 @@
...
+    description: '',
+    createdAt: '',
+    createdBy: null,
+    updatedAt: '',
+    updatedBy: null,
+    lastUsed: null,
+    expiresAt: DateTime.utc().plus({ days: 7 }).toISO(),
+    allowedFields: [],
+    role: 'user',
+  }
```
mastercactapus, line 98: Let's move this out of the list and have the dialogs own the data:
- Split ActionDialog into Create and Edit dialogs
- Instead of TokenDialog, just have it be part of the create dialog's code to display the token at the end
- EditDialog should just take the key ID (similar to the Delete dialog)

web/src/app/admin/admin-api-keys/AdminAPIKeysDeleteDialog.tsx
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  return (
+    <FormDialog
+      title='Are you sure?'
+      confirm
+      subTitle={'This will delete the API Key ' + apiKeyName + '.'}
+      loading={deleteAPIKeyStatus.fetching}
+      errors={nonFieldErrors(deleteAPIKeyStatus.error)}
+      disableBackdropClose
```
mastercactapus, line 72: Backdrop close is OK for delete dialogs; the only place we want to disable it is when we're displaying a token the user can't find again

web/src/app/admin/admin-api-keys/AdminAPIKeysDrawer.tsx
```diff
@@ -0,0 +1,138 @@
...
+import AdminAPIKeysDeleteDialog from './AdminAPIKeysDeleteDialog'
+import { Time } from '../../util/Time'
+
+// property for this object
+interface Props {
+  onClose: () => void
+  apiKey: GQLAPIKey
+  setCreate: (param: boolean) => void
+  setAPIKey: (param: GQLAPIKey) => void
+  setOpenActionAPIKeyDialog: (param: boolean) => void
```
mastercactapus, line 26: Can we just store edit and delete as true/false in state and disconnect/remove these?

The dialogs are already modal, so I don't think we need to control that from here

mastercactapus (changes requested):

web/src/app/admin/admin-api-keys/AdminAPIKeysDrawer.tsx
1ddo: The implementation has been updated when Edit and Create dialog has been separated.

web/src/app/admin/admin-api-keys/AdminAPIKeysDeleteDialog.tsx
1ddo: Noted. Flag has been removed.

web/src/app/admin/AdminAPIKeys.tsx
1ddo: Noted. The items has been implemented.

mastercactapus (approved):

Forfold (changes requested):
A few things. When using the Create API Key dialog, I'm seeing the error:

```
useControlled.js:35 MUI: A component is changing the default value state of an uncontrolled Autocomplete after being initialized. To suppress this warning opt to use a controlled Autocomplete. 
```

After submitting when viewing the API key to copy, I get the error:

```
<p> cannot appear as a descendant of <p>.
```

Also, can we fix the following:
- Decrease the max-width on the copy API key dialog by a level
- Update "No Data Available" to "No Results"
- Match the whitespace for the header and the no results text

Screenshot 2023-10-16 at 12 35 14 PM

- Have list go to full width of card so there's no margin when hovering/selected:

Screenshot 2023-10-16 at 12 30 47 PM

- Center time text in list, or add a secondary `<OtherActions />` for editing and deleting, to match other list pages:

Screenshot 2023-10-16 at 12 33 16 PM

- Format delete dialog with Cancel, Submit, and Title Case "Delete API Key":

Screenshot 2023-10-16 at 12 33 38 PM

- The transition throws an error when selecting/deselecting a key:

```
Failed prop type: Invalid prop `in` of type `object` supplied to `ForwardRef(Fade2)`, expected `boolean`.
```

Forfold:
I can work on the above fixes ^

Forfold (approved):
Optimistic updates on creating and deleting would be nice, but it's looking good to go 👍

mastercactapus (approved):

PR #3339, opened 2023-10-04
Smtp logger

### LINKED ISSUE #3335: Excessive log noise from go-smtp dependency due to AWS Load Balancer healthchecks
**Describe the Bug:**
When GoAlert's SMTP server handled AWS Load Balancer health checks, a lot of unnecessary log entires are being generated. These health checks are simply establishing a TCP connection, without sending any data or receiving SMTP protocol greeting, resulting in unnecessary log entries. In addition, these entries are not using the structured logging library. 

**Steps to Reproduce:**
1. Start the SMTP server to listen on localhost:9025
2. Issue `nc -vz localhost 9025` in a terminal
3. Observe the log entries

Specifically, this can be seen in an environment similar to Kubernetes hosted on AWS. Error messages like `handler error: read tcp 127.0.0.1:9025->127.0.0.1:55372: read: connection reset by peer` can be seen.

**Expected Behavior:**
There should be no log entries for health checks (TCP connection with no data to the SMTP port) and when errors do occur they should go through the normal GoAlert logger. The logging should be minimal and only pertinent entries should be there.

**Observed Behavior:**
Non-essential log entries are being created due to AWS Load Balancer health checks which aren't utilizing structured logging. Additionally, the logs get spammed with these non-essential entries at regular intervals.

**Application Version:**
This bug is present in GoAlert version 0.31.1 and latest (master).

**Additional Context:**
https://github.com/emersion/go-smtp/issues/236 for this already exists in the go-smtp project repository as the root cause seems to be the handling of the TCP health checks at that level. But for GoAlert's perspective, it's causing unnecessary log noise and it does not adhere to the structured logging mechanism, which needs to be addressed.

nimjor:

**Description:**
Ensures built-in SMTP server for generating alerts uses GoAlert's structured logger instead of stdout from go-smtp's Server type.  Filters out excessive logging of ECONNRESET errors that result from common TCP probe healthchecks from load balancers--see https://github.com/emersion/go-smtp/issues/236

Fixes #3335 

**Out of Scope:**
N/A

**Screenshots:**
N/A

N/A

N/A

**Additional Info:**
Once `emersion/go-smtp` handles zero-byte connections without logging an error, the conditional filter added here can be removed.

### DIFF
--- a/smtpsrv/server.go
+++ b/smtpsrv/server.go
@@ -12,25 +12,31 @@ import (
 	"github.com/target/goalert/util/log"
 )
 
-// SMTPLogger implements the smtp.Logger interface using the main app Logger
+// SMTPLogger implements the smtp.Logger interface using the main app Logger.
 type SMTPLogger struct {
 	logger *log.Logger
 }
 
-// Printf adheres to smtp.Server's Logger interface while filtering out ECONNRESET errors caused by TCP health checks
+// Printf adheres to smtp.Server's Logger interface.
 func (l *SMTPLogger) Printf(format string, v ...interface{}) {
 	s := fmt.Sprintf(format, v...)
-	if !strings.Contains(s, "read: connection reset by peer") {
-		l.logger.Error(context.Background(), errors.New(s))
+	// TODO: Uses string compare to filter out errors caused by TCP health checks,
+	// remove once https://github.com/emersion/go-smtp/issues/236 has been fixed.
+	if strings.Contains(s, "read: connection reset by peer") {
+		return
 	}
+	l.logger.Error(context.Background(), errors.New(s))
 }
 
-// Print adheres to smtp.Server's Logger interface while filtering out ECONNRESET errors caused by TCP health checks
+// Print adheres to smtp.Server's Logger interface.
 func (l *SMTPLogger) Println(v ...interface{}) {
 	s := fmt.Sprint(v...)
-	if !strings.Contains(s, "read: connection reset by peer") {
-		l.logger.Error(context.Background(), errors.New(s))
+	// TODO: Uses string compare to filter out errors caused by TCP health checks,
+	// remove once https://github.com/emersion/go-smtp/issues/236 has been fixed.
+	if strings.Contains(s, "read: connection reset by peer") {
+		return
 	}
+	l.logger.Error(context.Background(), errors.New(s))
 }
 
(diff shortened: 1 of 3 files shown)


smtpsrv/server.go
```diff
@@ -2,12 +2,37 @@ package smtpsrv
...
+	"github.com/target/goalert/util/log"
 )
 
+// SMTPLogger implements the smtp.Logger interface using the main app Logger
+type SMTPLogger struct {
+	logger *log.Logger
+}
+
+// Printf adheres to smtp.Server's Logger interface while filtering out ECONNRESET errors caused by TCP health checks
+func (l *SMTPLogger) Printf(format string, v ...interface{}) {
```
mastercactapus, line 21: Let's add a TODO for these
```suggestion
// Printf adheres to smtp.Server's Logger interface.
//
// TODO: Uses string compare to filter out errors caused by TCP health checks,
// remove once https://github.com/emersion/go-smtp/issues/236 has been fixed.
func (l *SMTPLogger) Printf(format string, v ...interface{}) {
```

smtpsrv/server.go
```diff
@@ -2,12 +2,37 @@ package smtpsrv
...
+	s := fmt.Sprintf(format, v...)
+	if !strings.Contains(s, "read: connection reset by peer") {
+		l.logger.Error(context.Background(), errors.New(s))
+	}
+}
```
mastercactapus, line 26: nit: prefer early return, invert/short circuit stuff like this so the main work happens in the outer body

> The code reads well if the successful flow of control runs down the page, eliminating error cases as they arise.
>
> https://go.dev/doc/effective_go

```suggestion
	if strings.Contains(s, "read: connection reset by peer") {
		return // ignore connection reset for now
	}

	l.logger.Error(context.Background(), errors.New(s))
}
```

---

alternatively the TODO could be put here

nimjor: Good catch, thank you for picking that nit.  Also added the TODO here rather than above the func definition since these functions will still be needed to bring go-smtp logging under the main GoAlert logger.

smtpsrv/server.go
nimjor: Added TODO at conditional under each func

mastercactapus (approved):

ethan-haynes (approved):

PR #3361, opened 2023-10-12
feat: service rule filter struct translation

### LINKED ISSUE #3321: Service Rule Filter Structs + Translation
Add in logic to convert filters between expr string and frontend/user facing structs. This is needed for user CRUD operations as well as graphql queries. On insert should convert structs to expr string and verify. On retrieval should convert expr strings to structs and sent back to user.

Converting the expr string to a struct:
- Split the raw expr string into multiple filters by splitting by the delimeter " && "
- Each filter struct should have a `field`, `operand`, and `value`
  - for example, `Temp >= 4` becomes: `{Field: "Temp", Operand: ">=", Value: "4"}`
  - `Message contains "Warning"` becomes: `{Field: "Message", Operand: "contains", Value: "Warning"}`

Notes:
- the `Operand` should be limited to the expr operators ">", "<", "<=", ">=", "==", "!=", and "contains" for now
- numerical operands should have number validation
- the `contains` operand in an expr string requires the value to be in quotes. When parsing and serializing, the quotes should be added and removed as needed

Part of #3318

andrewbenington:

**Description:**
- Added Filter structs and translation between expr strings and those structs
  - Interactions with the GraphQL API always use structs
  - Filters are stored in the database as expr strings (and stay that way when checking if a rule applies)
  - Parsing logic uses the `expr` library's own abstract syntax tree parsing logic
- Filter values can be strings, numbers or boolean
  - this limits the operators available (e.g. strings can use `contains` while numbers can use `>=`

Fixes #3321 

**Screenshots:**
Screenshot 2023-10-12 at 9 43 54 AM
Screenshot 2023-10-12 at 9 43 36 AM

GraphQL returns filters as their own objects instead of the expr string


allending313 (approved):
lgtm

PR #3395, opened 2023-10-30
Implement Duplicate Functionality for API Keys

mastercactapus:
**Description:**
This PR introduces a convenient "Duplicate" option for API keys within the GoAlert application. Users can now effortlessly create a new API key based on an existing one, streamlining key management and improving user efficiency. This feature applies to both active and expired keys, ensuring a continuous workflow. With user experience in mind, the duplication process is designed to be intuitive and seamless, mirroring the expiration characteristics of the source key relative to the current time.

This is part of #3007 

**Out of Scope:**
Visual updates for expired keys and special sorting considerations for nearly expired keys are out of scope for this PR.

**Screenshots:**
image

Users will notice a new "Duplicate" option when managing API keys. This option allows the generation of a new key with an incremented name and an expiration period set to an equivalent duration from the current moment, just as the original key was set from its creation time.

N/A

**Additional Info:**
This PR also includes a fix for the previously unsorted API keys. Now, keys will be sorted by name, recognizing numeric values which will improve the findability and management of multiple keys. The "Duplicate" feature respects the expiration paradigm of the original key, maintaining consistency in key lifecycle management.

**Note:** This feature is behind an experimental flag, start with `make start EXPERIMENTAL=gql-api-keys`

### DIFF
--- a/web/src/app/admin/AdminAPIKeys.tsx
+++ b/web/src/app/admin/AdminAPIKeys.tsx
@@ -52,14 +52,10 @@ const useStyles = makeStyles((theme: Theme) => ({
 export default function AdminAPIKeys(): JSX.Element {
   const classes = useStyles()
   const [selectedAPIKey, setSelectedAPIKey] = useState<GQLAPIKey | null>(null)
-  const [createAPIKeyDialogClose, onCreateAPIKeyDialogClose] = useState(false)
+  const [createDialog, setCreateDialog] = useState<boolean>(false)
+  const [createFromID, setCreateFromID] = useState('')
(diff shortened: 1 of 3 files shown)


Forfold (changes requested):
I spy an AI talking 👀 Functionally works but I'm noticing the state isn't clearing on exit, so when I go to create a new key, the previous data is still in the dialog

ethan-haynes (approved):

Forfold (approved):

ethan-haynes (approved):

Issue #3405, opened 2023-11-02
Integrate GoAlert as a Destination for "Signal" Integration


ethan-haynes:
**Problem Description:**
Currently, the "Signal" feature set in GoAlert lacks the capability to route requests to a GoAlert destination based on predefined filter requirements. Users need the ability to configure a "Signal" integration key to route requests to GoAlert if specific key-value pairs in the inbound JSON body match predefined conditions set by the user. Additionally, users should be able to define templates in the "Signal" rule engine to populate the "Summary" and "Details" fields for a traditional GoAlert, allowing the routed alerts to behave as if they were generated in the traditional way.

**Proposed Solution:**
Enhance the "Signal" feature set to allow users to configure the routing of requests to GoAlert based on predefined filter requirements. Implement the following steps:
1. Enable users to define filter conditions for key-value pairs in the inbound JSON body.
2. Implement a "Signal" rule engine where users can specify templates for populating the "Summary" and "Details" fields for traditional GoAlerts.
3. Ensure that routed alerts, once created, behave in a manner consistent with traditional alerts.

**Implementation Steps:**
1. Implement the ability for users to configure filter conditions for "Signal" integration keys.
2. Develop a "Signal" rule engine where users can set up templates for "Summary" and "Details" fields.
3. Ensure that alerts routed through this process behave as if they were generated in the traditional way within GoAlert.

**Alternatives Considered:**
N/A

**Additional Context:**
This enhancement is aimed at providing users with more flexibility and control over routing alerts based on specific conditions. It enables the integration of the "Signal" feature set with GoAlert in a way that seamlessly integrates routed alerts into the existing alerting workflow. 

This GitHub issue is created to track the development and implementation of this configurable routing feature for "Signal" feature set in GoAlert.

PR #3427, opened 2023-11-06
Go signal to alert logic

ethan-haynes:
added signal to alert logic in rules engine

### DIFF
--- a/signalapi/signal.go
+++ b/signalapi/signal.go
@@ -2,12 +2,14 @@ package signalapi
 
 import (
 	"encoding/json"
+	"fmt"
 	"io"
 	"mime"
 	"net/http"
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/target/goalert/alert"
 	"github.com/target/goalert/permission"
 	"github.com/target/goalert/retry"
 	"github.com/target/goalert/service/rule"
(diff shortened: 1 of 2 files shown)


signalapi/signal.go
```diff
@@ -74,11 +80,46 @@ func (h *Handler) ServeCreateSignals(w http.ResponseWriter, r *http.Request) {
 	signals := []signal.Signal{}
 
 	for _, rule := range rules {
-		if rule.SendAlert {
-			// TODO: implement create alert logic
-			continue
-		}
 		for _, action := range rule.Actions {
+			if rule.SendAlert && action.DestType == "goalert" {
```
tony-tvu, line 84:
```suggestion
			if rule.SendAlert && strings.EqualFold(action.DestType, "alert") {
```
Latest UI code is setting destType as 'ALERT' when 'Create Alert' is toggled. `strings.EqualFold` for case-insensitivity.

ethan-haynes:
for this issue: https://github.com/target/goalert/issues/3405

tony-tvu (changes requested):
Just 1 change request. 

We'll also need to handle custom alert fields either in this PR or another eventually. Custom fields are set in the `action.contents` e.g.
`{"dest_type":"ALERT","contents":[{"prop":"summary","value":"custom summary text"},{"prop":"details","value":"custom detail text"}]}`

PR #3431, opened 2023-11-07
ui/urql: update admin queries and mutations

Forfold:
**Description:**
Updates `AdminConfig`, `AdminLimits`, `AdminNumberLookup`, `AdminSMSSend`, and `SlackActions` to use `urql` instead of `@apollo/client` for their queries and mutations

Part of #3240

### DIFF
--- a/web/src/app/admin/AdminNumberLookup.tsx
+++ b/web/src/app/admin/AdminNumberLookup.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useState, useEffect } from 'react'
 import { Form } from '../forms'
 import {
   Button,
@@ -15,7 +15,7 @@ import {
   ListItemText,
   Tooltip,
 } from '@mui/material'
-import { useMutation, useQuery, ApolloError, gql } from '@apollo/client'
+import { useMutation, useQuery, gql, CombinedError } from 'urql'
 import CopyText from '../util/CopyText'
 import TelTextField from '../util/TelTextField'
 import LoadingButton from '../loading/components/LoadingButton'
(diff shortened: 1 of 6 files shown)


web/src/app/admin/AdminNumberLookup.tsx
```diff
@@ -50,24 +50,27 @@ const numInfoQuery = gql`
 export default function AdminNumberLookup(): JSX.Element {
   const [number, setNumber] = useState('')
   const [staleCarrier, setStaleCarrier] = useState(true)
-  const [lastError, setLastError] = useState(null as null | ApolloError)
 
-  const { data: numData } = useQuery(numInfoQuery, {
+  const [{ data: numData, error: queryError }] = useQuery({
+    query: numInfoQuery,
```
Forfold, line 55: Note that `pollInterval: 0,` has been removed. I did not see an equivalent for `urql`, though this may be fine to leave at the default.

mastercactapus: IIRC, it was there because, initially, carrier info was part of the query (and we'd get charged for each request). But I don't think it's necessary anymore since we broke that out into a mutation in the end.

web/src/app/admin/AdminNumberLookup.tsx
```diff
@@ -50,24 +50,27 @@ const numInfoQuery = gql`
...
+  ] = useMutation(carrierInfoMut)
   const carrInfo = carrData?.debugCarrierInfo as DebugCarrierInfo
 
+  const [lastError, setLastError] = useState<CombinedError>()
+  useEffect(() => {
+    setLastError(queryError)
+  }, [queryError])
+  useEffect(() => {
+    setLastError(mutationError)
+  }, [mutationError])
```
mastercactapus, line 72: We need a conditional on each to only set the error if there is one.

i.e. `if (queryError) setLastError(queryError)`

web/src/app/admin/AdminNumberLookup.tsx
```diff
@@ -50,24 +50,27 @@ const numInfoQuery = gql`
...
-      onError: (err) => setLastError(err),
-    },
-  )
+  const [
+    { data: carrData, fetching: carrLoading, error: mutationError },
+    commit,
+  ] = useMutation(carrierInfoMut)
   const carrInfo = carrData?.debugCarrierInfo as DebugCarrierInfo
 
+  const [lastError, setLastError] = useState<CombinedError>()
```
mastercactapus, line 66: nit: Use `null` for explicit empty instead of `undefined`
```suggestion
  const [lastError, setLastError] = useState<CombinedError | null>(null)
```

web/src/app/admin/AdminSMSSend.tsx
```diff
@@ -98,7 +87,19 @@ export default function AdminSMSSend(): JSX.Element {
...
+          commit({
+            input: {
+              from: fromNumber,
+              to: toNumber,
+              body,
+            },
+          }).then((res) => {
+            if (res.error) {
+              setShowErrorDialog(true)
+            }
```
mastercactapus, line 99:
```suggestion
              setShowErrorDialog(true)
              return
            }
```

web/src/app/admin/AdminSMSSend.tsx
```diff
@@ -98,7 +87,19 @@ export default function AdminSMSSend(): JSX.Element {
...
+            console.log('res.data: ', res.data)
```
mastercactapus, line 100:
```suggestion
```

web/src/app/admin/SlackActions.tsx
```diff
@@ -63,7 +63,7 @@ export default function SlackActions(): JSX.Element {
           {
             label: 'App Manifest',
             handleOnClick: () => {
-              getManifest()
+              commit()
```
mastercactapus, line 66: This causes a re-fetch (since it's `useQuery`); the original behavior of this is code was not to send the query until clicked.

However, there's no reason to wait; this handler can just set `setShowManifest`
```suggestion
```

mastercactapus (changes requested):

web/src/app/admin/AdminNumberLookup.tsx
Forfold: It's same thing in my eyes with more characters added to the file, but overall, sure 👍

web/src/app/admin/SlackActions.tsx
Forfold: Gotcha, hard to tell when there's no comments for business logic

mastercactapus (approved):

tony-tvu (approved):
lgtm

PR #3433, opened 2023-11-08
Go signal to alert

ethan-haynes:
adding in the template variable injection for summary and details

### DIFF
--- a/signalapi/signal.go
+++ b/signalapi/signal.go
@@ -187,63 +187,54 @@ func buildOutgoingPayload(action rule.Action, incomingPayload map[string]interfa
 	return outgoingPayload
 }
 
-func buildOutgoingAlertPayload(action rule.Action, incomingPayload map[string]interface{}) (sigAlert SignalAlertMapper, err error) {
-	if incomingPayload["action"] != nil {
-		val, ok := incomingPayload["action"].(string)
-		if !ok {
-			return sigAlert, fmt.Errorf("Field 'action' is not a string: %s", incomingPayload["action"])
-		}
-		sigAlert.Action = val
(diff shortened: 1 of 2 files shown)


signalapi/signal.go
```diff
@@ -185,41 +187,66 @@ func buildOutgoingPayload(action rule.Action, incomingPayload map[string]interfa
...
+	}
 	sigAlert.Status = alert.StatusTriggered
 	if sigAlert.Action == "close" {
 		sigAlert.Status = alert.StatusClosed
 	}
-
 	return sigAlert, nil
 }
+
+func InjectTemplateValues(tmplStr string, data map[string]interface{}) (string, error) {
```
tony-tvu, line 241:
```suggestion
func injectTemplateValues(tmplStr string, data map[string]interface{}) (string, error) {
```
Does this func need to be public?

signalapi/signal.go
```diff
@@ -185,41 +187,66 @@ func buildOutgoingPayload(action rule.Action, incomingPayload map[string]interfa
...
+		return "", err
 	}
-	if incomingPayload["dedup"] != nil {
-		val, ok := incomingPayload["dedup"].(string)
-		if !ok {
-			return sigAlert, fmt.Errorf("Field 'dedup' is not a string: %s", incomingPayload["dedup"])
+	if val == "" {
+		for _, content := range action.Contents {
+			if content.Prop == fieldName {
+				return InjectTemplateValues(content.Value, payload)
```
tony-tvu, line 210:
```suggestion
				return injectTemplateValues(content.Value, payload)
```
Does this func need to be public?

tony-tvu (approved):
lgtm

Issue #3406, opened 2023-11-02
Enhance UI Form Validation for "Signal" Rule Creation


ethan-haynes:
**Problem Description:**
The current user interface (UI) flow for creating rules in the "Signal" rules engine lacks adequate form validation for input values. Users can input values that may not meet the required format or constraints, such as improperly formatted email addresses or missing '#' for Slack channel names. This can lead to user errors and difficulties in configuring rules. To improve the user experience, the UI should inform users when they input values that aren't allowed and provide helpful messages to guide users regarding the expected field input types.

**Proposed Solution:**
Enhance the UI form validation for creating rules in the "Signal" rules engine. Implement the following steps:
1. Implement validation checks for various input fields to ensure they meet the required format and constraints.
2. Provide informative error messages to alert users when they enter values that do not adhere to the expected format or requirements.
3. Offer guidance to users about the expected field input types, helping them input valid data.

**Implementation Steps:**
1. Develop and integrate validation checks for input fields within the "Signal" rule creation form.
2. Design and display informative error messages when users input invalid values.
3. Include tooltips or hints to guide users regarding the expected input types for each field.

**Alternatives Considered:**
N/A

**Additional Context:**
This enhancement is aimed at improving the user experience when configuring rules in the "Signal" rules engine. By providing proper form validation and informative error messages, users can create rules more effectively and avoid common input errors.

This GitHub issue is created to track the development and implementation of enhanced UI form validation for "Signal" rule creation in GoAlert.

mastercactapus:
All destination fields are now validated on submission and displayed in the UI if in error.

PR #3436, opened 2023-11-08
Signals UI validation

tony-tvu:
**Description:**
Implement Validation for Email Addresses and Webhook URLs in Signal Forms

**Issue:**
https://github.com/target/goalert/issues/3406

**Out of Scope:**
Resolved issue: Form Availability Consistency – Now, querying all Slack channels no longer triggers errors, ensuring uninterrupted access to the form for all users.

### DIFF
--- a/web/src/app/services/ServiceRules/ServiceRuleCreateDialog.tsx
+++ b/web/src/app/services/ServiceRules/ServiceRuleCreateDialog.tsx
@@ -12,7 +12,6 @@ import {
 } from '../../../schema'
 import ServiceRuleForm, { ServiceRuleValue, destType } from './ServiceRuleForm'
 import Spinner from '../../loading/components/Spinner'
-import { GenericError } from '../../error-pages'
 
 const mutation = gql`
   mutation ($input: CreateServiceRuleInput!) {
--- a/web/src/app/services/ServiceRules/ServiceRuleCreateDialog.tsx
+++ b/web/src/app/services/ServiceRules/ServiceRuleCreateDialog.tsx
@@ -158,12 +158,11 @@ export default function ServiceRuleCreateDialog(props: {
   const [actionsError, setActionsError] = useState<boolean>(false)
   const [createRuleStatus, commit] = useMutation(mutation)
 
-  const [{ fetching, error, data }] = useQuery({
+  const [{ fetching, data }] = useQuery({
     query,
     variables: {},
   })
   if (fetching && !data) return <Spinner />
-  if (error) return <GenericError error={error.message} />
 
   return (
     <FormDialog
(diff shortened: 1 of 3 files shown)


web/src/app/services/ServiceRules/ServiceRuleCreateDialog.tsx
```diff
@@ -158,12 +157,11 @@ export default function ServiceRuleCreateDialog(props: {
   const [actionsError, setActionsError] = useState<boolean>(false)
   const [createRuleStatus, commit] = useMutation(mutation)
 
-  const [{ fetching, error, data }] = useQuery({
+  const [{ fetching, data }] = useQuery({
     query,
     variables: {},
   })
   if (fetching && !data) return <Spinner />
-  if (error) return <GenericError error={error.message} />
```
Forfold, line 166: We could perhaps pass this error to the FormDialog so this doesn't get missed in any edge cases:
`errors={nonFieldErrors(createRuleStatus.error || error)}`

web/src/app/services/ServiceRules/ServiceRuleForm.tsx
```diff
@@ -300,6 +300,53 @@ export default function ServiceRuleForm(
...
+        (action.destType === destType.SLACK && c.prop === 'channel') ||
+        (action.destType === destType.SLACK && c.prop === 'channel_id')
+      ) {
+        return <React.Fragment key={contentIdx} />
+      }
+      return (
+        <Grid
+          key={c.prop}
+          item
+          style={{ flexGrow: 1, marginTop: '1em' }}
```
Forfold, line 318: nit: use the sx prop on MUI components moving forward. It should work the same but uses mui's newer `@emotion` code instead
```suggestion
          sx={{ flexGrow: 1, marginTop: '1em' }}
```

Forfold (changes requested):

web/src/app/services/ServiceRules/ServiceRuleForm.tsx
```diff
@@ -300,6 +300,48 @@ export default function ServiceRuleForm(
...
+  const renderContents = (
+    actionIdx: number,
+    action: ServiceRuleActionInput,
+  ): React.ReactNode => {
+    return action.contents.map((c: Content, contentIdx: number) => {
+      if (
+        (action.destType === destType.SLACK && c.prop === 'channel') ||
+        (action.destType === destType.SLACK && c.prop === 'channel_id')
+      ) {
+        return <React.Fragment key={contentIdx} />
```
Forfold, line 312: let's supply the key with some additional info on top of the index, to ensure each key is always unique and can be properly referenced:

```suggestion
        return <React.Fragment key={contentIdx+c.prop} />
```

tony-tvu: I went back and did this for all service rule files that are using a key :thumbsup:.

tony-tvu:
Closing since signals UI destinations implementation from this PR is much different than how destinations is currently working on master branch today.

Issue #2557, opened 2022-08-11
loading design spike


Forfold:
- https://material.io/components/progress-indicators
- https://material.io/design/communication/launch-screen.html
- https://material.io/design/communication/empty-states.html
- https://material.io/design/motion/speed.html

Thoughts

- **React Suspense** to show **progress indicator** under top app bar. Wait until whole page loads to render anything.
- If using a **progress indicator** at the top, wait until the new page is loaded before going navigating to it. for github, nothing on the page changes until the new page loads (i.e. the url)

- When opening dialogs, do a 200ms wait before showing potential loading states. Load text before opening dialogs. Images and icons load in placeholder states.

mastercactapus:
![Loading](https://user-images.githubusercontent.com/595010/184194012-d2e288ad-466b-420f-b807-31af24bdcf2f.png)

mastercactapus:
Needs working session/SPIKE/POC to evaluate options

Forfold:
Going to work on a POC for this using Suspense/Context/Lazy loading as needed

mastercactapus:
Closing due to age; we have, however, implemented suspense and largely removed spinners across the application.

PR #3440, opened 2023-11-08
ui: Suspenseful loading

mastercactapus:
**Description:**
Updates the UI to enable `suspense` for all `urql` queries.

As components are migrated from apollo to urql, they will automatically get tracked by suspense.

Notably, <Suspense> boundaries were added:
- Around the entire app for the Login switch, with a new `RequireAuth` component (instead of the redux mess)
- Around the title bar text
- Around the page content
- Around groups of dialog boxes

The `QuerySelect` tool explicitly turns off suspenseful loading as it handles loading state internally and shouldn't trigger a page-level (or dialog-level) spin.

Part of #2557

**Out of Scope:**
- apollo queries will not trigger Suspense, so some pages still aren't as nice as they will be
- `fetching`/`isLoading` logic will still need to be removed/cleaned up separately
- Cleaning up redux usage around auth detection

**Screenshots:**

<video src="https://github.com/target/goalert/assets/595010/7cdc0a07-0153-41a1-a94f-ce5c34d5ba18"></video>

- Application-wide spinner will resolve once config has been loaded, or the Login page is ready
- Dialogs will not appear until their data has loaded (`urql` only)
- Title bar will be blank while the name is fetched, if missing (no "User" -> "Abigail" flicker)
- Pages using `urql` (and any nested components doing so) will spin until _all_ content is loaded

N/A

**Additional Info:**
With `suspense` enabled, a `useQuery` will never return `fetching: true` you will always get `data` or `error`.

### DIFF
--- a/web/src/app/schedules/calendar/ScheduleCalendar.tsx
+++ b/web/src/app/schedules/calendar/ScheduleCalendar.tsx
@@ -384,11 +384,7 @@ function ScheduleCalendar(props: ScheduleCalendarProps): JSX.Element {
             onView={() => {}} // stub to hide false console err
             components={{
               // @ts-expect-error Property 'children' does not exist on type - yes it does
-              eventWrapper: ({ children, event }) => (
-                <ScheduleCalendarEventWrapper event={event}>
-                  {children}
-                </ScheduleCalendarEventWrapper>
-              ),
+              eventWrapper: ScheduleCalendarEventWrapper,
               toolbar: () => null,
             }}
           />
--- a/web/src/app/selection/QuerySelect.js
+++ b/web/src/app/selection/QuerySelect.js
@@ -30,6 +30,8 @@ const asArray = (value) => {
 const mapValueQuery = (query, index) =>
   mapInputVars(fieldAlias(query, 'data' + index), { id: 'id' + index })
 
+const noSuspenseContext = { suspense: false }
+
 // makeUseValues will return a hook that will fetch select values for the
 // given ids.
 function makeUseValues(query, mapNode) {
@@ -61,6 +63,7 @@ function makeUseValues(query, mapNode) {
       pause: !value.length,
       variables,
       requestPolicy: 'cache-first',
+      context: noSuspenseContext,
     })
 
     if (!value.length) {
@@ -100,6 +103,7 @@ function makeUseOptions(query, mapNode, vars, defaultVars) {
 
       pollInterval: 0,
       errorPolicy: 'all', // needs to be set explicitly for some reason
+      context: noSuspenseContext,
     })
 
     let result = []
--- a/web/src/app/util/TelTextField.tsx
+++ b/web/src/app/util/TelTextField.tsx
@@ -26,6 +26,8 @@ const useStyles = makeStyles({
   },
 })
 
+const noSuspense = { suspense: false }
+
 export default function TelTextField(
   props: TextFieldProps & { value: string },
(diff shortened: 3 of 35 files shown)


web/src/app/util/TelTextField.tsx
```diff
@@ -46,6 +48,7 @@ export default function TelTextField(
     variables: { number: '+' + phoneNumber },
     requestPolicy: 'cache-first',
     pause: !phoneNumber || props.disabled,
+    context: noSuspense,
```
Forfold, line 51:
```suggestion
    context: React.useMemo(() => noSuspense, []),
```

web/src/app/selection/QuerySelect.js
```diff
@@ -61,6 +63,7 @@ function makeUseValues(query, mapNode) {
       pause: !value.length,
       variables,
       requestPolicy: 'cache-first',
+      context: noSuspenseContext,
```
Forfold, line 66: These 2 will also need to be behind a memo

web/src/app/util/TelTextField.tsx
mastercactapus: `useMemo` isn't necessary in these cases because the object is declared a const outside the render function, so the reference never changes.

web/src/app/selection/QuerySelect.js
mastercactapus: Not actually needed in this case, otherwise, it'd render in an infinite loop.

Forfold: I'm basing this off of their documentation, where when using context to pair that with useMEmo, and having talked to their devs when asking for help:

https://github.com/urql-graphql/urql/discussions/3421#discussioncomment-7430479

mastercactapus: Adding context/ref. from offline convo:

Under the covers, this is basically the "issue" on why it needs stable references:
```js
console.log({} === {}) // false
```

Each time `useQuery` gets a new context object (even if it has the exact same props) it will re-fetch detecting a change. `useMemo` is a way to return a _new_ object, ONLY if a dependency changes.

There's logic that's basically if `if (newContext !== oldContext) refetch()`. So if you inline `{ suspense: false }` it's actually creating a new object reference every time the function runs, and it will **always** appear to be changing without something like `useMemo` (and creating an infinite render loop).

In our case, since we just need to set an option with no dependencies or changes, we can skip that step by just declaring a const at the module level. That will never change, and so we don't get the render issue.

Forfold: Thank you for sharing here! I'll resolve the comments.

Forfold (dismissed):
LGTM, let's test in CI

Forfold (approved):

web/src/app/schedules/calendar/ScheduleCalendar.tsx
```diff
@@ -384,11 +384,7 @@ function ScheduleCalendar(props: ScheduleCalendarProps): JSX.Element {
             onView={() => {}} // stub to hide false console err
             components={{
               // @ts-expect-error Property 'children' does not exist on type - yes it does
```
Forfold, line 386: can this comment be removed now?

Forfold (approved):

KatieMSB (dismissed):

KatieMSB (approved):

Forfold (approved):

PR #3441, opened 2023-11-09
adr: api key architecture

mastercactapus:
**Description:**
Add ADR document outlining the API key architecture

### DIFF
--- a/docs/adrs/000005-api-key-architecture.md
+++ b/docs/adrs/000005-api-key-architecture.md
@@ -1,0 +1,92 @@
+# API Key Architecture for System Integrations
+
+- Status: accepted
+- Date: 2023-07-20
+
+**Acceptance Criteria**:
+
+- Clear Consensus: Simple Majority of requested reviewers AND no rejections
+
+Technical Story: #3007
+
+## Context and Problem Statement
+
+With the need for secure system integrations, we evaluated the best approach to provide API access. We require a method that leverages our existing GraphQL infrastructure, maintains our security posture, and avoids the complexity of substantial refactoring or maintaining parallel API systems.
+
+## Decision Drivers
+
+- Security: Ensuring that the API keys cannot be easily exploited.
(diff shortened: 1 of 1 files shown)


KatieMSB (approved):

docs/adrs/000005-api-key-architecture.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Unofficial use of user session tokens for API access
+
+- ❌ Bad, because it is not secure and not officially supported.
+
+### Restricted API keys with predefined queries and operations
+
+- ✅ Good, because it aligns with existing authorization checks and infrastructure.
+- ✅ Good, because it offers a secure and scalable solution for system integrations.
+- ✅ Good, because it avoids the need for a significant system overhaul.
```
Forfold, line 67: Are there any cons of this accepted solution that we want to document and stay aware of?

Forfold (approved):

docs/adrs/000005-api-key-architecture.md
mastercactapus: Hey, thanks for bringing this up. Quick thoughts on the cons:

- **GraphQL Know-How:** You _do_ need to know GraphQL for these keys, but since they're for GraphQL API only, that's kinda given.
- **All-Access Key Temptation:** Some admins might want an easier all-access key, but that's exactly what we're trying to avoid for security reasons.
- **Extra Work vs. Session Tokens:** True, it was more work than sticking with session tokens, but we already did that part. Plus, it wasn't really a big deal in our discussions since we agreed to go down this road.
- **Manual Key Rotation:** We don't have an auto-rotation setup yet, but we agreed to keep it simple for now and handle it manually.

Overall, we were pretty much on the same page during the discussions, and I don't remember any big concerns about this approach that we need to flag in the ADR. We can always circle back if something pops up later.

Issue #3456, opened 2023-11-15
Incorrect Autoupdate of Start and End Dates in Override Form


mastercactapus:
**Describe the Bug:**
Changing the day input in the "Start" or "End" time fields in the Override form automatically updates the other respective field. This makes changing the exact date and time for a future schedule nearly impossible via typing in numbers.

**Steps to Reproduce:**
1. Go to an existing schedule.
2. Click "Override".
3. Click "Cover Someone's Shifts".
4. Click on the Day in the "Start" or "End" time input fields.
5. Change the Day via keyboard input (E.g. input "15" for the 15th).
6. Observe that the other date changes.
- If you change Start, then the End time changes.
- If you change End, then the Start time changes.

**Expected Behavior:**
The day input for "Start" or "End" time field should be able to be changed independently, without automatically updating the other field.

**Observed Behavior:**
Changing the start or end day leads to the automatic update of the other respective field, making it difficult to set a specific future date/time.

**Application Version:**
This issue is seen across all versions of GoAlert.

**Desktop/Browser:**
This issue has been reproduced across all browsers.

**Additional Context**
Possibly related to, or compounded by: https://github.com/target/goalert/issues/3426

Forfold:
Via typing is a duplicate of #3426, however there is also the issue of the date picker -> when selecting a month in the future (with the intention of setting the date backwards) results in the end being updates without the user wanting.

Removing validation will fix both use-cases.

PR #3457, opened 2023-11-15
ui/time pickers: remove ensureInterval, falling back to backend validation

### LINKED ISSUE #3426: ui/time pickers: allow temporary invalid state if typing
When updating an ISOTimePicker `end` date via typing, say with a start date set to 9/14 and an end date of 9/30, if the user wants to set the end date to 9/15 by typing "1" then "5" in the day field, currently nothing happens as "9/1" is less than "9/14"

Forfold:
**Description:**
Removes the use of `ensureInterval` (with fixes to broken references to components always expecting a valid interval)

This allows the user to type any value in the date pickers, with validation falling back to any GraphQL errors returned.

Since the temporary schedule experience is always expecting a valid range, I decided to not render anything on the right until the start/end is valid. Also doesn't render anything if start/end is an interval greater than 6 months, since that's too many subheaders to render while staying performant.

Fixes #3426 and #3456
Part of #3240

**Additional changes:**
- Converts `web/src/app/schedules/useOverrideNotices.ts` to use `urql`
- Fixes a console error in `StrictMode` where `findDOMNode` is deprecated when using the `CSSTransition` component from `react-transition-group` (by adding refs)


Forfold:
Seeing some weird behavior as much of the UI is anticipating the start and end to always be valid. Will need some more work in ensuring states are always good.

KatieMSB (approved):
lgtm

PR #3479, opened 2023-11-22
ui/message-logs: use a sane limit for the max num of graph points

Forfold:
Reduces the limit from 1000 intervals on the graph to a max of 75. This still allows for weekly intervals for a full year (52) and all the minutes in an hour (60), but not by day for a full year (365), which is non-performant especially when segmentation is introduced in #3296, where more data is being loaded for each interval duration.


KatieMSB (approved):

tony-tvu (approved):

mastercactapus:
should pass on rebuild now that it has the fix from master

KatieMSB (approved):

Forfold:
What fix are you referring to @mastercactapus?

Issue #3425, opened 2023-11-06
Admin-Level Label Configuration, Expansion of Search Syntax and Chip Display on Details Page


mastercactapus:
**What problem would you like to solve? Please describe:**
Users can currently add arbitrary labels to services for better search and grouping, but these are not visible unless the labels subpage is visited. Furthermore, there's no way to search for services not containing a specific key.

**Describe the solution you'd like:**
Introduce a configurable admin-level option to specify service labels which would then be visible during service creation and editing, as well as displayed as chips on the service details page. The admin should be able to add the label key, a brief description, and ideally a regex for validation. Also, propose new syntax that doesn't break the existing compatibility for searching services without a specific key.

**Describe alternatives you've considered:**
An alternative could be to manually write and share guidelines for using the current system throughout the organization. However, this could lead to inconsistencies and confusion.

mastercactapus:
```
┌──────────────────────────────────────────┐
│                                          │
│  Create New Service                      │
│                                          │
│  ┌────────────────────────────────────┐  │
│  │ Name                               │  │
│  └────────────────────────────────────┘  │
│                                          │
│  ┌────────────────────────────────────┐  │
│  │ Description                        │  │
│  │                                    │  │
│  │                                    │  │
│  │                                    │  │
│  │                                    │  │
│  │                                    │  │
│  └────────────────────────────────────┘  │
│                                          │
│  ┌────────────────────────────────────┐  │
│  │ Escalation Policy                  │  │
│  └────────────────────────────────────┘  │
│                                          │
│  ┌────────────────────────────────────┐  │
│  │ My Custom Thing                    │  │
│  └────────────────────────────────────┘  │
│                                          │
│                                ┌───────┐ │
│                                │ Submit│ │
│                                └───────┘ │
│                                          │
└──────────────────────────────────────────┘
```

PR #3530, opened 2023-12-12
add label chips to service details

mastercactapus:
**Description:**
This PR adds chips on the service details page for configured labels.

Part of #3425 

**Screenshots:**
image

Service details page will now display any labels under the escalation policy.

N/A


Forfold (commented):
I think below the escalation policy is a good placement for it, but perhaps we should put it under the icon? that way it won't feel as constrained to fit, and wrapping won't be an issue if multiple labels are present

mastercactapus:
@Forfold, I moved them to their own section in the `DetailsPage` component. It seemed fitting since other types will be adopting labels in the future, too.

Forfold (changes requested):
Two small things related to padding, can we:
- Reduce the bottom/top padding of the description text, and
- Add padding to the chips (maybe put in a grid?)

I can push this up if you want!

Screenshot 2023-12-26 at 8 20 51 PM

mastercactapus:
@Forfold, that would be great, ty :)

Forfold (approved):
Here is the updated padding:

Screenshot 2023-12-28 at 9 07 19 AM

LGTM!

KatieMSB (approved):

PR #3531, opened 2023-12-12
config: add Service.RequiredLabels

mastercactapus:
**Description:**
Adds a `Required Labels` field for services.

- Creating a new service will require the user to input a value for the label
- Editing a service will allow updating required label(s), but not force

Part of #3425 

**Out of Scope:**
- Displaying labels on details page (in another PR)
- Reporting on services missing required labels
- Enforcing a hard requirement via API or otherwise
- Validating the input beyond empty/not-empty

**Screenshots:**
image
image

- Required service labels will now appear in the ServiceForm
- Create Service dialog will require a user to set a value

N/A


tony-tvu (changes requested):
A couple things that should be addressed:
- There's no validation on admin configs page for invalid label keys.
- There's no error message when editing a service config's label and submitting.

mastercactapus:
@tony-tvu good catch, I added the validation for Config and updated the form to display label value errors. you can validate it by adding trailing space to the value in one of the dialogs

config/config.go
```diff
@@ -34,6 +34,10 @@ type Config struct {
 		DisableCalendarSubscriptions bool   `public:"true" info:"If set, disables all active calendar subscriptions as well as the ability to create new calendar subscriptions."`
 	}
 
+	Services struct {
+		RequiredLabels []string `public:"true" info:"List of label names to require new services to define."`
+	}
+
```
Forfold, line 40: Can we move this in general? If not, can we move related things like "Disable Label Creation" here?

Forfold (changes requested):
It would be nice to see this hidden requirement before submitting and receiving an error. Also, the error message displayed could definitely be improved (likely very out of scope though) since it's using the `fieldName+index:` and not the `Field Name:`

Screenshot 2023-12-28 at 11 07 51 AM

Forfold:
One more UX related topic, can we make it so we know that we are filling out a label when creating a service? Maybe have the field name be Service Label, and the actual set label can be a fixed left-adornment within the text field.

Old:

Screenshot 2023-12-28 at 11 11 14 AM

New:

Screenshot 2023-12-28 at 11 16 18 AM

mastercactapus:
@Forfold 
- for the validation bit, it's not ideal but yeah probably out of scope since it's just following the existing pattern.
- I agree with updating the text to include the format though, I'll work on that.
- Hmm, I never thought about an adornment, I do like that better, not sure about `Service Label` though on every field -- there could be several; how would that look if there were 5?

I'd like to switch to the adornment thing if we can figure out how to label the fields in a way that's not too repetitive

Forfold:
I think I would be okay with it since each would have a unique adornment.

Or if there's more than 1 we could put in the effort to make it like the admin config text field:

```
-Service Labels--------------------
| first/key:                      |
-----------------------------------
-----------------------------------
| second/key:                     |
-----------------------------------
-----------------------------------
| third/key:                      |
-----------------------------------
```

Forfold:
Updated the label ux:

(one label required)
Screenshot 2024-01-04 at 9 44 08 AM

(more than one label required)
Screenshot 2024-01-04 at 9 44 45 AM

Forfold (approved):

tony-tvu (approved):

config/config.go
mastercactapus: Offline convo summary for ref:
- Keeping it under services as other things may be added service-specific later
- Things like DisableLabelCreation will be removed/deprecated

PR #3533, opened 2023-12-13
dev: Split GraphQL schema into multiple files

mastercactapus:
**Description:**
This PR splits the graphql schema (>1500 lines) into a few different files, setting the groundwork to do more of this in the future.

- the gql api keys types, query and mutations have been moved to their own file
- directives live in their own file
- top-level Query and Mutation are each in their own file
- the `gqltsgen` tool has been updated to sort it's output and fixed some bugs uncovered with multiple files
- the `schema.go` logic has been simplified by using built-in generated code (it was previously parsing the schema on it's own)

**Out of Scope:**
- Only Query, Mutation and gqlapi keys code was split (as an example)

N/A

N/A - the resulting schema is identical, this work just makes the source code easier to manage and review going forward.


devtools/gqltsgen/run.go
```diff
@@ -5,28 +5,36 @@ import (
...
 
 func main() {
 	out := flag.String("out", "", "Output file.")
 	flag.Parse()
 	log.SetFlags(log.Lshortfile)
 	var src []*ast.Source
-	for _, file := range flag.Args() {
-		data, err := os.ReadFile(file)
+	for _, pattern := range flag.Args() {
+		files, err := filepath.Glob(pattern)
```
mastercactapus, line 21: Needed to add globbing support, works the same otherwise

devtools/gqltsgen/run.go
```diff
@@ -5,28 +5,36 @@ import (
...
+			}
+			src = append(src, &ast.Source{
+				Name:  file,
+				Input: string(data),
+			})
+		}
 	}
 
-	doc, err := parser.ParseSchemas(src...)
+	sch, err := gqlparser.LoadSchema(src...)
```
mastercactapus, line 37: LoadSchema interprets things like `extend type` whereas ParseSchemas keeps it as a document and fails to normalize things like that

devtools/gqltsgen/run.go
```diff
@@ -65,16 +73,60 @@ func main() {
 
 	fmt.Fprintf(w, "// Code generated by devtools/gqltsgen DO NOT EDIT.\n\n")
 
-	for _, def := range doc.Definitions {
+	var types []*ast.Definition
+	for _, typ := range sch.Types {
```
mastercactapus, line 77: sch.Types is a map, so putting it into a slice so we can sort it, ensuring consistent output

graphql2/graph/_Mutation.graphqls
```diff
@@ -0,0 +1,101 @@
+
+type Mutation {
```
mastercactapus, line 2: This was moved as-is with no changes other than removing the GQL API key stuff

graphql2/graph/_Query.graphqls
```diff
@@ -0,0 +1,132 @@
+
+type Query {
```
mastercactapus, line 2: Same as mutation, cut and paste with no changes other than removing the gql API key stuff

graphql2/graph/gqlapikeys.graphqls
```diff
@@ -0,0 +1,48 @@
...
+  gqlAPIKeys: [GQLAPIKey!]!
+}
+
+extend type Mutation {
+  createGQLAPIKey(input: CreateGQLAPIKeyInput!): CreatedGQLAPIKey!
+  updateGQLAPIKey(input: UpdateGQLAPIKeyInput!): Boolean!
+  deleteGQLAPIKey(id: ID!): Boolean!
+}
+
+type CreatedGQLAPIKey {
```
mastercactapus, line 11: This is all cut-and-paste exactly as it was in the original graphql.schema doc

graphql2/schema.go
```diff
@@ -7,34 +7,16 @@ import (
...
 var schemaFields []string
 var astSchema *ast.Schema
 
-// Schema will return the GraphQL schema.
-func Schema() string {
-	return schema
-}
-
 func init() {
-	schDoc, err := parser.ParseSchema(&ast.Source{Input: schema})
```
mastercactapus, line 26: This was all unecessary complexity since we can call directly into the generated code

KatieMSB (approved):

ethan-haynes (approved):

PR #3560, opened 2023-12-28
dev: add playwright component tests

mastercactapus:
**Description:**
Adds component testing via `make test-components` that can be used to run tests on react components.

- An initial test for the `TelTextField` was added


tony-tvu (dismissed):
LGTM

tony-tvu (approved):

tony-tvu (dismissed):

.gitignore
```diff
@@ -72,4 +72,11 @@ smoketest_db_dump
...
 /.pnp.*
-
+/test-results/
+/playwright-report/
+/blob-report/
+/playwright/.cache/
+/test-results/
+/playwright-report/
+/blob-report/
+/playwright/.cache/
```
Forfold, line 82:
```suggestion
/test-results/
/playwright-report/
/blob-report/
/playwright/.cache/
```

web/src/app/util/TelTextField.tsx
```diff
@@ -45,7 +45,7 @@ export default function TelTextField(
   // check validation of the input phoneNumber through graphql
   const [{ data }] = useQuery({
     query: isValidNumber,
-    variables: { number: '+' + phoneNumber },
```
Forfold, line 48: Is the `+` sign w/country code no longer needed?

Forfold (changes requested):

web/src/app/util/TelTextField.tsx
tony-tvu: This was a bug we found that currently exists in production. We're already appending a `+` on line 99 in the `handleChange` func, so it's been getting sent back to API as `++15555555`.

tony-tvu (approved):

Forfold (approved):

PR #3561, opened 2023-12-29
ui/schedules: create confirmation component for editing a temporary schedule

### LINKED ISSUE #3497: ui/schedules: add diff confirmation to temp sched edit
Confidence when making changes to a temporary schedule

Add a diff feature to track changes in shifts. Changes include added or removed shifts, and added or removed users.

Diffing via JSON serialization was touched on during a meeting.

**Additional context:**
1. User adds all shifts to temp sched, wants to finish & create
2. User clicks Submit
3. Dialog transitions to new screen, old (current) shift set on the left, new shift set on the right.
  - Shifts should line up one-to-one so differences are easily identifiable.
  - Shifts that are the same can be dimmed?
  - Shifts that are different can be accentuated?

Forfold:
**Description:**
This PR adds a confirmation screen to the temporary schedule dialog (when editing). The experience should remain the same when creating a new temporary schedule.

When a temporary schedule is edited, after clicking on "Next", a new component will render displaying the current temporary schedule's shifts on the left, with the proposed changes on the right. If everything looks good, the user can submit- otherwise, they can click "Back" and continue to make changes.

Gap coverage errors will be handled before this screen.

Closes #3497

**Out of Scope:**
N/A

**Screenshots:**

**Additional Info:**
An information icon has been added to the gap coverage list items to let the user know they can click on the coverage gap to automatically fill in the times for that range

### DIFF
--- a/web/src/app/lists/FlatList.tsx
+++ b/web/src/app/lists/FlatList.tsx
@@ -44,7 +44,8 @@ import { ExpandLess, ExpandMore } from '@mui/icons-material'
 const useStyles = makeStyles({
   alertAsButton: {
     width: '100%',
-    margin: '0.5rem 0 0.5rem 0',
+    marginTop: '6px',
+    marginBottom: '6px',
     '&:hover, &.Mui-focusVisible': {
       filter: 'brightness(90%)',
     },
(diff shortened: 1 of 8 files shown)


mastercactapus (approved):

tony-tvu (approved):
LGTM!

PR #3563, opened 2023-12-29
playwright-ct: Add GQLMock for mocking individual graphql operations

mastercactapus:
**Description:**
This PR adds a helper for mocking out individual GraphQL operations.

Since the playwright API only allows per-path handlers, this can be used to switch off of the `operationName` field of a GraphQL query.

- The existing `TelTextField.ct-spec.tsx` test has been updated to use the new helper.
- A default response was added by default for the `ConfigProvider` component, this can be overridden/added to as-needed.


tony-tvu (approved):

mastercactapus:
Closing in favor of Storybook.js work

Issue #3578, opened 2024-01-08
tech debt: reduce time-to-test and noise with storybook


Forfold:
Introduced in https://github.com/target/goalert/pull/3571, Storybook works well but only if you have a newer computer.

It would be nice if we could reduce the time to test `make test-components`, which is currently sitting around 1-2 minutes for me. Storybook also takes at least a minute to render the components for `make storybook`, once the browser automatically opens.

![image](https://github.com/target/goalert/assets/11381794/78417b2a-ce85-4a97-8cf4-e737bbb6eb82)

Also, the build output is super noisy both locally and in CI so it would be great to get that sorted out as well.

Forfold:
It appears that Storybook also attempts to access too many files at once, and consistently fails after being open for 5-10 minutes.

![Screenshot 2024-01-16 at 11 14 27 AM](https://github.com/target/goalert/assets/11381794/df8ccc45-d9ea-4854-8b7b-962ccc164893)

Forfold:
As is Storybook so far is a 👎 from me with how it runs on my machine..

mastercactapus:
@Forfold is this still an issue with the latest versions of everything?

PR #3579, opened 2024-01-08
dev: pipe storybook warnings to null

mastercactapus:
**Description:**
Updating the Makefile to pipe storybook build stderr to `/dev/null`. Fixes the flood of logs when running `test-components`


Forfold (approved):
lgtm, part of #3578

PR #3590, opened 2024-01-10
ui: convert CalendarSubscribe dialogs to ts and to use urql instead of apollo

Forfold:
Part of https://github.com/target/goalert/issues/3240 and #2318

Converts `CalendarSubscribeCreateDialog`, `CalendarSubscribeEditDialog`, and `CalendarSubscribeDeleteDialog` to typescript and updates to use `urql` instead of `@apollo/client`


web/src/app/schedules/calendar-subscribe/CalendarSubscribeEditDialog.tsx
```diff
@@ -36,27 +43,31 @@ export function CalendarSubscribeEditDialogContent(props) {
...
+      onSubmit={() =>
+        commit(
+          {
+            input: {
+              id: props.data.id,
+              name: value.name,
+              fullSchedule: value.fullSchedule,
+            },
+          },
+          { additionalTypenames: ['User'] },
```
tony-tvu, line 62:
```suggestion
          { additionalTypenames: ['User', 'UserCalendarSubscription'] },
```
I believe this should include `UserCalendarSubscription` to avoid the update lag. You can replicate this if you edit a subscription's title and then immediately edit it again - it'll show the old title for 2-5 seconds in the edit dialog before it updates to the new value.

tony-tvu (changes requested):

web/src/app/schedules/calendar-subscribe/CalendarSubscribeEditDialog.tsx
Forfold: I'm not seeing this behavior- if I edit and save it updates after just a moment for me

mastercactapus: @Forfold that's likely just timing, since our update mutations return a bool -- there is no immediate update _unless_ we add the object type here.

most likely without it we'll see flaky tests. 2-5 seconds usually means the catch-all-refetch is the only reason you're seeing the correct data come in

it should update almost instantly locally if it's configured correctly.

Forfold: This information is nested under the User query though, no? This query is only used by the edit dialog by the looks of it, which closes and resets all fields on complete

Forfold: After testing, both solutions work just fine, but we don't need both. I pushed up the requested change. @mastercactapus @tony-tvu

mastercactapus (dismissed):

mastercactapus (approved):

tony-tvu (approved):

Issue #3492, opened 2023-11-29
Enhancement to Temporary Schedules Feature


mastercactapus:
**What problem would you like to solve? Please describe:**
Improvements are sought for the Temporary Schedules feature based on recent user feedback. 

**Describe the solution you'd like:**

1. **History and Change Confidence**
     - Add a diff feature to track changes in shifts. Changes include added or removed shifts, and added or removed users. 

2. **Timezones**
     - Implement an app wide timezone setting defaulted to local. Always display time in local but provide an option to hover to view schedule or rotation-specific timezone.

3. **Overlapping Schedules**
     - Change the handling of overlapping temporary schedules to trim existing schedules instead of merging. Existing schedules that are overlapped by a new one will be truncated or split. 

4. **Shift Editing User Experience Improvements**
     - Implement fixed shifts configuration for temporary schedules. 
     - Make user selection a required field unless 'allow gaps' is checked. 
     - Add a new feature to select fallback user or schedule to automatically fill gaps.

**Additional context:**
This updates are expected to enhance user efficiency and confidence when managing temporary schedules.

mastercactapus:
Mind map from discussion:
![Temp  Schedules](https://github.com/target/goalert/assets/595010/890134e4-5dd2-45fb-a96a-af5c02675081)

PR #3606, opened 2024-01-16
ui/temp-sched: Add fixed shifts for temporary schedules

KatieMSB:

**Description:**
Adds new fields in the temporary schedule dialogs to have fixed shifts for the entire schedule. Shift duration defaults to 1 Day for all shifts but shifts can still be added with custom durations in the Add Shift form. Coverage gaps are also now rendered by the set shift duration, and more clearly show the time frame without coverage.

Part of #3492

**Screenshots:**
Screenshot 2024-01-18 at 12 27 23 PM
Screenshot 2024-01-18 at 1 59 09 PM


Forfold (changes requested):
Looks like the type/padding is getting messed up for the day interval here (this is on open without making any changes)

Screenshot 2024-01-18 at 7 58 47 AM

Also, can we make it so the shift length can be temporarily null? With it set to 1 day, I tried to change it to 6 hours by typing Backspace then 1, but it wouldn't let me so I had to use the arrow keys. But with the arrow keys, there was a bit of lag for each digit as it went up to render everything on the right

Forfold (approved):
LGTM! Working well & as intended

ethan-haynes (approved):

PR #3610, opened 2024-01-19
statusmgr: expire status subscriptions after 7 days

### LINKED ISSUE #3245: Status Updates for Closed Alerts Lack Max-Age
**Describe the Bug:**
Status updates for closed alerts don't seem to have a max-age, causing confusion for users who receive updates for alerts with old original notifications. This appears particularly prevalent with alerts left open for extended periods.

**Steps to Reproduce:**
1. Open an alert.
2. Leave the alert open for a long period (e.g., a year).
3. Close the alert.
4. See the status update pertaining to the closed alert.

**Expected Behavior:**
A user would typically expect not to receive status updates for alerts that were activated a long time ago and have since been closed. 

**Observed Behavior:**
Users are receiving status updates for closed alerts irrespective of when the original notification was issued.

**Application Version:**
This report is based on user feedback and hasn't been tied to a specific GoAlert version or platform.

**Additional Context:**
Implementing a max-age for statuses could mitigate confusion, improving user experience significantly. Users are primarily disrupted when receiving updates about alerts triggered a long time ago.

mastercactapus:
**Description:**
This PR updates the statusmgr to expire/delete alert status subscriptions after 7 days of inactivity.

Closes #3245

N/A

N/A

**Additional Info:**
A new `updated_at` column was added, and the engine will now update the timestamp each time a status update is processed. Rows where `updated_at` is more than 7 days old are deleted/cleaned up, no longer requiring processing and no longer resulting in update messages.

If an alert escalates and a _new_ message is sent, a new subscription would be created and behave as it does today. Effectively this means you may receive updates to a notification alert for up to 7 days as an absolute maximum.

NOTE: Existing status subscriptions will be set to the current time, so the 7 day maximum will take effect in full 7 days after this update is deployed.


tony-tvu (dismissed):
LGTM

tony-tvu (approved):

Forfold (approved):

Issue #2639, opened 2022-09-29
notification: plug-able providers


mastercactapus:
**What problem would you like to solve? Please describe:**
Adding a new notification provider currently requires MANY touch-points in code and a lot of overhead: DB migrations, new types in several packages, updating various switch statements, etc...

**Describe the solution you'd like:**
Make destinations generic so that things like listing Slack channels, sending messages to them, SMS, voice, etc... can be done through a separate service that can also be built/packaged with the core app.

**Describe alternatives you've considered:**
As it stands, the only way to extend providers is to update a lot of spider-webby code.

**Additional context:**
This will provide a clean way to add new providers quickly: #2290, #2102, #2049, #871, #552

mastercactapus:
At a high level, here's what this might look like for an SMS provider. The plugin could also be built into GoAlert, but the overall functionality will be the same.

![image](https://user-images.githubusercontent.com/595010/204911266-eac1639c-51df-4912-89d8-6399ee5bf572.png)

Additionally, optional interfaces for listing/enumeration (e.g., Slack channels) would be defined.

tong-bluehill:
+1 for this feature. 
In the above flow chart, can the class sms be a more general class like NotificationRequest?
Just curious about the priority. Is there a tool to see all features progress? 

Thank you!

PR #3700, opened 2024-02-21
dest: add API support to create EP step with destinations

tony-tvu:

**Description:**
This PR adds support for creating an escalation policy step using generic destinations when the experimental flag is set make start EXPERIMENTAL=dest-types.

part of https://github.com/target/goalert/issues/2639

**Out of Scope:**
N/A

**Screenshots:**
N/A

N/A

Added `CreateEscalationPolicyStepInput` resolver. Also added `favoritesFirst` field to `DestinationFieldSearchInput` schema to show favorites first when searching for rotations, schedules, and users.

**Additional Info:**
N/A


graphql2/graph/destinations.graphqls
```diff
@@ -37,6 +37,7 @@ input DestinationFieldSearchInput {
   omit: [String!] # values/ids to omit from results
   after: String # cursor to start search from
   first: Int = 15 # number of results to return
+  favoritesFirst: Boolean = false # sort favorite rotations first
```
mastercactapus, line 40: false is already implied
```suggestion
  favoritesFirst: Boolean # sort favorited items (if supported) first
```

---

That said, is this necessary? Would we ever _not_ want to return favorites first in this context?

I think everywhere we return users we already happen to do fav. first, but in this context it's limited even further to be search-select situations only -- so we can probably leave out the option and just always set favorites first behind the scenes

mastercactapus (changes requested):

graphql2/graph/destinations.graphqls
tony-tvu: I agree. I've updated it so that we're setting `favoritesFirst: true ` when searching for users, rotations, and schedules in `destinationtypes.go`. Although, since `&graphql2.UserSearchOptions` requires a bool pointer, I had to set `favFirst := true ` on line 128 and use its pointer address, which I suspect isn't best practice. I'd be open to suggestions on how to do this more efficiently.

graphql2/graphqlapp/escalationpolicy.go
```diff
@@ -19,12 +19,30 @@ import (
...
 
+func (a *App) CreateEscalationPolicyStepInput() graphql2.CreateEscalationPolicyStepInputResolver {
+	return (*CreateEscalationPolicyStepInput)(a)
+}
+
+func (a *CreateEscalationPolicyStepInput) Actions(ctx context.Context, input *graphql2.CreateEscalationPolicyStepInput, actions []graphql2.DestinationInput) error {
+	tgts := make([]assignment.RawTarget, len(actions))
+	var err error
+	for i, action := range actions {
+		tgts[i], err = CompatDestToTarget(action)
```
mastercactapus, line 37: Let's call ValidateDestination here before mapping to target, that will give us proper error code responses if any of the destinations are invalid.

Make sure the `fieldName` includes the index

graphql2/graphqlapp/escalationpolicy.go
```diff
@@ -19,12 +19,33 @@ import (
...
+	return (*CreateEscalationPolicyStepInput)(a)
+}
+
+func (a *CreateEscalationPolicyStepInput) Actions(ctx context.Context, input *graphql2.CreateEscalationPolicyStepInput, actions []graphql2.DestinationInput) error {
+	tgts := make([]assignment.RawTarget, len(actions))
+	var err error
+	for i, action := range actions {
+		if ok, _ := (*App)(a).ValidateDestination(ctx, "input.actions", &action); !ok {
+			return validation.NewFieldError("actions["+strconv.Itoa(i)+"].", "invalid DestInput")
+		}
```
mastercactapus, line 39:
```suggestion
	for i, action := range actions {
		if ok, _ := (*App)(a).ValidateDestination(ctx, fmt.Sprintf("input.actions.%d.dest", i), &action); !ok {
			return nil // ValidateDestination already sets error state
		}
```

mastercactapus (approved):

ethan-haynes (approved):

PR #3717, opened 2024-02-27
dest/escalation-policies: use destinations for EP step form

tony-tvu:

**Description:**
This PR adds an escalation policy step form component that utilizes generic destinations. This component will have no effect on the current state of the UI, even with experimental flag set to `dest-types`. 

Part of https://github.com/target/goalert/issues/2639

**Out of Scope:**
N/A

**Screenshots:**
Screenshot 2024-02-27 at 10 21 37 AM

N/A

N/A

**Additional Info:**
N/A


graphql2/graphqlapp/destinationdisplayinfo.go
```diff
@@ -34,6 +34,11 @@ func (a *Destination) DisplayInfo(ctx context.Context, obj *graphql2.Destination
 func (a *Query) DestinationDisplayInfo(ctx context.Context, dest graphql2.DestinationInput) (*graphql2.DestinationDisplayInfo, error) {
 	app := (*App)(a)
 	cfg := config.FromContext(ctx)
+
+	if err := app.ValidateDestination(ctx, "input.action", &dest); err != nil {
```
mastercactapus, line 38: For this query the `input` variable is the `DestinationInput`

```suggestion
	if err := app.ValidateDestination(ctx, "input", &dest); err != nil {
```

web/src/app/escalation-policies/PolicyStepFormDest.stories.tsx
```diff
@@ -0,0 +1,130 @@
...
+              data: {
+                destinationDisplayInfo: {
+                  text: vars.input.values[0].value,
+                  iconURL: 'builtin://phone-voice',
+                  iconAltText: 'Voice Call',
+                },
+              },
+            }
+          }
+          return HttpResponse.json(body)
```
mastercactapus, line 66: nit: I think this would be clearer to just have `return HttpResponse.json({})` in the first part and then not need the `else` to keep it all flat. Or we could do a switch.

In any case, we should probably switch off of `.type` instead of picking up a field value.
```
switch (vars.input.type) {
case "webhook":
  return HttpResponse.json({data: ... })
case "phone":
  return HttpResponse.json({data: ... })
default:

  return HttpResponse.json({error: ... })
}
```

web/src/app/escalation-policies/PolicyStepFormDest.stories.tsx
```diff
@@ -0,0 +1,130 @@
...
+        'https://target2.com',
+      )
+    })
+
+    await userEvent.click(await canvas.findByText('Add Action'))
+
+    // expect to see action added
+    await expect(await canvas.findByText('https://target2.com')).toBeVisible()
+    await expect(await canvas.findByTestId('WebhookIcon')).toBeVisible()
+  },
```
mastercactapus, line 129: I feel like this is really just testing the DestField is working.

I'd like to see an interaction test for:
- Click Add Action, and show an error on dest type, the field, and a generic error (i.e., unknown, to the left of the button)
- Editing an input field clears the errors
- With no validation error, Add Action adds the chip to the list and clears it's form

tony-tvu: I was able to add assertions for all of these, except for showing an error on the destination type and a generic error (one of these is covered in https://github.com/target/goalert/pull/3718). This PR already consists of 440 lines of code changes. We should consider increasing the PR size threshold if we want to add these extra assertions because attempting to add all these test cases and trying to stay under 500 lines of code change has been a (seemingly unnecessary) challenge.

mastercactapus (approved):

ethan-haynes (approved):

PR #3720, opened 2024-02-28
user contact method edit dialog dest

ethan-haynes:
Description:
Adds the Edit dialog for user contact methods using the new dest API.

Which issue(s) this PR fixes:
Part of https://github.com/target/goalert/issues/2639

Screenshots:
Screenshot 2024-02-28 at 2 31 50 PM


graphql2/schema.graphql
```diff
@@ -1233,8 +1233,13 @@ input CreateUserNotificationRuleInput {
 
 input UpdateUserContactMethodInput {
   id: ID!
+  type: ContactMethodType
+
+  dest: DestinationInput @experimental(flagName: "dest-types")
```
mastercactapus, line 1238: it is never valid to update the destination of a contact method, only the name, and status update preference

`value` is deprecated for that reason, as a new "destination" requires another consent/validation flow.

`type` should also have the deprecation warning, not sure why that was missed.

mastercactapus (changes requested):
Everything's looking good code wise for the most part, but I noticed 2 issues functionally validating:

- <details><summary>name field errors aren't showing up on the field</summary>
  
  image</details>
- The contact method list doesn't update on success (likely need `additionalTypeNames: UserContactMethod`)

mastercactapus (dismissed):

web/src/app/users/UserContactMethodEditDialogDest.tsx
```diff
@@ -0,0 +1,112 @@
...
+  }
+`
+
+const mutation = gql`
+  mutation UpdateUserContactMethod($input: UpdateUserContactMethodInput!) {
+    updateUserContactMethod(input: $input)
+  }
+`
+
+export default function UserContactMethodCreateDialogDest(props: {
```
Forfold, line 41:
```suggestion
export default function UserContactMethodEditDialogDest(props: {
```

web/src/app/users/UserContactMethodEditDialog.tsx
```diff
@@ -96,3 +102,15 @@ export default function UserContactMethodEditDialog({
     />
   )
 }
+
+export default function UserContactMethodEditDialogSwitch(
```
Forfold, line 106: not sure what's happening but the whole page flickers when i open the edit dialog from profile. seeing in stage too so not blocking but would be nice to have fixed

web/src/app/users/UserContactMethodEditDialog.tsx
```diff
@@ -96,3 +102,15 @@ export default function UserContactMethodEditDialog({
...
+  props: UserContactMethodEditDialogProps,
+): React.ReactNode {
+  const isDestTypesSet = useExpFlag('dest-types')
+
+  if (isDestTypesSet) {
+    return <UserContactMethodEditDialogDest {...props} />
```
Forfold, line 112: something in here is giving the following console error:
`Warning: validateDOMNesting(...): <p> cannot appear as a descendant of <p>.`

Forfold (changes requested):

mastercactapus:
Also to note: let's switch to use `canvasElement` instead of `screen` -- check some of the other examples for how to setup `disablePortal` to make that work

web/src/app/users/UserContactMethodEditDialog.tsx
ethan-haynes: Looks like this issue is on prod as well. created an issue for it since it is unrelated to this PR https://github.com/target/goalert/issues/3751

web/src/app/users/UserContactMethodEditDialog.tsx
ethan-haynes: issue was due to nesting another component in Typography component in UserContactMethodListDest.tsx. Pushed Fix

web/src/app/users/UserContactMethodEditDialogDest.tsx
```diff
@@ -0,0 +1,115 @@
...
+          fieldID
+          value
+          label
+        }
+      }
+      value
+      statusUpdates
+    }
+  }
+`
```
mastercactapus, line 33: I think we can remove type & value; we'll likely mark them as deprecated soon (in favor of only using dest)

```suggestion
const query = gql`
  query userCm($id: ID!) {
    userContactMethod(id: $id) {
      id
      name
      dest {
        type
        values {
          fieldID
          value
          label
        }
      }
      statusUpdates
    }
  }
`
```

mastercactapus (dismissed):

mastercactapus (approved):

Issue #3653, opened 2024-02-05
Add Metadata to Alerts for Traceability and Integration


mastercactapus:
**What problem would you like to solve? Please describe:**
Difficulties in tracing and integrating with other systems due to the lack of ability to associate arbitrary metadata with an alert upon its creation and retrieval.

**Describe the solution you'd like:**
- Implement a new `meta` in alert creation endpoints to accommodate a set of `key=value` pairs. This should also apply to the `createAlert` mutation input.
- Create a new `alert_data` table for storing the new data, with an `alert_id` column and a data column (`jsonb`).
- Update the UI to display these key-value pairs.
- Extend Webhook notifications to include the new `meta` values.

Example of the Graphql input and struct formats:

```graphql
input AlertMetadataInput {
  key: String!
  value: String!
}

input CreateAlertInput {
  // ...
  meta: [AlertMetadataInput!]
}

type AlertMetadata {
  key: String!
  value: String!
}

type Alert {
  // ...
  meta: [AlertMetadata!]
}
```

```go
type AlertMetaValue struct {
	Key string `json:"key"`
	Value string `json:"value"`
}

const TypeAlertMetaV1 = "alert_meta_v1"

type AlertMeta struct {
	Type string `json:"type"`
	AlertMetaV1 []AlertMetaValue `alert_meta_v1`
}
```

**Additional context:**
This feature can enhance the handling of alerts to make them compatible with query parameters as well as GraphQL.

Forfold:
@mastercactapus What do you envision `Type string `json:"type"`` being within `AlertMeta`?

mastercactapus:
@Forfold That struct is specifically what's going to Postgres -- the idea is to have type/version info embedded in it so we have something to key off of in any migrations/changes we need to make down the line.

It wouldn't be included in any of the API/external stuff, just internal

Forfold:
@mastercactapus Gotcha, so the external integration wouldn't ever see this value, as it would be a hardcoded string set in the function at time of insert/update?

mastercactapus:
@Forfold yep, exactly. Not only that, but the `alert.Store` method shouldn't even expose it, it's just to ensure integrity within the store method to the DB (since things change over time).

mastercactapus:
Also, a nice addition would be adding a `metaValue(key: string): string` field to `Alert` in the GraphQL schema. It would play nice with GraphQL, allowing mapping (potentially multiple) metadata fields in the return structure, for example:
```graphql
[...]

PR #3764, opened 2024-03-21
Add alert metadata

shivanishendge:
- [] Added appropriate tests for any new functionality.
- [] All new and existing tests passed.

**Description:**
- Implement a new meta in alert creation endpoints to accommodate a set of key=value pairs. 
- Create a new alert_data table for storing the new data, with an alert_id column and a data column (jsonb).
- Extend Webhook notifications to include the new meta values.

[Add Metadata to Alerts for Traceability and Integration](https://github.com/target/goalert/issues/3653)

**Out of Scope:**
Display of metadata in goAlert UI

**Screenshots:**
n/a

None

Users can add 'meta' field in the create alert requests to track additional metadata(kv pairs)


alert/alert.go
```diff
@@ -25,6 +26,7 @@ type Alert struct {
 	ServiceID string    `json:"service_id"`
 	CreatedAt time.Time `json:"created_at"`
 	Dedup     *DedupID  `json:"dedup"`
+	Meta      AlertMeta `json:"meta"`
```
mastercactapus, line 29: I think this should be a separate method for the store rather than bundling it all together on the struct -- we only need to make the second call if metadata is provided, and the only way to retrieve it is GraphQL which we can handle that via resolver

alert/alert.go
```diff
@@ -57,12 +59,21 @@ func (a Alert) Normalize() (*Alert, error) {
 	}
 	a.Summary = strings.Replace(a.Summary, "\n", " ", -1)
 	a.Summary = strings.Replace(a.Summary, "  ", " ", -1)
+
+	var validateMeta error
+	for k := range a.Meta.AlertMetaV1 {
+		if k == "" {
+			validateMeta = validation.NewFieldError("Meta", "key must be non empty string")
+		}
+	}
```
mastercactapus, line 68: If we use a separate method, we can handle validation there.

That aside, we should set sane limits on the number of keys and key & value lengths also. We should also probably limit the keys to ASCII values or some sort of subset.

Lastly, just a nit, but we should break the loop if we find this.

alert/meta.go
```diff
@@ -0,0 +1,26 @@
...
+func ToAlertMeta(meta AlertMetaInput) AlertMeta {
+	alertMeta := AlertMetaData{}
+	for _, v := range meta {
+		alertMeta[v.Key] = v.Value
+	}
+	return AlertMeta{
+		Type:        TypeAlertMetaV1,
+		AlertMetaV1: alertMeta,
+	}
+}
```
mastercactapus, line 26: This belongs in the GraphQL layer rather than here

alert/store.go
```diff
@@ -527,6 +529,14 @@ func (s *Store) _create(ctx context.Context, tx *sql.Tx, a Alert) (*Alert, *aler
...
 	}
 
+	// insert alert metadata
+	err = s.insertMetaData(tx, ctx, a.ID, a.Meta.AlertMetaV1)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	ctx = log.WithFields(ctx, log.Fields{"AlertID": a.ID, "Meta": a.Meta.AlertMetaV1})
+	log.Logf(ctx, "Alert created.")
```
mastercactapus, line 539: I think we already have log entries for this, also, I don't think we want to log metadata since it's specific to the client and potentially large

alert/meta.go
```diff
@@ -0,0 +1,26 @@
+package alert
+
+const TypeAlertMetaV1 = "alert_meta_v1"
+
+type AlertMeta struct {
+	Type        string        `json:"type"`
+	AlertMetaV1 AlertMetaData `json:"alert_meta_v1"`
+}
+
+type AlertMetaData map[string]string
```
mastercactapus, line 10: Metadata is one word, also because we're in the context of the `alert` package `Alert` stutters, so this should just be `Metadata`

```suggestion
type Metadata map[string]string
```

alert/meta.go
```diff
@@ -0,0 +1,26 @@
+package alert
+
+const TypeAlertMetaV1 = "alert_meta_v1"
+
+type AlertMeta struct {
+	Type        string        `json:"type"`
+	AlertMetaV1 AlertMetaData `json:"alert_meta_v1"`
+}
```
mastercactapus, line 8: These types should be private, and are only used going to/from the DB layer

genericapi/handler.go
```diff
@@ -89,6 +89,17 @@ func (h *Handler) ServeCreateAlert(w http.ResponseWriter, r *http.Request) {
...
+	var md alert.AlertMetaInput
+	var meta alert.AlertMeta
+	if metaData != "" {
+		err = json.Unmarshal([]byte(metaData), &md)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+		meta = alert.ToAlertMeta(md)
+	}
```
mastercactapus, line 102: We don't want to encode JSON inside the URL or form values, let's stick to form encoding

```suggestion

	metaData := make(alert.Metadata)
	for _, v := range r.Form["meta"] {
		key, val, ok := strings.Cut(s, "=")
		if !ok {
			continue
		}
		metaData[key] = val
	}
```

genericapi/handler.go
```diff
@@ -100,6 +111,7 @@ func (h *Handler) ServeCreateAlert(w http.ResponseWriter, r *http.Request) {
 
 		var b struct {
 			Summary, Details, Action, Dedup *string
+			Meta                            alert.AlertMetaInput
```
mastercactapus, line 114: We don't have the same constraints here we do with GraphQL, so the simplest API would be just to require an object of key/value strings directly.

Since this is an API handler, using `map[string]string` instead of `alert.Metadata` here just serves to remove indirection so someone looking at this code can immediately see the expected type without having to switch context and look at the alert package.

```suggestion
			Meta                            map[string]string
```

graphql2/graphqlapp/alert.go
```diff
@@ -403,6 +403,17 @@ func (m *Mutation) CreateAlert(ctx context.Context, input graphql2.CreateAlertIn
...
+	if input.Meta != nil {
+		alertMeta := alert.AlertMetaData{}
+		for _, v := range input.Meta {
+			alertMeta[v.Key] = v.Value
+		}
+		a.Meta = alert.AlertMeta{
+			Type:        alert.TypeAlertMetaV1,
+			AlertMetaV1: alertMeta,
+		}
+	}
```
mastercactapus, line 415: Use `withContextTx` and create the alert, and then if this is `!= nil` call the AlertStore method directly (part of separating Meta from the base Alert struct)

graphql2/graphqlapp/alert.go
```diff
@@ -578,3 +589,27 @@ func (m *Mutation) UpdateAlertsByService(ctx context.Context, args graphql2.Upda
...
+func (a *Alert) Meta(ctx context.Context, alert *alert.Alert) ([]graphql2.AlertMetadata, error) {
+	md, err := a.AlertStore.Metadata(ctx, alert.ID)
+	if err != nil {
+		return nil, err
+	}
+	var alertMeta []graphql2.AlertMetadata
+	for k, v := range md {
+		alertMeta = append(alertMeta, graphql2.AlertMetadata{Key: k, Value: v})
+	}
+	return alertMeta, nil
```
mastercactapus, line 602: Before returning, sort this slice by keys so it's not random each time

graphql2/graphqlapp/alert.go
```diff
@@ -578,3 +589,27 @@ func (m *Mutation) UpdateAlertsByService(ctx context.Context, args graphql2.Upda
...
+func (a *Alert) MetaValue(ctx context.Context, alert *alert.Alert, key string) (*string, error) {
+	md, err := a.AlertStore.Metadata(ctx, alert.ID)
+	val := ""
+	if err != nil {
+		return &val, err
+	}
+	if value, ok := md[key]; ok {
+		val = value
+	}
+	return &val, nil
```
mastercactapus, line 614: nit: let's simplify this, also if we get a not-found error, return empty string
```suggestion
	md, err := a.AlertStore.Metadata(ctx, alert.ID)
	if errors.Is(err, sql.ErrNoRows) {
		return new(string), nil // note: if we change the schema to `String!` we can clean this up
	}
	if err != nil {
		return nil, err
	}
	return &md[key], nil
```

graphql2/graphqlapp/dataloaders.go
```diff
@@ -174,6 +175,7 @@ func (app *App) FindOnePolicy(ctx context.Context, id string) (*escalation.Polic
 func (app *App) FindOneService(ctx context.Context, id string) (*service.Service, error) {
 	loader, ok := ctx.Value(dataLoaderKeyService).(*dataloader.Loader[string, service.Service])
 	if !ok {
+		fmt.Println("using service store")
```
mastercactapus, line 178: leftover?

graphql2/schema.graphql
```diff
@@ -904,6 +911,10 @@ type Alert {
   metrics: AlertMetric
 
   noiseReason: String
+
+  meta: [AlertMetadata!]
+
+  metaValue(key: String!): String
```
mastercactapus, line 917: Let's just return an empty string for missing values

```suggestion
  metaValue(key: String!): String!
```

migrate/migrations/20240212125328-alert-metadata.sql
```diff
@@ -0,0 +1,12 @@
...
+	alert_id integer NOT NULL,
+	id uuid DEFAULT gen_random_uuid() NOT NULL,
+	metadata jsonb,
+	CONSTRAINT alerts_metadata_alert_id_fkey FOREIGN KEY (alert_id) REFERENCES alerts(id) ON DELETE CASCADE,
+	CONSTRAINT alerts_metadata_pkey PRIMARY KEY (id, alert_id),
+	CONSTRAINT alerts_metadata_id UNIQUE (alert_id)
+);
+
+-- +migrate Down
+DROP TABLE alerts_metadata;
```
mastercactapus, line 12: Let's follow the same pattern we use for schedules

```suggestion
CREATE TABLE alert_data (
	alert_id BIGINT PRIMARY KEY REFERENCES alerts (id) ON DELETE CASCADE,
	metadata jsonb
);

-- +migrate Down
DROP TABLE alert_data;
```

notification/alert.go
```diff
@@ -9,11 +9,17 @@ type Alert struct {
 	Details     string
 	ServiceID   string
 	ServiceName string
+	Meta        []AlertMeta
```
mastercactapus, line 12: The []{key, value} stuff is just for GraphQL because the language does not support any kind of map.

```suggestion
	Meta        alert.Metadata
```

mastercactapus (changes requested):

genericapi/handler.go
shivanishendge: Okay👍
Q: Do you think the notification webhook should also send the meta values of type map[string]string?

alert/alert.go
shivanishendge: Okay, Is there any convention followed in limiting these values. I saw that `label Key has limit of 255 in the codebase`, should the meta `Key` in meta also follow the same?

mastercactapus: Yeah, I think that's reasonable for the key -- it's possible we may index the keys in the future, so why don't we go with:


- keys: validate.ASCII 1-255 characters
- values: just limit the length to 2k/2048 for now, (see note below)
- total number of key/value pairs: let's start with 16 (you can do `len()` on the map) (see note below)

note:
For values and # of keys, we don't have validation functions for those. For values we don't want to use `validate.Text` because it restricts whitespace and we want to preserve the integrity/exactness of metadata.

Here's what that may look like and how to create a validation error for the value

```go
if len(meta) > 16 { // meta is of type map[string]string
	return validation.NewFieldError("Meta", "must not have more than 16 keys")
}

var err error
for k, v :=range meta {
	fieldName := fmt.Sprintf("Meta[%s]", k)
	err = validate.Many(err, validate.ASCII(fieldName, k))
	if len(v) > 2048 {
		err = validate.Many(err,
			validation.NewFieldError(fieldName, "cannot be more than 2048 characters"),
		)
	}
}
if err != nil {
	return err
}
```

mastercactapus: One thing I'm wondering is if we should truncate long values, rather than rejecting. Keys can be pre-defined, but values may likely be a pass-through from other systems.

Since value may be arbitrary, it'd likely be better to truncate and create an alert, than have certain alerts fail to create (i.e., due to unexpected data)

genericapi/handler.go
mastercactapus: Yes, I think that makes the most sense, and makes it the easiest for downstream systems to use the data.

alert/alert.go
shivanishendge: I've added new method `CreateMetaDataTx` and we can pass the meta and alertId as an agrument.

alert/alert.go
shivanishendge: sounds good!

alert/meta.go
shivanishendge: removed the json typing

genericapi/handler.go
shivanishendge: okay

graphql2/graphqlapp/dataloaders.go
shivanishendge: removed the print statement

genericapi/handler.go
```diff
@@ -158,7 +158,7 @@ func (h *Handler) ServeCreateAlert(w http.ResponseWriter, r *http.Request) {
 	}
 
 	err = retry.DoTemporaryError(func(int) error {
-		createdAlert, isNew, err := h.c.AlertStore.CreateOrUpdate(ctx, a, meta)
+		createdAlert, isNew, err := h.c.AlertStore.CreateOrUpdate(ctx, a)
```
shivanishendge: I understand that we're handling it with a specific store method, But I couldn't to find any references to the store method `SetMetadataTx` for generic api call.
Is the plan is to support adding meta only via graphql?

mastercactapus: That's not a bad idea. The method to set metadata is in `alert/metadata.go` in the latest commit.

What you have here looks good, two thoughts:
- Have the `CreateOrUpdate` method just call the new *Meta version with `nil`
- Only set metadata if `isNew` is true so we can treat it as write-once (i.e., only on alert creation) at least to start out

shivanishendge: Do you mean just the update the current `CreateOrUpdate` method to have new arg *Meta? In that case, I'll have to update all the other clients(grafana, prometheus, ...) to update it to have nil values, Is that okay?

weefatboi (approved):

mastercactapus (dismissed):

mastercactapus (dismissed):

mastercactapus (approved):

mastercactapus (approved):

weefatboi (approved):

PR #3775, opened 2024-03-26
util/sqlutil: fix handling of block delimiters in SplitQuery

mastercactapus:
**Description:**
The `SplitQuery` function wrongly assumed `$$` was the only possible block delimiter, when it can be a user-defined tag like `$foobar$`

This PR fixes the bug and adds a test case for the more complex case.

More info:
https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-SYNTAX-DOLLAR-QUOTING

### DIFF
--- a/util/sqlutil/splitquery.go
+++ b/util/sqlutil/splitquery.go
@@ -32,6 +32,7 @@ func sqlSplitBlock(delim []byte, blockIdx int, data []byte, atEOF bool) (advance
 	return next + advance, data[:next+len(token)], nil
 }
 
+// https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-SYNTAX-DOLLAR-QUOTING
 var splitRx = regexp.MustCompile(`\$[^$]*\$`)
 
 func sqlSplitQuery(data []byte, atEOF bool) (advance int, token []byte, err error) {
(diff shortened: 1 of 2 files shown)


util/sqlutil/splitquery.go
```diff
@@ -31,15 +32,19 @@ func sqlSplitBlock(blockIdx int, data []byte, atEOF bool) (advance int, token []
...
 	}
 	semiIdx := bytes.IndexRune(data, ';')
-	blockIdx := bytes.Index(data, []byte("$$"))
-	if blockIdx != -1 && (semiIdx == -1 || semiIdx > blockIdx) {
+
+	idx := splitRx.FindIndex(data)
+	if idx != nil && (semiIdx == -1 || semiIdx > idx[0]) {
 		// have block start and it comes before semi (or no semi)
-		return sqlSplitBlock(blockIdx, data, atEOF)
+		return sqlSplitBlock(data[idx[0]:idx[1]], idx[0], data, atEOF)
```
ethan-haynes, line 47: we should probably have variable names that are more self-explanatory. Or comments that explain why we are doing this slice operation in place.

ethan-haynes (changes requested):

ethan-haynes (approved):

allending313 (approved):

PR #3779, opened 2024-03-28
dest: Destination display error handling

mastercactapus:
**Description:**
Fixes error handling with the new destination API. Namely, it prevents the entire page from breaking (e.g., the escalation policy details page) if any single action is invalid (e.g., archived or invalid slack channel). Instead individual chips/line items will render an error.

The purpose is to ensure in such an event, a user is able to interact with the page and take action to resolve the issue.

- gql typescript generator tool was updated to support GraphQL unions

**Screenshots:**
<details>
<summary>Escalation Policy Details</summary>

TODAY
image

This PR
image
image
</details>

<details>
<summary>User Contact Methods</summary>

TODAY
image

This PR
image
</details>

<details>
<summary>Schedule On-Call Notifications</summary>

TODAY
image

This PR
image
image
</details>

N/A fixing broken behavior

!! Breaking changes to the destination's API !!

- `FieldValuePair` no longer contains `label` and `isFavorite` these were moved to `FieldSearchResult` to separate the two
- `Destination.displayInfo` return type changed from `DestinationDisplayInfo` to a union of `DestinationDisplayInfo` and `DestinationDisplayInfoError` -- rather than failing the entire query with a global error, errors are now returned inline
- `destinationFieldSearch` return type changed to `FieldSearchConnection`, which was renamed from `FieldValueConnection` to match `FieldSearchResult`

**Additional Info:**
The affected APIs are only used when the `dest-types` experimental flag is enabled.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -245,7 +245,7 @@ test-components:  $(NODE_DEPS) bin/waitfor
 	yarn playwright install chromium
 	yarn concurrently -k -s first -n "SB,TEST" -c "magenta,blue" \
 		"yarn http-server storybook-static -a 127.0.0.1 --port 6008 --silent" \
-		"./bin/waitfor tcp://localhost:6008 && yarn test-storybook --ci --url http://127.0.0.1:6008"
+		"./bin/waitfor tcp://localhost:6008 && yarn test-storybook --ci --url http://127.0.0.1:6008 --maxWorkers 2"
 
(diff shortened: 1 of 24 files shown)


allending313 (approved):

ethan-haynes (approved):
Doesn't have to be a part of this PR, but maybe we should write some additional tests to catch this type of behavior from happening with other areas we are making changes in.

PR #3785, opened 2024-04-02
ui/sched: always use Dest API

mastercactapus:
**Description:**
Updates schedule on-call notifications always to use the new dest API.

Part of #2639 

N/A

N/A

**Additional Info:**
- Fixed issue with dest dialogs where incorrect times were recorded when sched. TZ was different than local


mastercactapus:
Waiting for #3779 to merge

allending313 (dismissed):

Forfold (commented):
I'm not seeing a quick link appear for on-call notifications, however if I go to the link manually, it exists. It looks like we hide the link if slack/webhooks is disabled. Is this intended behavior at this point with destinations being dynamic?

Screenshot 2024-05-14 at 11 54 16 AM

```
        ].concat(
          // only slack/webhook supported atm, so hide the link if disabled
          slackEnabled || webhookEnabled
            ? [
                {
                  label: 'On-Call Notifications',
                  url: 'on-call-notifications',
```

Forfold:
Looks like we hide the link if slack/webhooks is disabled. Is this intended behavior at this point with destinations being dynamic?

mastercactapus:
@Forfold It's existing behavior untouched by this work. I agree it should be changed but maybe do that in it's own PR since there may be a bit of refactoring and dialog updates to handle no valid types.

Forfold:
Gotcha. As long as we don't lose track of it..

Forfold (approved):

PR #3792, opened 2024-04-03
gql-api-keys: only setState once when duplicating key

### LINKED ISSUE #3791: API key duplicate resetting schema on poll
API key duplicate resetting schema on poll

**Steps to Reproduce:**
1. Go to API Keys
2. Click on a key
3. Click duplicate
4. Remove query
5. See query field get re-populated

mastercactapus:
**Description:**
Fixes an issue where fields on the GQL API key create dialog would reset with each poll.

The new approach leverages the Suspense behavior by setting the initial state/value directly, rather than relying on a `useEffect` handler.

Closes #3791


Forfold (approved):
Functionally validated. Ty!

allending313 (approved):

PR #3795, opened 2024-04-03
webworker: add check on worker method name

mastercactapus:
**Description:**
It adds a check that the method name sent to the web worker is one of the expected/defined ones, and it prevents any unknown values from matching, e.g., the prototype.


Forfold (approved):
:soap:

PR #3815, opened 2024-04-15
service maint: allow specifying a custom end time

### LINKED ISSUE #3681: Extend Maintenance Mode to Allow Custom End Time
The current user interface for enabling maintenance mode only provides options to set a maintenance window of 1, 2, or 4 hours. This may not be adequate for use cases where the maintenance needs to happen for a different duration or when the return-to-normal time is known in advance.

Enable a custom end time option in the maintenance mode feature in the GoAlert UI. The custom end time should adhere to a maximum limit of 24 hours as per the backend validation.

Currently, in cases where the maintenance needs to happen over a longer period, users would have to either come back and extend the maintenance window manually or potentially constrict their maintenance to shorter windows. Having a custom end time option would eliminate the need for these workarounds.

mastercactapus:
**Description:**
Updates the maintenance dialog to allow specifying a custom end time up to 24 hours in the future.

Closes #3681 

**Screenshots:**
Screenshot 2024-04-15 at 10 28 04 AM

- A 4th option in the dialog allows specifying an exact end time.

N/A

### DIFF
--- a/web/src/app/services/ServiceMaintenanceDialog.tsx
+++ b/web/src/app/services/ServiceMaintenanceDialog.tsx
@@ -45,9 +45,10 @@ function ServiceMaintenanceForm(props: {
         <FormControlLabel value={1} control={<Radio />} label={label(1)} />
         <FormControlLabel value={2} control={<Radio />} label={label(2)} />
         <FormControlLabel value={4} control={<Radio />} label={label(4)} />
-        <FormControlLabel value={-1} control={<Radio />} label='Specify' />
+        <FormControlLabel value={-1} control={<Radio />} label='Until...' />
       </RadioGroup>
       <ISODateTimePicker
+        label='Select a date'
         value={props.value.toISO()}
         disabled={selectedOption !== -1}
         onChange={(iso) => props.onChange(DateTime.fromISO(iso))}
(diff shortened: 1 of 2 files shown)


web/src/app/services/ServiceMaintenanceDialog.tsx
```diff
@@ -66,9 +83,13 @@ export default function ServiceMaintenanceModeDialog(
       subTitle={
         <React.Fragment>
```
Forfold, line 84: this should be a span/typography so it isn't in times new roman

web/src/app/services/ServiceMaintenanceDialog.tsx
```diff
@@ -25,19 +26,35 @@ function label(hours: number): JSX.Element {
...
-        onChange={(e) => props.onChange(parseInt(e.target.value, 10))}
+        value={selectedOption}
+        onChange={(e) => handleOption(parseInt(e.target.value, 10))}
       >
         <FormControlLabel value={1} control={<Radio />} label={label(1)} />
         <FormControlLabel value={2} control={<Radio />} label={label(2)} />
         <FormControlLabel value={4} control={<Radio />} label={label(4)} />
+        <FormControlLabel value={-1} control={<Radio />} label='Specify' />
       </RadioGroup>
+      <ISODateTimePicker
```
Forfold, line 51: I think some left and top padding would help here. Left padding to align with the radio buttons and top padding just to nudge it down slightly.

Forfold (approved):

KatieMSB (approved):

PR #3836, opened 2024-05-02
add type universal to integration keys

Forfold:
**Description:**
Adds type `universal` to integration keys, as well as everywhere else key types are used.

Unsure

**Out of Scope:**
Href value


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/target/goalert
 
-go 1.22.0
+go 1.22
```
Forfold, line 3: the `.0` was causing my intellisense to stop working

Forfold:
Do I need to put this behind an experimental flag @mastercactapus? Seems like there's quite a few places where it's being introduced (types, sources, schema, and hard coded in the wizard UI)

go.mod
mastercactapus: IIRC adding it was to workaround a separate bug that caused it to break

graphql2/graphqlapp/config.go
```diff
@@ -16,6 +16,7 @@ func (a *Query) IntegrationKeyTypes(ctx context.Context) ([]graphql2.Integration
 		{ID: "grafana", Name: "Grafana", Label: "Grafana Webhook URL", Enabled: true},
 		{ID: "site24x7", Name: "Site 24x7", Label: "Site24x7 Webhook URL", Enabled: true},
 		{ID: "prometheusAlertmanager", Label: "Alertmanager Webhook URL", Name: "Prometheus Alertmanager", Enabled: true},
+		{ID: "universal", Label: "Universal Integration Key", Name: "Universal Integration Key", Enabled: true},
```
mastercactapus, line 19: This one needs to be behind the flag, we don't want to allow users to create a key that doesn't work (unless the exp flag is set)

migrate/migrations/20240502114803-universal-integration-key.sql
```diff
@@ -0,0 +1,8 @@
+-- +migrate Up notransaction
+ALTER TYPE enum_integration_keys_type
+  ADD VALUE IF NOT EXISTS 'universal';
+
+ALTER TYPE enum_alert_source
+  ADD VALUE IF NOT EXISTS 'universal';
```
mastercactapus, line 6: duplicate
```suggestion
```

web/src/app/wizard/WizardForm.jsx
```diff
@@ -188,6 +188,10 @@ export default function WizardForm(props) {
                 label: 'Email',
                 value: 'email',
               },
+              {
+                label: 'Universal',
+                value: 'universal',
+              },
```
mastercactapus, line 194: I'd say just leave it out from this until/unless we update the wizard to use the API/types hook

web/src/app/admin/admin-service-metrics/AdminServiceFilter.tsx
```diff
@@ -139,6 +139,7 @@ function AdminServiceFilter(): JSX.Element {
                     'site24x7',
                     'prometheusAlertmanager',
                     'email',
+                    'universal',
```
mastercactapus, line 142: should be conditional on the flag, we can do so by updating this component to use the types hook instead of hard-coding them

web/src/app/wizard/utilTestData.js
```diff
@@ -23,6 +23,10 @@ const keys = [
     label: 'Email',
     value: 'email',
   },
+  {
+    label: 'Universal',
+    value: 'universal',
+  },
```
mastercactapus, line 29: can leave this out since it's wizard specific and probably not worth doing the experimental flag check

web/src/cypress/support/service.ts
```diff
@@ -221,6 +221,7 @@ function createIntKey(
       'grafana',
       'site24x7',
       'prometheusAlertmanager',
+      'universal',
```
mastercactapus, line 224: we can leave this out at this stage. UIK will get it's own test entirely since it's a very different flow & use case

mastercactapus (changes requested):
@Forfold
- Anything listing keys should hide "universal" unless the exp flag is set (this should be automatic in the UI `useFeatures().integrationKeyTypes`
- Creating an integration key should guard against the Universal type unless the exp flag is set

migrate/migrations/20240502114803-universal-integration-key.sql
Forfold: I was following the pattern the prometheus migration uses where it sets `enum_integration_keys_type` as well as `enum_alert_source`. Is only `enum_integration_keys_type` needed?

web/src/app/wizard/WizardForm.jsx
Forfold: yeah I think all of the UI should be using the context to grab this list, since it already exists

web/src/app/admin/admin-service-metrics/AdminServiceFilter.tsx
Forfold: https://github.com/target/goalert/pull/3837

web/src/app/wizard/utilTestData.js
Forfold: https://github.com/target/goalert/pull/3837

web/src/cypress/support/service.ts
Forfold: https://github.com/target/goalert/pull/3837

web/src/app/wizard/WizardForm.jsx
Forfold: https://github.com/target/goalert/pull/3837

migrate/migrations/20240502114803-universal-integration-key.sql
mastercactapus: ohhhhhh, totally missed that they were different. That's embarrassing. don't mind me

graphql2/graphqlapp/config.go
```diff
@@ -4,19 +4,26 @@ import (
...
 		{ID: "email", Name: "Email", Label: "Email Address", Enabled: cfg.EmailIngressEnabled()},
 		{ID: "generic", Name: "Generic API", Label: "Generic Webhook URL", Enabled: true},
 		{ID: "grafana", Name: "Grafana", Label: "Grafana Webhook URL", Enabled: true},
 		{ID: "site24x7", Name: "Site 24x7", Label: "Site24x7 Webhook URL", Enabled: true},
 		{ID: "prometheusAlertmanager", Label: "Alertmanager Webhook URL", Name: "Prometheus Alertmanager", Enabled: true},
-	}, nil
+	}
+
+	if expflag.ContextHas(ctx, expflag.UnivKeys) {
+		typeInfo = append(typeInfo, graphql2.IntegrationKeyTypeInfo{ID: "universal", Label: "Universal Integration Key", Name: "Universal Integration Key", Enabled: true})
```
mastercactapus, line 23: These field names aren't great, but "Label" is for the copy link
```suggestion
		typeInfo = append(typeInfo, graphql2.IntegrationKeyTypeInfo{ID: "universal", Label: "Universal Integration Key URL", Name: "Universal Integration Key", Enabled: true})
```

integrationkey/store.go
```diff
@@ -80,6 +81,10 @@ func (s *Store) Create(ctx context.Context, dbtx gadb.DBTX, i *IntegrationKey) (
 		return nil, err
 	}
 
+	if i.Type == TypeUniversal && expflag.ContextHas(ctx, expflag.UnivKeys) {
+		validate.OneOf("Type", i.Type, TypeUniversal)
+	}
```
mastercactapus, line 86: We want to check if the type is universal, and if the flag is _not_ set, then return an error.

```suggestion
	if i.Type == TypeUniversal && !expflag.ContextHas(ctx, expflag.UnivKeys) {
		return nil, validation.NewGenericError("experimental flag not enabled")
	}
```

mastercactapus (changes requested):

integrationkey/store.go
Forfold: Gotcha, I thought the intention was to validate it's the correct type if set as Universal

mastercactapus (approved):

allending313 (approved):

PR #3852, opened 2024-05-08
chore: pkg imported more than once

testwill:
**Description:**
Removes redundant import in the test harness


mastercactapus (changes requested):
good catch -- let's go with the non-aliased import though; the `pgx5` alias was leftover from the upgrade

mastercactapus (approved):

KatieMSB (approved):

PR #3867, opened 2024-05-13
ui/policy: always use Dest API

mastercactapus:
**Description:**
Updates escalation policy step form to always to use the new dest API.

Part of #2639 

The new destination form will be used

N/A


web/src/app/escalation-policies/PolicyStepsCard.tsx
```diff
@@ -122,17 +116,10 @@ export default function PolicyStepsCard(
...
-              escalationPolicyID={props.escalationPolicyID}
-              onClose={resetCreateStep}
-            />
-          ) : (
-            <PolicyStepCreateDialog
-              escalationPolicyID={props.escalationPolicyID}
-              onClose={resetCreateStep}
-            />
-          )}
+          <PolicyStepCreateDialogDest
```
Forfold, line 119: I think I would be fine having these take the old names now that we're removing it, for sake of brevity 🙂 Thoughts?

```suggestion
          <PolicyStepCreateDialog
```

web/src/app/escalation-policies/PolicyStepsQuery.tsx
```diff
@@ -2,26 +2,6 @@ import React from 'react'
...
-          name
-          type
-        }
-      }
-    }
-  }
-`
 
 export const policyStepsQueryDest = gql`
   query stepsQueryDest($id: ID!) {
```
Forfold, line 7: same here

web/src/app/escalation-policies/PolicyStepsCard.tsx
mastercactapus: How would you feel about doing a rename after all at once? if we do them at the same time, git will think we're making a bunch of changes instead of just deleting the old versions. in hindsight we could have renamed the old to `*Legacy` or something so it's clear they would be deleted in the future

Forfold: Yeah that works 👍

Forfold (commented):
Not seeing any console logs in your proposed changes, but looks like this got through at some point when opening the "Edit Step" dialog:

Screenshot 2024-05-14 at 12 20 41 PM

There's also padding issues with this that we'll need to address:

Screenshot 2024-05-14 at 12 22 00 PM

Forfold (approved):

Forfold (approved):

Forfold (dismissed):

Forfold (approved):

Forfold (dismissed):

Forfold (approved):

allending313 (approved):

PR #3871, opened 2024-05-14
profile: optimize on-call query

### LINKED ISSUE #3870: User Profile Access Denied when Creating Too Many Services
**Describe the Bug:**
The issue arises when a user is assigned to too many escalation policies. There's an error note "access denied: exceeded auth check limit" that prevents the user and others from accessing/editing the problematic user's profile page. This problem appears to be related to the 'assignedTo' part of the query from the 'profileInfo' query.

**Steps to Reproduce:**
1. Create many services (user will be assigned by default)
2. Go to User Profile Page
3. Attempt to view or edit user's profile
4. See error "access denied: exceeded auth check limit"

**Expected Behavior:**
Users should be able to create services up to a reasonable limit without encountering errors and access should remain unaffected for them and other users to view/edit a particular user's profile.

**Observed Behavior:**
When a user creates too many services, both that user and others become unable to access or edit the problematic user's profile receiving the error "access denied: exceeded auth check limit".

**Potential Solution:**
Look into optimization in the 'assignedTo' part of 'profileInfo' query to resolve this. Alternatively, for the profile page, it may be better to add a field that simply returns the number of assigned services (since this is the actual needed data) which could be handled far more efficiently in the backend.

**Application Version:**
This issue affects the master branch of GoAlert.

**Additional Context:**
There's no defined practical limit for creating services. As a temporary solution, removing the user from escalation policies/schedules/rotations associated with services will restore the profile.

mastercactapus:
**Description:**
Makes a single DB call for on-call assignments instead of the recursive resolvers it previously used.

Fixes #3870

- Added `onCallOverview` to the User type
- Deprecated `onCallSteps`

### DIFF
--- a/graphql2/schema.graphql
+++ b/graphql2/schema.graphql
@@ -1210,7 +1210,7 @@ type OnCallOverview {
 
 type OnCallServiceAssignment {
   """
-  stepNumber is the escalation step this assignment is from (begining with 0).
+  stepNumber is the escalation step this assignment is from (beginning with 0).
   """
   stepNumber: Int!
 
--- a/graphql2/schema.graphql
+++ b/graphql2/schema.graphql
@@ -1209,6 +1209,9 @@ type OnCallOverview {
 }
 
 type OnCallServiceAssignment {
+  """
+  stepNumber is the escalation step this assignment is from (begining with 0).
+  """
   stepNumber: Int!
 
   escalationPolicyID: ID!
--- a/graphql2/schema.graphql
+++ b/graphql2/schema.graphql
@@ -1178,6 +1178,9 @@ type User {
   sessions: [UserSession!]!
 
   onCallSteps: [EscalationPolicyStep!]!
+    @deprecated(reason: "Use `onCallOverview` instead.")
+
+  onCallOverview: OnCallOverview!
 
   isFavorite: Boolean!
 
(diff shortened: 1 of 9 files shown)


graphql2/schema.graphql
```diff
@@ -1200,6 +1203,24 @@ type UserNotificationRule {
...
 }
 
+type OnCallOverview {
+  serviceCount: Int!
+  serviceAssignments: [OnCallServiceAssignment!]!
+}
+
+type OnCallServiceAssignment {
+  """
+  stepNumber is the escalation step this assignment is from (begining with 0).
```
Forfold, line 1213:
```suggestion
  stepNumber is the escalation step this assignment is from (beginning with 0).
```

web/src/app/users/UserOnCallAssignmentList.tsx
```diff
@@ -56,22 +34,20 @@ interface Service {
...
-      if (!svcs[svc.id]) {
-        svcs[svc.id] = {
-          id: svc.id,
-          name: svc.name,
-          policyName: s.escalationPolicy.name,
-          policySteps: [s.stepNumber],
-        }
-      } else {
-        svcs[svc.id].policySteps.push(s.stepNumber)
+  ;(onCallSteps || []).forEach((s) => {
```
Forfold, line 39: Not your code, but I imagine the ` || []` is not needed since we are defaulting to `[]` already in the function parameters

```suggestion
  ;(onCallSteps).forEach((s) => {
```

Forfold (changes requested):

web/src/app/users/UserOnCallAssignmentList.tsx
mastercactapus: If the backend returns null, the function defaults won't come into play. IIRC, those are only used if you call `services()` without an argument. As written, I don't think the default value would ever get used.

Forfold: Gotcha, I was under the impression `null` would trigger defaulting to `[]` in the params if present. Thanks, Javascript.

Forfold (approved):

Forfold (approved):

PR #3892, opened 2024-05-20
ui/on-call notifications: always show quick link, show error on no days

Forfold:
This PR makes a couple small changes to the Schedule On-Call Notifications experience:

- The quick link from the Schedule page should always show now. This was previously conditionally rendered if slack or webhooks were enabled in the config
- Shows an error and disables the user from submitting if "Specify at time and day of week" is enabled, but all days are unchecked, saying that at least one day is required. Right now, nothing happens and the dialog closes.


Forfold:
Wrong base branch

PR #3894, opened 2024-05-21
uik: add token rotation dialogs and controls

mastercactapus:
**Description:**
Adds token rotation dialogs and functionality, along with integration test for universal integration keys


web/src/app/services/UniversalKey/UniversalKeyPage.tsx
```diff
@@ -1,16 +1,138 @@
...
+      keyID: props.keyID,
+      serviceID: props.serviceID,
+    },
+  })
+
+  // Redirect to the correct service if the key is not in the service
+  if (
+    q.data &&
+    q.data.integrationKey &&
+    q.data.integrationKey.serviceID !== props.serviceID
```
Forfold, line 57: we should be able to simplify this:

```
if (q?.data?.integrationKey?.serviceID !== props.serviceID) {
```

Forfold (dismissed):

Forfold:
`make test` passes locally- rerunning on CI

allending313 (approved):

web/src/app/loading/components/LoadingButton.tsx
```diff
@@ -34,7 +34,7 @@ const LoadingButton = (props: LoadingButtonProps): JSX.Element => {
         disabled={loading || disabled}
         type={noSubmit ? 'button' : 'submit'}
       >
-        {buttonText || (!attemptCount ? 'Confirm' : 'Retry')}
+        {!attemptCount ? buttonText || 'Confirm' : 'Retry'}
```
Forfold, line 37: This is not preferable moving forward as we want the button to say a specific text ALWAYS, and not have it change to "Retry" on the first click.

mastercactapus: This is handled by the `alert` prop on FormDialog -- once we have a token we set that prop to `true` which forces the button to always say `DONE`

mastercactapus: Changing this logic causes playwright tests to fail as other parts of the application count on `Retry` overriding the `buttonText` after the first attempt

Forfold (approved):

PR #3899, opened 2024-05-22
storybook: allow for per-story graphql mocking

mastercactapus:
**Description:**
Adds support for a new parameter `graphql` which allows specifying payloads for named GraphQL operations. Works per-story.

The decorator was updated so that each component/story renders with it's own unique prefix path to the `/api/graphql` endpoint, and it's handlers are registered only for that component.

Under the hood it continues to use `msw` with the following benefits:
- Requests continue to show up in the network tab and any GraphQL browser extensions
- Requests to different components can be differentiated by the unique URL
- Existing `msw` config will continue to work, making this change backwards-compatible

This change allows for stories to display/test different GraphQL response states per story without having to rely on conditional returns and interaction


web/src/app/urql.ts
```diff
@@ -104,15 +105,19 @@ const apolloRefetchExchange: Exchange = ({ forward }) => {
...
-    refetchExchange(),
-    cacheExchange,
-    apolloRefetchExchange,
-    retryExchange({}) as Exchange,
-    fetchExchange,
-  ],
-  requestPolicy: 'cache-and-network',
-})
+export function newClient(url: string): Client {
+  return createClient({
```
Forfold, line 109: is there any benefit from breaking this variable up? seems to just add lines from the previous definition

.storybook/preview.ts
```diff
@@ -1,15 +1,54 @@
...
+  parameters: {
+    graphql?: GraphQLParams
+  }
+}
+
+// graphqlLoader is a loader function that sets up the GraphQL mocks for the component. It stores them in a global object, componentConfig, which is used by the mswHandler to resolve the mocks.
+//
+// We need to do this because the browser will render all components at once, and so we need to handle all GraphQL mocks at once.
+//
+// The way this works is that each component get's a unique URL for GraphQL (see decorators.tsx). This URL is used to store the GraphQL mocks for that component in componentConfig.
```
Forfold, line 46: nit: these lines look pretty long- breaking up into more readable function comment

```suggestion
/**
 * graphqlLoader is a loader function that sets up the GraphQL mocks for the
 * component. It stores them in a global object, componentConfig, which is used
 * by the mswHandler to resolve the mocks.
 *
 * We need to do this because the browser will render all components at once,
 * and so we need to handle all GraphQL mocks at once.
 *
 * The way this works is that each component get's a unique URL for GraphQL
 * (see decorators.tsx). This URL is used to store the GraphQL mocks for that
 * component in componentConfig.
 */
```

web/src/app/urql.ts
mastercactapus: not sure I know what you mean, for the new decorator we need to be able to create a new client, but in all other cases we need a single instance.

storybook uses `newClient` and the normal goalert webapp uses the const

.storybook/preview.ts
```diff
@@ -1,15 +1,54 @@
...
-  handleExpFlags,
-} from '../web/src/app/storybook/graphql'
+import { defaultConfig } from '../web/src/app/storybook/graphql'
+import { graphql, HttpResponse } from 'msw'
 
 initialize({
   onUnhandledRequest: 'bypass',
 })
 
+export type GraphQLRequestHandler = (variables: unknown) => object | object
```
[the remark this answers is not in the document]
mastercactapus: I have no idea, fixed

web/src/app/urql.ts
Forfold: The code is exactly the same- I didn't notice the `export` being the only difference. Makes sense. 👍

Forfold (approved):

allending313 (approved):

PR #3909, opened 2024-05-28
ui: add DynamicActionField component

mastercactapus:
**Description:**
Adds the DynamicActionField component for configuring actions.

**Screenshots:**
image

N/A adding a component for future use

DestinationField has a new simpler field for errors that includes fieldID and message, the old one was marked as deprecated.

### DIFF
--- a/web/src/app/selection/DestinationField.tsx
+++ b/web/src/app/selection/DestinationField.tsx
@@ -13,7 +13,15 @@ export type DestinationFieldProps = {
 
   disabled?: boolean
 
+  /* deprecated */
   destFieldErrors?: DestFieldValueError[]
+
+  fieldErrors?: DestFieldError[]
+}
+
+export interface DestFieldError {
+  fieldID: string
+  message: string
 }
 
 function capFirstLetter(s: string): string {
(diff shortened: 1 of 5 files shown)


web/src/app/selection/DynamicActionField.tsx
```diff
@@ -0,0 +1,218 @@
...
+  return (err as StaticFieldError).section === 'static-params'
+}
+function isInput(err: BaseError): err is InputError {
+  return (err as InputError).section === 'input'
+}
+function isDynamic(err: BaseError): err is DynamicParamError {
+  return (err as DynamicParamError).section === 'dynamic-params'
+}
+
+export default function DynamicActionField(
```
Forfold, line 121: Per our Slack convo- let's name this `DynamicActionFields`, since it's rendering more than 1 field.

```suggestion
export default function DynamicActionFields(
```

web/src/app/selection/DestinationField.tsx
```diff
@@ -46,11 +65,7 @@ export default function DestinationField(
           props.onChange(newValues)
         }
```
Forfold, line 67: Same with this file- since it renders more than 1 field, let's make it plural: `DestinationFields`. This was a big source of confusion for me coming in new this quarter.

web/src/app/selection/DynamicActionField.tsx
mastercactapus: This component operates as a single value that happens to have 1 or more fields, but not all the same, so how about `DynamicActionForm`? I think that's more in-line with how we do "manage multiple inputs" elsewhere

allending313 (approved):

Forfold (approved):

allending313 (approved):

PR #3939, opened 2024-06-25
Add ErrorConsumer

mastercactapus:
**Description:**
Adds `ErrorConsumer` and hook from `uik-editor` branch


allending313 (approved):

Forfold (approved):
We'll probably want to start an issue to migrate error handing to this, similar to the urql and typescript migrations. When we will actually have the time to do it though is another story 😬

PR #3940, opened 2024-06-25
DestinationField StringMap update

mastercactapus:
**Description:**
Updates DestinationField and dependants to use the `args` field with StringMap

**Out of Scope:**
UserContactMethodForm and related components currently have a placeholder solution and will be updated in the next PR.

N/A

N/A


web/src/app/users/UserContactMethodForm.tsx
```diff
@@ -123,6 +123,19 @@ export default function UserContactMethodForm(
...
+                value: v,
+              }))
+            }}
+            mapValue={(v: FieldValueInput[]): StringMap => {
+              const map: StringMap = {}
+              v.forEach((fv) => {
+                map[fv.fieldID] = fv.value
+              })
+              return map
+            }}
```
Forfold, line 138: Potential cleanup:

```suggestion
            mapValue={(v: FieldValueInput[]): StringMap => (
              v.map((i) => ({
                [i.fieldID]: i.value
              }))
            )}
```

mastercactapus: This will return an array of objects like this `[{foo: 'bar'}, {bin: 'baz'}]` rather than `{foo: 'bar', bin: 'baz'}`

Forfold (approved):

web/src/app/users/UserContactMethodForm.tsx
Forfold: Oops! I missed that in the variable definition.

mastercactapus: np, renamed the var for clairity

allending313 (approved):

PR #3943, opened 2024-06-26
dest: update user forms to use StringMap instead of FieldValuePair[]

mastercactapus:
**Description:**
Updates the user contact method forms to use the `args` field of DestinationInput

Follow up to #3940


web/src/app/users/UserContactMethodCreateDialog.tsx
```diff
@@ -84,7 +81,13 @@ export default function UserContactMethodCreateDialog(props: {
...
     <UserContactMethodForm
       disabled={createCMStatus.fetching}
-      errors={formErrors}
+      nameError={errs.getErrorByPath('createUserContactMethod.input.name')}
+      destTypeError={errs.getErrorByPath(
+        'createUserContactMethod.input.dest.type',
+      )}
+      destFieldErrors={errs.getErrorMap(
+        /createUserContactMethod.input.dest(.args)?/,
+      )}
```
Forfold, line 90: Would it be worth just passing the errors object to the form, and the form can decide the names of the fields for error messaging?

mastercactapus: The form does that by the prop names (e.g., `nameError`). The path is specific to the GraphQL request, which the dialog owns, so it's responsibility is to determining what error from the response is for the `name` and the form controls how to display an error on the `name` field.

mastercactapus: Think of how an input field just has an `error` prop to indicate it's in error, and the form just has separate `*error` props for the individual fields. Both don't care/know where the error comes from.

mastercactapus: This was intentional to decouple our forms from the API in favor of being explicit and handling all API mapping (both input and output) in whatever component is making the API call

Forfold: Sounds good- that makes sense. I was thinking in the sense of decoupling the Form from the Dialog, since right now it's more separate just to keep the lines down in the file (can't have one without the other), which is why I thought maybe moving that down a level would be fine, to save on adding to props/interfaces.

This works though 👍 I'll resolve.

Forfold (approved):

allending313 (approved):

PR #3971, opened 2024-07-08
engine: fix for notification channel deduplication migration

mastercactapus:
**Description:**
Fixes a big issue with the notification channel deduplication migration.

- Introduces the ability to add data to the `migrations_test.go` tests.
- Updates the dedup migration so other updates will retain the dedup table
- Uses the dedup table to do a 1-time migration of the `schedule_data` table as necessary
- Adds a smoketest that reproduces the bug (try disabling the schedule manager migration code to see the fkey error)

**Out of Scope**
- For instances updated between the problematic migration and this fix (nightly builds June 29th - July 9th) another PR will be opened with a fix to bypass the error so schedules don't get stuck.


engine/schedulemanager/queries.sql
```diff
@@ -0,0 +1,35 @@
+-- name: SchedMgrGetNCIDs :many
+SELECT
+    id
+FROM
+    notification_channels;
```
Forfold, line 5: doesn't seem to be used when I search

engine/schedulemanager/scheduledata.go
```diff
@@ -0,0 +1,97 @@
...
+			}
+			data.V1.OnCallNotificationRules[i].ChannelID = newID
+			hadUpdate = true
+		}
+		if !hadUpdate {
+			db.migrateSchedIDs = db.migrateSchedIDs[1:]
+			continue
+		}
+
+		newData, err := json.Marshal(data.V1.OnCallNotificationRules) // important that we are only updating the V1.OnCallNotifications field.
```
Forfold, line 76: can you briefly add why it is important?

Forfold (changes requested):

Forfold (dismissed):

allending313 (approved):

Forfold (approved):

allending313 (approved):

PR #3975, opened 2024-07-09
engine/signal: Add signal manager module and support for sending signal messages

mastercactapus:
**Description:**
It adds the signal manager, which adds pending signals to the outgoing messages queue and supports sending signals to Slack.


engine/sendmessage.go
```diff
@@ -168,8 +171,29 @@ func (p *Engine) sendMessage(ctx context.Context, msg *message.Message) (*notifi
 			ScheduleID:   msg.ScheduleID,
 			Users:        onCallUsers,
 		}
+	case notification.MessageTypeSignalMessage:
+		id, err := uuid.Parse(msg.ID)
+		if err != nil {
+			return nil, errors.Wrap(err, "parse signal message id")
+		}
+		rawParams, err := gadb.New(p.b.db).EngineGetSignalParams(ctx, uuid.NullUUID{Valid: true, UUID: id})
```
Forfold, line 179: had to go up & expand almost 200 lines to see that `p` = `Engine`. Still not sure what the `b` is 😬

engine/signalmgr/queries.sql
```diff
@@ -0,0 +1,34 @@
...
+SET
+    message_id = $2
+WHERE
+    id = $1;
+
+-- name: SignalMgrDeleteStale :exec
+-- Delete stale pending signals.
+DELETE FROM pending_signals
+WHERE message_id IS NULL
+    AND created_at < NOW() - INTERVAL '1 hour';
```
Forfold, line 33: I feel like this may have unexpected consequences down the road, but hopefully a user is never in this situation

mastercactapus: Mitigating factors:
- The data in this table is only useful/needed until a message is sent
- If a message is over an hour old, it means there is at least an hour worth of more-important messages being sent in the meantime
- Signals represent a way to potentially have a lot more messages coming through, so it serve as a cap on the backlog as well

engine/sendmessage.go
mastercactapus: Yes, the engine code badly needs refactor and to be broken apart, the receiver methods are far too long.

IIRC the `b` is just a bunch of prepared statements, a small PR converting them to sqlc is likely in order too.

allending313 (approved):

PR #3977, opened 2024-07-10
dest: use gadb.DestV1 for Destination and DestinationInput

mastercactapus:
**Description:**
Switches from generated models to use `gadb.DestV1` for GraphQL to simplify code and future validation.

### DIFF
--- a/graphql2/graphqlapp/destinationdisplayinfo.go
+++ b/graphql2/graphqlapp/destinationdisplayinfo.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/nyaruka/phonenumbers"
 	"github.com/target/goalert/config"
+	"github.com/target/goalert/gadb"
 	"github.com/target/goalert/graphql2"
 	"github.com/target/goalert/util/errutil"
 	"github.com/target/goalert/util/log"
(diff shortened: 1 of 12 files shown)


graphql2/graphqlapp/destinationdisplayinfo.go
```diff
@@ -21,40 +22,15 @@ type (
...
-	m := make(map[string]string, len(obj.Values))
-	for _, val := range obj.Values {
-		m[val.FieldID] = val.Value
-	}
-
-	return m, nil
-}
-
-func (a *Destination) Values(ctx context.Context, obj *graphql2.Destination) ([]graphql2.FieldValuePair, error) {
+func (a *Destination) Values(ctx context.Context, obj *gadb.DestV1) ([]graphql2.FieldValuePair, error) {
```
Forfold, line 33: I don't see any error checks here, perhaps we can set the return type to only `[]graphql2.FieldValuePair`?

test/smoke/actionvalidate_test.go
```diff
@@ -26,10 +25,17 @@ func TestActionValid(t *testing.T) {
 	check := func(destType string, dest, dyn params) harness.QLResponse {
 		t.Helper()
 		var vars struct {
-			Input graphql2.ActionInput `json:"input"`
+			Input struct {
```
Forfold, line 28: should we be using `DestV1` here?

graphql2/graphqlapp/destinationdisplayinfo.go
mastercactapus: This is a GraphQL resolver (i.e., compatibility resolver), so it needs to have that specific signature (args and return values) even if it doesn't have an error condition.

test/smoke/actionvalidate_test.go
mastercactapus: No, since this is a test, it should be self-contained so that it fails if something changes in an incompatible way. Before, if ActionInput changed in a breaking way, this test wouldn't detect the problem since it re-used that type.

Forfold (approved):

PR #4013, opened 2024-07-23
dest: switch UI and alertlog from type/val to dest

mastercactapus:
**Description:**
- Switch alert log to use dest registry and sqlc
- move UI away from targets to dest where appropriate


alert/alertlog/store.go
```diff
@@ -316,46 +310,34 @@ func (s *Store) logEntry(ctx context.Context, tx *sql.Tx, _type Type, meta inter
...
 		case permission.SourceTypeNotificationCallback:
 			r.subject._type = SubjectTypeUser
-			var dt notification.ScannableDestType
-			err = txWrap(ctx, tx, s.lookupCallbackType).QueryRowContext(ctx, src.ID).Scan(&dt.CM, &dt.NC)
+			r.subject.userID = permission.UserNullUUID(ctx)
+
+			id, err := uuid.Parse(src.ID)
 			if err != nil {
-				return nil, errors.Wrap(err, "lookup notification type for callback ID")
+				return nil, errors.Wrap(err, "parse channel ID")
```
Forfold, line 329: is this always a channel (thinking as in slack channel)?
```suggestion
				return nil, errors.Wrap(err, "parse destination source ID")
```

Forfold: NVM realizing this is probably meaning notification channel

mastercactapus: yes, regardless of type it's the UUID of a row in the `notification_channels` table

Forfold (dismissed):

Forfold (approved):

allending313 (approved):

PR #4071, opened 2024-09-19
http: Add Secure Headers

mastercactapus:
**Description:**
This PR adds secure headers to responses, using a nonce value for CSP.

If necessary, they can be disabled with `--disable-secure-headers`


mastercactapus:
note: this needs to be tested with a caching reverse-proxy, it seems the index page headers may need to be adjusted

KatieMSB (approved):

PR #4116, opened 2024-11-01
engine/schedulemanager: convert to sqlc

mastercactapus:
**Description:**
Transitions the schedule manager to sqlc
- most IDs are now handled natively as UUIDs
- Queries are unchanged


m17ch (approved):
LGTM 👍

PR #4124, opened 2024-11-08
logging: introduce slog.Logger

mastercactapus:
**Description:**
Introduced `*slog.Logger` to the `App` struct as an eventual replacement for the custom logging package.

This can also be used as a logger to pass into other tools that take an *slog.Logger argument.


KatieMSB (dismissed):

KatieMSB (approved):

m17ch (approved):
LGTM 👍

PR #4164, opened 2024-11-26
engine: migrate status update manager to job queue

mastercactapus:
**Description:**
Migrates status updates to use the job queuing system.

Previously, the status manager would run maintenance operations every tick, followed by each subscription in a tight loop (up to 100 per tick) keeping a slice of IDs to omit (which was passed to the DB for each call).

The new approach:
- hourly periodic job to trim old subscriptions
- every 5 seconds a look-for-work periodic job that schedules jobs for individual subscriptions
- each subscription runs in it's own job in the queue, up to 5 concurrently (management tasks take priority)

**Out of Scope:**
- Adding a trigger from the CreateOrUpdate method of alerts to immediately schedule things like status updates
- With the above ^ we could update the "look for work" job to run once per minute instead

N/A

N/A


engine/statusmgr/cleanup.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"github.com/riverqueue/river"
+	"github.com/target/goalert/gadb"
+)
+
+type CleanupArgs struct{}
+
+func (CleanupArgs) Kind() string { return "status-manager-cleanup" }
+
+func (db *DB) cleanup(ctx context.Context, j *river.Job[CleanupArgs]) error {
```
AlaricWhitney, line 15: add functional comment to explain the purpose of the function

app/initriver.go
```diff
@@ -42,6 +42,40 @@ type noopWorker struct{}
...
 
+type ignoreCancel struct {
+	h slog.Handler
+}
+
+func (i *ignoreCancel) Enabled(ctx context.Context, level slog.Level) bool {
+	return i.h.Enabled(ctx, level)
+}
+
+func (i *ignoreCancel) Handle(ctx context.Context, rec slog.Record) error {
```
AlaricWhitney, line 53: add functional comment to explain the purpose of the function

engine/statusmgr/lookforwork.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/jackc/pgx/v5"
+	"github.com/riverqueue/river"
+	"github.com/target/goalert/gadb"
+)
+
+type LookForWorkArgs struct{}
+
+func (LookForWorkArgs) Kind() string { return "status-manager-look-for-work" }
+
+func (db *DB) lookForWork(ctx context.Context, j *river.Job[LookForWorkArgs]) error {
```
AlaricWhitney, line 18: add functional comment to explain the purpose of the function

engine/statusmgr/setup.go
```diff
@@ -0,0 +1,36 @@
...
+const (
+	QueueName           = "status-manager"
+	PriorityLookForWork = 2
+	PriorityCleanup     = 3
+	PriorityProcess     = 4
+)
+
+var _ processinglock.Setupable = &DB{}
+
+func (db *DB) Setup(ctx context.Context, args processinglock.SetupArgs) error {
```
AlaricWhitney, line 20: add functional comment to explain the purpose of the function

AlaricWhitney (changes requested):

AlaricWhitney (approved):

AlaricWhitney (approved):

KatieMSB (dismissed):

AlaricWhitney (approved):

ethan-haynes (approved):

PR #4173, opened 2024-12-03
engine: add support for triggering all periodic jobs during tests

mastercactapus:
**Description:**
This PR adds a wrapper for defining periodic jobs so that they can be triggered during tests. Many tests are currently dependant on the "1-cycle = work complete" and aren't written to wait for real-world time to elapse. This hook allows triggering jobs immediately, but still using the same mechanisms, arguments, and code to keep tests as close to real-world as is reasonable.

To avoid side effects and scheduling issues with the job runner, the cypress "fast-forward" code has been modified to run the backend in an `--api-only` mode while fast-forwarding through time. This allows using the API to insert data for the test requirements, but keeping engine processing and job scheduling isolated from a moving clock.


test/smoke/harness/backend.go
```diff
@@ -45,7 +52,7 @@ func (h *Harness) watchBackendLogs(r io.Reader) {
 				// ignore printed SQL errors
 				continue
 			}
-			h.t.Errorf("Backend: %s(%s) %s", strings.ToUpper(entry.Level), entry.Source, entry.Error)
+			h.t.Errorf("Backend: %s(%s) %s: %s\n%s", strings.ToUpper(entry.Level), entry.Source, entry.Error, entry.Message, string(raw))
```
AlaricWhitney, line 55: would a newline cause visualization issues?  I know with TAP the error message should be on a single line, or each line gets written to a new log entry, leading to confusion.

AlaricWhitney (commented):
just one question

test/smoke/harness/backend.go
mastercactapus: This is only used by the harness when reporting in tests (`testing.T.Errorf`); never comes from the application itself

If a test fails, you'll see the short error message inline, followed by the JSON blob on it's own that may have more details on what went wrong

AlaricWhitney (approved):

KatieMSB (approved):

PR #4186, opened 2024-12-09
engine/cleanupmanager: move alert cleanup to job queue and sqlc

mastercactapus:
**Description:**
Adds the boilerplate job `Setup` method to cleanup manager and delegates alert auto-closing and alert deletion routines to the job queue.

**Out of Scope:**
- Other cleanup operations, including alert logs, are omitted.

**Additional Info:**
This was broken out, rather than all at once, to keep the overall change reviewable.

Note: Middleware was added to make the config struct available in worker functions

Addl: Cleanup tests were altered slightly now that the logic is no longer tied to engine cycles. We now restart GoAlert with the new config (ensuring RunAtStart jobs run) and assert that cleanup operations _eventually_ run (albeit within 15 seconds).


engine/cleanupmanager/setup.go
```diff
@@ -0,0 +1,37 @@
...
+
+	"github.com/riverqueue/river"
+	"github.com/target/goalert/engine/processinglock"
+)
+
+var _ processinglock.Setupable = &DB{}
+
+const QueueName = "cleanup-manager"
+
+func (db *DB) Setup(ctx context.Context, args processinglock.SetupArgs) error {
```
AlaricWhitney, line 16: add function comment to describe its purpose

AlaricWhitney (changes requested):
LGTM.  Just need to address one functional comment

AlaricWhitney (approved):

util/contextsleep.go
```diff
@@ -0,0 +1,16 @@
+package util
+
+import (
+	"context"
+	"time"
+)
+
+// ContextSleep will sleep for the specified duration or until the context is canceled.
+func ContextSleep(ctx context.Context, d time.Duration) error {
```
CameronJHall, line 9: **question:** this is a nice way to cleanly handle both of: checking for context cancel and delaying the next loop execution. I do have slight concerns that the intent of the function is not totally clear. Are there other contexts that we would use this other than a retry or wait loop?

CameronJHall (changes requested):
I have a few questions. I don't think there are necessarily any changes that need to be made, just want to make sure that I'm not missing anything.

test/smoke/harness/harness.go
```diff
@@ -343,6 +346,38 @@ func (h *Harness) Start() {
 	}
 }
 
+// RestartGoAlertWithConfig will restart the backend with the provided config.
+func (h *Harness) RestartGoAlertWithConfig(cfg config.Config) {
```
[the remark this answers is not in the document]
mastercactapus: No, each instance started with a harness is fully isolated, including its own unique Postgres database.

engine/cleanupmanager/alerts.go
```diff
@@ -0,0 +1,63 @@
...
+			}
+
+			return len(ids) < 100, nil
+		})
+		if err != nil {
+			return fmt.Errorf("auto close alerts: %w", err)
+		}
+	}
+
+	if cfg.Maintenance.AlertCleanupDays > 0 {
```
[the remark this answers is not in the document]
mastercactapus: They're done sequentially because we want the cleanup/delete operation to benefit from alerts that were potentially closed by auto-close (since it will ONLY delete closed alerts). However, these queries are done in "chunks" of 100 until complete, so I laid it out so that we auto-close _all_ alerts before performing the cleanup.

That said, if there were an error, River (the job queue library) will retry up to 25 times with backoff, alternatively the cleanup operation will run again in an hour anyhow.

None of this is super critical, and one of the benefits of the job queue is we can allow jobs to take as long as they need; so we don't have need for them to happen concurrently either. If either were to fail, there would be no impact to critical operations. Especially since both of these settings are 0/disabled by default.

util/contextsleep.go
mastercactapus: It's basically a context-aware version of `time.Sleep`. When a request is cancelled, connection failed, or app is shutting down; time.Sleep will continue to wait. This just allows for an early exit as well as a way to break the loop if the context is cancelled.

There are actually a number of loops in GoAlert that implement this, or something very similar. But the only one I can think of that isn't a retry/wait loop is in Basic auth where we implement a random delay for failed password attempts that we'd want to immediately abort when the request is cancelled.

CameronJHall: Sounds great, hopefully we can start reusing this in some of those places too. It's nice that this can just use `time.After` now too, cleaning up timers was a pain.

PR #4192, opened 2024-12-12
engine/cleanupmgr: Migrate shifts to job queue

mastercactapus:
**Description:**
Migrates shift cleanup to use the job queue


engine/cleanupmanager/shifts.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/target/goalert/config"
+	"github.com/target/goalert/gadb"
+)
+
+type ShiftArgs struct{}
+
+func (ShiftArgs) Kind() string { return "cleanup-manager-alerts" }
+
+// CleanupShifts will automatically cleanup old shift and override records.
+func (db *DB) CleanupShifts(ctx context.Context, j *river.Job[ShiftArgs]) error {
```
ethan-haynes, line 18:
```suggestion
// CleanupShifts will automatically cleanup old shift and override records.
func (db *DB) CleanupShifts(ctx context.Context, j *river.Job[ShiftArgs]) error {
    cfg := config.FromContext(ctx)
    if cfg.Maintenance.ScheduleCleanupDays <= 0 {
        return nil
    }

    type cleanupFunc func(context.Context, *sql.Tx, int64) (int, error)

    runCleanup := func(fn cleanupFunc) error {
        return db.whileWork(ctx, func(ctx context.Context, tx *sql.Tx) (done bool, err error) {
            count, err := fn(ctx, tx, int64(cfg.Maintenance.ScheduleCleanupDays))
            if err != nil {
                return false, err
            }
            return count < 100, nil
        })
    }

    cleanupTasks := []cleanupFunc{
        gadb.New(tx).CleanupMgrDeleteOldScheduleShifts,
        gadb.New(tx).CleanupMgrDeleteOldOverrides,
        gadb.New(tx).CleanupMgrDeleteOldStepShifts,
    }

    for _, task := range cleanupTasks {
        if err := runCleanup(task); err != nil {
            return fmt.Errorf("cleanup task failed: %w", err)
        }
    }

    return nil
}
```

ethan-haynes: Not the end of the world, but there is a lot of duplicate code here. Maybe we create cleanup tasks and loop through them all at the end? Or we could separate this out into a helper function.

ethan-haynes: or maybe something like this:

```suggestion
type cleanupFunc func(context.Context, *sql.Tx, int64) (int, error)

// CleanupShifts will automatically cleanup old shift and override records.
func (db *DB) CleanupShifts(ctx context.Context, j *river.Job[ShiftArgs]) error {
    cfg := config.FromContext(ctx)
    scheduleCleanupDays := cfg.Maintenance.ScheduleCleanupDays
    if scheduleCleanupDays <= 0 {
        return nil
    }

    cleanupTasks := []struct {
        fn       cleanupFunc
        taskName string
    }{
        {gadb.New(ctx).CleanupMgrDeleteOldScheduleShifts, "delete old shifts"},
        {gadb.New(ctx).CleanupMgrDeleteOldOverrides, "delete old overrides"},
        {gadb.New(ctx).CleanupMgrDeleteOldStepShifts, "delete old step shifts"},
    }

    for _, task := range cleanupTasks {
        if err := runCleanupTask(ctx, db, scheduleCleanupDays, task.fn, task.taskName); err != nil {
            return err
        }
    }

    return nil
}

func runCleanupTask(ctx context.Context, db *DB, days int, cleanupFunc func(context.Context, *sql.Tx, int64) (int, error), taskName string) error {
    return db.whileWork(ctx, func(ctx context.Context, tx *sql.Tx) (done bool, err error) {
        count, err := cleanupFunc(ctx, tx, int64(days))
        if err != nil {
            return false, fmt.Errorf("%s: %w", taskName, err)
        }
        return count < 100, nil
    })
}
```

mastercactapus: I think this is a case where a "a little copying is better than a little dependency" or "clear is better than clever"

I considered compressing the code almost exactly like you have it, but decided it wasn't worth the indirection to save the lines and doesn't make it more readable/scannable. Not all the cleanup operations will have the same signature or have a count of 100 forever (this PR just uses it as-is for the most part) so as soon as something doesn't "fit" would require a refactor.

Happy to discuss further, as I went back and forth a lot while writing this.

ethan-haynes (dismissed):

KatieMSB (approved):

KatieMSB (approved):

PR #4203, opened 2024-12-17
engine/cleanupmgr: move schedule data cleanup to job queue

mastercactapus:
**Description:**
This pull request moves the schedule data cleanup operations to the job queue. It is in charge of:
- Removing shifts for invalid/deleted users
- Removing temporary schedules that occur in the past

**Out of Scope:**
- API key cleanup will be moved in a separate PR
- Alert log cleanup will be moved in a separate PR

**Additional Info:**
- Cleanup operation will run once per day, and each schedule is processed once per month
- A catch was put in so that if a schedule has invalid JSON it will be logged and skipped (rather than perpetual failing)
- Added missing periodic job for schedule shift cleanup


engine/cleanupmanager/queries.sql
```diff
@@ -77,3 +77,48 @@ WHERE id = ANY (
...
+-- CleanupMgrScheduleData will find the next schedule data that needs to be cleaned up.
+SELECT
+    schedule_id,
+    data
+FROM
+    schedule_data
+WHERE
+    data NOTNULL
+    AND (last_cleanup_at ISNULL
+        OR last_cleanup_at <= now() - '1 month'::interval)
```
CameronJHall, line 90: **question:** should this use the same time range input as the other queries in the file here? It seems odd that this is hard coded and doesn't use the configuration value for `ScheduleCleanupDays`. I don't know how the mechanisms might differ and why that might require this to be set to 1 month.

ref: https://github.com/target/goalert/blob/b79c639430a69edd22533420ee8c0f2abfe99a33/engine/cleanupmanager/queries.sql#L1-L16

CameronJHall (changes requested):
One question, the changes look like they accomplish what they're meant to though

engine/cleanupmanager/queries.sql
mastercactapus: This isn't a cutoff time for cleanup; that's handled in code. It is a "time since last cleanup." The difference is that this operation is more expensive (we pull JSON data for each individual schedule and then validate that individual users still exist, etc.) and so we don't want to do it all the time.

So the difference is this ensures that each schedule is only **processed** once per month. ScheduleCleanupDays has been used for how long to keep historical schedule shift information, which is handled separately.

This bit, for temporary schedules, actually deletes the instant the EndTime is in the past, since at that point we use historical data to fill in gaps in the calendar.

mastercactapus: Put another way, `cleanup_days` is for things that have a "how long do we keep this data (alerts, shift history) around for" and this query is "how often should we validate schedule config data"

CameronJHall: Ah, okay I understand the distinction. I misinterpreted the last cleanup time as something like last update time. I think that re-using the variable is fine, I'd just recommend that we add additional context to the function doc comment.

Something that highlights the flow - the last_updated row is used to control how often we remove outdated schedules and do the more expensive validation all of the users, and finally for any temp schedules that were processed and weren't deleted that the last cleanup time is set to now and will be evaluated again once enough time has passed. Probably without the run-on sentence 😄

CameronJHall (approved):
The name refactoring helps distinguish the different inputs, thanks for making those changes

KatieMSB (approved):

PR #4211, opened 2024-12-18
engine/cleanupmgr: move alert log operations to job queue

mastercactapus:
**Description:**
The alert log cleanup manually scans the alert_logs table for entries belonging to alerts that no longer exist. Because of the volume, there is no `ON DELETE CASCADE` trigger (it would make it impossible to delete some services, for example); this is the garbage collection operation to compensate.

This PR moves the alert log cleanup operation to the job queue system.
- The look-for-work component breaks the alert logs table into blocks of 100k, and schedules a job for each block
- Each of these jobs will scan/cleanup 5k rows at a time until complete
- This operation will run once per week, or if restarting GoAlert, up to once per day.
- The priority is set to the lowest of all cleanup operations.


cgray (approved):

app/initriver.go
```diff
@@ -100,6 +89,7 @@ func (app *App) initRiver(ctx context.Context) error {
 		Queues: map[string]river.QueueConfig{
 			river.QueueDefault: {MaxWorkers: 100},
 		},
+		RescueStuckJobsAfter: 5 * time.Minute,
```
CameronJHall, line 92: **question:** each of the batches is executed as a separate job so nothing here should be hitting that 5m timeout. is there anything else that could potentially end up hitting this across any of the other exiting cleanup? or does this end up being a kind of guideline for how large the amount of work per job should be?

CameronJHall (dismissed):
I had one question but it's mostly a curiosity and not a blocking question

app/initriver.go
mastercactapus: The jobs themselves will actually timeout after 60 seconds (context will be canceled) but it's expected that the worker do the cleanup/retry logic. This is the fail-safe if that fails.

If a process is terminated or crashes while a job is executing, this would be how long before it would "recover" and retry.

mastercactapus: For some extra context/reference, this is global; so for all jobs (by default):
- Timeout after 60 seconds (the library default), at which point they will go through retry (up to 25 times, increasing backoff); this is handled by context.Context in the worker itself
- _If_ there is a DB/network/execution failure where it gets "stuck" (and likely no longer processing) then after 5min it will get rescheduled even though it's reported as active. Another worker can pick it up.
- In a happy-path scenario where a job is taking too long, it would have 4 minutes to react to the context cancelation (which should result in tx being rolled back immediately, etc...). When things are working properly, the retry could happen right away, this timeout is just for "how long do we wait before we consider it 'stuck'"

The default is 1 hour, which is far too long for us, even if things are going wrong. Cleanup itself isn't super critical, but imagine we had a crash while sending messages and so according to the DB that job is still executing, even though it's not; this setting means we'd retry after 5mins maximum in the worst case.

CameronJHall: Yeah, definitely makes sense. Waiting an hour to catch those weird cases is a waste of stuck connections.

CameronJHall (approved):

cgray (approved):

Issue #4224, opened 2024-12-22
Webhook Signatures and Validation


EvanMerlock:
**What problem would you like to solve? Please describe:**
It should be possible to validate that a given webhook payload came from a specific GoAlert instance and that it wasn't tampered with over the wire.

**Describe the solution you'd like:**
By signing a webhook payload and sending that signature as an HTTP header, we can validate that (so long as the transport protocol is secure, e.g. HTTPS) a webhook payload came from a specific GoAlert instance. This would also ensure tampering doesn't happen.

**Describe alternatives you've considered:**
Including a bearer token, basic auth, or other authN/Z protocol as part of the webhook delivery mechanism would give stronger guarantees than "this payload came from this GoAlert instance" (e.g. "this payload came from this GoAlert delivery method") but would involve potentially storing credentials for each webhook delivery method.

Similarly, having signing keys for each webhook delivery method (per user, etc) would cause the same "credential bloat" but would avoid the risk of end-user credential rotation (like basic AuthN/Z or another form of end-user provided credential solution). I think it is a safe assumption to make that an end-user should "trust" all webhooks originating from a given GoAlert instance. Whether or not the application receiving the webhook should act upon a received webhook once it has been validated is likely not a concern of GoAlert which sends the webhook.

**Additional context:**
Sketching out a solution, it is likely that the simplest approach is to use a single `Keyring` per instance that can be manually rotated via the API and sign the webhook as it gets delivered. The public key could be accessed via the UI or API (proof that it "came from this specific GoAlert instance") and used by the receiving party to validate. I'm going to try writing this up.

EvanMerlock:
> For rotation, since the secret is never shared -- and rotating would break everything; is it necessary?

I tend to agree it is unnecessary to be able to individually rotate the single key-ring...

> Instead, it may make sense to have a top-level GoAlert option that is "reset ALL keyrings" in case of database + encryption key compromise but where you want to keep your config.

Agreed. Not sure it's a prerequisite of this functionality, it'd be important either way

mastercactapus:
I did some digging around existing examples, this is the most active/comprehensive RFC I could find for existing work solving this problem in the wild:
https://datatracker.ietf.org/doc/rfc9421/

Looks like there's a tool out there to experiment with a list of libraries:
https://httpsig.org/

TL;DR; uses two headers:
- `Signature-Input` which specifies things like which headers to use in the calculation (incl. special ones for `@method` and `@target-uri`)
[...]

PR #4225, opened 2024-12-22
notification/webhook: all webhooks are signed by a new set of signing keys

EvanMerlock:

**Description:**
This PR:
- initializes a new set of signing keys for the webhook delivery method
- ensures that every outgoing webhook is signed
- ensures that the signature is included in `X-Webhook-Signature`, converting the raw binary signature to base64

Begins to fix https://github.com/target/goalert/issues/4224

**Out of Scope:**

- API changes to expose the public signing key
- API changes to allow for the rotation of the signing key
- UI changes to allow for getting the signing key of the instance

**Screenshots:**

very basic webhook signatures

Users will notice a new `X-Webhook-Signature` header on any outgoing webhook.

They will _not_ be able to validate the webhook signature with just this PR as the public key will not have been exposed.

N/a

**Additional Info:**
Additional PRs for API/UI changes to come with confirmation this is the approach we want to take for resolving #4224

### DIFF
--- a/app/initstores.go
+++ b/app/initstores.go
@@ -126,6 +126,18 @@ func (app *App) initStores(ctx context.Context) error {
 		return errors.Wrap(err, "init API keyring")
 	}
 
+	if app.WebhookKeyring == nil {
+		app.WebhookKeyring, err = keyring.NewDB(ctx, app.cfg.LegacyLogger, app.db, &keyring.Config{
+			Name:       "webhook-keys",
+			MaxOldKeys: 100,
+			Keys:       app.cfg.EncryptionKeys,
+		})
+
+		if err != nil {
+			return errors.Wrap(err, "init webhook keyring")
+		}
+	}
+
 	if app.AuthLinkStore == nil {
 		app.AuthLinkStore, err = authlink.NewStore(ctx, app.db, app.AuthLinkKeyring)
 	}
--- a/app/app.go
+++ b/app/app.go
@@ -129,6 +129,7 @@ type App struct {
 	SessionKeyring  keyring.Keyring
 	APIKeyring      keyring.Keyring
 	AuthLinkKeyring keyring.Keyring
+	WebhookKeyring  keyring.Keyring
 
 	NonceStore    *nonce.Store
 	LabelStore    *label.Store
--- a/app/startup.go
+++ b/app/startup.go
@@ -94,7 +94,7 @@ func (app *App) startup(ctx context.Context) error {
 	app.DestRegistry.RegisterProvider(ctx, app.slackChan)
 	app.DestRegistry.RegisterProvider(ctx, app.slackChan.DMSender())
 	app.DestRegistry.RegisterProvider(ctx, app.slackChan.UserGroupSender())
-	app.DestRegistry.RegisterProvider(ctx, webhook.NewSender(ctx))
+	app.DestRegistry.RegisterProvider(ctx, webhook.NewSender(ctx, app.WebhookKeyring))
 	if app.cfg.StubNotifiers {
 		app.DestRegistry.StubNotifiers()
 	}
--- a/notification/webhook/sender.go
+++ b/notification/webhook/sender.go
@@ -3,6 +3,7 @@ package webhook
 import (
 	"bytes"
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"github.com/target/goalert/keyring"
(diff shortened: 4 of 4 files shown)


test/smoke/webhook_signing_test.go
```diff
@@ -0,0 +1,85 @@
...
+	h := harness.NewHarness(t, sql, "webhook-user-contact-method-type")
+	defer h.Close()
+
+	alert := <-ch
+
+	// convert alert.Signature from base64 to byte slice
+	signatureBytes, err := base64.StdEncoding.DecodeString(alert.Signature)
+	if !assert.NoError(t, err) {
+		return
+	}
```
mastercactapus, line 78:
```suggestion
	require.NoError(t, err)
```

test/smoke/webhook_signing_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	valid, _ := h.App().WebhookKeyring.Verify(alert.Body, signatureBytes)
+
+	if !assert.True(t, valid, "webhook signature invalid") {
+		return
+	}
```
mastercactapus, line 84: Adding the GraphQL/API bit (but not the UI changes) to grab the public key might be worthwhile so that the smoke test can demonstrate getting the key and verifying. 

It's also essential we _do NOT_ use any in-app code to verify the signature -- if the in-app keyring ever changes, the test won't detect the compatibility break; we need to insulate the two. This test should use it's own validation code, which should match whatever we'd document users use.

This will also help us understand how we expect verification to work externally.

Our existing keyring works well for internal GoAlert stuff. Still, we also need to do a little bit of digging into libraries of other languages to ensure validation isn't going to be painful if we use that too.

app/initstores.go
```diff
@@ -126,6 +126,18 @@ func (app *App) initStores(ctx context.Context) error {
 		return errors.Wrap(err, "init API keyring")
 	}
 
+	if app.WebhookKeyring == nil {
+		app.WebhookKeyring, err = keyring.NewDB(ctx, app.cfg.LegacyLogger, app.db, &keyring.Config{
+			Name:       "webhook-keys",
```
mastercactapus, line 131: nit: just to clairify
```suggestion
			Name:       "webhook-signature-keys",
```

mastercactapus (changes requested):
Can we get some info (PR description is fine):
- how the signature is calculated
- what is covered by the signature
- what is the signing algorithm
- how would to verify it manually

I know this is currently using our in-app code; but since this is going to be an external API, we'll need to talk through that before introducing it.

If we change anything about in-app signatures, this will need to remain as-is; so let's flush it out a bit.

Also, as mentioned in a comment, I think we need to include the GraphQL API for at least the public key to start so we can validate before enabling this.

Alternatively, we could put it behind an experimental flag until it's 100%.

test/smoke/webhook_signing_test.go
EvanMerlock: I think the best method of ensuring external use-cases would be to allow the keyring to execute [`SignASN1`](https://pkg.go.dev/crypto/ecdsa@go1.23.4#SignASN1) and MAYBE [`VerifyASN1`](https://pkg.go.dev/crypto/ecdsa@go1.23.4#VerifyASN1) operations rather than sending `r`/`s` along side the signature itself; most libraries support ASN1 verification:

- [Java](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/security/Signature.html) via [BouncyCastle](https://downloads.bouncycastle.org/java/docs/bcprov-jdk18on-javadoc/org/bouncycastle/jcajce/provider/asymmetric/ec/SignatureSpi.ecDSA256.html) (unclear to me if it supports ECDSA with SHA512/224 digest as we currently use without some extra work)
- Python via [Cryptography](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/ec/) (allows for precomputed hashes)
- Golang via crypto/ecdsa (requires digest input)
- Rust through RustCrypto or OpenSSL bindings (requires digest input)
- Erlang/Elixir through `:crypto` (can accept digest input)

EvanMerlock:
I think if we want to flesh this out a bit we should move to ECDSA P256 with SHA512 hashing rather than 512/224; it looks to me like BouncyCastle (Java, C#) supports 512 more than 512/224 (and I'd imagine 512 alone is more common than 512/224 for support in other languages that can take digests).

Another strong consideration would be moving to Ed25519 for this; it would remove a large amount of complexity here (ASN1 vs. manual r/s, selection of hashing algorithm, etc) but would add additional complexity to the codebase (we would likely need to implement another `keyring.Keyring` for Ed25519, although the original interface defn would work)

mastercactapus:
I think using a more established combo is a good idea. We use the 224 mostly just for size since we sign and verify our own tokens internally.

Since implementation may change a bit based on this convo, let's continue in the issue so it's easier to find/reference later: https://github.com/target/goalert/issues/4224#issuecomment-2563099378

PR #4269, opened 2025-02-10
engine: refactor rotations to process through job queue

### LINKED ISSUE #4244: Migrate Rotation Manager to Use the New Job Queue System
The rotation manager currently processes **all** rotations in a single transaction every 5 seconds, which can block or delay updates for **every** rotation if just one rotation encounters an issue. There is a batch size limit, but it still **fetches all rotations** each time. Additionally, tuning the interval isn’t practical since all legacy engine modules share the same loop. As GoAlert grows, this “all-or-nothing” approach becomes more inefficient and harder to manage.

**Describe the solution you’d like:**
- **Job Queue Integration**: Switch rotation updates to the fine-grained River Queue system.
  - **On-Demand Updates**: When a rotation changes (e.g., adding participants), enqueue a job specifically for that rotation.
  - **Future Scheduling**: Each job schedules the rotation’s **next** update based on its interval.
  - **Missed Updates Fallback**: Include a mechanism to detect and recover from missed updates or failures (e.g., legacy versions, DB restores, crashes), ensuring rotations never become stuck.
- **Isolated Transactions**: Each rotation update runs in its own transaction, so one failure does not block all updates.
- **Scalable & Resilient**: Remove the need to process every rotation every 5 seconds, reducing both transaction size and performance bottlenecks.

**Describe alternatives you’ve considered:**
1. **Continue Interval-Based Updates**: Tuning intervals or splitting batches still relies on one shared loop and scanning all rotations each time.
2. **Hybrid Batching**: Partially batching updates reduces transaction size but continues fetching all rotations, leaving potential for blocking on failures.
3. **Event-Driven + Periodic Sweeps**: The job queue already handles event-driven updates; additional sweeps would add complexity without the benefits of the proven queue approach.

**Additional context:**
- **Rotation Manager**: Governs on-call handoff by updating `rotation_state` to the next user at set intervals.
- **Previously Migrated Modules**: Cleanup manager, signal manager, and status update manager are already using the job queue successfully.
- **Concurrency & Multiple Engine Instances**: 
  - The old interval-based approach runs on every active engine instance, causing duplicate work and potential contention (e.g., two instances each fetch all rotations every 5 seconds).
  - The new job queue ensures that **only one instance** handles a job at a time, vastly improving resilience and performance. Multiple engine instances will be more practical once **all** modules are on the queue system.
- **Value**: By isolating each rotation update, we minimize the risk of widespread failures and enable large-scale rotation schedules to run more smoothly.

mastercactapus:
**Description:**
Migrates rotation processing to the job queue. Uses the existing advancement calculation function without any changes to logic.

Notes:
- Adds an `entity_updates` table to track changes to rotations via DB trigger as a fallback
- Adds an event for the `updateRotation` mutation to schedule a job immediately
- Adds a once-per-minute periodic job to look for changed rotations
- Adds an update-rotation job that schedules itself for the next rotation advance
- Deletes the old 5-sec-polling cycle updates
- WaitAndAssertOnCallUsers test helper was updated to wait for eventual correctness -- since the job queue needs time to run, previously it only checked once after a single engine cycle.
- `calcAdvance` was updated to abort for unknown rotation state versions

Closes #4244 

**Additional Info:**
The previous approach updated _all_ rotations every 5 seconds; this one will update rotations immediately on change and when the individual rotation is meant to advance.


engine/rotationmanager/lookforwork.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/riverqueue/river"
+	"github.com/target/goalert/gadb"
+)
+
+type LookForWorkArgs struct{}
+
+func (LookForWorkArgs) Kind() string { return "rotation-manager-lfw" }
+
+// cleanup is a worker function that will remove any stale subscriptions.
+func (db *DB) lookForWork(ctx context.Context, j *river.Job[LookForWorkArgs]) error {
```
AlaricWhitney, line 19: update functional comment

engine/rotationmanager/setup.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+const (
+	QueueName         = "rotation-manager"
+	PriorityScheduled = 1
+	PriorityEvent     = 2
+	PriorityLFW       = 4
+)
+
+var _ processinglock.Setupable = &DB{}
```
AlaricWhitney, line 21: is this empty setup correct?

engine/rotationmanager/updaterotation.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+type UpdateArgs struct {
+	RotationID uuid.UUID
+}
+
+func (UpdateArgs) Kind() string { return "rotation-manager-update" }
+
+// cleanup is a worker function that will remove any stale subscriptions.
+func (db *DB) updateRotation(ctx context.Context, j *river.Job[UpdateArgs]) error {
```
AlaricWhitney, line 24: update functional comment

AlaricWhitney (changes requested):

engine/rotationmanager/setup.go
mastercactapus: Yes, but I'll add a comment and switch to the nil version.

https://go.dev/doc/effective_go#blank_implements

AlaricWhitney (dismissed):

KatieMSB (approved):

migrate/migrations/20250311113136-track-rotation-updates.sql
```diff
@@ -0,0 +1,67 @@
+-- +migrate Up
+CREATE TABLE entity_updates(
+    id bigserial PRIMARY KEY,
+    entity_type text NOT NULL,
+    entity_id uuid NOT NULL,
+    created_at timestamptz NOT NULL DEFAULT now()
+)
+WITH (
+    autovacuum_vacuum_threshold = 50, -- Lower threshold to trigger vacuum sooner
```
Forfold, line 9: question: the only other reference to this also says it's using a lower threshold, but the value is set to `5`. what are we lowering these from?

engine/rotationmanager/lookforwork.go
```diff
@@ -0,0 +1,63 @@
...
+
+// lookForWork will schedule jobs for rotations in the entity_updates table.
+func (db *DB) lookForWork(ctx context.Context, j *river.Job[LookForWorkArgs]) error {
+	var hadWork bool
+	err := db.lock.WithTxShared(ctx, func(ctx context.Context, tx *sql.Tx) error {
+		g := gadb.New(tx)
+
+		rotations, err := g.RotMgrFindWork(ctx)
+		if errors.Is(err, sql.ErrNoRows) {
+			// done, no more work
```
Forfold, line 26:
```suggestion
			// job's done
```

image

engine/rotationmanager/queries.sql
```diff
@@ -0,0 +1,79 @@
...
+    state.shift_start AS state_shift_start,
+    ARRAY (
+        SELECT
+            p.id
+        FROM
+            rotation_participants p
+        WHERE
+            p.rotation_id = rot.id
+        ORDER BY
+            position)::uuid[] AS participants
```
Forfold, line 17: has this already been linted?
```suggestion
            position
      )::uuid[] AS participants
```

Forfold (approved):

engine/rotationmanager/queries.sql
mastercactapus: it has to have the cast so that `sqlc` infers the correct type -- it's usually pretty good but any coalesce or nested queries (like this one) tend to cause trouble without an explicit cast

migrate/migrations/20250311113136-track-rotation-updates.sql
mastercactapus: We're actually not lowering the threshold itself from Postgres’s default (which is 50 rows by default). We're explicitly setting it here just for clarity. The meaningful change we're introducing is reducing autovacuum_vacuum_scale_factor from its default of 0.2 (20%) to 0.05 (5%), causing vacuum operations to trigger sooner relative to table size.

Both values (autovacuum_vacuum_threshold and autovacuum_vacuum_scale_factor) are explicitly defined here just for clarity and readability, even though only the scale factor is actually changing from the default behavior.

PR #4271, opened 2025-02-13
fix(Dockerfile): replace CMD with ENTRYPOINT

tongpu:
Switching from CMD to ENTRYPOINT makes is easier to provide commandline flags using the args flag, without having to add goalert to the args.

**Description:**
Switching from CMD to ENTRYPOINT makes is easier to provide commandline flags using the `args` key in Kubernetes, without having to add `goalert` to the args.

n/a

**Out of Scope:**
n/a

**Screenshots:**
n/a

Existing deployments which already use `args` in the following format
```yaml
args:
  - goalert
  - --public-url
  - https://goalert.example.com
```
could be facing issues with this change, since the results command line would be `/usr/bin/goalert goalert --public-url https://goalert.example.com`, which would potentially fail to start.

n/a

**Additional Info:**
n/a

### DIFF
--- a/devtools/ci/dockerfiles/goalert/Dockerfile
+++ b/devtools/ci/dockerfiles/goalert/Dockerfile
@@ -7,7 +7,7 @@ FROM docker.io/library/alpine
 RUN apk --no-cache add ca-certificates
 ENV GOALERT_LISTEN :8081
 EXPOSE 8081
-CMD ["/usr/bin/goalert"]
+ENTRYPOINT ["/usr/bin/goalert"]
 
 COPY --from=build /build/bin/build/goalert-linux-amd64/goalert/bin/* /usr/bin/
 RUN /usr/bin/goalert self-test


mastercactapus:
Point well taken on passing arguments, however, there are a few reasons why this won't work for the GoAlert container:
1. **Environment Variables as the Preferred Method**: Our recommended approach for configuring GoAlert in a container is through environment variables (e.g., GOALERT_PUBLIC_URL) rather than command-line flags. This method provides clearer configuration management and aligns with our documentation and best practices.

2. **Multiple Binaries Support**: The GoAlert container includes multiple binaries, with plans to incorporate more through plugins. Switching from CMD to ENTRYPOINT would limit the flexibility to execute these binaries directly, as it would override the default executable context.

3. **Ensuring Safe Updates**: A core principle of GoAlert is to ensure that updates do not disrupt existing deployments. As you've noted, changing to ENTRYPOINT could cause issues for deployments that currently use the args array in their configurations, potentially leading to startup failures.

mastercactapus:
No worries! We _do_ have them in the [Getting Started](https://github.com/target/goalert/blob/master/docs/getting-started.md#cli-flags) guide, but I think we could do a better job of making them more obvious -- esp. the important ones like the DB URL and public URL.

Any thoughts, suggestions, contributions, or feedback you have on that document or README/other would be appreciated -- we try to treat documentation confusion as a bug. The hard part is you only get to see it for the first time once :)

PR #4324, opened 2025-03-27
messages: record and display full message status history for notificaitons

mastercactapus:
**Description:**
Adds a `message_status_history` table to track the history of messages over time.

- Cleaned up along with `outgoing_messages` via foreign key
- Alert details page updated to allow viewing full message history

**Out of Scope:**
- Admin message logs has not been updated to display the new history.
- No metrics are emitted or aggregated for things like time-to-delivery.
- Alert log messages are added when _sent_ but not when initially scheduled, this change makes fixing that a natural improvement, but is not in this PR.

**Screenshots:**
image

- Alert logs now have an expansion option for sent notifications

New query for getting the history of a message.


KatieMSB (approved):

AlaricWhitney (approved):
LGTM

PR #4330, opened 2025-04-02
dev: switch to `go tool` for most tools

mastercactapus:
**Description:**
Switches to using `go tool` from the combination of make targets and `go run` paths.
- replaces most usages of `bin/<tool>`, `bin/tools/<tool>`, `go run ...` with `go tool <tool>`
- replaces all tool-related make targets with the `tool` metapath `go install tool` (which builds binaries for ALL defined tools and places them in `GOBIN`
- removed all old ci tasks/scripts that are no longer used
- migrated `sqlc` to be version tracked via `go.mod` and used via `go tool` rather than manually downloading the binary
- removed old `tool.go` now that everything is managed via `go tool`
- updated `go:generate` commands to use `go tool` removing the need to do relative paths
- updated Procfile* and cypress helper to use `go tool`


KatieMSB (dismissed):

Forfold (approved):
🚀

KatieMSB (approved):

PR #4381, opened 2025-07-02
dataloader: refactor for simplicity and param support

mastercactapus:
**Description:**
Refactors the dataloader implementation to be simpler to follow and adds support for parameterized batches.

**Additional Info:**
Depends on #4380


weefatboi (approved):
much simpler, nice! 

does this need a docs usage update too, or does this codebase do it comment-based / self-documenting style?

Issue #4187, opened 2024-12-10
Finalize Universal Integration Keys for Stable Release


mastercactapus:
**What problem would you like to solve? Please describe:**

We need to finalize and polish the Universal Integration Keys (UIK) feature so it can be moved out of the experimental state. The current pain points include:

- Missing guidance on how to use a UIK.
- No links or info on writing Expr expressions.
- No straightforward way to test conditions and actions.
- Adding and editing actions is confusing.
- No visibility into whether recent requests worked, failed, or even occurred.
- Handling of invalid data returns a `500` instead of a `400` error.

**Describe the solution you'd like:**

**Must-have enhancements:**  
- [x] Display a copyable POST endpoint URL on the Edit page, and clarify that requests should use `Authorization: Bearer <token>` (PR #4189).  
- [x] Support multi-line edit fields with syntax highlighting.  
- [x] Add query parameters to the Expr evaluation environment.  
- [x] Include an info icon indicating that the expressions use the Expr language, with a link to relevant documentation.  
- [ ] Provide a way to test an individual expression using a given JSON payload and query string.
- [ ] Improve the action editing flow to make it more intuitive.

**Additional improvements:**  
- [ ] Allow testing the entire key with a given JSON payload and query string, and show which rules evaluated and how.  
- [ ] Log recent requests (both successes and failures) to aid in debugging.  
- [ ] Support saving test payloads for quick reuse.  

**Additional context:**
- The top-level card should include instructions on how to make a request, preferably in a sidebar or prominent location.
- In edit dialogs, add a `<< Debug` button that expands a panel to the left. This panel should allow selecting/editing a test request. All Expr editors should use the test request data as input and display evaluated output in real-time (e.g., in a bottom overlay) when fields are focused.

**Action editing/adding flow:**
- Show only an "Add Action" button initially. When clicked, reveal a screen for choosing the destination type and disable other controls until a choice is made.
- In this "Add Action" view, provide two buttons: "Save this Action" and "Delete this Action".
- On the main action list (chips), replace the delete icon with an edit icon. Clicking edit enters the same flow as add, and from there the action can be deleted if needed (no direct delete from the chip list).

mastercactapus:
Possible idea for the "info icon" placement
![image](https://github.com/user-attachments/assets/d1159e41-99b0-4200-ade3-990b0593de8d)

And from there link to https://expr-lang.org/docs/language-definition or a tooltip with the tagline from the Expr homepage (including syntax link)
[...]

PR #4395, opened 2025-07-22
uik: save most recent request to universal integration key

KatieMSB:

**Description:**
Adds new `uik_logs` table for storing the most recent requests to universal integration keys. 

Part of #4187 

**Out of Scope:**
Follow up work will be to make logs viewable in the UI to aid in debugging integrations with universal integration keys.

**Additional Considerations:**
This implementation focuses on capturing and storing the most recent request log for each universal integration key to support basic debugging functionality as a MVP.
In the future, we may want to expand this system to store multiple logs per key, with smarter deduplication strategies or time-based retention. Ideas include:
- Retaining the last N unique logs, with a configurable maximum (e.g. 15 per integration key), potentially deduplicated by payload and result type.
- Tracking both the most recent successful request and the most recent error, as different error types (e.g. parse, execution, send) may need to be surfaced separately.
- Adding support for aggregate metadata, such as first/last seen timestamps and occurrence counts, similar to Kubernetes’ event log behavior.
- Setting TTLs for stored logs, such as keeping full payloads for 48 hours while retaining metadata longer.
- Future UI support could show a breakdown of what actions were generated, including any errors encountered, and could allow for test payload replay and pinning important logs.


expflag/flags.go
```diff
@@ -7,11 +7,13 @@ type Flag string
 const (
 	Example  Flag = "example"
 	UnivKeys Flag = "univ-keys"
+	UikLogs  Flag = "uik-logs"
```
mastercactapus, line 10: nit: https://go.dev/wiki/CodeReviewComments#initialisms
```suggestion
	UIKLogs  Flag = "uik-logs"
```

also, I'm wondering if there's a different way to describe this than "logs" and "logs support" since it doesn't have anything to do with application logs, and the table itself only records the last status/result, and maintains no running "log" of events.

Maybe `uik_requests` for the table, and the expflag I'm less sure, maybe `uik-debug`?

integrationkey/queries.sql
```diff
@@ -48,6 +48,27 @@ WHERE
...
+    $1, $2, $3, $4, $5, $6, $7
+)
+ON CONFLICT (integration_key_id) DO UPDATE
+SET
+    content_type = EXCLUDED.content_type,
+    user_agent = EXCLUDED.user_agent,
+    raw_body = EXCLUDED.raw_body,
+    status = EXCLUDED.status,
+    error_message = EXCLUDED.error_message,
+    received_at = EXCLUDED.received_at;
```
mastercactapus, line 70: For testing/consistency, we'll want to use DB time -- if we pass in the `received_at` time, we'll need to do a separate query to get the current time. Instead, we can just rely on the default of `now()` and on the conflict update set the value to the current DB time.

```suggestion
-- name: IntKeyLog :exec
INSERT INTO uik_logs (
    integration_key_id,
    content_type,
    user_agent,
    raw_body,
    status,
    error_message
) VALUES (
    $1, $2, $3, $4, $5, $6
)
ON CONFLICT (integration_key_id) DO UPDATE
SET
    content_type = EXCLUDED.content_type,
    user_agent = EXCLUDED.user_agent,
    raw_body = EXCLUDED.raw_body,
    status = EXCLUDED.status,
    error_message = EXCLUDED.error_message,
    received_at = now();
```

mastercactapus (changes requested):

PR #4397, opened 2025-07-29
auth-check: replace "auth check" mechanism with proper external call counting

mastercactapus:
**Description:**
- Adds `calllimiter` which will limit external calls (DB query & HTTP request)
- Switches to a passed `http.Client` rather than relying on DefaultClient in-app
- Adds caching for individual Slack channel info with a fill mechanism
- Removes auth check code from permissions package
- Adds test for UIK config with large number of Slack channels (previously would fail due to too many auth checks)
- Replace incorrect use of `.EqualValues` where order isn't guaranteed with `.ElementsMatch`

**Out of Scope:**
- Not migrating the rest of Slack methods to use the new `GetOrFill` method

N/A

N/A

### DIFF
--- a/engine/message/dedupalerts_test.go
+++ b/engine/message/dedupalerts_test.go
@@ -33,7 +33,7 @@ func TestDedupAlerts(t *testing.T) {
 	res, err := dedupAlerts(messages, func(parentID string, duplicates []string) error {
 		assert.Equal(t, "4", parentID)
 		sort.Strings(duplicates)
-		assert.EqualValues(t, []string{"3", "5"}, duplicates)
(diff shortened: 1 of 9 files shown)


notification/slack/cache.go
```diff
@@ -56,3 +70,38 @@ func (c *ttlCache[K, V]) Get(key K) (val V, ok bool) {
 
 	return val, false
 }
+
+func (c *ttlCache[K, V]) GetOrFill(key K, fn func() (V, error)) (val V, err error) {
```
AlaricWhitney, line 74: can you add a functional comment to explain the purpose of the function is?

Issue #4084, opened 2024-09-25
proposal: Merging Schedules and Rotations


mastercactapus:
Some initial thoughts on deprecating Rotations in favor of making Schedules "smarter".

# RFC: Merging Schedules and Rotations in GoAlert

## Overview
GoAlert currently manages on-call responsibilities through two separate systems: **Schedules** and **Rotations**. Schedules consist of rules (assignments) such as day-of-week filters and optional time windows, which can target either a user or a rotation. Rotations are simple lists of users that rotate based on time intervals.

### Current Issues
1. **Always-active rotations**: No way to restrict rotations to specific days or hours, requiring manual management.
2. **User confusion**: Schedules and rotations can both be assigned to escalation policy steps, and users sometimes get confused about which to use, especially since only schedules have temporary overrides.
3. **Complex rotation code**: Rotations rely on **database triggers**, which makes them hard to maintain and prone to deadlocks. Joining across multiple tables adds complexity when calculating schedules.
4. **Limited future on-call data**: Rotations don’t expose future on-call information, which limits integration with external systems that need this data.

### Proposed Solution
I propose consolidating rotations into schedules, enhancing schedule functionality to cover all use cases of rotations. The benefits of this solution include:
1. **Deprecating rotations** while maintaining compatibility for existing systems, ensuring gradual migration.
2. **Simplifying scheduling logic** by storing all configuration in a single JSONB field in `schedule_data`, removing the need for table joins and database triggers.
3. **Improving future on-call data visibility** by consolidating all scheduling logic under schedules.

### Long-Term Compatibility Plan
To minimize disruption, rotations will remain in the codebase for a period of time:
1. **Opt-in configuration**: Introduce a configuration option to **disable new rotations** creation. Initially, this will be opt-in, allowing early adopters to transition to schedules.
2. **Opt-out configuration**: In a future release, the creation of new rotations will be **disabled by default**, with the option to opt back in.
3. **Complete removal**: Eventually, rotations will be removed entirely, and a newer version of GoAlert will refuse to start or migrate if rotations are detected.
4. **Auto convert/migrate**: Include an admin option for **bulk migration** of existing rotations to schedules when the auto-convert functionality is built.

## Migration Strategy
1. **Deprecate Rotations**: 
   - Add a confirmation before assigning a rotation to an escalation policy.
   - Provide a "Convert to Schedule" link to migrate rotations into schedules.
   
2. **Enhance Schedules**:
   - Allow assigning multiple users to a schedule at once.
[...]

PR #4448, opened 2025-12-08
fixed daily rotation participation distribution for schedules with day restrictions

### LINKED ISSUE #4447: [BUG] Rotation with 14 participants only shows 10 in schedule when using weekday-only rule
### Description:

When I assign a rotation with 14 participants to a schedule with a weekday only rule(`09:00-17:00`) only 10 of the 14 participants appear in the schedule's calendar. The `rotation_participants` table stores all 14 participants and the `participant_count` column has value 14 in the `rotations` table.

### Steps to reproduce:

1. Create a rotation with 14 participants
2. Set rotation type to Daily with shift_length = 1
3. Create a schedule assign the rotation with Start as 09:00 and End as 17:00 and Days: Monday through Friday only
4. Navigate to the schedule calendar view
5. View future months
6. Observe that only 10 unique participants appear in the calendar

### Expected Behavior:

All 14 rotation participants should appear in the schedule calendar as the rotation cycles through. Over 20 weekdays in a month, each of the 14 participants should get at least one shift, with some getting two shifts as the rotation cycles.

### Observed Behavior:

Only 10 of 14 participants appear in future schedule shifts.

thulasirajkomminar:
- [x] All new and ~existing tests passed~.

**Description:**

When a daily rotation is combined with a schedule which has restriction(eg: active only on weekdays) the rotation advances through calendar time and not with active days. Due to this some participants fell on the inactivate days and never appeared on the schedule.

This PR fixes it by advancing rotations based on active day instead of calendar time.

Fixes #4447 

**Out of Scope:**

**Screenshots:**

**Additional Info:**

### DIFF
--- a/oncall/singlerulecalculator.go
+++ b/oncall/singlerulecalculator.go
@@ -63,32 +63,11 @@ func (t *TimeIterator) NewSingleRuleCalculator(loc *time.Location, rule Resolved
 			rotCopy := *rule.Rotation
 			rotCopy.Users = append([]string{}, rule.Rotation.Users...)
 
-			// For daily rotations with day restrictions, we need to pause the rotation
+			// for daily rotations with day restrictions, we need to pause the rotation
 			// on inactive days to ensure fair distribution (otherwise some participants
 			// would never appear if their turn always falls on inactive days).
 			if !rule.IsAlways() && rotCopy.Type == rotation.TypeDaily {
-				// For daily rotation with day restrictions, advance position per active day
-				activeStart := rule.StartTime(t.Start().In(loc))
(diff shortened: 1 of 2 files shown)


mastercactapus:
Thanks for putting this together! Unfortunately I can't merge this as-is—the change updates the calendar display calculation but doesn't change the actual rotation behavior (which is controlled by the engine evaluating on-demand and updating the DB). This would result in the calendar showing different people than who would actually be on-call when alerts fire.

I left more context on #4447 about the underlying architecture. Proper fix is tracked in #4084.

thulasirajkomminar:
Closing this PR since a larger refactor is planned.

PR #4483, opened 2026-05-04
build(deps): update golang and goalert build-env to v1.26.3

mastercactapus:
**Description:**
Update to Go 1.26.3 as well as all Go dependencies to latest versions


devtools/gettool/getlint.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	url := fmt.Sprintf("https://github.com/golangci/golangci-lint/releases/download/v%s/golangci-lint-%s-%s-%s.%s",
+		version, version, osName, runtime.GOARCH, ext,
+	)
+	fd, n, err := fetchFile(url)
+	if err != nil {
+		return fmt.Errorf("fetch: %w", err)
+	}
+	defer fd.Close()
```
AlaricWhitney, line 32: question:  is this needed if golangci-lint is listed as a tool within go.mod?

AlaricWhitney (approved):

devtools/gettool/getlint.go
mastercactapus: I removed that in favor of this as there were countless build issues with different dependencies and golanglint-ci explicitly recommends against it.

image

KatieMSB (approved):

PR #4522, opened 2026-07-10
Fix auto closing

KatieMSB:

**Description:**
When creating an alert with a universal integration key the config allows for setting a dedup key based on the received alert payload, but the key was never passed into the created alert config on create/update so any alerts created via UIK used the default dedup hash created from the summary + details. 

This surfaced as upstream systems unable to close alerts automatically due to the dedupe hash of the opened alert not matching the cleared alert request.

This pr adds the uik configured dedup key to the alert on create/update.


CameronJHall (approved):
Good catch

weefatboi (approved):

PR #4565, opened 2026-09-17
Update Validity Period from 10s to 30s

lekhacsl:
**Description:**
Increases the RCS ValidityPeriod from 10 seconds to 30 seconds to address the higher number of RCS delivery errors.

The increased validity period gives messages additional time to be delivered before they expire, helping reduce failures caused by the current 10-second window.


KatieMSB (approved):

CameronJHall (approved):
Is this something that we'd want to tune without making a code change - i.e. should it be configurable?

nahumguess (approved):

PR #4566, opened 2026-09-18
fix(webhook): handle HTTP error responses

### LINKED ISSUE #4558: Outgoing webhook notifications marked sent on HTTP error responses
**Describe the Bug:**
Outgoing webhook notifications are treated as successfully sent as long as the HTTP request completes, even when the destination returns an error status (500, 404, 429, etc.).

That means GoAlert will not retry, the outgoing message is recorded as sent, and the alert log looks like the page went through. If the webhook is the only notification path (or the one that actually pages the on-call system), a failed delivery is silent.

This is in `notification/webhook/sender.go` `SendMessage`. After building the POST it does:

```go
_, err = http.DefaultClient.Do(req)
if err != nil {
    return nil, err
}

return &notification.SentMessage{State: notification.StateSent}, nil
```

Three problems in that path:

1. The response status is never checked. `http.Client.Do` only returns an error for transport failures (timeout, DNS, connection refused). A 500 from the receiver is a nil error, so the message is marked `StateSent`.
2. The `*http.Client` stored on `Sender` is never used. Startup passes the app client (`webhook.NewSender(ctx, app.httpClient)` in `app/startup.go`), which wraps the transport with `calllimiter.RoundTripper`. Using `http.DefaultClient` bypasses that limiter and any other client config.
3. The response body is never closed. On a successful `Do`, that leaks the connection.

The engine already has the right failure handling. In `engine/message/db.go`, `StateFailedTemp` sets `next_retry_at` (up to 3 retries, 15 seconds apart) and `StateFailedPerm` fails the message for good. Returning `StateSent` skips all of that.

Invalid/disallowed URLs already return `StateFailedPerm` just above this. Error HTTP statuses should go through the same state machine instead of being reported as sent.

**Steps to Reproduce:**
1. Enable Webhooks in Admin config.
2. Add a webhook destination (contact method or escalation-policy channel) pointing at an endpoint that returns HTTP 500, or any non-2xx.
3. Create an alert that notifies that destination.
4. Watch the destination: it receives the POST and responds 500.
5. Check the outgoing message / alert log in GoAlert.

**Expected Behavior:**
A non-success HTTP response should not be recorded as sent.

- 5xx and 429 should be `StateFailedTemp` so the existing retry path can try again (same as a transport error).
- Other 4xx should be `StateFailedPerm`, since retrying a 404/401 without a config change will not help.
- `SendMessage` should use `s.Client` (falling back only if it is nil), and should always close the response body.
- The outgoing message should show as failed, with the status code in `status_details`, so it is obvious the page did not land.

**Observed Behavior:**
The POST is made, the destination returns an error, and GoAlert still records the message as sent. There is no retry. The alert log shows a successful notification.

Transport failures (connection refused, timeout) do fail and retry, because those return an error from `Do`. Only HTTP-level failures are swallowed.

**Screenshots/Stack Traces:**
N/A. Confirmed in source on current master and the v0.34.1 tag. The send is the last lines of `SendMessage` in `notification/webhook/sender.go`.

**Application Version:**
v0.34.1 and current master. Same send path on both.

**Desktop:**
N/A (server-side send path, not UI-specific)

**Smartphone/Tablet:**
N/A

**Additional Context:**
`notification.StateSent` is documented as the message having been sent completely. Treating a 5xx as sent is the opposite of that, and it is worse for webhooks than for SMS/voice because there is no later delivery callback to correct the status.

Slack already uses the injected client (`cs.cfg.Client` in `notification/slack/withclient.go`). Webhooks should do the same.

Happy to open a PR: use `s.Client`, close/drain the body, map 2xx to `StateSent`, 5xx/429 to `StateFailedTemp`, and other 4xx to `StateFailedPerm`, with a unit test against an `httptest.Server`.

artgas1:
## Summary

- treat successful 2xx webhook responses as sent
- retry 429 and 5xx responses through the existing temporary-failure path
- mark other non-2xx responses as permanent failures
- include the HTTP status in failure details

GoAlert currently records every completed HTTP request as sent, even when the webhook endpoint returns an error. This can silently drop an on-call notification because the existing retry path is skipped.

## Test plan

- `go test ./notification/webhook -run '^TestSender_HTTPStatus$' -count=1`
- `go test ./notification/webhook -count=1`
- `make check`

QA-gate: critical-chain (outgoing webhook pages must not be reported as sent after HTTP errors)

Fixes #4558


artgas1:
Hi @mastercactapus, would you mind taking a look at this webhook retry fix when you have a chance? It addresses #4558, where non-2xx webhook responses were being marked as sent. This PR routes 429 and 5xx responses through the existing temporary-failure path and treats other non-2xx responses as permanent failures. I'd especially appreciate your take on those status classifications. Thanks!

